PR #9, opened 2025-05-05
Add base cache implementation

ThaminduDilshan:
## Purpose

This pull request introduces caching functionality defining a base cache and related operations.

### Caching System:

* [`internal/system/cache/basecache.go`](diffhunk://#diff-6ead129ecfa4f8b01cf7fe27eb01f587a4b5af8e06bc9243a3cb18adc76bdcffR1-R104): Added `BaseCache`, a thread-safe in-memory caching system with support for adding, retrieving, and clearing cache entries. Includes expiration handling for cache entries.

### DIFF
--- a/internal/system/cache/basecache.go
+++ b/internal/system/cache/basecache.go
@@ -1,0 +1,104 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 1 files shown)


internal/system/cache/basecache.go
```diff
@@ -0,0 +1,104 @@
...
+	entry, exists := bc.cache[CacheKey(key)]
+	if !exists || time.Now().After(entry.ExpiryTime) {
+		// Remove the expired entry.
+		delete(bc.cache, CacheKey(key))
+
+		return nil
+	}
+
+	return entry
+}
```
rajithacharith, line 88: We have used a read lock, 
But we are doing delete operation as well. 
Suggestion
```
func (bc *BaseCache) GetValueFromCache(key CacheKey) *CacheEntry {
	bc.mu.RLock()
	entry, exists := bc.cache[key]
	if !exists || time.Now().After(entry.ExpiryTime) {
		bc.mu.RUnlock() // Unlock read lock before acquiring write lock

		// Acquire write lock to delete expired entry
		bc.mu.Lock()
		defer bc.mu.Unlock()

		// Double-check if the entry still exists and is expired (in case of race)
		if entry, exists := bc.cache[key]; exists && time.Now().After(entry.ExpiryTime) {
			delete(bc.cache, key)
		}
		return nil
	}

	bc.mu.RUnlock()
	return entry
}
```

ThaminduDilshan: Addressed with https://github.com/asgardeo/thunder/pull/29

PR #149, opened 2025-06-16
Fix attribute collector with federated flows and update docs

ThaminduDilshan:
## Purpose

This pull request introduces updates to the handling of authenticated user attributes and runtime data across various authentication and flow processing components. Additionally, it standardizes the default UserID placeholder used in federated scenarios. The changes aim to enhance attribute management and improve consistency in user-related data handling.

Additionally this PR adds the auth flow customization guide.

### DIFF
--- a/backend/internal/executor/attributecollect/attributecollector.go
+++ b/backend/internal/executor/attributecollect/attributecollector.go
@@ -83,7 +83,10 @@ func (a *AttributeCollector) Execute(ctx *flowmodel.NodeContext) (*flowmodel.Exe
 		log.String(log.LoggerKeyFlowID, ctx.FlowID))
 	logger.Debug("Executing attribute collect executor")
 
-	execResp := &flowmodel.ExecutorResponse{}
+	execResp := &flowmodel.ExecutorResponse{
+		AdditionalData: make(map[string]string),
+		RuntimeData:    make(map[string]string),
+	}
 
 	if !ctx.AuthenticatedUser.IsAuthenticated {
 		logger.Debug("User is not authenticated, cannot collect attributes")
--- a/backend/internal/executor/attributecollect/attributecollector.go
+++ b/backend/internal/executor/attributecollect/attributecollector.go
@@ -34,7 +34,6 @@ import (
 const (
 	loggerComponentName   = "AttributeCollector"
 	userAttributeUserID   = "userID"
-	userAttributeUsername = "username"
 	userAttributePassword = "password"
 )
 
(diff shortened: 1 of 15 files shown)


backend/internal/flow/engine/flowengine.go
```diff
@@ -169,8 +169,18 @@ func updateContextWithNodeResponse(engineCtx *model.EngineContext, nodeResp *mod
 
 	// Handle authenticated user from the node response
 	if nodeResp.AuthenticatedUser.IsAuthenticated {
+		authnUserAttrs := engineCtx.AuthenticatedUser.Attributes
 		engineCtx.AuthenticatedUser = nodeResp.AuthenticatedUser
 
+		// If engine context already had authenticated user attributes, merge them with the new ones
```
darshanasbg: Why we need to merge? Isn't the latest one is the most uptodate one?

backend/internal/executor/attributecollect/attributecollector.go
```diff
@@ -368,11 +412,16 @@ func (a *AttributeCollector) getInputAttributes(ctx *flowmodel.NodeContext) map[
 
 	for _, inputAttr := range requiredInputAttrs {
 		// Skip special attributes that shouldn't be stored in the user profile
-		if inputAttr.Name == "username" || inputAttr.Name == "userID" || inputAttr.Name == "password" {
+		if inputAttr.Name == userAttributeUsername || inputAttr.Name == userAttributeUserID ||
```
darshanasbg: Setting username in a attribute collection step (when username is not already set) is a valid use case.. So lets avoid skipping username.

backend/internal/flow/engine/flowengine.go
ThaminduDilshan: Here actually the attributes from node response is added to the engineCtx. If the same attribute exists in both, nodeCtx one overrides the engine ctx attribute.

Updated the comment with 0a74469574fb98bae85241d073fb528a763728a1

backend/internal/executor/attributecollect/attributecollector.go
ThaminduDilshan: Fixed with 0a74469574fb98bae85241d073fb528a763728a1

backend/internal/flow/engine/flowengine.go
darshanasbg: One in the engineCtx is probably from the previous step.. So, IMO we can depend on the latest one, rather combining two. But we can revisit this later.

darshanasbg (approved):

backend/internal/flow/engine/flowengine.go
ThaminduDilshan: We actually depend on the latest one.

PR #171, opened 2025-06-24
Update release workflow to bump sample app version

ThaminduDilshan:
## Purpose
$subject

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,11 +9,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Thunder product version (e.g., v1.0.0)'
-        required: true
-        type: string
-      sample_version:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -9,7 +9,11 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: 'Release version (e.g., v1.0.0)'
+        description: 'Thunder product version (e.g., v1.0.0)'
+        required: true
+        type: string
+      sample_version:
```
darshanasbg: shall we use the same version, rather requesting two inputs?

I don't see a need to have two versions as of now, and requesting two things which needs to be same for the most of the time, would be error prone.

ThaminduDilshan: Fixed with 83601c389e931dc753e56cead6891cbc9b354305

darshanasbg (approved):

PR #220, opened 2025-07-03
Fix SMS auth flow and add all option registration graph

ThaminduDilshan:
## Purpose

- Fixes the SMS auth executor to get user id from context and skip user identification when the user is already authenticated. This fixes the authentication failure when the mobile number is prompted from the user during the auth flow.
- Add the customized all option registration flow graph

### DIFF
--- a/backend/cmd/server/repository/resources/graphs/auth_flow_config_basic_google_github_sms.json
+++ b/backend/cmd/server/repository/resources/graphs/auth_flow_config_basic_google_github_sms.json
@@ -127,6 +127,11 @@
                     "name": "lastName",
                     "type": "string",
                     "required": false
+                },
+                {
+                    "name": "mobileNumber",
+                    "type": "string",
+                    "required": false
                 }
             ],
             "executor": {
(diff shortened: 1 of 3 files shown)


backend/cmd/server/repository/resources/graphs/auth_flow_config_basic_google_github_sms.json
```diff
@@ -127,6 +127,11 @@
                     "name": "lastName",
                     "type": "string",
                     "required": false
+                },
+                {
```
ThaminduDilshan, line 131: This is required. Otherwise mobile will not be saved in user profile in MFA flows

backend/cmd/server/repository/resources/graphs/registration_flow_config_basic_google_github_sms.json
```diff
@@ -0,0 +1,164 @@
...
+                {
+                    "name": "username",
+                    "type": "string",
+                    "required": false
+                },
+                {
+                    "name": "password",
+                    "type": "string",
+                    "required": false
+                },
```
ThaminduDilshan, line 125: Prompt optional username & password to allow basic login

darshanasbg (approved):

PR #249, opened 2025-07-21
Refactor the authentication flow

ThaminduDilshan:
## Purpose

This pull request refactors the authentication handling logic by moving the service level logic in AuthenticationHandler to a a new AuthenticationService implementation. The changes aim to improve modularity and simplify flow execution and error handling. Additionally, minor updates were made to constants to support the new service structure.

## Related Issues
- https://github.com/asgardeo/thunder/issues/82

## Related PR:
- https://github.com/asgardeo/thunder/pull/238

PR #238, opened 2025-07-17
Modify authorization code flow to work with flow execution

ThaminduDilshan:
## Purpose

This pull request removes the legacy authenticators and integrates the authorization code flow with new flow engine/execution. This introduces the following changes.
- Modify the Authentication endpoint (`flow/authn`) to only accept POST requests and define a contract similar to flow execution endpoint (simply will be a wrapper for flow execution endpoint)
- Authorization handler redirects the initial authorization request to the configured gate client login page.
- Initial authentication request contains the `sessionDataKey` and upon receiving it, server will remove the previous session and creates a new one with the generated flowId.
- Gate client should execute the authentication flow by invoking the authentication endpoint and at the end will receive a redirect url with the COMPLETE status. Redirection happens to the authorize endpoint with the new session data key.
- Gate client should then perform the redirection which will again be redirected to the application callback with the authorization code.
- This change also modifies the sample gate client to handle the auth flow as per the above implementation, however only supports local authentications (basic).

### OAuth Application Enhancements:
* Added a new `ID` field (`app_id`) to the `OAuthApplication` model and updated related query logic to include `app_id` in database operations. (`backend/internal/application/model/oauthapp.go` [[1]](diffhunk://#diff-296b99e6ee7eed869a234d8fa13b2594be72e637afa713e5ba7d77261ee1d419R31) `backend/internal/application/store/constants.go` [[2]](diffhunk://#diff-97834acd579649526d90b2f2113c0f3a9b96b58615c0dfd8785ecd4c61be5793L27-R27)
* Updated the `GetOAuthApplication` method to retrieve and handle `app_id` from the database. (`backend/internal/application/service/applicationservice.go` [[1]](diffhunk://#diff-184bf5335b299c8ffcabe997d2252f1818e2a249d828d4fd2f2aafa891589fb5L84-R86) [[2]](diffhunk://#diff-184bf5335b299c8ffcabe997d2252f1818e2a249d828d4fd2f2aafa891589fb5L106-R113)

### Parsing and Validation Improvements:
[...]

Issue #82, opened 2025-05-29
Integrate OAuth flow with auth executors and Remove legacy authenticators


ThaminduDilshan:
### Current Limitation

In early stages thunder was using outbound authenticators [1] that can be configured via the `deployment.yaml` file [2]. This is mainly used for configuring OIDC outbound authenticators such as Github login. We recently have introduced support for the new authentication flow engine model that aligns with the flow orchestration architecture via [3] and [4] deprecating the existing legacy authenticator implementation. However still the OAuth flow utilizes this legacy authenticator model.

[1] https://github.com/asgardeo/thunder/tree/main/backend/internal/outboundauth
[2] https://github.com/asgardeo/thunder/blob/c7978352a5c1d2b980e0308b0ff385bbf911e0bb/backend/cmd/server/repository/conf/deployment.yaml#L33-L39
[3] https://github.com/asgardeo/thunder/pull/41
[4] https://github.com/asgardeo/thunder/pull/77

### Suggested Improvement

Improve OAuth flow to support new auth executors via the authentication flow engine (service) and remove the legacy authenticators implementation.

PR #254, opened 2025-07-22
Refactor AuthN: Remove the authn and introduce authz back channel endpoint

ThaminduDilshan:
## Purpose

This pull request significantly refactors the previous authentication handling logic as below.
- Removes the `flow/authn` endpoint and related service completely. With this there won't be an authentication endpoint for the server.
- Introduces a new POST resource to the `authorize` endpoint with the ability to handle authorize requests after flow completion. This endpoint acts as a backchannel authorize call and accepts a `sessionDataKey` and a `assertion` in json body params. It returns a json response containing a `redirect_uri` to the application callback (if SUCCESS) or to the configured gate client error page (if FAILURE).
- Update the gate client to handle the new authz/authn changes.

In an authorization code flow, the authentication flow happens as below.
1. The application sends a GET request for the initial authorization for the server authorize endpoint.
2. Server redirects the client (browser) to the gate client login page with `applicationID` and `sessionDataKey` query params.
3. Gate client invokes the app native `flow/execution` endpoint with the received app ID and receive the assertion at end of the authentication.
4. Gate client sends a POST request to the server `/authorize` endpoint with the sessionDataKey and the assertion (JWT token).
5. Server performs the necessary validations and responds with the `redirect_uri` in the response body.
6. Gate client does the redirection to the application callback (with code) or error page as per the received redirect uri.

The changes aim to improve modularity and simplify flow execution and error handling. Additionally, minor updates were made to constants to support the new structure.

## Related Issues
- https://github.com/asgardeo/thunder/issues/82

## Related PR:
- https://github.com/asgardeo/thunder/pull/238
- https://github.com/asgardeo/thunder/pull/249 (deprecates this PR)


backend/internal/oauth/oauth2/authz/utils.go
```diff
@@ -0,0 +1,270 @@
+/*
```
darshanasbg: Do we really need this to be separate out to a `utils.go`? Since these are not reused can't we maintain these functions in the  `authzhandler.go` itself?

backend/internal/oauth/oauth2/authz/utils.go
```diff
@@ -0,0 +1,270 @@
...
+
+	sessionDataKey := authZReq.SessionDataKey
+	sessionDataStore := sessionstore.GetSessionDataStore()
+	ok, sessionData := sessionDataStore.GetSession(sessionDataKey)
+	if !ok {
+		return nil, fmt.Errorf("session data not found for session data key: %s", sessionDataKey)
+	}
+
+	// Remove the session data after retrieval.
+	sessionDataStore.ClearSession(sessionDataKey)
```
darshanasbg: An an improvement, we better to avoid using generalized SessionDataStore and have more specific storage implementation.

backend/internal/oauth/oauth2/authz/utils.go
ThaminduDilshan: Just extracted out them to increase the readability of the authzhandler.go file. Ideally these are utils, but shouldn't be able to reuse in other packages. Hence removed the previous utils package and moved it to this location to define private functions.

darshanasbg: These are not used anywhere except for the `authzhandler`.. And the functions like `getOAuthMessage()` is directly part of the core logic of the `authzhandler`..

Functions like `decodeAttributesFromAssertion()` we can keep in the utils (even it don't have multiple usages as of now), since its functionality not attached to `authzhandler`.. 

But separating out the functions to a util that is tightly bound to the caller (`authzhandler`) doesn't look like a good option.

If the `authzhandler` doing too many things, we should check on bringing suitable logical layer, like `request validator`, `storage`, `service`, and offload logic to that, rather than extracting things to general place like `util`.

If the code is, structured properly, IMO the no of lines in the file isn't not a much of a concern for readability.

ThaminduDilshan: Addressed the comment

backend/internal/oauth/oauth2/authz/utils.go
ThaminduDilshan: Will track as an improvement

ThaminduDilshan: Created an issue to track https://github.com/asgardeo/thunder/issues/255

darshanasbg (approved):

PR #304, opened 2025-08-08
Changes to make frontend build easier. No functional changes.

PasinduRavimal:
In this commit following changes were made,
1. Modified apps/gate/esling.config.mjs to silent linter errors.
2. Modified apps/gate/src/app/login/content.tsx to make prettier linter happy. No functional modifications.
3. Modified package.json to use pnpm version 8 when installing to make sure the lock file is honoured.

All these changes were tested and verified in 2 debian distributions, Ubuntu 24 and Kali.

### DIFF
--- a/frontend/apps/gate/eslint.config.mjs
+++ b/frontend/apps/gate/eslint.config.mjs
@@ -73,6 +73,9 @@ const baseConfigs = [
       ],
       '@typescript-eslint/typedef': 'off',
       // TODO: Temporarily disable this rule until the plugin 
+      '@typescript-eslint/no-unused-vars': 'off', // Disable unused-vars
+      '@typescript-eslint/explicit-function-return-type': 'off', // Disable return type checks
+      'prefer-const': 'off', // Disable prefer-const rule
       // false positives are resolved
       'headers/header-format': 'off'
     }
--- a/frontend/apps/gate/src/app/login/content.tsx
+++ b/frontend/apps/gate/src/app/login/content.tsx
@@ -55,14 +55,19 @@ const LoginPageContent = function (): ReactElement {
     setInsecureWarning(params.get('showInsecureWarning') === 'true');
 
     if (key) {
-      axios.post(AppConfig.flowExecutionEndpoint, { applicationId: appId, flowType: "AUTHENTICATION" }, {
-        headers: {
-          Accept: 'application/json',
-          'Content-Type': 'application/json',
-        },
-        withCredentials: true,
-      })
-        .then((res) => {
+      axios
+        .post(
+          AppConfig.flowExecutionEndpoint,
+          { applicationId: appId, flowType: 'AUTHENTICATION' },
+          {
+            headers: {
+              Accept: 'application/json',
+              'Content-Type': 'application/json',
+            },
+            withCredentials: true,
+          },
+        )
+        .then(res => {
           setFlowId(res.data.flowId);
           setInputs(res.data.data?.inputs || []);
           setLoading(false);
@@ -81,7 +86,7 @@ const LoginPageContent = function (): ReactElement {
   };
 
   const handleTogglePasswordVisibility = () => {
-    setShowPassword((prev) => !prev);
+    setShowPassword(prev => !prev);
   };
 
   const handleSubmit = async (e: React.FormEvent) => {
@@ -92,14 +97,18 @@ const LoginPageContent = function (): ReactElement {
     let continueFlow = true;
     while (continueFlow) {
       try {
-        const res = await axios.post(AppConfig.flowExecutionEndpoint, {
-          flowId: currentFlowId,
-          inputs: currentInputs,
-        }, {
-          headers: { 'Content-Type': 'application/json' },
-          withCredentials: true,
-          validateStatus: () => true,
-        });
+        const res = await axios.post(
+          AppConfig.flowExecutionEndpoint,
+          {
+            flowId: currentFlowId,
+            inputs: currentInputs,
+          },
+          {
+            headers: { 'Content-Type': 'application/json' },
+            withCredentials: true,
+            validateStatus: () => true,
+          },
+        );
         if (res.status !== 200) {
           setError(res.data?.message || `Error response status: ${res.status}`);
           continueFlow = false;
@@ -111,14 +120,18 @@ const LoginPageContent = function (): ReactElement {
         } else if (res.data.flowStatus === 'COMPLETE') {
           const assertion = res?.data?.assertion;
 
-          const authZRes = await axios.post(AppConfig.authorizationEndpoint, {
-            sessionDataKey,
-            assertion,
-          }, {
-            headers: { 'Content-Type': 'application/json' },
-            withCredentials: true,
-            validateStatus: () => true,
-          });
+          const authZRes = await axios.post(
+            AppConfig.authorizationEndpoint,
+            {
+              sessionDataKey,
+              assertion,
+            },
+            {
+              headers: { 'Content-Type': 'application/json' },
+              withCredentials: true,
+              validateStatus: () => true,
+            },
+          );
 
           if (authZRes.status !== 200) {
             setError(authZRes.data?.message || `Authorization failed with status: ${authZRes.status}`);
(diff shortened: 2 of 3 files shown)


darshanasbg:
Thanks for the contribution @PasinduRavimal!

> 1. Modified apps/gate/esling.config.mjs to silent linter errors.
> 2. Modified apps/gate/src/app/login/content.tsx to make prettier linter happy. No functional modifications.
> 3. Modified package.json to use pnpm version 8 when installing to make sure the lock file is honoured. This is needed since the latest pnpm version ignores the lock file and generate errors.

Regarding the above changes,
* We can accept the 2 above.. 
* Regarding 1, turning off linter rules doesn't seems looks like the right thing to do
* Regarding 3, currently we have the gate app excluded from the product build. But when we used to have that, we had `PNPM 10+` as a pre-requisite. So the improvement 3 above is redundant and prefer to avoid having specific version hardcoded in the build script.

Also, we prefer to have minimal commits in the PR, so lets the merge commit excluded in the PR by rebasing.

Looping @jeradrutnam to share any comments.

frontend/package.json
```diff
@@ -4,11 +4,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "start": "pnpm run",
-    "build": "pnpm i --yes && nx run-many --target=build --all",
+    "build": "npx pnpm@8 install && nx run-many --target=build --all",
```
jeradrutnam: What's the reason behind downgrading pnpm version?

frontend/apps/gate/eslint.config.mjs
```diff
@@ -73,6 +73,9 @@ const baseConfigs = [
       ],
       '@typescript-eslint/typedef': 'off',
       // TODO: Temporarily disable this rule until the plugin 
+      '@typescript-eslint/no-unused-vars': 'off', // Disable unused-vars
+      '@typescript-eslint/explicit-function-return-type': 'off', // Disable return type checks
+      'prefer-const': 'off', // Disable prefer-const rule
```
jeradrutnam: Any reason to disable these rules?

frontend/apps/gate/src/app/login/content.tsx
```diff
@@ -55,14 +55,19 @@ const LoginPageContent = function (): ReactElement {
...
-      axios.post(AppConfig.flowExecutionEndpoint, { applicationId: appId, flowType: "AUTHENTICATION" }, {
-        headers: {
-          Accept: 'application/json',
-          'Content-Type': 'application/json',
-        },
-        withCredentials: true,
-      })
-        .then((res) => {
+      axios
+        .post(
```
jeradrutnam:
```suggestion
      axios.post(
```

jeradrutnam (changes requested):
Thank you for the contribution. 

Can you also please explain how the code formatting impact the "... frontend build easier" ?

frontend/package.json
jeradrutnam: image

pnpm is a prerequisite. We can install pnpm as a "pre-build" if want. But not a lower version.

jeradrutnam: Ok just noticed @thiva-k has override the pnpm version. We have to revert that change.

thiva-k: Reverted via #307

PasinduRavimal:
Pardon me for my wording there. I tried to mean "Changes to build frontend without errors".

frontend/apps/gate/eslint.config.mjs
PasinduRavimal: I get following errors otherwise:

      Failed to compile.
      
      ./src/app/login/content.tsx
      39:10  Error: 'appId' is assigned a value but never used.  @typescript-eslint/no-unused-vars
      84:70  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
      88:45  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
      92:51  Error: Missing return type on function.  @typescript-eslint/explicit-function-return-type
      95:9  Error: 'currentFlowId' is never reassigned. Use 'const' instead.  prefer-const
      96:9  Error: 'currentInputs' is never reassigned. Use 'const' instead.  prefer-const
      157:16  Error: 'err' is defined but never used.  @typescript-eslint/no-unused-vars
      
Since this is my very first contribution and I am not aware of the current development, I didn't want to correct those (because I thought this file is being actively developed). But since it is in the main branch I thought to disable these rules so the build process won't fail.
     
If this need to be rebased, I can address these issues and update the PR.

jeradrutnam: @ThaminduDilshan we need to fix these.

jeradrutnam:
> Pardon me for my wording there. I tried to mean "Changes to build frontend without errors".

No worries. Thanks for explaining. And once again thanks for the PR.

frontend/apps/gate/eslint.config.mjs
jeradrutnam: No worries. Yes @PasinduRavimal let's fix the issues instead of disabling the rules. Because if we disable the rules it will impact the future lint validations as well.

PasinduRavimal: I have updated the PR with the requested changes.

Although there is a merge to resolve pnpm lock file issue, the lock file is still ignored by the latest pnpm version. In fact,  now, even pnpm v8 complains that the new lock file is broken. I had to use the previous version of lock file with pnpm downgraded to v8 to build the frontend.

frontend/package.json
```diff
@@ -4,11 +4,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "start": "pnpm run",
-    "build": "pnpm i --yes && nx run-many --target=build --all",
+    "build": "pnpm install && nx run-many --target=build --all",
```
jeradrutnam, line 7: Any issue that you came across to remove `--yes` flag? It was added to skip confirmation messages prompt during next app builds.

frontend/apps/gate/eslint.config.mjs
jeradrutnam: Thanks for the update. Let's try to upgrade to new pnpm version globally and keep the existing lockfile. You can delete and update to the latest if you install the latest pnpm. Double check with `pnpm -v`

frontend/package.json
PasinduRavimal: ```--yes``` flag is not needed. ```pnpm install``` skips the confirmation message as it is but generates an error if ```--yes``` is used. In fact, ```--yes``` flag is not yet implemented, I guess.

frontend/apps/gate/eslint.config.mjs
PasinduRavimal: To be exact, Using the latest pnpm works. It just ignores the lock file and installs the dependencies with the latest versions. But the problem comes when you build the project. The project just builds for 20 minutes or so and fails with a SIGINT.

frontend/apps/gate/eslint.config.mjs
```diff
@@ -74,6 +74,7 @@ const baseConfigs = [
       '@typescript-eslint/typedef': 'off',
       // TODO: Temporarily disable this rule until the plugin
```
jeradrutnam, line 75: better move this after below rule. As it is not valid for below rule.

frontend/apps/gate/src/app/login/content.tsx
```diff
@@ -141,7 +153,7 @@ const LoginPageContent = function (): ReactElement {
           continueFlow = false; // Stop loop, wait for user to submit again
           setError(''); // Clear any previous error
         }
-      } catch (err) {
+      } catch (_err) {
```
jeradrutnam:
```suggestion
      } catch {
```

frontend/package.json
jeradrutnam: If that is that case then fine. Can you also provide a reference if you have?

PasinduRavimal: > If that is that case then fine. Can you also provide a reference if you have?

Refer the following,
[https://github.com/pnpm/pnpm/issues/8554](https://github.com/pnpm/pnpm/issues/8554)

frontend/apps/gate/eslint.config.mjs
PasinduRavimal: Lockfile issue is now resolved.

frontend/package.json
jeradrutnam: Hi @PasinduRavimal thanks for providing the reference. As per the reference it mention that the `--yes` flag is not working at all at the moment. But it seems it's a undocumented feature. As when using `pnpm i --yes` it skips confirmation messages prompts during the `next build` in CI.

This was added to bypass `The modules directories will be removed and reinstalled from scratch. Proceed? (Y/n) › true` etc. messages in during the build. So, I think better leave this as it is.

PasinduRavimal: Hi @jeradrutnam, Thank you for the reply. I tried the building process multiple times and I didn't get the above message.  But if I keep the ```--yes``` argument I get the following error:

image

It might be a problem only I have. So, if you want to leave it as it is, I will update the PR again.

jeradrutnam: you are using the latest pnpm right? If you still get the error, no need to revert. Let's proceed with this.

PasinduRavimal: Yeah it is the latest version.

PR #368, opened 2025-09-08
Remove seed script dependency

thiva-k:
## Purpose
Eliminate hardcoded seed script-based dependencies and replace with API-based entity creation for better flexibility and maintainability.

## Changes Made
- Update build process to remove seeding logic
- Clean up seed script files (`seed_data_*.sql`, `seed_data.sh`)
- Refactor documentation to use API-based entity creation based on requirements
- Fix Postman samples to use correct API formats and variables

## Related issue
- https://github.com/asgardeo/thunder/issues/370

### DIFF
--- a/samples/api/postman/idp.json
+++ b/samples/api/postman/idp.json
@@ -37,7 +37,7 @@
         "header": [],
         "body": {
           "mode": "raw",
-          "raw": "{\n  \"name\": \"Github\",\n  \"description\": \"Login with Github\",\n  \"client_id\": \"client1\",\n  \"client_secret\": \"secret1\",\n  \"redirect_uri\": \"https://localhost:8090/flow/authn\",\n  \"scopes\": [\n      \"user:email\",\n      \"read:user\"\n    ]\n}",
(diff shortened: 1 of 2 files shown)


thiva-k:
> A quick question about the first run experience, without seed data, there’s no straightforward way for a fresh environment to go from zero to a working state right?
> 
> Would it make sense to provide a Postman collection that chains the three main steps so a user can get started faster?
> 
> * Create an Organization Unit → save {{organization-unit-id}}
> * Create an IdP (using the new properties[] payload) → save {{idp-id}}
> * Create a User that references the OU → save {{user-id}}

Creating the Local IdP explicitly is a temporary work around since we are adding the Local IdP fallback at the code-level soon (until the Local IdP handling is finalized). Also the hardcoded organization for user registration flow will be removed with the finalization of org concept. 

But agree that Postman collections would be a good add-on for the current setup.

darshanasbg:
> * Create an IdP (using the new properties[] payload) → save {{idp-id}}

We have data seeding through API planned as followup improvement.

darshanasbg (approved):

Issue #370, opened 2025-09-08
Remove seed script dependency


thiva-k:
### Current Limitation

Thunder currently relies on seed scripts to populate initial data (users, applications, OUs, IDPs) which creates hardcoded dependencies and makes the system less flexible.

With this approach, users must perform additional seeding steps during startup, adding complexity and reducing the out-of-box experience. This approach also creates security concerns with hardcoded credentials, makes testing more complex as integration tests depend on specific seeded entity IDs (#367), and requires ongoing maintenance of seed scripts alongside code changes.

### Suggested Improvement

- Eliminate seed scripts and replace with API-based entity creation wherever needed to improve system maintainability. 
- Update README with step-by-step API calls for creating required entities
- Implement code-level fallback to automatically configure **Local IDP** if not created via APIs (since basic authentication is an integral feature). 

This approach will eliminate hardcoded dependencies, enable dynamic entity management, and make the system more secure by removing hardcoded credentials from seed scripts while reducing maintenance overhead.

### Version

v0.6.0

PR #383, opened 2025-09-14
Introduce default fallback config for Local IdP

thiva-k:
## Purpose
1. Add a fall back config for `Local` IdP, since explicitly creating a IdP for basic authentication reduces the out of the box experience.
2. Update readme to remove local IdP creation and update Google/GitHub IdP-related steps.

## Related issue
- https://github.com/asgardeo/thunder/issues/370


backend/internal/flow/utils/utils.go
```diff
@@ -294,15 +296,36 @@ func getIDP(idpName string) (*idpmodel.IDP, error) {
...
 
 	return idp, nil
 }
 
+// getDefaultLocalIDP retrieves the default Local IDP configuration.
+func getDefaultLocalIDP() (*idpmodel.IDP, error) {
+	runtimeConfig := config.GetThunderRuntime()
+	if runtimeConfig == nil {
+		return nil, fmt.Errorf("runtime configuration not available")
+	}
```
ThaminduDilshan: This check is redundant as `GetThunderRuntime()` function panics when the runtimeConfig is nil

backend/internal/flow/utils/utils.go
```diff
@@ -294,15 +296,36 @@ func getIDP(idpName string) (*idpmodel.IDP, error) {
 	idpSvc := idpservice.GetIDPService()
 	idp, err := idpSvc.GetIdentityProviderByName(idpName)
 	if err != nil {
-		return nil, fmt.Errorf("error while getting IDP with the name %s: %w", idpName, err)
-	}
-	if idp == nil {
+		if idpName == "Local" && errors.Is(err, idpmodel.ErrIDPNotFound) {
```
ThaminduDilshan: Rather than handling this with an error from the IDP service, we can improve the code in one of the following ways.

1. Follow a lazy approach to create local IDP in the database. I.e. when trying to retrieve local IDP from the IDP service and if it doesn't exists, create and return the IDP.
2. Prevent creating local IDP in the database and retrieve it from the configs. So we can improve the IDP service to check if it's the Local IDP and then return the IDP from configs. This way we won't be able to allow configuring local IDP properties organization wide.

@darshanasbg WDYT?

backend/internal/flow/utils/utils.go
```diff
@@ -294,15 +296,36 @@ func getIDP(idpName string) (*idpmodel.IDP, error) {
...
 }
 
+// getDefaultLocalIDP retrieves the default Local IDP configuration.
+func getDefaultLocalIDP() (*idpmodel.IDP, error) {
+	runtimeConfig := config.GetThunderRuntime()
+	if runtimeConfig == nil {
+		return nil, fmt.Errorf("runtime configuration not available")
+	}
+
+	if !runtimeConfig.Config.IDP.LocalFallbackEnabled {
```
ThaminduDilshan: Feels we don't need this fallback config anyway

backend/internal/flow/utils/utils.go
thiva-k: Regarding 1,
Is it ok adding the local IdP to the database directly if it does not exist? 
In this case if a local IdP is created explicitly via API (after a local IdP has been lazily added), there will be multiple Local IdPs in the DB (no unique constraint on name currently) and cause a runtime error when retrieving. We should then allow to only update the pre-created Local IdP in this case and return error for creation requests right?

backend/internal/flow/utils/utils.go
thiva-k: Removed

backend/internal/flow/utils/utils.go
thiva-k: Added if local idp fallback should not be used and error should be thrown if it is not explicitly created. Will remove if this case is not needed.

backend/internal/flow/utils/utils.go
ThaminduDilshan: Then anyway that issue will be there if someone execute the same API call twice. Unique validation should be there in the IDP service. Seems we have missed adding this.
Improvement will be tracked through https://github.com/asgardeo/thunder/issues/386

backend/internal/flow/utils/utils.go
ThaminduDilshan: No need to add unnecessary configs. Let's stick to one approach based on the agreement on my previous comment

backend/internal/flow/utils/utils.go
darshanasbg: Can we remove the need of the "Local" IDP?

As the alternate option, its better to avoid performing modifications in a get request. So i'm aligned to follow the 2nd option. Any needed configs of Local IDP we can pull from the configs.. Rather making this multiple tenant/org compatible, we should solve the need of Local IDP for these kind of requests.

backend/internal/flow/utils/utils.go
thiva-k: Removed

backend/internal/flow/utils/utils.go
ThaminduDilshan: Yes we can remove local IDP. It will just drop the consistency of having an IDP attached to each authenticator (auth executor). But we can special case local authenticators and handle without having the local IDP

thiva-k: Updated current PR to remove Local IDP requirement and use default `Local` id and name in Basic and SMSOTP executors.

darshanasbg (approved):

Issue #324, opened 2025-08-15
Improve database connection handling


ThaminduDilshan:
### Current Limitation

Currently we have written a database client that wraps around standard sql.db. Database connections are obtained in other packages by invoking the `GetDBClient` method on the DB provider. This calls `sql.Open()` which returns a database connection [1]. After using it we're calling `Close()` method on db client to close the connection.

However as per the sql package, it maintains a connection pool underline and doesn't require closing the connections.

Quoted from `sql.go`;
```
// The returned [DB] is safe for concurrent use by multiple goroutines
// and maintains its own pool of idle connections. Thus, the OpenDB
// function should be called just once. It is rarely necessary to
// close a [DB].
```

[1] https://github.com/asgardeo/thunder/blob/6291abec37fcd9f9d153689fbae4b2f93a91aada/backend/internal/system/database/provider/dbprovider.go#L69
[2] https://github.com/asgardeo/thunder/blob/6291abec37fcd9f9d153689fbae4b2f93a91aada/backend/internal/system/database/client/dbclient.go#L136

### Suggested Improvement

Refactor database client usages to prevent calling close method. With this need to refactor dbprovider to return a singleton instance. This will optimize the db connection usages underline with a connection pool.

PR #395, opened 2025-09-17
Centralizing DB Client creation and closing & Introduce database connection pooling

thiva-k:
## Purpose

1. Improve to use a singleton instance of DBProvider that returns identity and runtime DBClients with configured connection pooling.
2. Stop closing DBClients at each store-level and close only during shutting down of server.

## Related issue
- https://github.com/asgardeo/thunder/issues/324

### DIFF
--- a/backend/cmd/server/main.go
+++ b/backend/cmd/server/main.go
@@ -25,13 +25,16 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"os/signal"
 	"path"
+	"syscall"
 	"time"
 
 	"github.com/asgardeo/thunder/internal/cert"
 	"github.com/asgardeo/thunder/internal/flow"
 	"github.com/asgardeo/thunder/internal/system/cache"
 	"github.com/asgardeo/thunder/internal/system/config"
+	"github.com/asgardeo/thunder/internal/system/database/provider"
 	"github.com/asgardeo/thunder/internal/system/jwt"
 	"github.com/asgardeo/thunder/internal/system/log"
 	"github.com/asgardeo/thunder/internal/system/managers"
@@ -53,6 +56,8 @@ func main() {
 	}
 
 	initFlowService(logger)
+
+	closeDBConnections(logger)
 
 	if cfg.Server.HTTPOnly {
 		logger.Info("TLS is not enabled, starting server without TLS")
@@ -193,4 +198,19 @@ func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux
 	}
 
 	return server, serverAddr
+}
+
+// closeDBConnections sets up signal handling for graceful shutdown
+func closeDBConnections(logger *log.Logger) {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-c
+		dbProvider := provider.GetDBProvider()
+		if err := dbProvider.Close(); err != nil {
+			logger.Error("Error closing database connections", log.Error(err))
+		}
+		logger.Info("Database connections closed successfully")
+	}()
 }
--- a/backend/cmd/server/main.go
+++ b/backend/cmd/server/main.go
@@ -25,16 +25,13 @@ import (
 	"fmt"
 	"net/http"
 	"os"
-	"os/signal"
 	"path"
-	"syscall"
 	"time"
 
 	"github.com/asgardeo/thunder/internal/cert"
 	"github.com/asgardeo/thunder/internal/flow"
 	"github.com/asgardeo/thunder/internal/system/cache"
 	"github.com/asgardeo/thunder/internal/system/config"
-	"github.com/asgardeo/thunder/internal/system/database/provider"
 	"github.com/asgardeo/thunder/internal/system/jwt"
 	"github.com/asgardeo/thunder/internal/system/log"
 	"github.com/asgardeo/thunder/internal/system/managers"
@@ -56,8 +53,6 @@ func main() {
 	}
 
 	initFlowService(logger)
-
-	closeDBConnections(logger)
 
 	if cfg.Server.HTTPOnly {
 		logger.Info("TLS is not enabled, starting server without TLS")
@@ -198,19 +193,4 @@ func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux
 	}
 
 	return server, serverAddr
-}
-
-// closeDBConnections sets up signal handling for graceful shutdown
-func closeDBConnections(logger *log.Logger) {
-	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
-
-	go func() {
-		<-c
-		dbProvider := provider.GetDBProvider()
-		if err := dbProvider.Close(); err != nil {
-			logger.Error("Error closing database connections", log.Error(err))
-		}
-		logger.Info("Database connections closed successfully")
-	}()
 }
--- a/backend/internal/application/store/store.go
+++ b/backend/internal/application/store/store.go
@@ -27,7 +27,6 @@ import (
 	"github.com/asgardeo/thunder/internal/application/constants"
 	"github.com/asgardeo/thunder/internal/application/model"
 	oauth2const "github.com/asgardeo/thunder/internal/oauth/oauth2/constants"
-	"github.com/asgardeo/thunder/internal/system/database/client"
 	dbmodel "github.com/asgardeo/thunder/internal/system/database/model"
 	"github.com/asgardeo/thunder/internal/system/database/provider"
 	"github.com/asgardeo/thunder/internal/system/log"
@@ -101,18 +100,10 @@ func (st *ApplicationStore) CreateApplication(app model.ApplicationProcessedDTO)
 
 // GetTotalApplicationCount retrieves the total count of applications from the database.
 func (st *ApplicationStore) GetTotalApplicationCount() (int, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		return 0, fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func(dbc client.DBClientInterface) {
-		err := dbc.Close()
-		if err != nil {
-			logger.Error("Failed to close database client", log.Error(err))
-		}
-	}(dbClient)
 
 	results, err := dbClient.Query(QueryGetApplicationCount)
 	if err != nil {
@@ -135,17 +126,11 @@ func (st *ApplicationStore) GetTotalApplicationCount() (int, error) {
 func (st *ApplicationStore) GetApplicationList() ([]model.BasicApplicationDTO, error) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationPersistence"))
 
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		logger.Error("Failed to get database client", log.Error(err))
 		return nil, fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func(dbc client.DBClientInterface) {
-		err := dbc.Close()
-		if err != nil {
-			logger.Error("Failed to close database client", log.Error(err))
-		}
-	}(dbClient)
 
 	results, err := dbClient.Query(QueryGetApplicationList)
 	if err != nil {
@@ -171,17 +156,11 @@ func (st *ApplicationStore) GetApplicationList() ([]model.BasicApplicationDTO, e
 func (st *ApplicationStore) GetOAuthApplication(clientID string) (*model.OAuthAppConfigProcessedDTO, error) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
 
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		logger.Error("Failed to get database client", log.Error(err))
 		return nil, fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func() {
-		if closeErr := dbClient.Close(); closeErr != nil {
-			logger.Error("Failed to close database client", log.Error(closeErr))
-			err = fmt.Errorf("failed to close database client: %w", closeErr)
-		}
-	}()
 
 	results, err := dbClient.Query(QueryGetOAuthApplicationByClientID, clientID)
 	if err != nil {
@@ -275,17 +254,11 @@ func (st *ApplicationStore) getApplicationByQuery(query dbmodel.DBQuery, param s
 	*model.ApplicationProcessedDTO, error) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
 
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		logger.Error("Failed to get database client", log.Error(err))
 		return nil, fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func() {
-		if closeErr := dbClient.Close(); closeErr != nil {
-			logger.Error("Failed to close database client", log.Error(closeErr))
-			err = fmt.Errorf("failed to close database client: %w", closeErr)
-		}
-	}()
 
 	results, err := dbClient.Query(query, param)
 	if err != nil {
@@ -347,17 +320,11 @@ func (st *ApplicationStore) UpdateApplication(existingApp, updatedApp *model.App
 func (st *ApplicationStore) DeleteApplication(id string) error {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
 
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		logger.Error("Failed to get database client", log.Error(err))
 		return fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func() {
-		if closeErr := dbClient.Close(); closeErr != nil {
-			logger.Error("Failed to close database client", log.Error(closeErr))
-			err = fmt.Errorf("failed to close database client: %w", closeErr)
-		}
-	}()
 
 	_, err = dbClient.Execute(QueryDeleteApplicationByAppID, id)
 	if err != nil {
@@ -675,16 +642,10 @@ func buildApplicationFromResultRow(row map[string]interface{}) (model.Applicatio
 func executeTransaction(queries []func(tx dbmodel.TxInterface) error) error {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
 
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
 	if err != nil {
 		return fmt.Errorf("failed to get database client: %w", err)
 	}
-	defer func() {
-		if closeErr := dbClient.Close(); closeErr != nil {
-			logger.Error("Failed to close database client", log.Error(closeErr))
-			err = fmt.Errorf("failed to close database client: %w", closeErr)
-		}
-	}()
 
 	tx, err := dbClient.BeginTx()
 	if err != nil {
--- a/backend/internal/group/store/store.go
+++ b/backend/internal/group/store/store.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/asgardeo/thunder/internal/group/constants"
 	"github.com/asgardeo/thunder/internal/group/model"
-	"github.com/asgardeo/thunder/internal/system/database/client"
 	dbmodel "github.com/asgardeo/thunder/internal/system/database/model"
 	"github.com/asgardeo/thunder/internal/system/database/provider"
 	"github.com/asgardeo/thunder/internal/system/log"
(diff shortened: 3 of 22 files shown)


backend/internal/system/database/provider/dbprovider.go
```diff
@@ -108,3 +167,29 @@ func getDBConfig(dataSource config.DataSource) dbConfig {
...
+			errs = append(errs, fmt.Errorf("failed to close runtime client: %w", err))
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("errors closing database clients: %v", errs)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
thiva-k: Added

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,110 @@ type dbConfig struct {
...
+		return
+	}
+	d.identityClient = identityClient
+
+	runtimeClient, err := d.createClient(config.Runtime, "runtime")
+	if err != nil {
+		d.runtimeClient = nil
+		return
+	}
+	d.runtimeClient = runtimeClient
```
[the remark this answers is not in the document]
thiva-k: Restructured to use if-else without returning. We are not handling error logging here.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -108,3 +167,29 @@ func getDBConfig(dataSource config.DataSource) dbConfig {
 
 	return dbConfig
 }
+
+// Close gracefully closes all database connections.
+func (d *DBProvider) Close() error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
```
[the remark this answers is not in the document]
thiva-k: Added read locks to GetDBClient()

backend/cmd/server/main.go
```diff
@@ -194,3 +199,19 @@ func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux
...
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-c
+		dbProvider := provider.GetDBProvider()
+		if err := dbProvider.Close(); err != nil {
+			logger.Error("Error closing database connections", log.Error(err))
+		}
+		logger.Info("Database connections closed successfully")
+		os.Exit(0)
```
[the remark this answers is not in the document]
thiva-k: Removed os.Exit(0) from closeDBConnections

backend/cmd/server/main.go
```diff
@@ -194,3 +199,18 @@ func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux
...
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+
+	go func() {
+		<-c
+		dbProvider := provider.GetDBProvider()
+		if err := dbProvider.Close(); err != nil {
+			logger.Error("Error closing database connections", log.Error(err))
+		}
+		logger.Info("Database connections closed successfully")
```
ThaminduDilshan:
```suggestion
		logger.Debug("Database connections closed successfully")
```
@darshanasbg debug log would be sufficient for this right?

backend/cmd/server/main.go
```diff
@@ -54,6 +57,8 @@ func main() {
 
 	initFlowService(logger)
```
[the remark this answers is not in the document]
thiva-k: closeDBConnections listens for exit signals in a separate go routine as expected. Cannot call closeDBConnections after server initialization since it is a blocking function.

backend/internal/application/store/store.go
```diff
@@ -101,18 +100,10 @@ func (st *ApplicationStore) CreateApplication(app model.ApplicationProcessedDTO)
 
 // GetTotalApplicationCount retrieves the total count of applications from the database.
 func (st *ApplicationStore) GetTotalApplicationCount() (int, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationStore"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
ThaminduDilshan, line 103: Shall we define `DBProvider` inside `ApplicationStore` struct to allow dependency injection (similar to `cert/store`)?

backend/internal/group/store/store.go
```diff
@@ -35,17 +35,10 @@ const loggerComponentName = "GroupStore"
 
 // GetGroupListCount retrieves the total count of root groups.
 func GetGroupListCount() (int, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
ThaminduDilshan, line 37: Same as previous comment: We should define `DBProvider` inside `Store` struct to allow dependency injection (similar to `cert/store`).
However since the code base doesn't follow the struct and receiver method pattern, let's do this in a separate PR

backend/internal/idp/store/store.go
```diff
@@ -52,17 +52,10 @@ func NewIDPStore() IDPStoreInterface {
 
 // CreateIdentityProvider handles the IdP creation in the database.
 func (s *IDPStore) CreateIdentityProvider(idp model.IdpDTO) error {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "IdPPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
ThaminduDilshan, line 55: Same as previous comment: Shall we define `DBProvider` inside `IDPStore` struct to allow dependency injection (similar to `cert/store`)?

Shall we fix all the places?

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,110 @@ type dbConfig struct {
...
 }
 
 // GetDBClient returns a database client based on the provided database name.
 func (d *DBProvider) GetDBClient(dbName string) (client.DBClientInterface, error) {
-	// Create the database connection string based on the configured database type.
-	config := config.GetThunderRuntime().Config.Database
-	var dbConfig dbConfig
+	d.once.Do(func() {
+		d.initializeClients()
+	})
```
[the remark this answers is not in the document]
thiva-k: This is to ensure that DB clients are initialized once in the DBProvider. This initialization should happen lazily when GetDBClient() is called, not globally when the DBProvider itself is being initialized for better separation of concerns.

backend/cmd/server/main.go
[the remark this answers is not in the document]
thiva-k: Repeated comment. Addressed above.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,113 @@ type dbConfig struct {
...
+	} else {
+		d.identityClient = identityClient
+	}
+
+	runtimeClient, err := d.createClient(config.Runtime, "runtime")
+	if err != nil {
+		d.runtimeClient = nil
+	} else {
+		d.runtimeClient = runtimeClient
+	}
```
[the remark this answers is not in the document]
thiva-k: Errors handled in parent GetDBClient() function

backend/internal/userschema/store/store.go
```diff
@@ -31,17 +31,10 @@ import (
 
 // GetUserSchemaListCount retrieves the total count of user schemas.
 func GetUserSchemaListCount() (int, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "UserSchemaPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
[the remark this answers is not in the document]
thiva-k: No need for logging after removal of dbClient.Close(). We do not need unnecessary logger imports.

backend/internal/userschema/store/store.go
```diff
@@ -94,17 +82,10 @@ func GetUserSchemaList(limit, offset int) ([]model.UserSchemaListItem, error) {
 
 // CreateUserSchema creates a new user schema.
 func CreateUserSchema(userSchema model.UserSchema) error {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "UserSchemaPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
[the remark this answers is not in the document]
thiva-k: No need for logging after removal of dbClient.Close(). We do not need unnecessary logger imports.

backend/internal/userschema/store/store.go
```diff
@@ -116,17 +97,10 @@ func CreateUserSchema(userSchema model.UserSchema) error {
 
 // GetUserSchemaByID retrieves a user schema by its ID.
 func GetUserSchemaByID(schemaID string) (model.UserSchema, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "UserSchemaPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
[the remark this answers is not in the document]
thiva-k: No need for logging after removal of dbClient.Close(). We do not need unnecessary logger imports.

backend/internal/userschema/store/store.go
```diff
@@ -168,17 +135,10 @@ func GetUserSchemaByName(name string) (model.UserSchema, error) {
 
 // UpdateUserSchemaByID updates a user schema by its ID.
 func UpdateUserSchemaByID(schemaID string, userSchema model.UserSchema) error {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "UserSchemaPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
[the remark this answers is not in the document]
thiva-k: No need for logging after removal of dbClient.Close(). We do not need unnecessary logger imports.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,113 @@ type dbConfig struct {
...
+// GetDBProvider returns the instance of DBProvider.
+func GetDBProvider() DBProviderInterface {
+	once.Do(func() {
+		instance = &DBProvider{}
+	})
+	return instance
 }
 
 // GetDBClient returns a database client based on the provided database name.
 func (d *DBProvider) GetDBClient(dbName string) (client.DBClientInterface, error) {
```
ThaminduDilshan: Let's add a comment saying we shouldn't/ generally not needed to close the client

backend/internal/userschema/store/store.go
```diff
@@ -142,17 +116,10 @@ func GetUserSchemaByID(schemaID string) (model.UserSchema, error) {
 
 // GetUserSchemaByName retrieves a user schema by its name.
 func GetUserSchemaByName(name string) (model.UserSchema, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "UserSchemaPersistence"))
-
-	dbClient, err := provider.NewDBProvider().GetDBClient("identity")
+	dbClient, err := provider.GetDBProvider().GetDBClient("identity")
```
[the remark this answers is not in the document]
thiva-k: No need for logging after removal of dbClient.Close(). We do not need unnecessary logger imports.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,113 @@ type dbConfig struct {
...
 	}
+}
+
+// initializeClients initializes the database clients.
+func (d *DBProvider) initializeClients() {
+	config := config.GetThunderRuntime().Config.Database
+
+	identityClient, err := d.createClient(config.Identity, "identity")
+	if err != nil {
+		d.identityClient = nil
```
ThaminduDilshan, line 158: shouldn't we throw an error here?

backend/internal/system/database/provider/dbprovider.go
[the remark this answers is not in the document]
ThaminduDilshan: Do we need to have two mutexes here? Will there be an issue if we remove the `mu` and `once` defined inside the struct and initialize the two db clients when initializing the provider?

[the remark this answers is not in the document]
thiva-k: `once` is to lazily initialize the DB Client and to prevent it being initialized together with Provider in `GetDBProvider`. Mutex is to prevent access for DB client when GetDBClient() is called when Close() is in process. But is redundant if we are only closing during server exit. Will remove both for simplicity.

backend/internal/system/database/provider/dbprovider.go
thiva-k: Since we are calling `initializeClients` within `once.Do`, we can either log the error within `once.Do` (similar to https://github.com/asgardeo/thunder/blob/main/backend/internal/system/crypto/cryptoservice.go#L64) or use a variable from outer scope to capture it right ? 
A error will anyways be thrown from  `GetDBClient` if the client instances are nil.

Should we log this error (which will be logged in `GetDBProvider()` if we move initializing the DB Clients logic also into it) ?

backend/internal/system/database/provider/dbprovider.go
thiva-k: Added

backend/internal/group/store/store.go
thiva-k: Will move to this pattern in a follow-up PR.

backend/internal/idp/store/store.go
thiva-k: Will move to this pattern in a follow-up PR.

backend/internal/application/store/store.go
thiva-k: Will move to this pattern in a follow-up PR.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -108,3 +163,28 @@ func getDBConfig(dataSource config.DataSource) dbConfig {
...
 	return dbConfig
 }
+
+// Close gracefully closes all database connections.
+func (d *DBProvider) Close() error {
+	var errs []error
+
+	if d.identityClient != nil {
+		if err := d.identityClient.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to close identity client: %w", err))
```
ThaminduDilshan: We can use `errors.Join` to concat multiple errors.  Shall we use that instead of current approach?

Also saw some other place where you are string concatinating two errors. Shall we fix that as well?

backend/internal/system/database/provider/dbprovider.go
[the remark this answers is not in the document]
ThaminduDilshan: I think we should make it safe against unexpected Close() calls if we're making the Close method public. If someone implementing a new code calls the Close method on either the DB provider or one of the client's, it will give a runtime error during someother DB operation right?

To fix the DB client, we can move all the db codebase to a flat hierarchy and make the client's Close method private

backend/cmd/server/main.go
darshanasbg: +1 to make this a debug log.

backend/internal/system/database/provider/dbprovider.go
darshanasbg: IMO, we should throw the error from `initializeClients()`.. Even CryptoService does that..

How `we log the error` or `where it logs` is outside of the contract of this.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,106 @@ type dbConfig struct {
...
+)
 
-// NewDBProvider creates a new instance of DBProvider.
-func NewDBProvider() DBProviderInterface {
-	return &DBProvider{}
+// GetDBProvider returns the instance of DBProvider.
+func GetDBProvider() DBProviderInterface {
+	once.Do(func() {
+		instance = &DBProvider{}
+		instance.initializeClients()
```
darshanasbg: We need to cater the scenario where, db service is not up and running when the thunder is starting.. but it available later so that thunder can continue.

Having initializeClients() here, make it impossible to initialize clients later.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -42,53 +44,106 @@ type dbConfig struct {
...
+	})
+	return instance
 }
 
 // GetDBClient returns a database client based on the provided database name.
+// Do not close the returned client manually since it manages its own connection pool.
```
darshanasbg: IMO, we shouldn't have to have this.. Rather make `DBClientInterface.Close()` private.

backend/cmd/server/main.go
```diff
@@ -194,3 +199,18 @@ func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux
 
 	return server, serverAddr
 }
+
+// closeDBConnections sets up signal handling for graceful shutdown
+func closeDBConnections(logger *log.Logger) {
```
darshanasbg: Shall we,
1. Move `closeDBConnections()` impl to `dbprovider` so we can make the `dbprovider.Close()` private. This will further access restriction trying to do with the [1].
2. Call `closeDBConnections()` from the `once.Do` block of the `GetDBProvider()`. So it will make sure, to close the clients, right after initializing them.


[1] https://github.com/asgardeo/thunder/pull/395/files#r2362703315

backend/cmd/server/main.go
```diff
@@ -54,6 +57,8 @@ func main() {
 
 	initFlowService(logger)
 
+	closeDBConnections(logger)
```
darshanasbg: We may want to `dbprovider.GetDBProvider()` here to initialize client in the server startup, rather lazy loading it.

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -62,7 +66,24 @@ var (
...
+
+		go func() {
+			<-c
+			logger := log.GetLogger()
+			if err := instance.close(); err != nil {
+				logger.Debug("Error closing database connections", log.Error(err))
+			} else {
+				logger.Info("Database connections closed successfully")
+			}
+		}()
```
darshanasbg: Its better to extract this to a function.

backend/internal/system/database/provider/dbprovider.go
thiva-k: Updated

backend/cmd/server/main.go
thiva-k: Updated

backend/cmd/server/main.go
thiva-k: Updated

backend/internal/system/database/provider/dbprovider.go
thiva-k: Updated to log within `once`

backend/internal/system/database/provider/dbprovider.go
thiva-k: Updated

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -41,54 +47,174 @@ type dbConfig struct {
...
+)
 
-// NewDBProvider creates a new instance of DBProvider.
-func NewDBProvider() DBProviderInterface {
-	return &DBProvider{}
+// GetDBProvider returns the instance of DBProvider.
+func GetDBProvider() DBProviderInterface {
+	once.Do(func() {
+		instance = &DBProvider{}
+		instance.closeDBConnections()
```
darshanasbg: lets rename this to something like `closeOnInterrupt()`

backend/internal/system/database/provider/dbprovider.go
ThaminduDilshan: We can slightly modify the comment saying it's not required to close this connection (based on our approach to fix DB.close). IMO it would be good to have such a comment as we may be exposing this as a package that otherrs can reuse.

darshanasbg: +1 to indicate `it's not required to close this connection`

backend/internal/system/database/provider/dbprovider.go
thiva-k: Moved the initialization logic to `GetDBClient` with thread-safe lazy initialization

backend/internal/system/database/provider/dbprovider.go
thiva-k: Modified accordingly

backend/internal/system/database/provider/dbprovider.go
thiva-k: Renamed

backend/internal/system/database/provider/dbprovider.go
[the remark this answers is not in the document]
thiva-k: Moved the DBClient logic to `provider` package. Thereby, made `close()` calls from both DBProvider and DBClient private

backend/internal/system/database/provider/dbprovider.go
```diff
@@ -41,54 +47,174 @@ type dbConfig struct {
...
+	go func() {
+		<-c
+		logger := log.GetLogger()
+		if err := d.close(); err != nil {
+			logger.Error("Error closing database connections", log.Error(err))
+		} else {
+			logger.Debug("Database connections closed successfully")
+		}
+	}()
 }
```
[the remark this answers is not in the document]
thiva-k: Only a single Go routine will be spawned here with a single focus on closing DB connections on interrupt received. Having it in other packages would require publicly exposing `close` methods which is not recommended anyways.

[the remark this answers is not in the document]
darshanasbg: Its a valid comment, if we came across to cleanup multiple things.. Due to the above comment, having it in the main will do more harm than good.. So lets proceed in this way, till we get the requirement to cleanup multiple things.

darshanasbg:
@thiva-k Seems there bunch of duplicated code to handle things on the two database types.. We are planning to onboard a third so this duplication will be increased.. Shall we give a prompt to reduce the code duplication and see how it goes?

darshanasbg:
> @thiva-k Seems there bunch of duplicated code in `dbprovider.go` to handle things on the two database types.. We are planning to onboard a third so this duplication will be increased.. Shall we give a prompt to reduce the code duplication and see how it goes?

Since this is only a minor improvement, lets do this in a separate PR.. Since this PR touches component, proceeding with merging this, to avoid conflicts and blocking other improvements.

darshanasbg (approved):

Issue #415, opened 2025-09-24
Encrypt secret properties of identity providers and notification senders


thiva-k:
### Current Limitation

Currently, Thunder stores sensitive configuration properties like client secrets, API keys, and authentication tokens for identity providers and notification senders in plain text within the database. These properties are marked with `IsSecret: true` in the database schema but are not encrypted, posing a significant security risk as anyone with database access can view these sensitive credentials.

### Suggested Improvement

Implement encryption for secret properties using Thunder's existing AES-GCM encryption infrastructure. Secret properties should be automatically encrypted when stored in the database and decrypted when accessed by the application. This enhancement should leverage the existing CryptoService and introduce a centralized Property model with lazy decryption capabilities. 

### Version

v0.8.0

PR #420, opened 2025-09-25
Add encryption to secret properties in identity providers and notification senders

thiva-k:
## Purpose

Integrate existing encryption capabilities to encrypt sensitive configuration properties of identity providers and notification senders.
Centralize `Property` model with functions to encrypt and decrypt, that can be reused across packages.

## Related issue

- https://github.com/asgardeo/thunder/issues/415


backend/internal/system/cmodels/property.go
```diff
@@ -0,0 +1,64 @@
...
+package cmodels
+
+import (
+	"fmt"
+
+	"github.com/asgardeo/thunder/internal/system/crypto"
+)
+
+// Property represents a generic property with Name, Value, and IsSecret fields.
+type Property struct {
```
ThaminduDilshan, line 29: With this we are always decrypting the secrets when retrieving, but haven't enforced the encryption. If someone didn't explicitly call the encrypt method, code will try to decrypt a unencrypted value.
IMO we have two options.

1. Expose property creation through a util method (i.e. `getProperty`, `createProperty`) and enforce the secret encryption. 
2. Define a private field to indicate whether the value is encrypted or not. This way we can only decrypt the encrypted values

backend/internal/notification/common/model.go
```diff
@@ -35,11 +37,8 @@ type OTP struct {
 }
 
 // SenderProperty represents a key-value property for a notification sender.
-type SenderProperty struct {
-	Name     string `json:"name"`
-	Value    string `json:"value"`
-	IsSecret bool   `json:"is_secret"`
-}
+// Deprecated: Use cmodels.Property instead
+type SenderProperty = cmodels.Property
```
ThaminduDilshan: Lets remove this. No need to preserve backward compatibility at this stage

backend/internal/idp/model/model.go
```diff
@@ -35,25 +37,22 @@ type BasicIdpDTO struct {
 }
 
 // IdpProperty represents a property of an identity provider.
-type IdpProperty struct {
-	Name     string `json:"name"`
-	Value    string `json:"value"`
-	IsSecret bool   `json:"is_secret"`
-}
+// Deprecated: Use cmodels.Property instead
+type IdpProperty = cmodels.Property
```
ThaminduDilshan: Let's remove this. No need to preserve backward compatibility at this stage

backend/internal/notification/common/model.go
thiva-k: Removed

backend/internal/idp/model/model.go
thiva-k: Removed

backend/internal/system/cmodels/property.go
thiva-k: Going with option 2, but with public field for `IsEncrypted` since we can't store private fields in DB. And option 1 would also require setting private fields which has the same concern.

ThaminduDilshan: Shall we keep `isEncrypted` field as private and define two receiver methods to check whether it's encrypted and set encrypted field?

thiva-k: Updated

ThaminduDilshan (approved):

PR #447, opened 2025-10-02
Improve secret properties handling

thiva-k:
## Purpose
This PR addresses the issue where `Property` struct exposes the `Value` field, allowing direct access to encrypted values and bypassing `GetValue()`. It encapsulates fields, adds constructors and getters, introduces a `PropertyDTO` for API communication, and moves encryption into the constructor. These changes enforce controlled access, preserve API compatibility, and improve error handling for encryption/decryption.

## Related Issue

- #448

### DIFF
--- a/backend/internal/system/cmodels/property.go
+++ b/backend/internal/system/cmodels/property.go
@@ -27,10 +27,56 @@ import (
 
 // Property represents a generic property with Name, Value, IsSecret, and isEncrypted fields.
 type Property struct {
-	Name        string `json:"name"`
-	Value       string `json:"value"`
-	IsSecret    bool   `json:"is_secret"`
+	name        string
+	value       string
+	isSecret    bool
 	isEncrypted bool
+}
+
+// PropertyDTO represents a property for API communication.
+type PropertyDTO struct {
+	Name     string `json:"name"`
+	Value    string `json:"value"`
+	IsSecret bool   `json:"is_secret"`
+}
+
+// NewRawProperty creates a new Property instance with the given parameters.
+func NewRawProperty(name, value string, isSecret, isEncrypted bool) *Property {
+	return &Property{
+		name:        name,
+		value:       value,
+		isSecret:    isSecret,
+		isEncrypted: isEncrypted,
+	}
+}
+
+// NewProperty creates a new Property instance with the given parameters.
+// If isSecret is true, the value will be automatically encrypted.
+func NewProperty(name, value string, isSecret bool) (*Property, error) {
+	property := &Property{
+		name:        name,
+		value:       value,
+		isSecret:    isSecret,
+		isEncrypted: false,
+	}
+
+	if isSecret && value != "" {
+		if err := property.Encrypt(); err != nil {
+			return nil, fmt.Errorf("failed to encrypt property %s: %w", name, err)
+		}
+	}
+
+	return property, nil
+}
+
+// GetName returns the name of the property
+func (p *Property) GetName() string {
+	return p.name
+}
+
+// IsSecret returns whether the property is a secret
+func (p *Property) IsSecret() bool {
+	return p.isSecret
 }
 
 // IsEncrypted returns whether the property value is encrypted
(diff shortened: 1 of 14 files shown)


backend/internal/system/cmodels/property.go
```diff
@@ -23,16 +23,63 @@ import (
...
+	property := &Property{
+		name:        name,
+		value:       value,
+		isSecret:    isSecret,
+		isEncrypted: false,
+	}
+
+	if isSecret && value != "" {
+		if err := property.Encrypt(); err != nil {
+			log.GetLogger().Error("failed to encrypt property", log.String("property", name), log.Error(err))
```
ThaminduDilshan: shall we return encrypt error here?

backend/internal/system/cmodels/property.go
```diff
@@ -23,16 +23,63 @@ import (
...
+		name:        name,
+		value:       value,
+		isSecret:    isSecret,
+		isEncrypted: isEncrypted,
+	}
+}
+
+// NewEncryptedProperty creates a new Property instance with the given parameters.
+// If isSecret is true, the value will be automatically encrypted.
+func NewEncryptedProperty(name, value string, isSecret bool) *Property {
```
ThaminduDilshan:
```suggestion
func NewProperty(name, value string, isSecret bool) *Property {
```

backend/internal/system/cmodels/property.go
```diff
@@ -23,16 +23,63 @@ import (
...
 
+// PropertyDTO represents a property for API communication.
+type PropertyDTO struct {
+	Name     string `json:"name"`
+	Value    string `json:"value"`
+	IsSecret bool   `json:"is_secret"`
+}
+
+// NewProperty creates a new Property instance with the given parameters.
+func NewProperty(name, value string, isSecret bool, isEncrypted bool) *Property {
```
ThaminduDilshan: Shall we think on a different name? May be `NewRawProperty`

backend/internal/system/cmodels/property.go
```diff
@@ -23,16 +23,63 @@ import (
...
+	return property
+}
+
+// GetName returns the name of the property
+func (p *Property) GetName() string {
+	return p.name
+}
+
+// GetIsSecret returns whether the property is a secret
+func (p *Property) GetIsSecret() bool {
```
ThaminduDilshan:
```suggestion
func (p *Property) IsSecret() bool {
```

backend/internal/system/cmodels/property.go
```diff
@@ -43,34 +90,63 @@ func (p *Property) SetEncrypted(encrypted bool) {
...
+// ToProperty converts PropertyDTO to Property.
+func (dto *PropertyDTO) ToProperty() *Property {
+	return NewEncryptedProperty(dto.Name, dto.Value, dto.IsSecret)
+}
+
+// ToPropertyDTO converts Property to PropertyDTO .
+func (p *Property) ToPropertyDTO() *PropertyDTO {
+	value, err := p.GetValue()
+	if err != nil {
+		log.GetLogger().Error("failed to get value for property", log.String("property", p.GetName()), log.Error(err))
```
ThaminduDilshan: We should return the error

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
```diff
@@ -23,16 +23,63 @@ import (
...
 
+// PropertyDTO represents a property for API communication.
+type PropertyDTO struct {
+	Name     string `json:"name"`
+	Value    string `json:"value"`
+	IsSecret bool   `json:"is_secret"`
+}
+
+// NewRawProperty creates a new Property instance with the given parameters.
+func NewRawProperty(name, value string, isSecret bool, isEncrypted bool) *Property {
```
ThaminduDilshan:
```suggestion
func NewRawProperty(name, value string, isSecret, isEncrypted bool) *Property {
```

backend/internal/system/cmodels/property.go
```diff
@@ -43,34 +90,64 @@ func (p *Property) SetEncrypted(encrypted bool) {
...
+// Returns an error if encryption fails for secret properties.
+func (dto *PropertyDTO) ToProperty() (*Property, error) {
+	return NewProperty(dto.Name, dto.Value, dto.IsSecret)
+}
+
+// ToPropertyDTO converts Property to PropertyDTO .
+func (p *Property) ToPropertyDTO() *PropertyDTO {
+	value, err := p.GetValue()
+	if err != nil {
+		log.GetLogger().Error("failed to get value for property", log.String("property", p.GetName()), log.Error(err))
```
ThaminduDilshan: Let's return the error here also. Ideally if there's an error while constructing any of the property structs, it should be returned appropriately (either from svc layer or API).

This specific implementation will lead to returning a empty value leading to confusions.

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

ThaminduDilshan (approved):

PR #466, opened 2025-10-09
Refactor encrypted property handling

thiva-k:
## Purpose
This PR refactors property storage from separate property tables to JSON columns in the main IDP and notification sender tables, simplifying the database schema. The `Property` struct's `isEncrypted` field is removed and property serialization logic is centralized in `cmodels/property.go` with reusable functions.

## Related issue
- https://github.com/asgardeo/thunder/issues/469

### DIFF
--- a/backend/internal/idp/store.go
+++ b/backend/internal/idp/store.go
@@ -56,12 +56,6 @@ func (s *idpStore) CreateIdentityProvider(idp IDPDTO) error {
 		return fmt.Errorf("failed to get database client: %w", err)
 	}
 
-	for _, property := range idp.Properties {
-		if property.GetName() == "" {
-			return fmt.Errorf("property name cannot be empty")
-		}
-	}
-
 	var propertiesJSON string
 	if len(idp.Properties) > 0 {
 		propertiesJSON, err = cmodels.SerializePropertiesToJSONArray(idp.Properties)
@@ -162,12 +156,6 @@ func (s *idpStore) UpdateIdentityProvider(idp *IDPDTO) error {
 	dbClient, err := s.dbProvider.GetDBClient("identity")
 	if err != nil {
 		return fmt.Errorf("failed to get database client: %w", err)
-	}
-
-	for _, property := range idp.Properties {
-		if property.GetName() == "" {
-			return fmt.Errorf("property name cannot be empty")
-		}
 	}
 
 	var propertiesJSON string
--- a/backend/internal/system/cmodels/property.go
+++ b/backend/internal/system/cmodels/property.go
@@ -22,7 +22,6 @@ package cmodels
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/asgardeo/thunder/internal/system/crypto"
 )
@@ -39,15 +38,6 @@ type PropertyDTO struct {
 	Name     string `json:"name"`
 	Value    string `json:"value"`
 	IsSecret bool   `json:"is_secret"`
-}
-
-// NewRawProperty creates a new Property instance with the given parameters.
-func NewRawProperty(name, value string, isSecret bool) *Property {
-	return &Property{
-		name:     name,
-		value:    value,
-		isSecret: isSecret,
-	}
 }
 
 // NewProperty creates a new Property instance with the given parameters.
@@ -109,38 +99,28 @@ func (p *Property) Encrypt() error {
 	return nil
 }
 
-// toJSONString returns the property as a JSON string
-func (p *Property) toJSONString() (string, error) {
-	propertyData := map[string]interface{}{
-		"name":      p.GetName(),
-		"value":     p.value,
-		"is_secret": p.IsSecret(),
-	}
-
-	jsonBytes, err := json.Marshal(propertyData)
-	if err != nil {
-		return "", fmt.Errorf("failed to marshal property to JSON: %w", err)
-	}
-
-	return string(jsonBytes), nil
-}
-
 // SerializePropertiesToJSONArray serializes an array of properties to a JSON array string
 func SerializePropertiesToJSONArray(properties []Property) (string, error) {
 	if len(properties) == 0 {
 		return "", nil
 	}
 
-	propertiesArray := make([]string, 0, len(properties))
+	propertyDTOs := make([]PropertyDTO, 0, len(properties))
 	for _, property := range properties {
-		propertyJSON, err := property.toJSONString()
-		if err != nil {
-			return "", fmt.Errorf("failed to serialize property %s to JSON: %w", property.GetName(), err)
+		propertyDTO := PropertyDTO{
+			Name:     property.GetName(),
+			Value:    property.value,
+			IsSecret: property.IsSecret(),
 		}
-		propertiesArray = append(propertiesArray, propertyJSON)
+		propertyDTOs = append(propertyDTOs, propertyDTO)
 	}
 
-	return "[" + strings.Join(propertiesArray, ",") + "]", nil
+	jsonBytes, err := json.Marshal(propertyDTOs)
+	if err != nil {
+		return "", fmt.Errorf("failed to serialize properties to JSON: %w", err)
+	}
+
+	return string(jsonBytes), nil
 }
 
 // DeserializePropertiesFromJSON deserializes an array of properties from JSON string
@@ -149,30 +129,19 @@ func DeserializePropertiesFromJSON(propertiesJSON string) ([]Property, error) {
 		return []Property{}, nil
 	}
 
-	var propertiesArray []map[string]interface{}
-	if err := json.Unmarshal([]byte(propertiesJSON), &propertiesArray); err != nil {
+	var propertyDTOs []PropertyDTO
+	if err := json.Unmarshal([]byte(propertiesJSON), &propertyDTOs); err != nil {
 		return nil, fmt.Errorf("failed to unmarshal properties JSON: %w", err)
 	}
 
-	properties := make([]Property, 0, len(propertiesArray))
-	for _, propertyData := range propertiesArray {
-		name, ok := propertyData["name"].(string)
-		if !ok {
-			return nil, fmt.Errorf("failed to parse property name as string")
+	properties := make([]Property, 0, len(propertyDTOs))
+	for _, propertyDTO := range propertyDTOs {
+		property := Property{
+			name:     propertyDTO.Name,
+			value:    propertyDTO.Value,
+			isSecret: propertyDTO.IsSecret,
 		}
-
-		value, ok := propertyData["value"].(string)
-		if !ok {
-			return nil, fmt.Errorf("failed to parse property value as string")
-		}
-
-		isSecret, ok := propertyData["is_secret"].(bool)
-		if !ok {
-			return nil, fmt.Errorf("failed to parse property is_secret as bool")
-		}
-
-		property := NewRawProperty(name, value, isSecret)
-		properties = append(properties, *property)
+		properties = append(properties, property)
 	}
 
 	return properties, nil


backend/internal/idp/store.go
```diff
@@ -64,42 +62,30 @@ func (s *idpStore) CreateIdentityProvider(idp IDPDTO) error {
...
+			if property.GetName() == "" {
 				return fmt.Errorf("property name cannot be empty")
 			}
 		}
 
-		propertyInsertQuery := queryInsertIDPProperties
-		propertyInsertQuery.Query = fmt.Sprintf(propertyInsertQuery.Query, strings.Join(queryValues, ", "))
-
-		_, err = tx.Exec(propertyInsertQuery.Query)
+		var err error
```
rajithacharith: Do we need this? 

Suggestion 
propertiesJSON, err := cmodels.SerializePropertiesToJSONArray(idp.Properties)

backend/internal/idp/store.go
```diff
@@ -192,12 +163,14 @@ func (s *idpStore) getIDP(query dbmodel.DBQuery, identifier string) (*IDPDTO, er
...
-	properties, err := s.getIDPProperties(idp.ID)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get idp properties: %w", err)
+	propertiesJSON, ok := row["properties"].(string)
+	if ok && propertiesJSON != "" {
+		properties, err := cmodels.DeserializePropertiesFromJSON(propertiesJSON)
+		if err != nil {
+			return nil, fmt.Errorf("failed to deserialize properties from JSON: %w", err)
+		}
+		idp.Properties = properties
```
rajithacharith: Can we do this part before line 159? 

Then we can return &IDPDTO{
		ID:          basicIDP.ID,
		Name:        basicIDP.Name,
		Description: basicIDP.Description,
		Type:        basicIDP.Type,
		Properties: properties,
	}

backend/internal/system/cmodels/property.go
```diff
@@ -122,10 +106,78 @@ func (p *Property) Encrypt() error {
...
 	p.value = encryptedValue
-	p.SetEncrypted(true)
 	return nil
 }
 
+// ToJSONString returns the property as a JSON string
+func (p *Property) ToJSONString() (string, error) {
+	propertyData := map[string]interface{}{
+		"name":      p.GetName(),
+		"value":     p.value,
```
rajithacharith: Do we need a getter for value? WDYT?

backend/internal/system/cmodels/property.go
```diff
@@ -122,10 +106,78 @@ func (p *Property) Encrypt() error {
 	}
 
 	p.value = encryptedValue
-	p.SetEncrypted(true)
 	return nil
 }
 
+// ToJSONString returns the property as a JSON string
+func (p *Property) ToJSONString() (string, error) {
+	propertyData := map[string]interface{}{
```
rajithacharith: Do we need to create this interface? Can we use the same property object like this 

```
prop := Property{
		Name:     "APIKey",
		Value:    "123456",
		IsSecret: true,
	}

	jsonData, err := json.Marshal(prop)
	if err != nil {
		fmt.Println("Error marshalling JSON:", err)
		return
	}
	```

File: backend/internal/system/cmodels/property.go [hunk 5]
```diff
@@ -122,10 +106,78 @@ func (p *Property) Encrypt() error {
...
+	for _, property := range properties {
+		propertyJSON, err := property.ToJSONString()
+		if err != nil {
+			return "", fmt.Errorf("failed to serialize property %s to JSON: %w", property.GetName(), err)
+		}
+		propertiesArray = append(propertiesArray, propertyJSON)
+	}
+
+	return "[" + strings.Join(propertiesArray, ",") + "]", nil
+}
```
rajithacharith, line 124: Can we just use json.Marshall ?
```suggestion
func SerializePropertiesToJSONArray(properties []Property) (string, error) {
    if len(properties) == 0 {
        return "[]", nil
    }

    jsonBytes, err := json.Marshal(properties)
    if err != nil {
        return "", fmt.Errorf("failed to serialize properties to JSON: %w", err)
    }

    return string(jsonBytes), nil
}
```

backend/internal/system/cmodels/property.go
```diff
@@ -122,10 +106,78 @@ func (p *Property) Encrypt() error {
...
+			return nil, fmt.Errorf("failed to parse property is_secret as bool")
+		}
+
+		property := NewRawProperty(name, value, isSecret)
+		properties = append(properties, *property)
+	}
+
+	return properties, nil
+}
+
```
rajithacharith, line 149:
```suggestion
func DeserializePropertiesFromJSON(propertiesJSON string) ([]Property, error) {
    if propertiesJSON == "" {
        log.Debug("Properties JSON string is empty, returning empty slice")
        return []Property{}, nil
    }

    var properties []Property
    if err := json.Unmarshal([]byte(propertiesJSON), &properties); err != nil {
        log.Error("Failed to unmarshal properties JSON", "error", err, "json_length", len(propertiesJSON))
        return nil, serviceerror.NewBadRequestError("invalid properties JSON format", err)
    }

    log.Debug("Successfully deserialized properties from JSON", "count", len(properties))
    return properties, nil
}

```

backend/internal/idp/store.go
```diff
@@ -64,42 +62,30 @@ func (s *idpStore) CreateIdentityProvider(idp IDPDTO) error {
...
-			retErr := fmt.Errorf("failed to execute query for inserting properties: %w", err)
-			if rollbackErr := tx.Rollback(); rollbackErr != nil {
-				retErr = errors.Join(retErr, fmt.Errorf("failed to rollback transaction: %w", rollbackErr))
-			}
-			return retErr
+			return fmt.Errorf("failed to serialize properties to JSON: %w", err)
 		}
 	}
 
+	_, err = tx.Exec(queryCreateIdentityProvider.Query, idp.ID, idp.Name, idp.Description, idp.Type, propertiesJSON)
```
ThaminduDilshan: Alternatively we can use `db.Query(...)` now as this is just a single db table and row update

backend/internal/notification/store.go
```diff
@@ -166,45 +161,56 @@ func (s *notificationStore) getSender(query dbmodel.DBQuery,
...
-				id,
-				prop.GetName(),
-				prop.GetStorageValue(),
-				sysutils.BoolToNumString(prop.IsSecret()),
-				sysutils.BoolToNumString(prop.IsEncrypted()),
-			)
-			return err
-		})
+
+	tx, err := dbClient.BeginTx()
```
ThaminduDilshan: Same as previous comment. We don't need transactions now

backend/internal/system/cmodels/property.go
```diff
@@ -20,17 +20,18 @@
...
 import (
+	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/asgardeo/thunder/internal/system/crypto"
 )
 
-// Property represents a generic property with Name, Value, IsSecret, and isEncrypted fields.
+// Property represents a generic property with Name, Value, and IsSecret fields.
```
ThaminduDilshan:
```suggestion
// Property represents a generic property with name, value, and isSecret fields.
```

backend/internal/idp/store.go
thiva-k: Using `:=` here would throw an error since `propertiesJSON` is out of the if scope.

backend/internal/system/cmodels/property.go
thiva-k: `GetValue()` returns decrypted value if it is encrypted to be used in service layers, but here we need to store encrypted values as it is.

backend/internal/system/cmodels/property.go
thiva-k: `Property` has private fields to prevent directly accessing encrypted values, so cannot be marshaled using json.Marshal

backend/internal/system/cmodels/property.go
thiva-k: `Property` has private fields to prevent directly accessing encrypted values, so cannot be marshaled using json.Marshal

backend/internal/system/cmodels/property.go
thiva-k: Similar reason as above. Would need to convert Property back to PropertyDTO and vice-versa if we are to use direct marshaling/unmarshaling.

backend/internal/idp/store.go
thiva-k: Updated

backend/internal/idp/store.go
thiva-k: Updated

backend/internal/notification/store.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

backend/internal/idp/store.go
```diff
@@ -59,53 +56,23 @@ func (s *idpStore) CreateIdentityProvider(idp IDPDTO) error {
...
-	if err != nil {
-		return fmt.Errorf("failed to begin transaction: %w", err)
-	}
-
-	_, err = tx.Exec(queryCreateIdentityProvider.Query, idp.ID, idp.Name, idp.Description, idp.Type)
-	if err != nil {
-		retErr := fmt.Errorf("failed to execute query: %w", err)
-		if rollbackErr := tx.Rollback(); rollbackErr != nil {
-			retErr = errors.Join(retErr, fmt.Errorf("failed to rollback transaction: %w", rollbackErr))
+	for _, property := range idp.Properties {
```
ThaminduDilshan: We don't need this loop as the validation is already done at the service layer [1] and store is private. 

[1] https://github.com/asgardeo/thunder/blob/b7d70e1822c8d8d42ba12345f9ffe32256cb35e5/backend/internal/idp/service.go#L215-L260

backend/internal/idp/store.go
```diff
@@ -209,64 +164,25 @@ func (s *idpStore) UpdateIdentityProvider(idp *IDPDTO) error {
...
-		}
-		return retErr
-	}
-
-	// delete existing properties for the IdP
-	if _, err := tx.Exec(queryDeleteIDPProperties.Query, idp.ID); err != nil {
-		retErr := fmt.Errorf("failed to execute query for deleting existing properties: %w", err)
-		if rollbackErr := tx.Rollback(); rollbackErr != nil {
-			retErr = errors.Join(retErr, fmt.Errorf("failed to rollback transaction: %w", rollbackErr))
+	for _, property := range idp.Properties {
```
ThaminduDilshan: Same as previous comment. We don't need this loop

backend/internal/system/cmodels/property.go
```diff
@@ -79,24 +77,14 @@ func (p *Property) IsSecret() bool {
...
-
-// SetEncrypted sets the encryption state of the property
-func (p *Property) SetEncrypted(encrypted bool) {
-	p.isEncrypted = encrypted
-}
-
-// GetStorageValue returns the value as is
-func (p *Property) GetStorageValue() string {
+// GetRawValue returns the raw value (encrypted for secret properties, plain for non-secret)
+func (p *Property) GetRawValue() string {
```
ThaminduDilshan:
```suggestion
func (p *Property) getRawValue() string {
```
Shall we make this method private?

backend/internal/system/cmodels/property.go
```diff
@@ -41,23 +41,21 @@ type PropertyDTO struct {
 }
 
 // NewRawProperty creates a new Property instance with the given parameters.
-func NewRawProperty(name, value string, isSecret, isEncrypted bool) *Property {
+func NewRawProperty(name, value string, isSecret bool) *Property {
```
ThaminduDilshan: Since now store layer uses util functions from the cmodel package, Do we need this construct anymore?
It seems only usage for this is from the `DeserializePropertiesFromJSON()` function. We should be able to  directly create the struct.

backend/internal/system/cmodels/property.go
thiva-k: It is needed in `SerializePropertiesToJSONArray` since it is not a receiver func. We can remove this if we are to do custom marshaling and remove converting to and from PropertyDTO to marshal

backend/internal/system/cmodels/property.go
thiva-k: Can be removed. Updated.

backend/internal/idp/store.go
thiva-k: Updated

backend/internal/idp/store.go
thiva-k: Updated

backend/internal/system/cmodels/property.go
thiva-k: Updated

PR #505, opened 2025-10-13
Add postgres automation testing workflow.

iff-sal:
## Purpose
This PR addresses the current limitation where the Thunder project's automated test workflows only validate functionality using the default Sqlite database. As Thunder supports both Sqlite and PostgreSQL, there is a need to onboard PostgreSQL compatibility testing to ensure full validation coverage.

Resolves #452

## Goals
Introduce a new GitHub workflow to automate running tests against a PostgreSQL database.
Ensure tests can be triggered on demand, daily, and during release events.

## Approach
The solution is implemented by adding a dedicated GitHub Actions workflow named "🐘 PostgreSQL Compatibility Tests". This workflow:
- Uses the latest PostgreSQL Docker image as a service for the test environment.
- Generates a PostgreSQL-specific test configuration dynamically for integration tests.
- follow the same pattern of sqlite testings.
- Is triggered manually (workflow_dispatch), on a daily schedule, and on new release creation.

This approach ensures PostgreSQL is continuously tested alongside Sqlite without interfering with existing workflows

## User stories
N/A

## Release note
Added a new GitHub Actions workflow to run integration tests against PostgreSQL.

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 - Integration tests - now includes postgres database.

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes
 - Ran FindSecurityBugs plugin and verified report? no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes

## Samples
N/A

## Related PRs
N/A

## Migrations (if applicable)
N/A

PR #506, opened 2025-10-13
Enhance Existing Workflow with Matrix Strategy for Database Compatibility (SQLite & PostgreSQL)

### LINKED ISSUE #452: Automate Thunder postgres compatibility
### Current Limitation

Thunder currently supports two databases - Sqlite & Postgres. Currently PR builder, release workflow etc validates functionality by running the tests, but since those workflows using the default db configurations, these functionality validations happens only on Sqlite.

### Suggested Improvement

Onboard capability run tests on Postgres. This workflow needs to be triggered,
- on demand
- daily
- during the release

### Version

_No response_

iff-sal:
## Purpose
Thunder currently supports both SQLite and PostgreSQL, but the CI/CD pipeline only tests against SQLite, which risks allowing PostgreSQL-specific bugs to go unnoticed until production or manual testing. This PR adds PostgreSQL tests to the CI/CD pipeline to catch issues early in development.

Resolves #452 

## Goals
This PR enhances the CI/CD pipeline to ensure database compatibility by testing against both SQLite and PostgreSQL databases before merging code or creating releases. Previously, tests only ran against SQLite, which could lead to PostgreSQL-specific issues being discovered late in the development cycle. This change provides early detection of database-specific bugs and ensures Thunder works correctly with both supported database backends.

## Approach
1. Modified .github/workflows/pr-builder.yml:

Added a matrix strategy to run tests against both sqlite and postgres databases in parallel.
Configured a PostgreSQL service container with health checks for the postgres matrix job.
Added database initialization steps that create and populate PostgreSQL databases using existing schema scripts from backend/dbscripts/ .
Integrated a new configuration script that dynamically generates test configuration based on the database type.

2. Modified .github/workflows/release.yml:

Applied the same matrix testing strategy to the release workflow's build-thunder job.
Tests now run against both databases before any release artifacts are created, ensuring database compatibility is verified before distribution

3. Created backend/tests/scripts/setup-test-config.sh:

New bash script that generates appropriate deployment.yaml configuration files based on the DB_TYPE environment variable
Reuses existing database initialization patterns from backend/scripts/init_script.sh .

## Release note
Enhanced existing CI workflow to run integration tests for both SQLite and PostgreSQL using matrix strategy.

## Automation tests
 - Integration tests
Integration tests now run in both SQLite and PostgreSQL contexts in a single workflow

## Related PRs
related to PR #505
This PR offers an alternative integrated approach.

 
## Learning
Investigated GitHub Actions matrix strategies and service containers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,29 +32,94 @@ env:
   GOFLAGS: "-mod=readonly"
 
 jobs:
-  build-thunder:
-    name: ⚡ Build Thunder
+  build-thunder:  
+    name: ⚡ Build Thunder  
+    runs-on: ubuntu-latest  
+    steps:  
+      - name: 📥 Checkout Code  
+        uses: actions/checkout@v4  
+        with:  
+          fetch-depth: 0  
+
+      - name: ✅ Validate Release Version  
+        run: |  
+          if ! [[ ${{ github.event.inputs.version }} =~ ^v[0-9]+\.[0-9]+(\.[0-9]+)?(-[0-9a-zA-Z.-]+)?(\+[0-9a-zA-Z.-]+)?$ ]]; then  
+            echo "❌ Error: Version '${{ github.event.inputs.version }}' does not follow format vX.Y[.Z][-PRERELEASE][+BUILD]"  
+            echo "❌ Version must start with 'v'"  
+            exit 1  
+          fi  
+          echo "✅ Version '${{ github.event.inputs.version }}' is valid"  
+
+      - name: 🏷️ Update Product Version  
+        run: |  
+          VERSION="${{ github.event.inputs.version }}"  
+          echo "$VERSION" > version.txt  
+
+      - name: ⚙️ Set up Go Environment  
+        uses: ./.github/actions/setup-go  
+
+      - name: 🗄️ Cache Go Modules  
+        uses: actions/cache@v4  
+        id: cache-go-modules  
+        with:  
+          path: |  
+            ~/.cache/go-build  
+            ~/go/pkg/mod  
+          key: ${{ runner.os }}-go-modules-${{ hashFiles('**/go.sum') }}  
+          restore-keys: |  
+            ${{ runner.os }}-go-modules-  
+
+      - name: 📦 Install Dependencies  
+        run: |  
+          cd backend  
+          go mod download  
+          cd ../tests/integration  
+          go mod download  
+
+      - name: 🧹 Clean Previous Builds  
+        run: make clean_all
+
+      - name: 🔨 Build Product with Coverage  
+        run: |  
+          set -e  
+          make build_with_coverage_only OS=$(go env GOOS) ARCH=$(go env GOARCH)
+          
+          # Find the built distribution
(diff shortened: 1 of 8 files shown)


iff-sal:
need some changes to the release workflow, there is a small problem with current release workflow. since we are using matrix, the build-thunder job now runs twice in parallel (once for sqlite and once for postgres), which means all the release-critical steps inside it—like creating the Git tag—also run twice. This creates a race condition that will cause the workflow to fail when the second job tries to create a tag that already exists. We might have to break that job into two separate parts or something, one for the parallel tests and another for the actual release which only needs to run once. need some feedback on this.

tests/integration/testutils/testutils.go
```diff
@@ -253,6 +255,12 @@ func copyDirectory(src, dest string) error {
 func RunInitScript(zipFilePattern string) error {
 	log.Println("Running init script...")
 
+	// Skip database initialization for PostgreSQL
```
ThaminduDilshan, line 262: This along is not enough. You also need to skip deployment.yaml and dbscript replace part inside `ReplaceResources()` function

iff-sal: I added an if condition to skip replacing dbscript. but currently the backend/tests/scripts/setup-test-config.sh i created generate respective deployment.yaml file inside tests/integration/resources/deployment.yaml. which, then will be used by replaceresources tocopy the yaml file to unzip server.

iff-sal:
@ThaminduDilshan  I ran the pr builder workflow to see if the changes i done far is correct. In one run(https://github.com/wasitha-john/thunder/actions/runs/18560057623/job/52906568171) the sqlite test passed while postgres one falied, so i looked in to the log and sumerize what failed in this gist (https://gist.github.com/iff-sal/e5659ba1e58b4d6520442fca8f257f0d) . within each testsuit some tests pass and some fail. is it exist for postgres or am  i missing something? the sqlite passed all the tests as can see from the action log.

Makefile
```diff
@@ -130,8 +140,8 @@ help:
 .PHONY: all prepare clean clean_all build build_samples package_samples run
 .PHONY: docker-build docker-build-latest docker-build-multiarch 
 .PHONY: docker-build-multiarch-latest docker-build-multiarch-push
-.PHONY: test_unit test_integration build_with_coverage test
-.PHONY: help go_install_tool
+.PHONY: test_unit test_integration build_with_coverage build_with_coverage_only test
+.PHONY: lint help go_install_tool golangci-lint
```
ThaminduDilshan: Let's fix this line. There seems to be conflicts

iff-sal: fixed it, removed duplicate lint and golangci-lint .

.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ on:
         type: boolean
         default: false
 
-# Add permissions to allow pushing tags and packages
+#Add permissions to allow pushing tags and packages
```
ThaminduDilshan:
```suggestion
# Add permissions to allow pushing tags and packages
```

ThaminduDilshan (approved):

ThaminduDilshan:
@iff-sal Thanks for your contribution!

Issue #504, opened 2025-10-13
Add ID token support


thiva-k:
### Problem

Thunder currently only generates access tokens and refresh tokens but does not support ID tokens as required by OpenID Connect. This prevents Thunder from being a fully compliant OIDC provider and limits integration with applications that require standardized user identity verification.

### Proposed Solution

Implement ID token generation when the `openid` scope is requested, including standard OIDC scope-to-claim mappings (profile, email, phone, address) and a UserInfo endpoint. Also support configurable claim mappings at the application level.

PR #521, opened 2025-10-15
Introduce initial ID token support

thiva-k:
## Purpose

Add initial ID token implementation with default OIDC scopes and application-based claims filtering.

## Related issue

- https://github.com/asgardeo/thunder/issues/504

### DIFF
--- a/backend/internal/oauth/oauth2/constants/constants.go
+++ b/backend/internal/oauth/oauth2/constants/constants.go
@@ -19,7 +19,11 @@
 // Package constants defines constants used across the OAuth2 module.
 package constants
 
-import "errors"
+import (
+	"errors"
+
+	"github.com/asgardeo/thunder/internal/oauth/oauth2/model"
+)
 
 // OAuth2 request parameters.
 const (
@@ -147,6 +151,7 @@ func (tam TokenEndpointAuthMethod) IsValid() bool {
 // OAuth2 token types.
 const (
 	TokenTypeBearer = "Bearer"
+	TokenTypeJWT    = "JWT"
 )
 
 // OAuth2 error codes.
@@ -164,3 +169,36 @@ const (
 
 // UnSupportedGrantTypeError is returned when an unsupported grant type is requested.
 var UnSupportedGrantTypeError = errors.New("unsupported_grant_type")
+
+// StandardOIDCScopes contains all standard OIDC scopes
+var StandardOIDCScopes = map[string]model.OIDCScope{
+	"openid": {
+		Name:        "openid",
+		Description: "REQUIRED scope for OpenID Connect authentication",
+		Claims:      []string{"sub"},
+	},
+	"profile": {
+		Name:        "profile",
+		Description: "Requests access to end-user's default profile claims",
+		Claims: []string{
+			"name", "family_name", "given_name", "middle_name",
+			"nickname", "preferred_username", "profile", "picture",
+			"website", "gender", "birthdate", "zoneinfo", "locale", "updated_at",
+		},
+	},
+	"email": {
+		Name:        "email",
+		Description: "Requests access to email and email_verified claims",
+		Claims:      []string{"email", "email_verified"},
+	},
+	"phone": {
+		Name:        "phone",
+		Description: "Requests access to phone_number and phone_number_verified claims",
+		Claims:      []string{"phone_number", "phone_number_verified"},
+	},
+	"address": {
+		Name:        "address",
+		Description: "Requests access to address claim",
+		Claims:      []string{"address"},
+	},
+}
--- a/backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
+++ b/backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
@@ -20,6 +20,7 @@ package granthandlers
 
 import (
 	"encoding/json"
+	"slices"
 	"strings"
 	"time"
 
@@ -165,19 +166,24 @@ func (h *authorizationCodeGrantHandler) HandleGrant(tokenRequest *model.TokenReq
 	// Get token configuration from OAuth app
 	iss := ""
 	validityPeriod := int64(0)
-	if oauthApp.Token != nil && oauthApp.Token.AccessToken != nil {
-		iss = oauthApp.Token.AccessToken.Issuer
-		validityPeriod = oauthApp.Token.AccessToken.ValidityPeriod
-	}
-	if iss == "" {
+
+	// Get issuer from token config
+	if oauthApp.Token != nil && oauthApp.Token.Issuer != "" {
+		iss = oauthApp.Token.Issuer
+	} else {
 		iss = config.GetThunderRuntime().Config.JWT.Issuer
 	}
-	if validityPeriod == 0 {
+
+	// Get validity period from access token config
+	if oauthApp.Token != nil && oauthApp.Token.AccessToken != nil {
+		validityPeriod = oauthApp.Token.AccessToken.ValidityPeriod
+	} else {
 		validityPeriod = config.GetThunderRuntime().Config.JWT.ValidityPeriod
 	}
 
-	userAttributes := map[string]interface{}{}
-	if len(oauthApp.Token.AccessToken.UserAttributes) > 0 {
+	var attrs map[string]interface{}
+	if len(oauthApp.Token.AccessToken.UserAttributes) > 0 ||
+		(oauthApp.Token != nil && oauthApp.Token.IDToken != nil && len(oauthApp.Token.IDToken.UserAttributes) > 0) {
 		user, svcErr := h.UserService.GetUser(authCode.AuthorizedUserID)
 		if svcErr != nil {
 			logger.Error("Failed to fetch user attributes",
@@ -188,7 +194,6 @@ func (h *authorizationCodeGrantHandler) HandleGrant(tokenRequest *model.TokenReq
 			}
 		}
 
-		var attrs map[string]interface{}
 		if err := json.Unmarshal(user.Attributes, &attrs); err != nil {
 			logger.Error("Failed to unmarshal user attributes", log.String("userID", authCode.AuthorizedUserID),
 				log.Error(err))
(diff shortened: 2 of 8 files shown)


docs/apis/application.yaml
```diff
@@ -123,6 +123,13 @@ paths:
                         issuer: "thunder"
                         validity_period: 3600
                         user_attributes: ["email", "username"]
+                      id_token:
+                        issuer: "thunder"
```
senthalan: Do we need different issuers for access token and id token ?

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
...
+		}
+	}
+
+	// Create ID token DTO
+	idTokenDTO := &model.TokenDTO{
+		Token:          idToken,
+		TokenType:      constants.TokenTypeJWT,
+		IssuedAt:       time.Now().Unix(),
+		ExpiresIn:      idTokenValidityPeriod,
+		Scopes:         authorizedScopes,
```
senthalan, line 385: Does the spec mandates having scopes in the ID Token ?

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
...
+				scopeClaims = defaultClaims
+			}
+		}
+
+		for _, claim := range scopeClaims {
+			claimAllowed := false
+			if len(idTokenUserAttributes) > 0 {
+				for _, allowedAttr := range idTokenUserAttributes {
+					if claim == allowedAttr {
+						claimAllowed = true
```
senthalan: Why we need this claimAllowed flag ?

Can't we do this
```suggestion
					if claim == allowedAttr && userAttributes[claim] != nil {
						claims[claim] = userAttributes[claim]
```

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
...
+
+		if scopeClaims == nil {
+			if defaultClaims, exists := constants.DefaultOIDCScopeClaims[scope]; exists {
+				scopeClaims = defaultClaims
+			}
+		}
+
+		for _, claim := range scopeClaims {
+			claimAllowed := false
+			if len(idTokenUserAttributes) > 0 {
```
senthalan: according to this, if the len(idTokenUserAttributes) > 0, we are not adding anything any userAttributes to the claim list. So shouldn't this checked in the top of the function at line 307 (before the for loops) and return early. (we can move the Add optional OIDC claims logic also top, which allows us to return early)

docs/apis/application.yaml
ThaminduDilshan: We have two levels to configure token attributes; root token config and the token config under `inbound_auth_config`. If someone doesn't need to modify token configurations between auth assertion and oauth tokens, they can just define the root token config (inbound auth token configs will be derived from the root token configs).

If we need to allow defying two issuers for access token and id token, we can consider this approach along with applying root level configs if not specified. Also can consider another level for id token configs; i.e. apply access token configs when id token configs are not defined.

However if there'a no requirement to allow defying two issuers for access token and id token, we can move the `issuer` to `inbound_auth_config.config.token`

backend/internal/application/service/applicationservice.go
```diff
@@ -129,11 +129,12 @@ func (as *ApplicationService) CreateApplication(app *model.ApplicationDTO) (*mod
 		Token:                     rootToken,
 	}
 	if inboundAuthConfig != nil {
-		// Wrap the finalOAuthToken in OAuthTokenConfig structure
+		// Wrap the finalOAuthAccessToken and finalOAuthIDToken in OAuthTokenConfig structure
 		var oAuthTokenConfig *model.OAuthTokenConfig
-		if finalOAuthToken != nil {
+		if finalOAuthAccessToken != nil || finalOAuthIDToken != nil {
```
ThaminduDilshan: We should have separate if conditions for this right?
Let's check whether we actually need this if condition(s)

backend/internal/application/service/applicationservice.go
```diff
@@ -192,9 +193,10 @@ func (as *ApplicationService) CreateApplication(app *model.ApplicationDTO) (*mod
 	if inboundAuthConfig != nil {
 		// Construct the return DTO with processed token configuration
 		var returnTokenConfig *model.OAuthTokenConfig
-		if finalOAuthToken != nil {
+		if finalOAuthAccessToken != nil || finalOAuthIDToken != nil {
```
ThaminduDilshan: Same as previous comment. Let's check other places as well

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: This is for the TokenDTO metadata (to reuse the existing common TokenDTO model). The ID token itself doesn't include scopes.

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -164,3 +165,34 @@ const (
...
+}
+
+// OIDCScopeDescriptions provides human-readable descriptions for OIDC scopes
+var OIDCScopeDescriptions = map[string]string{
+	OIDCScopeOpenID:  "REQUIRED scope for OpenID Connect authentication",
+	OIDCScopeProfile: "Requests access to end-user's default profile claims",
+	OIDCScopeEmail:   "Requests access to email and email_verified claims",
+	OIDCScopePhone:   "Requests access to phone_number and phone_number_verified claims",
+	OIDCScopeAddress: "Requests access to address claim",
+}
```
ThaminduDilshan, line 204: Instead of these individual const and vars, shall we define a struct for `OIDCScope` and add these different requirements as struct fields? Something like;

```go
type OIDCScope struct {
    Name
    Description
    DefaultClaims
}
```

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
...
+		if oauthApp.Token != nil && oauthApp.Token.IDToken != nil &&
+			oauthApp.Token.IDToken.ScopeClaims != nil {
+			if appClaims, exists := oauthApp.Token.IDToken.ScopeClaims[scope]; exists {
+				scopeClaims = appClaims
+			}
+		}
+
+		if scopeClaims == nil {
+			if defaultClaims, exists := constants.DefaultOIDCScopeClaims[scope]; exists {
+				scopeClaims = defaultClaims
```
ThaminduDilshan: .

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
...
+
+	// Add optional OIDC claims
+	now := time.Now().Unix()
+	claims["auth_time"] = now
+
+	return claims
+}
+
+// shouldGenerateIDToken checks if ID token should be generated based on requested scopes
+func shouldGenerateIDToken(scopes []string) bool {
```
ThaminduDilshan: Shall we simply use `slices.Contain` instead?

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
```diff
@@ -279,3 +293,107 @@ func validateAuthorizationCode(tokenRequest *model.TokenRequest,
 
 	return nil
 }
+
+// generateIDTokenClaims generates ID token claims based on scopes and application configuration
+func generateIDTokenClaims(scopes []string, userAttributes map[string]interface{},
```
ThaminduDilshan: Shall we consider a better name. It's not generating claims (Maybe populate or get)

backend/internal/application/service/applicationservice.go
thiva-k: Removed these conditions since they are redundant.

backend/internal/application/service/applicationservice.go
thiva-k: Removed these conditions since they are redundant.

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: Updated to return early if `len(idTokenUserAttributes) == 0`

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: Updated

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: Updated

backend/internal/oauth/oauth2/constants/constants.go
thiva-k: Updated

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: Updated to use `slices.Contains(..)` instead

docs/apis/application.yaml
thiva-k: `issuer` moved to token-level rather than having separate access and id token issuers

backend/internal/oauth/oauth2/granthandlers/authorizationcode.go
thiva-k: Scope-claim mapping would be handled through a separate API at server(organization)-level as discussed offline. Will be addressed in the follow-up PR.

ThaminduDilshan (approved):

Issue #158, opened 2025-06-20
Remove the usage of additional sqllite db script file in the tests and rename the postgress.sql to postgres.sql


darshanasbg:
### Current Limitation

1. Currently there are duplicate db scripts maintained.
Primary one - https://github.com/asgardeo/thunder/tree/main/backend/dbscripts
Duplicated one for tests - https://github.com/asgardeo/thunder/tree/main/tests/integration/resources/dbscripts

2. Rename the postgress.sql to postgres.sql

### Suggested Improvement

1. Remove the usage of additional sqllite db script file in the tests
2. Rename the postgress.sql to postgres.sql

ThaminduDilshan:
Refer https://github.com/asgardeo/thunder/pull/551#discussion_r2443942080

PR #551, opened 2025-10-20
Fix PostgreSQL query issues

ThaminduDilshan:
## Purpose

This pull request fixes the issues with PostgreSQL database by improving the handling of JSON properties and filtering in the codebase, with a focus on supporting both string and byte slice types (as returned by PostgreSQL for TEXT/JSON columns) and enhancing JSON path filtering for both simple and nested keys. It also adds comprehensive tests to ensure correct behavior for these scenarios.


backend/internal/system/database/utils/querybuilder.go
```diff
@@ -49,6 +50,26 @@ func BuildFilterQuery(
...
+
+// buildSQLiteJSONCondition builds a SQLite JSON filter condition.
+// For both nested and simple paths, it uses json_extract with dot notation.
+func buildSQLiteJSONCondition(columnName, key string) string {
+	return fmt.Sprintf(" AND json_extract(%s, '$.%s') = ?", columnName, key)
+}
+
 // validateKey ensures that the provided key contains only safe characters (alphanumeric and underscores).
 func validateKey(key string) error {
 	for _, char := range key {
```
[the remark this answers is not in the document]
ThaminduDilshan: This validation is only for the individual keys. Hence this is a invalid comment

.github/actions/run-integration-tests/action.yml
```diff
@@ -40,14 +40,14 @@ runs:
       run: |
         PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d thunderdb -c "CREATE DATABASE identitydb;"
         PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d thunderdb -c "CREATE DATABASE runtimedb;"
-        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d identitydb < backend/dbscripts/thunderdb/postgres.sql
-        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d runtimedb < backend/dbscripts/runtimedb/postgres.sql
+        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d identitydb < tests/integration/resources/dbscripts/thunderdb/postgres.sql
+        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d runtimedb < tests/integration/resources/dbscripts/runtimedb/postgres.sql
```
senthalan, line 44: Any specific reason to maintain separate file for the integration tests?  can't we reuse the `backend/dbscripts/thunderdb/postgres.sql` ?

With this developers have to update the schema in two places which might lead to mistakes

ThaminduDilshan: We have some queries for pre created test resources in that directory (Ex: notification sender id/ idp id's needs to be updated in flow graphs). This requires the following steps.

1. Start the server
2. Create resources (notification senders/ IDP, etc)
3. Replace flow graphs
4. Restart the server and continue integration tests

Hence currently we're following the above approach. Moved all the files there to fix the current tests.
We should be able to get rid of this once the flow configuration API is available.

senthalan: then we can execute it as two separate steps right ?
```
        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d identitydb < backend/dbscripts/thunderdb/postgres.sql 
        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d runtimedb < backend/dbscripts/runtimedb/postgres.sql
        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d identitydb < tests/integration/resources/dbscripts/thunderdb/postgres-init.sql
        PGPASSWORD=asgthunder psql -h localhost -p 5432 -U asgthunder -d runtimedb < tests/integration/resources/dbscripts/runtimedb/postgres-init.sql
```
Won't this work

So that the `backend/dbscripts/*` will only contain the production ready SQL commands and `tests/integration/resources/dbscripts/` will contains any additional commands needed for the tests

ThaminduDilshan: Modifying above action only fixes the workflow runs, however same approach needs to be applied to the local integration tests execution as well..
This needs some refactoring. Shall we check this separately as not related to $subject of this PR?

darshanasbg: Lets track the improvement and proceed.

ThaminduDilshan: Will be tracked in https://github.com/asgardeo/thunder/issues/158

Issue #510, opened 2025-10-14
Implement RBAC Foundation


senthalan:
Establish the foundational for the role-based access control system and integrate with auth code flow

Deliverables:
1. RBAC Implementation 
         - CURD APIs for Roles
         - Role to group or entry mapping
         - Role to permission mapping
2. Core authorization engine support
3. Integrate with token issuance

PR #552, opened 2025-10-20
Role API definition

senthalan:
API definition for the Role resource.

Part of https://github.com/asgardeo/thunder/issues/510

Supported operations
image

### DIFF
--- a/api/role.yaml
+++ b/api/role.yaml
@@ -1,0 +1,896 @@
+openapi: 3.0.3
+info:
+  title: Role Management API
+  version: "1.0"
+  description: This API is used to manage roles with permission-based access control.
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: https://{host}:{port}
+    variables:
+      host:
+        default: "localhost"
+      port:
+        default: "8090"
+
+tags:
+  - name: roles
+    description: Operations related to role management
+
+paths:
+  /roles:
+    get:
+      tags:
+        - roles
+      summary: List roles
+      parameters:
+        - $ref: '#/components/parameters/limitQueryParam'
+        - $ref: '#/components/parameters/offsetQueryParam'
+      responses:
+        "200":
+          description: List of roles
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RoleListResponse'
+              example:
+                totalResults: 5
+                startIndex: 1
+                count: 2
+                roles:
+                  - id: "3fa85f64-5717-4562-b3fc-2c963f66afa6"
+                    name: "front-desk-agent"
+                    description: "Front desk agent role with booking and customer management permissions"
+                    ouId: "a839f4bd-39dc-4eaa-b5cc-210d8ecaee87"
+                  - id: "257e528f-eb24-48b6-884d-20460e190957"
+                    name: "hotel-manager"
+                    description: "Hotel manager with full property access"
+                    ouId: "a839f4bd-39dc-4eaa-b5cc-210d8ecaee87"
+                links:
+                  - href: "roles?offset=2&limit=2"
+                    rel: "next"
+        "400":
+          description: Bad request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              examples:
+                invalid-request-format:
+                  summary: Invalid request format
+                  value:
+                    code: "ROL-1001"
+                    message: "Invalid request format"
+                    description: "The request body is malformed or contains invalid data"
+                invalid-limit:
+                  summary: Invalid limit parameter
+                  value:
+                    code: "ROL-1011"
+                    message: "Invalid limit parameter"
+                    description: "The limit parameter must be a positive integer"
+                invalid-offset:
+                  summary: Invalid offset parameter
+                  value:
+                    code: "ROL-1012"
+                    message: "Invalid offset parameter"
+                    description: "The offset parameter must be a non-negative integer"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "ROL-5000"
+                message: "Internal server error"
+                description: "An unexpected error occurred while processing the request"
+    
+    post:
+      tags:
+        - roles
+      summary: Create a new role
+      description: Creates a new role with optional initial permissions and assignments
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateRoleRequest'
+            example:
+              name: "front-desk-agent"
+              description: "Front desk agent role with booking and customer management permissions"
+              ouId: "a839f4bd-39dc-4eaa-b5cc-210d8ecaee87"
+              permissions:
+                - "create_reservation"
+                - "view_reservation"
+                - "update_reservation"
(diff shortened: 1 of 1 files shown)


api/role.yaml
```diff
@@ -0,0 +1,816 @@
...
+          format: uuid
+          description: "Unique identifier for the role"
+        name:
+          type: string
+          description: "Name of the role"
+        description:
+          type: string
+          nullable: true
+          description: "Optional description of the role"
+        organizationUnitId:
```
ThaminduDilshan: Shall we check if this name is consistent with the term in the user/ group APIs?

api/role.yaml
```diff
@@ -0,0 +1,816 @@
...
+                invalid-offset:
+                  summary: Invalid offset parameter
+                  value:
+                    code: "ROL-1012"
+                    message: "Invalid offset parameter"
+                    description: "The offset parameter must be a non-negative integer"
+        "500":
+          description: Internal server error
+          content:
+            text/plain:
```
ThaminduDilshan: Shall we have json response for 500 internal server errors?

@darshanasbg are we planning to migrate to `plain/text` for server errors?

api/role.yaml
```diff
@@ -0,0 +1,816 @@
...
+                - "view_reservation"
+                - "update_reservation"
+                - "cancel_reservation"
+                - "process_payment"
+                - "view_payment"
+                - "refund_payment"
+                - "view_customer_profile"
+                - "update_customer_profile"
+                - "create_customer_profile"
+              assignments:
```
ThaminduDilshan: Should we allow updating assignments from this endpoint, if we're not returning assignments in the `/roles/{id}` GET request?

api/role.yaml
senthalan: We are using mix of terms here
user API - `organizationUnit`
groups API - `organizationUnitId`
Authentication API - `organization_unit` 

Other APIs seems not have the Organization Unit reference

api/role.yaml
senthalan: yes, we allow updating the assignments through `PUT /roles/{id}`

senthalan: The `PUT /roles/{id}` will return the assignments list. The schema is '#/components/schemas/Role`. But the example of the GET needs to be updated, I will update.

api/role.yaml
senthalan: I have added the `application/json` content type as well

api/role.yaml
senthalan: updated the new schema to not to include the assignments in GET, PUT /roles/{id}

api/role.yaml
senthalan: We have discussed and agreed to go ahead with `ouId`

api/role.yaml
```diff
@@ -0,0 +1,877 @@
...
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "ROL-5000"
+                message: "Internal server error"
+                description: "An unexpected error occurred while processing the request"
+            text/plain:
+              schema:
+                type: string
+              example: "Internal server error"
```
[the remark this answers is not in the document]
senthalan: This is an intentional change. The our current implementation eventhough the specification only have application/text, we are returning application/json.

api/role.yaml
```diff
@@ -0,0 +1,877 @@
...
+          minItems: 1
+        assignments:
+          type: array
+          items:
+            $ref: '#/components/schemas/Assignment'
+          description: "Optional list of initial user and group assignments"
+
+    UpdateRoleRequest:
+      allOf:
+        - $ref: '#/components/schemas/CreateRoleRequest'
```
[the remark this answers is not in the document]
senthalan: RoleWithAssignments is only used in the response.

api/role.yaml
```diff
@@ -0,0 +1,931 @@
...
+    offsetQueryParam:
+      in: query
+      name: offset
+      required: false
+      description: |
+        Number of records to skip for pagination.
+      schema:
+        type: integer
+        minimum: 0
+        default: 0
```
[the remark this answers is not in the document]
senthalan: minimum: 0 is added.

api/role.yaml
```diff
@@ -0,0 +1,931 @@
...
+                  - "check_in_guest"
+                  - "check_out_guest"
+                  - "process_payment"
+                  - "view_payment"
+                  - "view_customer_profile"
+                  - "update_customer_profile"
+                assignments:
+                  - type: "user"
+                    id: "7a4b1f8e-5c69-4b60-9232-2b0aaf65ef3c"
+                    display: "John Doe"
```
ThaminduDilshan, line 145: Is it expected to return display here without using `include` parameter?

api/role.yaml
```diff
@@ -0,0 +1,931 @@
...
+      description: Returns all user and group assignments for the role. Use include=display to get display names.
+      parameters:
+        - in: path
+          name: id
+          required: true
+          schema:
+            type: string
+            format: uuid
+        - in: query
+          name: include
```
ThaminduDilshan, line 424: Should this be a comma separated value?

ThaminduDilshan (approved):

api/role.yaml
senthalan: We are returning it because the UI will need this display

api/role.yaml
senthalan: We don't need to support having multiple display attributes at this point of time, if there are any future improvements, we can improve it

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "ROL-5000"
+                message: "Internal server error"
+                description: "An unexpected error occurred while processing the request"
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

api/role.yaml
```diff
@@ -0,0 +1,896 @@
...
```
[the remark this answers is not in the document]
senthalan: This is an intentional change, we are planning to move response type from plain text to json.

PR #555, opened 2025-10-20
Role management implementation

senthalan:
Part of https://github.com/asgardeo/thunder/issues/510

This pull request introduces a new Role Management feature to the backend, including database schema changes, service initialization, error handling, data models, and supporting configuration for mocking and testing. The main focus is to enable creation, management, and assignment of roles, along with permissions and assignments to users and groups, with full API and database support.

**Key changes include:**

### Database Schema & Queries

- **New role management tables and indexes:** Added `ROLE`, `ROLE_PERMISSION`, and `ROLE_ASSIGNMENT` tables to both Postgres and SQLite schemas, including supporting indexes for efficient authorization queries. These tables support storing roles, their permissions, and assignments to users/groups. [[1]](diffhunk://#diff-c01456b5ea7b7cef68984dc1b755d9e255a28f488aa57ca3bdbe641206263983R58-R100) [[2]](diffhunk://#diff-478760829090a37dcf6fc4eed2791034ee3280fc878d05afbbafc26caec0f480R58-R100)
- **Role management SQL queries:** Introduced a comprehensive set of queries for CRUD operations on roles, permissions, and assignments, as well as checks for uniqueness and pagination support in `storeconstants.go`.

### Backend Service & API

- **Role service initialization and route registration:** Implemented the `Initialize` function for the role service, registering HTTP routes for role CRUD operations and assignment management, with appropriate CORS handling.
- **Service wiring:** Integrated the new role service into the main service manager, ensuring it is initialized with dependencies on user, group, and organization unit services. [[1]](diffhunk://#diff-36914807f52c25f79a7d09a3a5cca10a8adb6df10e3df582e89ea4c713143216R34) [[2]](diffhunk://#diff-36914807f52c25f79a7d09a3a5cca10a8adb6df10e3df582e89ea4c713143216L54-R56)

### Data Models

- **Role and assignment models:** Defined the core data structures for roles, assignments, and related API requests/responses, supporting role details, permissions, assignments, and pagination.

### Error Handling

- **Role management error constants:** Added detailed error types and codes for client and server errors related to role operations, including validation, not found, and conflict scenarios.

### Testing & Mocking

- **Mockery configuration updates:** Updated mockery configuration files to support mocking of new role, group, and organization unit interfaces for both private and public test scenarios. [[1]](diffhunk://#diff-aa648ea8e724297a9b8cbc5df07db2af3a4192c862e53db2efb6ef27a7db42a5R54-R59) [[2]](diffhunk://#diff-d6c48c1097bc5a5f933289e9cf77f735eb15dfe614dbf3d0eceffda01cc68050R239-R254)

These changes collectively enable robust role-based access control capabilities in the backend system.


backend/internal/role/handler.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJSON)
+	w.WriteHeader(http.StatusOK)
+
+	if err := json.NewEncoder(w).Encode(roleListResponse); err != nil {
+		logger.Error("Error encoding response", log.Error(err))
+		handleEncodingError(w, logger)
+		return
+	}
```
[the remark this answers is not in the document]
senthalan: When there is a encoding error, the response needs to be to 500.

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+		Name:               request.Name,
+		Description:        request.Description,
+		OrganizationUnitID: request.OrganizationUnitID,
+		Permissions:        request.Permissions,
+		Assignments:        request.Assignments,
+	}
+
+	if err := rs.roleStore.CreateRole(role); err != nil {
+		logger.Error("Failed to create role", log.Error(err))
+		return nil, &ErrorInternalServerError
```
thiva-k, line 151: Scenarios for duplicate assignments/permissions to a role should be treated as client-side errors. Currently, these seem not to be differentiated in pre-checks, so they will be treated as server errors due to DB constraints right ?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+
+		if len(invalidUserIDs) > 0 {
+			logger.Debug("Invalid user IDs found", log.Any("invalidUserIDs", invalidUserIDs))
+			return &ErrorInvalidAssignmentID
+		}
+	}
+
+	// Validate group IDs using group service
+	if len(groupIDs) > 0 {
+		for _, groupID := range groupIDs {
```
thiva-k: Can't we use the existing `validateGroupIDs` here by making it public, instead of looping?

backend/dbscripts/thunderdb/postgres.sql
```diff
@@ -55,6 +55,49 @@ CREATE TABLE GROUP_MEMBER_REFERENCE (
     FOREIGN KEY (GROUP_ID) REFERENCES "GROUP" (GROUP_ID) ON DELETE CASCADE
 );
 
+-- Table to store Roles
+CREATE TABLE "ROLE" (
+    ID                  INT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+    ROLE_ID             VARCHAR(36) UNIQUE NOT NULL,
```
thiva-k, line 61: It will be better to add a unique constraint on (OU_ID, NAME) at DB level too right ? Could reduce potential race conditions

backend/dbscripts/thunderdb/postgres.sql
```diff
@@ -55,6 +55,49 @@ CREATE TABLE GROUP_MEMBER_REFERENCE (
...
+    OU_ID               VARCHAR(36)        NOT NULL,
```
ThaminduDilshan: Shall we check for proper formatting in line 62-63?

backend/internal/role/handler.go
```diff
@@ -0,0 +1,407 @@
...
+
+	w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJSON)
+	w.WriteHeader(http.StatusCreated)
+
+	isErr := writeToResponse(w, createdRole, logger)
+	if isErr {
+		return
+	}
+
+	logger.Debug("Successfully created role", log.String("role id", createdRole.ID))
```
ThaminduDilshan:
```suggestion
	logger.Debug("Successfully created role", log.String("roleId", createdRole.ID))
```

Shall we use camel case for log parameters for consistency?
Check other places as well

backend/internal/role/handler.go
```diff
@@ -0,0 +1,407 @@
...
+			sanitized.Permissions[i] = sysutils.SanitizeString(permission)
+		}
+	}
+
+	if request.Assignments != nil {
+		sanitized.Assignments = make([]Assignment, len(request.Assignments))
+		for i, assignment := range request.Assignments {
+			sanitized.Assignments[i] = Assignment{
+				ID:   sysutils.SanitizeString(assignment.ID),
+				Type: assignment.Type,
```
ThaminduDilshan, line 379: Don't we need to sanitize this as well?

backend/internal/role/model.go
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+package role
+
+// AssigneeType represents the type of assignee entity.
+type AssigneeType string
+
+const (
+	// AssigneeTypeUser is the type for users.
+	AssigneeTypeUser AssigneeType = "user"
```
ThaminduDilshan, line 26: What if we define these type values as all caps for consistency?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+	"github.com/asgardeo/thunder/internal/system/utils"
+	"github.com/asgardeo/thunder/internal/user"
+)
+
+const loggerComponentName = "RoleMgtService"
+
+// RoleServiceInterface defines the interface for the role service.
+type RoleServiceInterface interface {
+	GetRoleList(limit, offset int) (*RoleListResponse, *serviceerror.ServiceError)
+	CreateRole(request CreateRoleRequest) (*Role, *serviceerror.ServiceError)
```
ThaminduDilshan: I'd prefer to define separate structs for service layer rather than using request models.
WDYT?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+			Href: fmt.Sprintf("%s?offset=%d&limit=%d", base, lastPageOffset, limit),
+			Rel:  "last",
+		})
+	}
+
+	return links
+}
+
+// getDisplayNameForAssignment retrieves the display name for a user or group assignment.
+func (rs *roleService) getDisplayNameForAssignment(assignment *Assignment) (string, error) {
```
ThaminduDilshan: Shall we handle service errors properly when calling other services from this method? Currently even client errors gets wrapped as a server error right?
Maybe we can return a service error from `getDisplayNameForAssignment` method too

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+
+	assignments, err := rs.roleStore.GetRoleAssignments(roleID, limit, offset)
+	if err != nil {
+		logger.Error("Failed to get role assignments", log.String("roleID", roleID), log.Error(err))
+		return nil, &ErrorInternalServerError
+	}
+
+	// Populate display names if requested
+	if includeDisplay {
+		for i := range assignments {
```
ThaminduDilshan: As discussed shall we add a note/ create an issue to improve display name retrieval?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+			userIDs = append(userIDs, assignment.ID)
+		case AssigneeTypeGroup:
+			groupIDs = append(groupIDs, assignment.ID)
+		}
+	}
+
+	// Validate user IDs using user service
+	if len(userIDs) > 0 {
+		invalidUserIDs, svcErr := rs.userService.ValidateUserIDs(userIDs)
+		if svcErr != nil {
```
ThaminduDilshan, line 546: Shall we handle service errors properly?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+			logger.Error("Failed to validate user IDs", log.String("error", svcErr.Error), log.String("code", svcErr.Code))
```
ThaminduDilshan: Is this line exceeding 120 characters?

backend/internal/role/service.go
```diff
@@ -0,0 +1,614 @@
...
+	// Validate group IDs using group service
+	if len(groupIDs) > 0 {
+		for _, groupID := range groupIDs {
+			_, svcErr := rs.groupService.GetGroup(groupID)
+			if svcErr != nil {
+				if svcErr.Code == group.ErrorGroupNotFound.Code {
+					logger.Debug("Invalid group ID found", log.String("groupID", groupID))
+					return &ErrorInvalidAssignmentID
+				}
+				logger.Error("Failed to validate group ID", log.String("groupID", groupID), log.String("error", svcErr.Error))
```
ThaminduDilshan: Check for line length

backend/internal/role/store.go
```diff
@@ -0,0 +1,610 @@
...
+	}
+
+	results, err := dbClient.Query(queryGetRoleAssignments, roleID, limit, offset)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get role assignments: %w", err)
+	}
+
+	assignments := make([]Assignment, 0)
+	for _, row := range results {
+		if assigneeID, ok := row["assignee_id"].(string); ok {
```
ThaminduDilshan: We should handle errors here right (when casting)?

backend/internal/role/store.go
```diff
@@ -0,0 +1,610 @@
...
+// getRolePermissions retrieves all permissions for a role.
+func (s *roleStore) getRolePermissions(dbClient provider.DBClientInterface, roleID string) ([]string, error) {
+	results, err := dbClient.Query(queryGetRolePermissions, roleID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get role permissions: %w", err)
+	}
+
+	permissions := make([]string, 0)
+	for _, row := range results {
+		if permission, ok := row["permission"].(string); ok {
```
ThaminduDilshan: Should we handle errors here?

backend/internal/role/store.go
```diff
@@ -0,0 +1,610 @@
...
+// getAllRoleAssignments retrieves all assignments for a role (without pagination).
+func (s *roleStore) getAllRoleAssignments(dbClient provider.DBClientInterface, roleID string) ([]Assignment, error) {
+	results, err := dbClient.Query(queryGetAllRoleAssignments, roleID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get role assignments: %w", err)
+	}
+
+	assignments := make([]Assignment, 0)
+	for _, row := range results {
+		if assigneeID, ok := row["assignee_id"].(string); ok {
```
ThaminduDilshan: Should we handle errors here?

backend/internal/role/store.go
```diff
@@ -0,0 +1,610 @@
...
+		return RoleSummary{}, fmt.Errorf("failed to parse role_id as string")
+	}
+
+	name, ok := row["name"].(string)
+	if !ok {
+		return RoleSummary{}, fmt.Errorf("failed to parse name as string")
+	}
+
+	description := ""
+	if desc, ok := row["description"].(string); ok {
```
ThaminduDilshan: Should we handle errors here? Shall we check other places as well

backend/internal/role/storeconstants.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	// queryDeleteRole deletes a role.
+	queryDeleteRole = dbmodel.DBQuery{
+		ID:    "RLQ-ROLE_MGT-06",
+		Query: `DELETE FROM "ROLE" WHERE ROLE_ID = $1`,
+	}
+
+	// queryCreateRolePermission creates a new role permission.
+	queryCreateRolePermission = dbmodel.DBQuery{
```
ThaminduDilshan, line 68: Query id numbers seems to be incorrect. Shall we check?

backend/internal/system/error/serviceerror/error.go
```diff
@@ -50,3 +50,14 @@ func CustomServiceError(svcError ServiceError, errorDesc string) *ServiceError {
 	}
 	return err
 }
+
+// Server errors
+var (
+	// EncodingError is the error returned when encoding the response.
+	ErrorEncodingError = ServiceError{
+		Type:             ServerErrorType,
+		Code:             "ENC-5001",
```
ThaminduDilshan: What if we have code as "ENC-5000"?

backend/.mockery.private.yml
```diff
@@ -51,4 +51,10 @@ packages:
       dir: internal/authn
       structname: '{{.InterfaceName}}Mock'
       pkgname: authn
+  github.com/asgardeo/thunder/internal/role:
```
ThaminduDilshan, line 55: Shall we add a new line for consistency?

tests/integration/role/roleapi_test.go
```diff
@@ -0,0 +1,933 @@
...
+	return &response, nil
+}
+
+func (suite *RoleAPITestSuite) updateRole(roleID string, request UpdateRoleRequest) (*Role, error) {
+	body, err := json.Marshal(request)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest("PUT", fmt.Sprintf("%s%s/%s", testServerURL, rolesBasePath, roleID), bytes.NewBuffer(body))
```
ThaminduDilshan: Shall we check line length?

tests/integration/role/roleapi_test.go
```diff
@@ -0,0 +1,933 @@
...
+	return nil
+}
+
+func (suite *RoleAPITestSuite) addAssignments(roleID string, request AssignmentsRequest) error {
+	body, err := json.Marshal(request)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest("POST", fmt.Sprintf("%s%s/%s/assignments/add", testServerURL, rolesBasePath, roleID), bytes.NewBuffer(body))
```
ThaminduDilshan: Shall we check line length

backend/internal/role/handler.go
senthalan: This is an AssigneeType. So don't need sanitize

backend/internal/role/model.go
senthalan: This is how it's been defined in the groups package

https://github.com/asgardeo/thunder/blob/main/backend/internal/group/model.go#L26

backend/internal/role/store.go
senthalan: if the type assertion fails (ok == false), it correctly defaults right?
Do you think we need to fail the flow in such cases?

backend/internal/role/service.go
senthalan: Updated the precheck to remove the duplicates

backend/dbscripts/thunderdb/postgres.sql
senthalan: Added  unique constraint on (OU_ID, NAME)

backend/internal/role/service.go
senthalan: Updated the getDisplayNameForAssignment function

backend/internal/system/error/serviceerror/error.go
```diff
@@ -50,3 +50,10 @@ func CustomServiceError(svcError ServiceError, errorDesc string) *ServiceError {
 	}
 	return err
 }
+
+// Server errors
+var (
+	// EncodingError is the error returned when encoding the response.
+	ErrorEncodingError = "{Code: \"ENC-5000\",Error: \"Encoding error\"," +
+		"ErrorDescription: \"An error occurred while encoding the response\"}"
+)
```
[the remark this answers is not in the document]
senthalan: This is intensional to send the json as string to the client incase of a encoding issue

backend/internal/role/storeconstants.go
senthalan: updated

backend/internal/role/internal/role/RoleServiceInterface_mock_test.go
rajithacharith, line 1: Is this directory structure intended? (internal/role/internal/role)

backend/internal/role/service.go
```diff
@@ -0,0 +1,699 @@
...
+	logger.Debug("Creating role", log.String("name", request.Name))
+
+	if err := rs.validateCreateRoleRequest(request); err != nil {
+		return nil, err
+	}
+
+	// Validate organization unit exists using OU service
+	_, svcErr := rs.ouService.GetOrganizationUnit(request.OrganizationUnitID)
+	if svcErr != nil {
+		if svcErr.Code == oupkg.ErrorOrganizationUnitNotFound.Code {
```
rajithacharith, line 128: Can we check for the exact error using errors.Is()? WDYT about returning the OU service error without having a new error object?

senthalan: We can't do `errors.Is()` because svcErr is not a type of go error

backend/internal/role/storeconstants.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	dbmodel "github.com/asgardeo/thunder/internal/system/database/model"
+)
+
+var (
+	// queryCreateRole creates a new role.
+	queryCreateRole = dbmodel.DBQuery{
+		ID:    "RLQ-ROLE_MGT-01",
+		Query: `INSERT INTO "ROLE" (ROLE_ID, OU_ID, NAME, DESCRIPTION) VALUES ($1, $2, $3, $4)`,
```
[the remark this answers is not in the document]
senthalan: Add a comment

backend/internal/role/errorconstants.go
```diff
@@ -0,0 +1,129 @@
...
+	ErrorEmptyAssignments = serviceerror.ServiceError{
+		Type:             serviceerror.ClientErrorType,
+		Code:             "ROL-1010",
+		Error:            "Empty assignments list",
+		ErrorDescription: "At least one assignment must be provided",
+	}
+	// ErrorMissingUserOrGroups is the error returned when both user ID and groups are missing.
+	ErrorMissingUserOrGroups = serviceerror.ServiceError{
+		Type:             serviceerror.ClientErrorType,
+		Code:             "ROL-1015",
```
ThaminduDilshan:
```suggestion
		Code:             "ROL-1011",
```

backend/internal/role/handler.go
ThaminduDilshan: Still this will be treated as a string right? AFAIK it doesn't guarantee this will be one of the defined AssigneeType.

backend/internal/role/handler.go
```diff
@@ -0,0 +1,513 @@
...
+	}
+
+	if limit == 0 {
+		limit = serverconst.DefaultPageSize
+	}
+
+	return limit, offset, nil
+}
+
+// Conversion helper methods between HTTP and service layer structs
```
ThaminduDilshan:
```suggestion
```

backend/internal/role/handler.go
```diff
@@ -0,0 +1,513 @@
...
+		ID:                 role.ID,
+		Name:               role.Name,
+		Description:        role.Description,
+		OrganizationUnitID: role.OrganizationUnitID,
+		Permissions:        role.Permissions,
+		Assignments:        httpAssignments,
+	}
+}
+
+func toHTTPLinks(links []Link) []LinkResponse {
```
ThaminduDilshan, line 498: Let's add a comment

backend/internal/role/model.go
ThaminduDilshan: I'd prefer to change them too. Maybe with a separate PR.
WDYT @darshanasbg ?

backend/internal/role/model.go
```diff
@@ -0,0 +1,198 @@
...
+// AssignmentListResponse represents the response for listing role assignments with pagination.
+type AssignmentListResponse struct {
+	TotalResults int                  `json:"totalResults"`
+	StartIndex   int                  `json:"startIndex"`
+	Count        int                  `json:"count"`
+	Assignments  []AssignmentResponse `json:"assignments"`
+	Links        []LinkResponse       `json:"links"`
+}
+
+// Internal service layer structs - used for business logic processing
```
ThaminduDilshan, line 117:
```suggestion
```

We don't need these comments right?

senthalan: I would prefer to keep it as a separator to differentiate the structs used in the handler and service layer

backend/internal/role/handler.go
senthalan: Since AssigneeType only has two valid values ("user" and "group"), and the service layer already validates this. Do we need to do string sanitize ?

If we want to do it, then the option is to Convert to string, sanitize, then convert back to AssigneeType
```
Type: AssigneeType(sysutils.SanitizeString(string(assignment.Type))),
```
I don't think that's good idea.

backend/internal/role/handler.go
senthalan: Added

backend/internal/role/service.go
senthalan: Created an issue to track this https://github.com/asgardeo/thunder/issues/639

PR #560, opened 2025-10-21
Update token endpoint auth method

thiva-k:
## Purpose
Update application API to have a single value for `token_endpoint_auth_method` to align with DCR specification[1] to support DCR compatibility.
Add validation to return error if no `token_endpoint_auth_method` is defined for an application.

[1] https://datatracker.ietf.org/doc/html/rfc7591#section-2

### DIFF
--- a/backend/internal/application/error_constants.go
+++ b/backend/internal/application/error_constants.go
@@ -114,13 +114,6 @@ var (
 		Error:            "Invalid response type",
 		ErrorDescription: "One or more provided response types are invalid",
 	}
-	// ErrorMissingTokenEndpointAuthMethod is the error returned when token endpoint auth method is not provided.
-	ErrorMissingTokenEndpointAuthMethod = serviceerror.ServiceError{
-		Type:             serviceerror.ClientErrorType,
-		Code:             "APP-1017",
-		Error:            "Missing token endpoint authentication method",
-		ErrorDescription: "Token endpoint authentication method is required and must be specified",
-	}
 	// ErrorInvalidTokenEndpointAuthMethod is the error returned when an invalid token endpoint auth method
 	// is provided.
 	ErrorInvalidTokenEndpointAuthMethod = serviceerror.ServiceError{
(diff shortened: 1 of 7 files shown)


backend/internal/application/error_constants.go
```diff
@@ -114,13 +114,20 @@ var (
 		Error:            "Invalid response type",
 		ErrorDescription: "One or more provided response types are invalid",
 	}
+	// ErrorMissingTokenEndpointAuthMethod is the error returned when token endpoint auth method is not provided.
+	ErrorMissingTokenEndpointAuthMethod = serviceerror.ServiceError{
```
ThaminduDilshan: Instead of an error, let's set a default value

tests/integration/testutils/mock_google_oidc_server.go
```diff
@@ -87,17 +87,17 @@ type GoogleUserInfo struct {
...
+	Issuer                           string   `json:"issuer"`
+	AuthorizationEndpoint            string   `json:"authorization_endpoint"`
+	TokenEndpoint                    string   `json:"token_endpoint"`
+	UserinfoEndpoint                 string   `json:"userinfo_endpoint"`
+	JwksURI                          string   `json:"jwks_uri"`
+	ResponseTypesSupported           []string `json:"response_types_supported"`
+	SubjectTypesSupported            []string `json:"subject_types_supported"`
+	IDTokenSigningAlgValuesSupported []string `json:"id_token_signing_alg_values_supported"`
+	ScopesSupported                  []string `json:"scopes_supported"`
+	TokenEndpointAuthMethodSupported string   `json:"token_endpoint_auth_method_supported"`
```
ThaminduDilshan: We don't need to change this. Let's revert it.
Check other mock servers as well

backend/internal/application/error_constants.go
thiva-k: Added fall backs for token endpoint method, grant and response type according to DCR specs to be consistent

tests/integration/testutils/mock_google_oidc_server.go
thiva-k: These are reverted

Issue #578, opened 2025-10-24
Analytics for Thunder


sahandilshan:
### Problem

Thunder doesn't have a way to check the below state
- No of logins
- No of success login attempts
- No of failures
- Duration during each login steps

1. Thunder should be able to publish relavant login events
2. Also should be able to view those events in detailed & abstract manner

### Proposed Solution

Solution will be discussed under https://github.com/asgardeo/thunder/discussions/532

PR #589, opened 2025-10-27
Add initial observability component

sahandilshan:
## Purpose
This component is introduce to manage all logs of Thunder. Initially this will work in pub sub way where all the logs will publish from publisher and there will be subscribers who listen to these publishers and have their own message format and output adapter which will be used to convert the event and publish it to their required destination.

## Approach
As mentioned in https://github.com/asgardeo/thunder/discussions/532#discussioncomment-14789330, this use a pub/sub approach to publish events of Thunder.

### Related Issue(s)
- https://github.com/asgardeo/thunder/issues/578

### DIFF
--- a/backend/cmd/server/main.go
+++ b/backend/cmd/server/main.go
@@ -31,11 +31,13 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/asgardeo/thunder/internal/observability"
 	"github.com/asgardeo/thunder/internal/system/cache"
 	"github.com/asgardeo/thunder/internal/system/cert"
 	"github.com/asgardeo/thunder/internal/system/config"
 	"github.com/asgardeo/thunder/internal/system/database/provider"
 	"github.com/asgardeo/thunder/internal/system/log"
+	"github.com/asgardeo/thunder/internal/system/middleware"
 )
 
 // shutdownTimeout defines the timeout duration for graceful shutdown.
@@ -53,6 +55,9 @@ func main() {
 
 	// Initialize the cache manager.
 	initCacheManager(logger)
+
+	// Initialize observability with console adapter and JSON format
+	initObservability(logger)
 
 	// Create a new HTTP multiplexer.
 	mux := http.NewServeMux()
@@ -136,6 +141,34 @@ func initCacheManager(logger *log.Logger) {
 	cm.Init()
 }
 
+// initObservability initializes the observability service with console adapter and JSON format.
+func initObservability(logger *log.Logger) {
+	// Configure observability to use console adapter with JSON format
+	observabilityCfg := &observability.Config{
+		Enabled: true,
+		Output: observability.OutputConfig{
+			Type:   "console", // Output to stdout
+			Format: "json",
+		},
+		Metrics: observability.MetricsConfig{
+			Enabled: true,
+		},
+		FailureMode: "graceful", // Don't fail if observability has issues
+	}
+
+	svc, err := observability.InitializeWithConfig(observabilityCfg)
+	if err != nil {
+		logger.Error("Failed to initialize observability service", log.Error(err))
+		return
+	}
+
+	if svc.IsEnabled() {
+		logger.Debug("Observability service initialized successfully with console adapter and JSON format")
+	} else {
+		logger.Warn("Observability service is disabled")
+	}
+}
+
 // loadCertConfig loads the certificate configuration and extracts the Key ID (kid).
 func loadCertConfig(logger *log.Logger, cfg *config.Config, thunderHome string) *tls.Config {
 	sysCertSvc := cert.NewSystemCertificateService()
@@ -198,15 +231,15 @@ func startHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux)
 
 // createHTTPServer creates and configures an HTTP server with common settings.
 func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux) (*http.Server, string) {
-	// Wrap the multiplexer with AccessLogHandler.
-	wrappedMux := log.AccessLogHandler(logger, mux)
+	handler := middleware.CorrelationIDMiddleware(mux)
+	handler = log.AccessLogHandler(logger, handler)
 
 	// Build the server address using hostname and port from the configurations.
 	serverAddr := fmt.Sprintf("%s:%d", cfg.Server.Hostname, cfg.Server.Port)
 
 	server := &http.Server{
 		Addr:              serverAddr,
-		Handler:           wrappedMux,
+		Handler:           handler,
 		ReadHeaderTimeout: 10 * time.Second, // Mitigate Slowloris attacks
 		WriteTimeout:      10 * time.Second,
 		IdleTimeout:       120 * time.Second,
@@ -225,6 +258,13 @@ func gracefulShutdown(logger *log.Logger, server *http.Server) {
 		logger.Error("Error during server shutdown", log.Error(err))
 	} else {
 		logger.Debug("HTTP server shutdown completed")
+	}
+
+	// Shutdown observability service
+	observabilitySvc := observability.GetService()
+	if observabilitySvc != nil {
+		observabilitySvc.Shutdown()
+		logger.Debug("Observability service shutdown completed")
 	}
 
 	// Close database connections
--- a/backend/internal/observability/publisher/publisher.go
+++ b/backend/internal/observability/publisher/publisher.go
@@ -1,0 +1,306 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package publisher provides the event bus mechanism for the observability system.
+package publisher
+
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/asgardeo/thunder/internal/observability/event"
+	"github.com/asgardeo/thunder/internal/observability/metrics"
+	"github.com/asgardeo/thunder/internal/observability/subscriber"
+	"github.com/asgardeo/thunder/internal/system/log"
+)
+
+// CategoryPublisher is an event bus interface with category-based routing.
+// Events are published synchronously to all interested subscribers.
+type CategoryPublisher interface {
+	// Publish publishes an event immediately to all interested subscribers.
+	Publish(event *event.Event)
+
+	// Subscribe adds a subscriber (subscriber decides what categories it wants).
+	Subscribe(sub subscriber.Subscriber)
+
+	// Unsubscribe removes a subscriber.
+	Unsubscribe(sub subscriber.Subscriber)
+
+	// GetActiveCategories returns categories that have at least one subscriber.
+	GetActiveCategories() []event.EventCategory
+
+	// Shutdown gracefully shuts down the publisher.
+	Shutdown()
+}
+
+// categoryEventPublisher implements category-based event bus with smart routing.
+// This is a simple synchronous event bus - no queuing, no async processing.
+// Events are published directly to subscribers who can decide whether to process them.
+type categoryEventPublisher struct {
+	subscribers           map[string]subscriber.Subscriber // subscriberID -> subscriber
+	subscribersByCategory map[event.EventCategory][]string // category -> []subscriberIDs
+	mu                    sync.RWMutex
+	isShutdown            bool
+	wg                    sync.WaitGroup // Tracks active goroutines for graceful shutdown
+	ctx                   context.Context
+	cancel                context.CancelFunc
+}
+
+const loggerComponentName = "CategoryEventPublisher"
+
+// NewCategoryPublisher creates a new category-based event bus.
+func NewCategoryPublisher() CategoryPublisher {
+	ctx, cancel := context.WithCancel(context.Background())
+	return &categoryEventPublisher{
+		subscribers:           make(map[string]subscriber.Subscriber),
+		subscribersByCategory: make(map[event.EventCategory][]string),
+		isShutdown:            false,
+		ctx:                   ctx,
+		cancel:                cancel,
+	}
+}
+
+// Publish publishes an event to all interested subscribers.
+// This method returns immediately without blocking the caller.
+// Each subscriber receives the event in its own goroutine, ensuring:
+// - No blocking of the main thread
+// - Parallel processing by all subscribers
+// - Isolated failures (one subscriber's panic doesn't affect others)
+func (p *categoryEventPublisher) Publish(evt *event.Event) {
+	if evt == nil {
+		return
+	}
+
+	// Validate event
+	if err := evt.Validate(); err != nil {
+		logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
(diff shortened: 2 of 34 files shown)


backend/internal/observability/examples/configuration_examples.go
```diff
@@ -0,0 +1,224 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package examples provides configuration examples for the analytics system.
+package examples
```
darshanasbg: Is this something needed?

backend/internal/observability/subscriber/defaultsubscriber/default_subscriber.go
```diff
@@ -0,0 +1,247 @@
...
+	if err := ds.adapter.Write(formattedData); err != nil {
+		ds.logger.Error("Failed to write event to adapter",
+			log.String("eventType", evt.Type),
+			log.String("eventID", evt.EventID),
+			log.String("adapterName", ds.adapter.GetName()),
+			log.Error(err))
+		return fmt.Errorf("failed to write to adapter: %w", err)
+	}
+
+	if ds.logger.IsDebugEnabled() {
```
rajithacharith: Let's remove IsDebugEnabled check

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+var (
+	serviceInstance *Service
+	serviceOnce     sync.Once
+)
+
+// GetService returns the singleton instance of the observability service.
+// Uses default configuration. For custom configuration, use InitializeWithConfig first.
+func GetService() *Service {
```
rajithacharith, line 69: What happens if this was called before calling InitializeWithConfig?

backend/cmd/server/main.go
```diff
@@ -136,6 +141,34 @@ func initCacheManager(logger *log.Logger) {
...
+	}
+
+	svc, err := observability.InitializeWithConfig(analyticsCfg)
+	if err != nil {
+		logger.Error("Failed to initialize analytics service", log.Error(err))
+		return
+	}
+
+	if svc.IsEnabled() {
+		logger.Info("Analytics service initialized successfully with console adapter and JSON format")
```
ThaminduDilshan:
```suggestion
		logger.Debug("Analytics service initialized successfully with console adapter and JSON format")
```

Do we need this info log?

backend/cmd/server/main.go
```diff
@@ -198,15 +231,18 @@ func startHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux)
 
 // createHTTPServer creates and configures an HTTP server with common settings.
 func createHTTPServer(logger *log.Logger, cfg *config.Config, mux *http.ServeMux) (*http.Server, string) {
-	// Wrap the multiplexer with AccessLogHandler.
-	wrappedMux := log.AccessLogHandler(logger, mux)
+	// Apply middleware chain:
+	// 1. CorrelationIDMiddleware - Extract/generate correlation ID and add to context
```
ThaminduDilshan: I don't think we need these additional comments. Your functions are self explanatory and can get the idea

backend/cmd/server/main.go
```diff
@@ -136,6 +141,34 @@ func initCacheManager(logger *log.Logger) {
...
+		logger.Error("Failed to initialize analytics service", log.Error(err))
+		return
+	}
+
+	if svc.IsEnabled() {
+		logger.Info("Analytics service initialized successfully with console adapter and JSON format")
+	} else {
+		logger.Warn("Analytics service is disabled")
+	}
+}
```
rajithacharith, line 170: Do we need this logging?
Can this be done inside the observarbility service

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+
+// InitializeWithConfig initializes the observability service with custom configuration.
+// Must be called before GetService to take effect.
+// Returns error only if FailureMode is "strict", otherwise logs error and disables observability.
+func InitializeWithConfig(cfg *Config) (*Service, error) {
+	var svc *Service
+
+	serviceOnce.Do(func() {
+		svc = newServiceWithConfig(cfg)
+		serviceInstance = svc
```
rajithacharith, line 84:
```suggestion
		serviceInstance = newServiceWithConfig(cfg)
```

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+	})
+
+	if serviceInstance == nil {
+		return nil, errors.New("failed to initialize observability service")
+	}
```
rajithacharith, line 89: Can't we remove this and the initialization part in the GetService() method as well?

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+
+	return serviceInstance, nil
+}
+
+// NewService creates and initializes a new observability service with the given configuration.
+func NewService(cfg *Config) *Service {
```
rajithacharith:
```suggestion
func newService(cfg *Config) *Service {
```
Let's have only one exported entrypoint

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+
+	svc := &Service{
+		publisher: pub,
+		logger:    logger,
+		config:    cfg,
+		enabled:   true,
+	}
+
+	// Initialize default subscribers
+	svc.initializeDefaultSubscribers()
```
rajithacharith: Shouldn't this be initializeSubscribers ?
```suggestion
	svc.initializeSubscribers()
```

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+	logger.Info("Observability service initialized successfully")
+	return svc
+}
+
+// initializeDefaultSubscribers sets up the default subscribers for analytics events.
+func (s *Service) initializeDefaultSubscribers() {
+	// Create formatter based on config
+	var fmtr formatter.Formatter
+	switch s.config.Output.Format {
+	case "json", "":
```
rajithacharith: Shouldn't empty string goes to default?

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+		fmtr = jsonformatter.NewJSONFormatter()
+	default:
+		fmtr = jsonformatter.NewJSONFormatter()
+		s.logger.Warn("Unknown output format, using JSON",
+			log.String("format", s.config.Output.Format))
+	}
+
+	// Create adapter based on config
+	var adptr adapter.OutputAdapter
```
rajithacharith, line 153: Can there be multiple output adapters?

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+	var err error
+
+	switch s.config.Output.Type {
+	case "file", "":
```
rajithacharith: Don't we need to handle the empty string as the default (console)?

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+		// Use file path from config or default
+		analyticsFile := s.config.Output.File.Path
+		if analyticsFile == "" || analyticsFile == "${THUNDER_HOME}/logs/analytics/analytics.log" {
```
rajithacharith: Are we expecting the users to set the file path in absolute manner with ${THUNDER_HOME}?

backend/internal/observability/service.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	s.publisher.Publish(evt)
+
+	if s.logger.IsDebugEnabled() {
+		s.logger.Debug("Event published",
+			log.String("eventType", evt.Type),
+			log.String("eventID", evt.EventID),
+			log.String("traceID", evt.TraceID))
+	}
```
rajithacharith: No need to use isDebugLogEnabled

backend/internal/observability/publisher/publisher.go
```diff
@@ -0,0 +1,277 @@
...
+// This is a simple synchronous event bus - no queuing, no async processing.
+// Events are published directly to subscribers who can decide whether to process them.
+type categoryEventPublisher struct {
+	subscribers           map[string]subscriber.Subscriber // subscriberID -> subscriber
+	subscribersByCategory map[event.EventCategory][]string // category -> []subscriberIDs
+	mu                    sync.RWMutex
+	isShutdown            bool
+}
+
+const publisherLogComponent = "CategoryEventBus"
```
rajithacharith: Does logComponent value needs to be changed to identify easily as this is coming from publisher?

backend/internal/observability/publisher/publisher.go
```diff
@@ -0,0 +1,277 @@
...
+func (p *categoryEventPublisher) Shutdown() {
+	p.mu.Lock()
+	if p.isShutdown {
+		p.mu.Unlock()
+		return
+	}
+	p.isShutdown = true
+
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, publisherLogComponent))
+	logger.Info("Shutting down event bus")
```
rajithacharith: Are we keeping the event bus references?

backend/internal/observability/publisher/publisher.go
```diff
@@ -0,0 +1,277 @@
...
+
+	// Get all subscribers
+	subscribers := make([]subscriber.Subscriber, 0, len(p.subscribers))
+	for _, sub := range p.subscribers {
+		subscribers = append(subscribers, sub)
+	}
+	p.mu.Unlock()
```
rajithacharith, line 289: This is used to copy the p.subscribers into another slice AFAIU?

Can't we change like this 

```suggestion
	p.mu.Lock()
    subscribers := make([]subscriber.Subscriber, len(p.subscribers))
    copy(subscribers, p.subscribers)
    p.mu.Unlock()
```

backend/internal/observability/subscriber/defaultsubscriber/default_subscriber.go
```diff
@@ -0,0 +1,247 @@
...
+	adapter    adapter.OutputAdapter
+	logger     *log.Logger
+	mu         sync.RWMutex
+}
+
+var _ subscriber.Subscriber = (*DefaultSubscriber)(nil)
+
+// NewDefaultSubscriber creates a new default subscriber with the given formatter and adapter.
+// By default, it subscribes to all categories.
+func NewDefaultSubscriber(fmt formatter.Formatter, adp adapter.OutputAdapter) *DefaultSubscriber {
```
rajithacharith:
```suggestion
func NewDefaultSubscriber(fmt formatter.Formatter, adp adapter.OutputAdapter) *subscriber.Subscriber {
```

backend/internal/observability/subscriber/defaultsubscriber/default_subscriber.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// SetCategories updates the categories this subscriber is interested in.
+func (ds *DefaultSubscriber) SetCategories(categories ...event.EventCategory) {
+	ds.mu.Lock()
+	defer ds.mu.Unlock()
+	ds.categories = categories
+}
+
+// AddEventTypeFilter adds an event type to the filter list.
```
rajithacharith: The method name does not descriptive enough for me. Feel like this can be improved

backend/internal/observability/subscriber/defaultsubscriber/default_subscriber.go
```diff
@@ -0,0 +1,247 @@
...
+	if len(ds.eventTypes) > 0 {
+		return ds.eventTypes[event.EventType(evt.Type)]
+	}
+
+	// If subscribed to all categories, process everything
+	for _, category := range ds.categories {
+		if category == event.CategoryAll {
+			return true
+		}
+	}
```
rajithacharith: This can be improved 
```suggestion
	if slices.Contains(ds.categories, event.CategoryAll) {
	    return true
	}
```

backend/internal/system/log/accesslog.go
```diff
@@ -42,9 +48,10 @@ func AccessLogHandler(logger *Logger, next http.Handler) http.Handler {
 			host = r.RemoteAddr
 		}
 
-		// Apache CLF-style format with additional fields. Last %d is the response time in milliseconds.
+		// Apache CLF-style format with correlation ID and response time
```
rajithacharith, line 51: [Discusstion] What is the place for correlationId in access logs? is this the correct place?

backend/internal/observability/adapter/opensearch/opensearch_adapter.go
```diff
@@ -0,0 +1,228 @@
...
+			"OpenSearch adapter initialized",
+			log.String("endpoint", cfg.Endpoint),
+			log.String("index", cfg.Index),
+		)
+	}
+
+	return adapter, nil
+}
+
+func selectHTTPClient(timeout time.Duration) httpclient.HTTPClientInterface {
```
rajithacharith: Is the `selectHTTPClient` correct method name?

backend/cmd/server/main.go
```diff
@@ -136,6 +141,34 @@ func initCacheManager(logger *log.Logger) {
 	cm.Init()
 }
 
+// initAnalytics initializes the analytics service with console adapter and JSON format.
+func initAnalytics(logger *log.Logger) {
+	// Configure analytics to use console adapter with JSON format
+	analyticsCfg := &observability.Config{
```
rajithacharith: Are we planning to get this from deployment.yaml or any other mounted config?

backend/internal/observability/publisher/publisher.go
```diff
@@ -0,0 +1,275 @@
...
+	for _, sub := range allSubscribers {
+		go func(s subscriber.Subscriber) {
+			startTime := time.Now()
+			defer func() {
+				if r := recover(); r != nil {
+					metrics.GetMetrics().IncrementSubscriberErrors()
+					logger.Error("Subscriber panicked while handling event",
+						log.String("subscriberID", s.GetID()),
+						log.Any("panic", r))
+				}
```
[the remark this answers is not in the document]
sahandilshan: Valid suggestion, will fix

backend/internal/observability/publisher/publisher.go
```diff
@@ -0,0 +1,275 @@
...
+		return
+	}
+
+	// Get ALL subscribers (they will filter themselves)
+	allSubscribers := make([]subscriber.Subscriber, 0, len(p.subscribers))
+	for _, sub := range p.subscribers {
+		allSubscribers = append(allSubscribers, sub)
+	}
+	p.mu.RUnlock()
+
```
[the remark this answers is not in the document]
sahandilshan: Not a valid suggestion

backend/cmd/server/main.go
sahandilshan: As of now we are planning to get this from deployment.yaml. But in future we have to read this from a different place. Probably let the admin enable disable this during the runtime

backend/internal/system/log/accesslog.go
sahandilshan: Do you mean to change the place where the co relation id is added?

backend/internal/observability/service.go
sahandilshan: We don't need to call the `InitializeWithConfig` before calling the getService. GetService will check if we need to load the config from the deployment.yaml or read by default

backend/cmd/server/main.go
sahandilshan: Having this here is make sense, right? Because this is getting initialised here.

backend/internal/observability/service.go
sahandilshan: Why do we need to remove this?

backend/internal/observability/service.go
sahandilshan: Per given subscriber, there can be only one adapter

backend/internal/observability/service.go
sahandilshan: Yes

backend/internal/observability/publisher/publisher.go
sahandilshan: Need to rename this. We should use some generic naming for this

PR #624, opened 2025-10-30
Initiate login flow from oauth before redirecting to gate

senthalan:
This pull request introduces a new flow initialization mechanism and integrates it into the OAuth authorization process. The changes allow the OAuth authorization handler to pre-initialize a flow with runtime data and store the resulting `flowId` in the session, improving extensibility for future authentication flows. The update also includes dependency injection improvements, new error handling, and test adjustments.

  Architecture Flow

  ```
  ┌─────────────────────────────────────────────────────────────┐
  │ OAuth Authorization Request                                 │
  │ GET /oauth2/authorize?client_id=X&scope=openid profile      │
  └──────────────────┬──────────────────────────────────────────┘
                     │
                     ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ OAuth Handler: handleInitialAuthorizationRequest()          │
  │ • Validates OAuth parameters                                │
  │ • Initiates flow: flowExecService.InitiateFlow({            │
  │     ApplicationID: appID,                                   │
  │     FlowType: "AUTHENTICATION",                             │
  │     RuntimeData: {"oauth_scope": "openid profile"}          │
  │   })                                                        │
  │ • Stores: SessionData{OAuthParams: ...}                     │
  └──────────────────┬──────────────────────────────────────────┘
                     │
                     ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Flow Engine: InitiateFlow()                                 │
  │ • Generates UUID for flowID                                 │
  │ • Creates EngineContext with RuntimeData                    │
  │ • Stores context to DB (CurrentNode = nil)                  │
  │ • Returns flowID                                            │
  └──────────────────┬──────────────────────────────────────────┘
                     │
                     ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Redirect to Gate App                                        │
  │ ?sessionDataKey=xxx                                         │
  │ &flowId=flow-456          ← Changed parameter               │
  └──────────────────┬──────────────────────────────────────────┘
                     │
                     ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Gate App                                                    │
  │ • Can use flowId to call POST /flow/execute directly        │
  └──────────────────┬──────────────────────────────────────────┘
                     │
                     ▼
  ┌─────────────────────────────────────────────────────────────┐
  │ Flow Execution: POST /flow/execute                          │
  │ • Loads context from DB using flowId                        │
  │ • Detects CurrentNode == nil → starts from beginning        │
  │ • OAuth scope available in RuntimeData["oauth_scope"]       │
  └─────────────────────────────────────────────────────────────┘

```

Depends on https://github.com/asgardeo/javascript/pull/229 for the Gate App to use the new flowId param

**Breaking change is for the deployments which hosts the Gate Application hosted outside(Currently, the oauth2/authorize will redirect the Gate with the applicationId, with this change the redirection from the oauth2/authorize will redirect the Gate with the flowId). In those deployments, the hosted applications `asgardeo/react` sdk needs to be updated to `0.5.33`**.

### DIFF
--- a/frontend/pnpm-lock.yaml
+++ b/frontend/pnpm-lock.yaml
@@ -7,8 +7,8 @@ settings:
 catalogs:
   default:
     '@asgardeo/react':
-      specifier: 0.5.32
-      version: 0.5.32
+      specifier: 0.5.33
+      version: 0.5.33
     '@asgardeo/react-router':
       specifier: 0.0.25
       version: 0.0.25
@@ -166,10 +166,10 @@ importers:
     dependencies:
       '@asgardeo/react':
         specifier: 'catalog:'
-        version: 0.5.32(@types/react@19.2.2)(react@19.2.0)
+        version: 0.5.33(@types/react@19.2.2)(react@19.2.0)
       '@asgardeo/react-router':
         specifier: 'catalog:'
-        version: 0.0.25(@asgardeo/react@0.5.32(@types/react@19.2.2)(react@19.2.0))(react-router@7.9.4(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(react@19.2.0)
+        version: 0.0.25(@asgardeo/react@0.5.33(@types/react@19.2.2)(react@19.2.0))(react-router@7.9.4(react-dom@19.2.0(react@19.2.0))(react@19.2.0))(react@19.2.0)
       '@emotion/react':
         specifier: 'catalog:'
         version: 11.14.0(@types/react@19.2.2)(react@19.2.0)
@@ -281,7 +281,7 @@ importers:
     dependencies:
       '@asgardeo/react':
         specifier: 'catalog:'
-        version: 0.5.32(@types/react@19.2.2)(react@19.2.0)
+        version: 0.5.33(@types/react@19.2.2)(react@19.2.0)
       '@emotion/react':
         specifier: 'catalog:'
         version: 11.14.0(@types/react@19.2.2)(react@19.2.0)
@@ -652,8 +652,8 @@ packages:
       react: '>=16.8.0'
       react-router: '>=6.0.0'
 
-  '@asgardeo/react@0.5.32':
-    resolution: {integrity: sha512-a53v/65F/WXKWxGBnGtJ972dUBcmT2odF/w+1cNwN+89aP/qk9mMOHR4iT3RX+1fcolQBEbpEKMcJ1Mgl0wvtg==}
+  '@asgardeo/react@0.5.33':
+    resolution: {integrity: sha512-B6ul8V+L9fzIDQ4nZ8fvEH4DvFRCOiBqEJ12v8KM2Cw7QFyTkmu9vGIjGeplsgN6a/pk7TtkjVsOxVxEzyhF3Q==}
     peerDependencies:
       '@types/react': '>=16.8.0'
       react: '>=16.8.0'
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: backend/internal/flow/common/constants/errorconstants.go [hunk 1]
```diff
@@ -252,3 +252,19 @@ var ErrorRetrievingContextInStore = serviceerror.ServiceError{
 	Error:            "Something went wrong",
 	ErrorDescription: "Error retrieving flow context from the store",
 }
+
+// ErrorInvalidFlowInitContext defines the error response for invalid flow init context.
+var ErrorInvalidFlowInitContext = serviceerror.ServiceError{
```
ThaminduDilshan: Shall we move this to line 82 where client errors are defined?

backend/internal/flow/flowexec/service.go
```diff
@@ -36,6 +36,7 @@ import (
 type FlowExecServiceInterface interface {
 	Execute(appID, flowID, actionID, flowType string, inputData map[string]string) (
 		*model.FlowStep, *serviceerror.ServiceError)
+	InitiateFlow(initContext *model.FlowInitContext) (string, *serviceerror.ServiceError)
```
ThaminduDilshan, line 39: Having a separate function could be misleading for a outside component which uses this flow execution service right?
In a normal flow execution, we should call `Execute()` function to both initialize and continue an existing flow. However with `InitiateFlow()`, someone could confuse which function to call when calling the service. Ideally this should only be called from an internal component.

frontend/pnpm-lock.yaml
```diff
@@ -7,8 +7,8 @@ settings:
 catalogs:
   default:
     '@asgardeo/react':
-      specifier: 0.5.32
```
ThaminduDilshan, line 10: Shall we revert these changes as it's not related to this PR?

frontend/pnpm-workspace.yaml
```diff
@@ -3,7 +3,7 @@ packages:
   - packages/*
 
 catalog:
-  '@asgardeo/react': 0.5.32
```
ThaminduDilshan, line 6: Shall we revert these changes as it's not related to this PR?

frontend/pnpm-lock.yaml
senthalan: This is to update the SDK change which would allow the gate app to work with the already initiated flowID. Now it only works with the applicationId

backend/internal/flow/flowexec/service.go
senthalan: As discussed offline, we will be refactoring later

backend/internal/oauth/oauth2/authz/handler_test.go
```diff
@@ -271,3 +278,94 @@ func (suite *AuthorizeHandlerTestSuite) TestGetErrorPageRedirectURL_Success() {
...
+	// Create a valid OAuth application with proper grant types and response types
+	app := suite.createTestOAuthApp()
+	suite.mockAppService.EXPECT().GetOAuthApplication("test-client-id").Return(app, nil)
+
+	// Mock flow exec service to return an error
+	expectedFlowInitCtx := &flowmodel.FlowInitContext{
+		ApplicationID: "test-app-id",
+		FlowType:      string(flowconstants.FlowTypeAuthentication),
+	}
+	mockError := &flowconstants.ErrorUpdatingContextInStore
```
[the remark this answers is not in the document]
senthalan: The InitiateFlow return is a pointer ServiceError. So when mocking we need to create a pointer	
`InitiateFlow(initContext *model.FlowInitContext) (string, *serviceerror.ServiceError)`

ThaminduDilshan (approved):

backend/internal/flow/flowexec/service.go
darshanasbg: Lets create an issue and track.. I would like to execute\initialize the flow, without needing the application id.

Issue #513, opened 2025-10-14
Implement Resource Model 


senthalan:
Resource Hierarchy Management
- CRUD APIs for ResourceGroups, Resources, and Actions

PR #666, opened 2025-11-04
Add Resource API definition and Implementation

senthalan:
### Purpose

Initial Implementation of https://github.com/asgardeo/thunder/issues/513

This pull request introduces a new resource management module to the backend, adding support for managing resource servers, resources, and actions via new APIs and database tables. It also integrates the resource service into the server initialization and adds error handling specific to resource management.

The most important changes are:

**Core Entities & Relationships**
```
ResourceServer (top-level container)
├── id (UUID, system-generated)
├── name (UPDATABLE, UNIQUE within system) 
├── description (updatable)
├── identifier (UPDATABLE, UNIQUE within system, optional) → user-provided for authorization requests) 
├── ouId (organization unit reference, updatable)
├── Actions (0..n) - can exist at server level
│   ├── id (UUID,  system-generated)
│   ├── name (updatable)
│   ├── handle (IMMUTABLE after creation, UNIQUE within resource server) → used for permission derivation
│   └── description (updatable)
└── Resources (0..n)
    ├── id (UUID, system-generated)
    ├── name (updatable)
    ├── handle (IMMUTABLE after creation, UNIQUE within parent(resource server or resource)) → used for permission derivation
    ├── description (updatable)
    ├── parent (IMMUTABLE after creation. UUID reference, nullable for top-level)
    ├── Sub-Resources (0..n, nested recursively via parent reference)
    └── Actions (0..n)
        ├── id (UUID,  system-generated)
        ├── name (updatable)
        ├── handle (IMMUTABLE after creation, UNIQUE within resource) → used for permission derivation
        └── description (updatable)
```

**Database Schema Additions:**

* Added new tables (`RESOURCE_SERVER`, `RESOURCE`, and `ACTION`) and associated unique indexes to both `postgres.sql` and `sqlite.sql` scripts to support resource server, resource, and action management, including constraints for uniqueness and referential integrity. [[1]](diffhunk://#diff-c01456b5ea7b7cef68984dc1b755d9e255a28f488aa57ca3bdbe641206263983R138-R223) [[2]](diffhunk://#diff-478760829090a37dcf6fc4eed2791034ee3280fc878d05afbbafc26caec0f480R140-R225)

**Resource Management Service Implementation:**

* Introduced the `internal/resource` package with initialization logic (`init.go`) that registers HTTP routes for resource server, resource, and action CRUD operations, including CORS middleware setup.
* Added a set of error constants for resource management operations in `error_constants.go` to standardize error handling and responses for resource-related endpoints.

**Integration and Service Registration:**

* Registered the new resource service in the main server initialization flow (`servicemanager.go`), ensuring it is available as part of the backend API. [[1]](diffhunk://#diff-36914807f52c25f79a7d09a3a5cca10a8adb6df10e3df582e89ea4c713143216R39) [[2]](diffhunk://#diff-36914807f52c25f79a7d09a3a5cca10a8adb6df10e3df582e89ea4c713143216R57-R59)

**Testing and Mocking Support:**

* Updated the `.mockery.private.yml` configuration to enable mock generation for interfaces in the new `internal/resource` package, supporting future unit testing.

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks


api/resource.yaml
```diff
@@ -0,0 +1,2244 @@
...
+          type: string
+          description: Optional identifier for authorization requests
+        ouId:
+          type: string
+          format: uuid
+          description: ID of the organization unit this resource server belongs to
+
+    UpdateResourceServerRequest:
+      type: object
+      required: [name, ouId]
```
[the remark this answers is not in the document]
senthalan: No, ouId is also mandatory. ResourceServer should exist within an OU

api/resource.yaml
```diff
@@ -0,0 +1,2244 @@
...
+          description: Optional identifier for authorization requests
+        parent:
+          type: string
+          format: uuid
+          nullable: true
+          description: Parent resource ID (null for top-level resources)
+
+    UpdateResourceRequest:
+      type: object
+      required: [name]
```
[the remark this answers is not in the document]
senthalan: There can be a Resource without description, identifier, parent

api/resource.yaml
```diff
@@ -0,0 +1,2244 @@
...
+        description:
+          type: string
+          description: Optional description of the action
+        handler:
+          type: string
+          description: Handler used for permission derivation
+
+    UpdateActionRequest:
+      type: object
+      required: [name]
```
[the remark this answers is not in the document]
senthalan: There can be an action with description

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
+openapi: 3.0.3
+info:
+  title: Resource Server Management API
+  version: "1.0"
+  description: |
+    This API is used to manage resource servers, resources, and actions for permission-based access control.
+    Resources can be nested hierarchically, and permissions are derived from the name-based path structure.
+    Resource and Action names are immutable after creation and used directly for permission derivation.
```
ThaminduDilshan: Do we need to make Resource/ Action name immutable? handle is the one being immutable right?

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+                    message: "Invalid offset parameter"
+                    description: "The offset parameter must be a non-negative integer"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "RES-5000"
```
ThaminduDilshan: Let's use the new server level InternalServerError struct instead of package specific error codes.
We should eventually get rid of other existing package level server errors as well.

Check other places and the implementation as well.

Ref: https://github.com/asgardeo/thunder/blob/4ed39980c0ed3efb93ee56cb51efa69da9927920/backend/internal/system/error/serviceerror/error.go#L57

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+      properties:
+        name:
+          type: string
+          description: Name of the resource server (used for permission derivation)
+        description:
+          type: string
+          description: Optional description of the resource server
+        identifier:
+          type: string
+          description: Optional identifier for authorization requests
```
ThaminduDilshan: If this supposed to be a uri, shall we update the description?

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+            type: string
+            format: uuid
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/UpdateResourceServerRequest'
+            example:
+              name: "booking-updated"
```
ThaminduDilshan: Should this resource server name needs to be unique as it's used during permission derivation?
Anyhow resource server identifier needs to be unique right?
If so, we need to have `409` response code also for the update request

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+                    code: "RES-1002"
+                    message: "Invalid request format"
+                    description: "Resource server ID is required"
+                cannot-delete:
+                  summary: Cannot delete resource server
+                  value:
+                    code: "RES-1006"
+                    message: "Cannot delete resource server"
+                    description: "Cannot delete resource server that has resources or actions"
+        "404":
```
ThaminduDilshan: Let's remove 404 error code as we're sticking to returning 204 when the resource doesn't exists.
This is the pattern we've been following throughout other APIs also.

Check other places as well

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+
+  /resource-servers/{rsId}/resources:
+    get:
+      tags:
+        - resources
+      summary: List resources in resource server
+      description: |
+        Returns a paginated list of resources. Use the parentId parameter to filter:
+        - No parentId: returns all resources (flat list)
+        - parentId=null or empty: returns only top-level resources
```
ThaminduDilshan: Shouldn't both of these (no parent id and parent id=nil) behave in the same manner? Feels not nice to change the behaviour based on a nil/ empty parameter.

Do we actually have a requirement to return all the resources including nested ones with this endpoint?

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+                    totalResults: 10
+                    startIndex: 1
+                    count: 3
+                    resources:
+                      - id: "7a4b1f8e-5c69-4b60-9232-2b0aaf65ef3c"
+                        name: "Reservations"
+                        description: "Reservation management"
+                        handler: "reservations"
+                        parent: null
+                        resourceServerId: "3fa85f64-5717-4562-b3fc-2c963f66afa6"
```
ThaminduDilshan: This feels redundant as it's already available as a path parameter. But fine with either

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+          application/json:
+            schema:
+              $ref: '#/components/schemas/CreateResourceRequest'
+            examples:
+              top-level-resource:
+                summary: Top-level resource
+                value:
+                  name: "Reservations"
+                  description: "Reservation management"
+                  handler: "reservations"
```
ThaminduDilshan: Should this be `handler` or `handle`? We're using `handle` in the ou APIs. Better to stick to one.

@darshanasbg WDYT?

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+                totalResults: 5
+                startIndex: 1
+                count: 2
+                actions:
+                  - id: "9c6d3g0g-7e8b-4d82-b454-4d2ccg87gh5e"
+                    name: "Create Booking"
+                    description: "Create a new booking"
+                    handler: "create"
+                    resourceServerId: "3fa85f64-5717-4562-b3fc-2c963f66afa6"
+                    resourceId: null
```
ThaminduDilshan: Shouldn't this be a list? If I'm not mistaken, Actions reside in the Resource Server and can be associated with 1 or more Resources. That means the same Action can be associated with multiple Resources right?

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Action'
+              example:
+                id: "9c6d3g0g-7e8b-4d82-b454-4d2ccg87gh5e"
+                name: "Create Booking"
+                description: "Create a new booking"
+                handler: "create"
+                resourceServerId: "3fa85f64-5717-4562-b3fc-2c963f66afa6"
+                resourceId: null
```
ThaminduDilshan: There's no point of returning the resourceId here if we're not allowing to set it while creating an action

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "RES-5000"
+                message: "Internal server error"
+                description: "An unexpected error occurred while processing the request"
+
+  /resource-servers/{rsId}/resources/{resourceId}/actions/{id}:
```
ThaminduDilshan, line 1375: I'm not sure whether we should allow creating/ updating/ deleting actions at this level. Actions actually reside in Resource Server and has only associations with the Resources.

1. Creating a Action under Resource gave me the impression that I'm creating an Action there. But actually this will be created in the resource server and will be accessible to other Resources as well. However from a UI/UX point of view, this is a valid scenario. Hence fine to have create here.
2. IMO update and delete at this level should be only for updating/ deleting the association of the Resource to the Action (Can be merged to a new endpoint). Not to update/ delete the actual Action.

However agreed that we should have GET action at this level.

WDYT @darshanasbg @senthalan

api/resource.yaml
```diff
@@ -0,0 +1,2200 @@
...
+                description: "An unexpected error occurred while processing the request"
+
+  /resource-servers/{id}/permissions:
+    get:
+      tags:
+        - permissions
+      summary: Get all derived permissions for resource server
+      description: |
+        Returns all derived permissions for the resource server, including:
+        - The resource server itself
```
ThaminduDilshan: Feels these bullet point are unnecessary

api/resource.yaml
ThaminduDilshan: As per the offline discussion, got to know that you have removed the handle from the implementation. However I feel like it's a valid used case to have a display name that can be modified for UI display purposes.

From a UI/UX perspective, UI can try to derive the handle based on the display name (replace spaces with underscore/ remove special chars, etc). Developer should also be able to change the generated handle name during resource/ action creation

api/resource.yaml
senthalan: Actions are not shareable. Actions can be created in two levels and belong  where they are created.
1. Actions can be in the resource server level - managed with `/resource-servers/{rsId}/actions`
2. Actions can be in a resource level - managed with `/resource-servers/{rsId}/resources/{resourceId}/actions`

ThaminduDilshan: This is due to a confusion I had with the model definition. Can be resolved

api/resource.yaml
ThaminduDilshan: Will be invalidated with the model clarification. Hence resolving

api/resource.yaml
ThaminduDilshan: The original comment will get invalidated with the model clarification. 

However returning both `resourceServerId` and `resourceId` in the response could be confusing. Only one of these will be applicable for a given Action right? 
What if we have a type to indicate whether it's a Action under a resource server or a resource and return the id in a field with the same key (Just an idea)

api/resource.yaml
senthalan: Updated to include the handler for resource and actions and made them immutable

api/resource.yaml
senthalan: Updated to include 409 in the definition

api/resource.yaml
senthalan: removed 404 for all the delete

api/resource.yaml
senthalan: settled to use `handle`

api/resource.yaml
senthalan: removed resourceServerId from the responses, 

but when we get actions of resource server and resource can't be retrieved in one response. 
When getting the actions at the,
- resource server level resourceId field won't be there. 
- resource level resourceId field will be available,

backend/dbscripts/thunderdb/sqlite.sql
```diff
@@ -137,3 +137,89 @@ CREATE TABLE CERTIFICATE (
...
+    DESCRIPTION TEXT,
+    IDENTIFIER VARCHAR(100),
+    PROPERTIES TEXT,
+    CREATED_AT TEXT DEFAULT (datetime('now')),
+    UPDATED_AT TEXT DEFAULT (datetime('now'))
+);
+
+-- Unique constraint: Resource server name must be unique within an OU
+CREATE UNIQUE INDEX uq_resource_server_ou_name
+    ON RESOURCE_SERVER(OU_ID, NAME);
```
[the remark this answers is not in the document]
senthalan: this correct. Name needs to be unique with the OU

api/resource.yaml
senthalan: removed

backend/internal/system/utils/uuidutil.go
```diff
@@ -47,6 +49,41 @@ func GenerateUUID() string {
...
+func GenerateUUIDv7() string {
+	var uuid [16]byte
+
+	// Get current Unix timestamp in milliseconds
+	now := time.Now()
+	unixMillis := uint64(now.UnixMilli()) // #nosec G115 -- safe conversion for UUID timestamp
+
+	// Set timestamp in first 48 bits (6 bytes)
+	binary.BigEndian.PutUint32(uuid[0:4], uint32(unixMillis>>16)) // #nosec G115 -- safe conversion for UUID timestamp
+	binary.BigEndian.PutUint16(uuid[4:6], uint16(unixMillis))     // #nosec G115 -- safe conversion for UUID timestamp
```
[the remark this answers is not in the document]
senthalan: Thank you for raising this concern about the UUIDv7 timestamp encoding. However, after careful analysis against RFC 9562 and concrete calculations, I can confirm that the **current implementation is correct** and the suggested change would actually introduce bugs.

## RFC 9562 Specification

According to [RFC 9562 Section 5.7](https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7), UUIDv7 encodes a **48-bit big-endian unsigned number** of the Unix Epoch timestamp in milliseconds. The specification shows the bit layout as:

```
 0                   1                   2                   3
 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|                           unix_ts_ms                          |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|          unix_ts_ms           |  ver  |       rand_a          |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

The `unix_ts_ms` field occupies **bits 0-47** (bytes 0-5) in big-endian format.

## Why Current Implementation (Shift by 16) is Correct

The current code properly encodes all 48 bits of the timestamp:

```go
binary.BigEndian.PutUint32(uuid[0:4], uint32(unixMillis>>16))  // Bits [47:16] → Bytes 0-3
binary.BigEndian.PutUint16(uuid[4:6], uint16(unixMillis))      // Bits [15:0]  → Bytes 4-5
```

### Concrete Example

Current Unix timestamp in milliseconds: `1737500000000` = `0x01947B9D7C00`

**With shift by 16 (current implementation):**
- `unixMillis >> 16` = `0x01947B9D` (32 bits) → bytes [0:3]
- `uint16(unixMillis)` = `0x7C00` (16 bits) → bytes [4:5]
- **Result:** `01 94 7B 9D 7C 00` ✅
- **All 48 bits preserved:** `0x01947B9D7C00`

**With shift by 8 (suggested change):**
- `unixMillis >> 8` = `0x01947B9D7C` (requires 33 bits)
- Cast to `uint32` truncates to 32 bits: `0x947B9D7C` (loses high byte `0x01`)
- `uint16(unixMillis)` = `0x7C00`
- **Result:** `94 7B 9D 7C 7C 00` ❌
- **Issues:**
  - Loses the most significant byte (`0x01`)
  - Creates byte duplication (`7C` appears twice)
  - Timestamp precision is lost

## Why Shift by 8 is Incorrect

Shifting by 8 instead of 16 creates bit overlap:
- High 32 bits would capture bits [39:8] of the timestamp
- Low 16 bits capture bits [15:0] of the timestamp
- This creates an 8-bit overlap (bits [15:8]) and loses the top 8 bits (bits [47:40])

The correct approach for big-endian 48-bit encoding from a 64-bit value is:
- Extract high 32 bits: `value >> 16` (bits [47:16])
- Extract low 16 bits: `value & 0xFFFF` or `uint16(value)` (bits [15:0])
- Total: Full 48 bits preserved in big-endian order

## Verification Against Reference Implementations

This approach aligns with established UUID libraries:
- The bit-shifting pattern `>>16` for 48-bit encoding is standard
- Libraries like `google/uuid` (Go) use identical logic
- Any deviation would break timestamp ordering and interoperability

## Conclusion

The current implementation at `backend/internal/system/utils/uuidutil.go:63` correctly follows RFC 9562's specification for UUIDv7 timestamp encoding. The shift by 16 bits is the mathematically correct approach to split a 48-bit value into 32-bit + 16-bit components for big-endian encoding.

**No changes are needed.** The suggested shift by 8 would break the implementation and violate the RFC 9562 specification.

## References
- RFC 9562 - Universally Unique IDentifiers (UUIDs): https://www.rfc-editor.org/rfc/rfc9562.html
- Section 5.7 - UUID Version 7: https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7

tests/integration/resources/dbscripts/thunderdb/sqlite.sql
```diff
@@ -138,6 +138,94 @@ CREATE TABLE USER_SCHEMAS (
...
+    WHERE IDENTIFIER IS NOT NULL;
+
+-- Table to store resources within resource servers.
+CREATE TABLE RESOURCE (
+    ID INTEGER PRIMARY KEY AUTOINCREMENT,
+    RESOURCE_ID VARCHAR(36) NOT NULL UNIQUE,
+    RESOURCE_SERVER_ID INTEGER NOT NULL,
+    PARENT_RESOURCE_ID INTEGER,
+    NAME VARCHAR(100) NOT NULL,
+    HANDLE VARCHAR(100) NOT NULL,
```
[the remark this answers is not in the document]
senthalan: Updated store.go:513

api/resource.yaml
```diff
@@ -0,0 +1,2113 @@
...
+          description: Bad request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "RES-1002"
+                message: "Invalid request format"
+                description: "Action ID is required"
+        "404":
```
ThaminduDilshan: Do we need 404 response code for this scenario? 

We may need to have 404 response to indicate resource server or resource not found when trying to delete an action. Then better to add it for the below two endpoints as well.

1. DELETE `/resource-servers/{rsId}/resources/{id}` <- resource server not found
2. DELETE `/resource-servers/{rsId}/actions/{id}` <- resource server not found

backend/cmd/server/servicemanager.go
```diff
@@ -53,6 +54,9 @@ func registerServices(mux *http.ServeMux, jwtService jwt.JWTServiceInterface) {
 	userSchemaService := userschema.Initialize(mux, ouService)
 	userService := user.Initialize(mux, ouService, userSchemaService)
 	groupService := group.Initialize(mux, ouService, userService)
+
+	// Initialize resource service.
```
```suggestion
```

backend/dbscripts/thunderdb/postgres.sql
```diff
@@ -135,3 +135,91 @@ CREATE TABLE CERTIFICATE (
...
+    NAME VARCHAR(100) NOT NULL,
+    DESCRIPTION TEXT,
+    IDENTIFIER VARCHAR(100),
+    PROPERTIES JSONB,
+    CREATED_AT TIMESTAMPTZ DEFAULT NOW(),
+    UPDATED_AT TIMESTAMPTZ DEFAULT NOW()
+);
+
+-- Unique constraint: Resource server name must be unique within an OU
+CREATE UNIQUE INDEX uq_resource_server_ou_name
```
ThaminduDilshan: Can't we define this constraint inside the table definition similar to the existing ones? Is there any advantage of following this pattern rather than defining it like below?

```
CREATE TABLE RESOURCE_SERVER (
    UNIQUE (OU_ID, NAME)
)
```

One difference is the index will have a known name

backend/internal/resource/error_constants.go
```diff
@@ -0,0 +1,148 @@
...
+		Code:             "RES-1014",
+		Error:            "Handle conflict",
+		ErrorDescription: "The same handle already exists within the specified resource",
+	}
+)
+
+// Internal error constants.
+var (
+	// ErrResourceServerNotFound is returned when the resource server is not found.
+	ErrResourceServerNotFound = errors.New("resource server not found")
```
ThaminduDilshan: These errors are only used internally to communicate errors between store and service layer right?. Shall we make them private?

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,719 @@
...
+		handleError(w, logger, svcErr)
+		return
+	}
+
+	w.WriteHeader(http.StatusNoContent)
+}
+
+// Helper functions
+
+func parsePaginationParams(query url.Values) (int, int, *serviceerror.ServiceError) {
```
ThaminduDilshan, line 469: Shall we add method comments to private functions as well for consistency and future references?

api/resource.yaml
senthalan: > Do we need 404 response code for this scenario? 

No, it was removed the implementation and missed to remove from the definition

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,734 @@
...
+		return
+	}
+
+	result, svcErr := h.resourceService.GetResourceServerList(limit, offset)
+	if svcErr != nil {
+		handleError(w, logger, svcErr)
+		return
+	}
+
+	response := toResourceServerListResponse(result)
```
rajithacharith, line 60: WriteSuccessResponse/WriteErrorResponse helper functions were added with https://github.com/asgardeo/thunder/pull/845. Jihan is currently working on refactoring existing handlers. Shall we align this with that approach?

backend/dbscripts/thunderdb/postgres.sql
```diff
@@ -189,3 +189,103 @@ CREATE TABLE CERTIFICATE (
...
+    NAME VARCHAR(100) NOT NULL,
+    HANDLE VARCHAR(100) NOT NULL,
+    DESCRIPTION TEXT,
+    PROPERTIES JSONB,
+    CREATED_AT TIMESTAMPTZ DEFAULT NOW(),
+    UPDATED_AT TIMESTAMPTZ DEFAULT NOW(),
+    UNIQUE (RESOURCE_ID, DEPLOYMENT_ID),
+
+    FOREIGN KEY (RESOURCE_SERVER_ID)
+        REFERENCES RESOURCE_SERVER(ID)
```
ThumulaPerera: Shouldn't these reference the UUIDs rather than int IDs ?

ThumulaPerera: When changing to UUIDs, since unique key is (RESOURCE_SERVER_ID, DEPLOYMENT_ID), we may have to update foreign key references to the same.

ThumulaPerera: This should also help get rid of some methods like `CheckResourceServerExistAndGetInternalID`, `CheckResourceExistAndGetInternalID` which do two things

backend/internal/resource/service_test.go
rajithacharith, line 1: Analyzed the tests with claude and claude suggested following test scenarios as improvements. We can validate and implement which is required.


## Overview
This document contains Priority test scenarios focusing on composite foreign keys and relationship integrity. These tests are essential for ensuring data integrity in the resource service.

---

## COMPOSITE FOREIGN KEYS

### 1. Composite Foreign Key Validation - Resources

#### 1.1 Cross-Server Resource Access
- **Test**: `TestGetResource_WrongServerID`
- **Scenario**: Resource created in server A, attempted retrieval using server B ID
- **Expected**: Should return `ErrorResourceNotFound`

#### 1.2 Cross-Server Resource Update
- **Test**: `TestUpdateResource_WrongServerID`
- **Scenario**: Attempt to update resource using wrong server ID
- **Expected**: Should return `ErrorResourceNotFound`

#### 1.3 Cross-Server Resource Deletion
- **Test**: `TestDeleteResource_WrongServerID`
- **Scenario**: Attempt to delete resource using wrong server ID
- **Expected**: Should be idempotent (return nil)

---

### 2. Composite Foreign Key Validation - Actions

#### 2.1 Cross-Resource Action Access
- **Test**: `TestGetActionAtResource_WrongResourceID`
- **Scenario**: Action created under resource X, attempted access under resource Y
- **Expected**: Should return `ErrorActionNotFound`

#### 2.2 Cross-Resource Action Update
- **Test**: `TestUpdateActionAtResource_WrongResourceID`
- **Scenario**: Attempt to update action using wrong resource parent
- **Expected**: Should return `ErrorActionNotFound`

#### 2.3 Cross-Resource Action Deletion
- **Test**: `TestDeleteActionAtResource_WrongResourceID`
- **Scenario**: Attempt to delete action using wrong resource parent
- **Expected**: Should be idempotent (return nil)

#### 2.4 Action Server Level vs Resource Level Confusion
- **Test**: `TestGetActionAtResourceServer_WhenActionBelongsToResource`
- **Scenario**: Action created at resource level, attempted retrieval at server level
- **Expected**: Should return `ErrorActionNotFound`

#### 2.5 Action Resource Level vs Server Level Confusion
- **Test**: `TestGetActionAtResource_WhenActionBelongsToServer`
- **Scenario**: Action created at server level, attempted retrieval at resource level
- **Expected**: Should return `ErrorActionNotFound`

---

### 3. Cross-Reference Validation

#### 3.1 Resource Parent from Different Server
- **Test**: `TestCreateResource_ParentFromDifferentServer`
- **Scenario**: Create resource in server A with parent from server B
- **Expected**: Should return `ErrorParentResourceNotFound`

#### 3.2 Action for Resource from Different Server
- **Test**: `TestCreateActionAtResource_ResourceFromDifferentServer`
- **Scenario**: Create action at resource that belongs to different server
- **Expected**: Should return `ErrorResourceNotFound`

#### 3.3 Complex Cross-Reference Chain
- **Test**: `TestCreateResource_ComplexCrossReference`
- **Scenario**: Resource A (server 1) → parent B (server 2) → parent C (server 1)
- **Expected**: All cross-server references should fail appropriately

---

## RELATIONSHIP INTEGRITY

### 4. Parent-Child Relationship Edge Cases

#### 4.1 Multi-Level Hierarchy Creation
- **Test**: `TestCreateResource_MultiLevelHierarchy`
- **Scenario**: Create resources with 3+ levels of nesting (root → child → grandchild)
- **Expected**: All creations should succeed

#### 4.2 Delete Parent with Children
- **Test**: `TestDeleteResource_WithChildren`
- **Scenario**: Attempt to delete parent resource when child resources exist
- **Expected**: Should return `ErrorCannotDelete`

#### 4.3 Delete Parent with Actions
- **Test**: `TestDeleteResource_WithActions`
- **Scenario**: Attempt to delete resource when actions are attached to it
- **Expected**: Should return `ErrorCannotDelete`

#### 4.4 Delete Chain (Child then Parent)
- **Test**: `TestDeleteResource_ChainDeletion`
- **Scenario**: Delete child resource first, then parent should succeed
- **Expected**: Both deletions should succeed

#### 4.5 Parent Immutability on Update
- **Test**: `TestUpdateResource_ParentImmutabilityStrict`
- **Scenario**: Verify parent field is preserved even if update request contains different parent
- **Expected**: Original parent should be maintained

#### 4.6 Parent from Different Server
- **Test**: `TestCreateResource_ParentFromDifferentServer`
- **Scenario**: Attempt to create resource with parent that belongs to different server
- **Expected**: Should return `ErrorParentResourceNotFound`

---

### 5. Handle Uniqueness Scope Tests

#### 5.1 Same Handle at Different Parent Levels
- **Test**: `TestCreateResource_SameHandleDifferentParents`
- **Scenario**: Create two resources with same handle but under different parents
- **Expected**: Both should succeed (different scope)

#### 5.2 Same Handle at Root and Child Level
- **Test**: `TestCreateResource_SameHandleRootAndChild`
- **Scenario**: Create resource "users" at root level and another "users" under a parent
- **Expected**: Both should succeed (different scope)

#### 5.3 Handle Conflict at Same Parent Level
- **Test**: `TestCreateResource_HandleConflictSameParent`
- **Scenario**: Create two resources with same handle under same parent
- **Expected**: Second creation should return `ErrorHandleConflict`

#### 5.4 Action Handle at Server Level vs Resource Level
- **Test**: `TestCreateAction_SameHandleDifferentScopes`
- **Scenario**: Create action "read" at server level, then create action "read" at resource level
- **Expected**: Both should succeed (different scope)

#### 5.5 Action Handle at Different Resources
- **Test**: `TestCreateAction_SameHandleDifferentResources`
- **Scenario**: Create action "read" at resource A, then create action "read" at resource B
- **Expected**: Both should succeed (different scope)

#### 5.6 Action Handle Conflict at Resource Level
- **Test**: `TestCreateActionAtResource_HandleConflict`
- **Scenario**: Create two actions with same handle under same resource
- **Expected**: Second creation should return `ErrorHandleConflict`

---

### 6. Resource Server Dependencies

#### 6.1 Server with Only Resources
- **Test**: `TestDeleteResourceServer_WithOnlyResources`
- **Scenario**: Resource server has resources but no actions
- **Expected**: Should return `ErrorCannotDelete`

#### 6.2 Server with Only Actions
- **Test**: `TestDeleteResourceServer_WithOnlyActions`
- **Scenario**: Resource server has actions but no resources
- **Expected**: Should return `ErrorCannotDelete`

#### 6.3 Server with Resources and Actions
- **Test**: `TestDeleteResourceServer_WithResourcesAndActions`
- **Scenario**: Resource server has both resources and actions
- **Expected**: Should return `ErrorCannotDelete`

#### 6.4 Server with Nested Resources
- **Test**: `TestDeleteResourceServer_WithNestedResources`
- **Scenario**: Resource server has parent-child resource hierarchy
- **Expected**: Should return `ErrorCannotDelete`

---

### 7. Resource Dependencies

#### 7.1 Resource with Only Children
- **Test**: `TestDeleteResource_WithOnlyChildResources`
- **Scenario**: Resource has child resources but no actions
- **Expected**: Should return `ErrorCannotDelete`

#### 7.2 Resource with Only Actions
- **Test**: `TestDeleteResource_WithOnlyActions`
- **Scenario**: Resource has actions but no child resources
- **Expected**: Should return `ErrorCannotDelete`

#### 7.3 Resource with Children and Actions
- **Test**: `TestDeleteResource_WithChildrenAndActions`
- **Scenario**: Resource has both child resources and actions
- **Expected**: Should return `ErrorCannotDelete`

---

## Summary

### Composite Foreign Keys
- **Total Tests**: 12
- **Focus**: Ensuring composite keys (server+resource, server+action) are properly validated
- **Impact**: Critical data integrity - prevents orphaned records and invalid references

###  Relationship Integrity
- **Total Tests**: 18
- **Focus**: Parent-child relationships, handle uniqueness scoping, dependency management
- **Impact**: Critical business logic - ensures hierarchical data consistency

### Total Priority Tests: 30

---

## Implementation Priority Order

1. **Composite Foreign Key Tests (1.1 - 1.3)** - Core data integrity
2. **Action Composite Key Tests (2.1 - 2.5)** - Action-resource relationship integrity
3. **Cross-Reference Tests (3.1 - 3.3)** - Cross-server validation
4. **Parent-Child Tests (4.1 - 4.6)** - Hierarchical relationship management
5. **Handle Uniqueness Tests (5.1 - 5.6)** - Scope-based uniqueness validation
6. **Dependency Tests (6.1 - 6.4, 7.1 - 7.3)** - Cascade delete prevention

---

## Why These Are Priority 1

### Data Integrity Risks
- **Orphaned Records**: Without composite key validation, actions/resources could become orphaned
- **Invalid References**: Cross-server references could create data inconsistencies
- **Broken Hierarchies**: Parent-child relationships could be corrupted

### Business Logic Risks
- **Accidental Deletions**: Without dependency checks, cascading deletes could occur
- **Duplicate Handles**: Without proper scoping, handle uniqueness could be violated
- **Data Corruption**: Invalid parent references could break the hierarchy

### Security Risks
- **Unauthorized Access**: Wrong server ID could bypass access controls
- **Data Leakage**: Accessing resources via wrong server could expose data

These tests ensure the fundamental integrity of the composite foreign key relationships that are central to the resource service architecture.

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,734 @@
...
+}
+
+// HandleResourceServerDeleteRequest handles deleting a resource server.
+func (h *resourceHandler) HandleResourceServerDeleteRequest(w http.ResponseWriter, r *http.Request) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, handlerLoggerComponentName))
+
+	id := r.PathValue("id")
+	svcErr := h.resourceService.DeleteResourceServer(id)
+	if svcErr != nil {
+		handleError(w, logger, svcErr)
```
ThaminduDilshan: Handling delete errors with the common helper method will return 404 for non existing case. Shouldn't we return 204 there?

Check other places as well

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,734 @@
...
+	w.WriteHeader(statusCode)
+
+	errResp := apierror.ErrorResponse{
+		Code:        svcErr.Code,
+		Message:     svcErr.Error,
+		Description: svcErr.ErrorDescription,
+	}
+
+	if err := json.NewEncoder(w).Encode(errResp); err != nil {
+		logger.Error("Error encoding error response", log.Error(err))
```
ThaminduDilshan: We should return `SSE-5001` for this right?

Ref: https://github.com/asgardeo/thunder/blob/3472f6ed1ca7e702f7cac57453ca37762f6c9880/backend/internal/system/error/serviceerror/error.go#L65

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,734 @@
...
+	}
+}
+
+// writeJSONResponse writes a JSON response with the given status code and data.
+func writeJSONResponse(w http.ResponseWriter, statusCode int, data interface{}, logger *log.Logger) {
+	w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJSON)
+	w.WriteHeader(statusCode)
+
+	if err := json.NewEncoder(w).Encode(data); err != nil {
+		logger.Error("Error encoding response", log.Error(err))
```
ThaminduDilshan: Same as previous comment. We should return `SSE-5001`

backend/internal/resource/handler.go
```diff
@@ -0,0 +1,734 @@
...
+	w.WriteHeader(http.StatusNoContent)
+}
+
+// Resource Handlers
+
+// HandleResourceListRequest handles listing resources.
+func (h *resourceHandler) HandleResourceListRequest(w http.ResponseWriter, r *http.Request) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, handlerLoggerComponentName))
+
+	rsID := r.PathValue("rsId")
```
ThaminduDilshan, line 146: Shall we define constants for these? Check other places as well

backend/internal/resource/handler_test.go
```diff
@@ -0,0 +1,1157 @@
...
+}
+
+// HandleActionPutAtResourceRequest Tests
+
+func (suite *HandlerTestSuite) TestHandleActionPutAtResourceRequest_Success() {
+	reqBody := UpdateActionRequest{
+		Description: "Updated description",
+	}
+
+	suite.mockService.On("UpdateActionAtResource", "rs-123", testResourceID, "action-123", mock.Anything).Return(&Action{
```
ThaminduDilshan: Line exceeds 120 characters. Shall we fix this?

backend/internal/resource/handler_test.go
```diff
@@ -0,0 +1,1157 @@
...
+	req.SetPathValue("id", "action-123")
+	w := httptest.NewRecorder()
+
+	suite.handler.HandleActionDeleteAtResourceRequest(w, req)
+
+	suite.Equal(http.StatusNoContent, w.Code)
+}
+
+func (suite *HandlerTestSuite) TestHandleActionDeleteAtResourceRequest_NotFound() {
+	suite.mockService.On("DeleteActionAtResource", "rs-123", testResourceID, "action-123").Return(&ErrorActionNotFound)
```
ThaminduDilshan: Line exceeds 120 characters. Shall we fix this? Check other places as well

backend/internal/resource/model.go
```diff
@@ -0,0 +1,186 @@
...
+	Description string `json:"description,omitempty"`
+}
+
+// UpdateActionRequest represents the request to update an action.
+type UpdateActionRequest struct {
+	Name        string `json:"name"`
+	Description string `json:"description,omitempty"`
+}
+
+// Internal service layer structs
```
ThaminduDilshan:
```suggestion
// Service layer structs
```

backend/internal/system/utils/uuidutil.go
```diff
@@ -47,6 +49,41 @@ func GenerateUUID() string {
...
+// Unix Epoch timestamp source, providing better database index locality and performance.
+func GenerateUUIDv7() string {
+	var uuid [16]byte
+
+	// Get current Unix timestamp in milliseconds
+	now := time.Now()
+	unixMillis := uint64(now.UnixMilli()) // #nosec G115 -- safe conversion for UUID timestamp
+
+	// Set timestamp in first 48 bits (6 bytes)
+	binary.BigEndian.PutUint32(uuid[0:4], uint32(unixMillis>>16)) // #nosec G115 -- safe conversion for UUID timestamp
```
ThaminduDilshan: Isn't it possible to get rid of these nosec comments with code fixes?

backend/.mockery.private.yml
```diff
@@ -196,3 +196,10 @@ packages:
       structname: '{{.InterfaceName}}Mock'
       pkgname: cert
       filename: "{{.InterfaceName}}_mock_test.go"
+  github.com/asgardeo/thunder/internal/resource:
```
ThaminduDilshan, line 200: Let's add an empty line for consistency

backend/internal/resource/store_constants.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+)
+
+// Resource Queries
+var (
+	// queryCreateResource creates a new resource.
+	queryCreateResource = dbmodel.DBQuery{
+		ID: "RSQ-RES_MGT-14",
+		Query: `INSERT INTO RESOURCE
+		        (RESOURCE_ID, RESOURCE_SERVER_ID, NAME, HANDLE, DESCRIPTION, PROPERTIES, PARENT_RESOURCE_ID, DEPLOYMENT_ID)
```
ThaminduDilshan, line 111: Check for line length. If this exceeds 120 characters, wondering why linter didn't detect it

backend/internal/resource/store_constants.go
```diff
@@ -0,0 +1,462 @@
...
+	queryCheckResourceServerIdentifierExistsExcludingID = dbmodel.DBQuery{
+		ID: "RSQ-RES_MGT-11",
+		Query: `SELECT COUNT(*) as count FROM RESOURCE_SERVER
+		        WHERE IDENTIFIER = $1 AND RESOURCE_SERVER_ID != $2 AND DEPLOYMENT_ID = $3`,
+	}
+
+	// queryCheckResourceServerHasDependencies checks if resource server has resources or actions.
+	queryCheckResourceServerHasDependencies = dbmodel.DBQuery{
+		ID: "RSQ-RES_MGT-12",
+		Query: `SELECT COUNT(*) as count FROM (
```
ThaminduDilshan, line 88: We don't need join queries here right? Both Resource and Action tables contains the resource server id

backend/internal/resource/store_constants.go
```diff
@@ -0,0 +1,462 @@
...
+		ID: "RSQ-RES_MGT-14",
+		Query: `INSERT INTO RESOURCE
+		        (RESOURCE_ID, RESOURCE_SERVER_ID, NAME, HANDLE, DESCRIPTION, PROPERTIES, PARENT_RESOURCE_ID, DEPLOYMENT_ID)
+		        VALUES ($1, $2, $3, $4, $5, $6, $7, $8)`,
+	}
+
+	// queryGetResourceByID retrieves a resource by ID.
+	queryGetResourceByID = dbmodel.DBQuery{
+		ID: "RSQ-RES_MGT-15",
+		Query: `SELECT r.RESOURCE_ID, rs.RESOURCE_SERVER_ID, r.NAME, r.HANDLE, r.DESCRIPTION,
```
ThaminduDilshan: Do we need join queries here also?

Shall we check all other queries and validate whether there's an actual requirement to have a join query in the usages?

backend/internal/resource/store_constants.go
```diff
@@ -0,0 +1,462 @@
...
+	}
+
+	// queryCheckResourceHandleExistsUnderParent checks if resource handle exists under same parent.
+	queryCheckResourceHandleExistsUnderParent = dbmodel.DBQuery{
+		ID: "RSQ-RES_MGT-25",
+		Query: `SELECT COUNT(*) as count
+		        FROM RESOURCE r
+		        JOIN RESOURCE_SERVER rs ON r.RESOURCE_SERVER_ID = rs.ID
+		        JOIN RESOURCE parent ON r.PARENT_RESOURCE_ID = parent.ID
+		        WHERE rs.RESOURCE_SERVER_ID = $1 AND r.HANDLE = $2 AND parent.RESOURCE_ID = $3 AND r.DEPLOYMENT_ID = $4`,
```
ThaminduDilshan: Seems to exceed 120 characters. Shall we check other places as well?

backend/internal/resource/store.go
```diff
@@ -0,0 +1,1043 @@
...
+	CreateResource(uuid string, resourceServerInternalID int, parentInternalID *int, res Resource) error
+	GetResource(id, resourceServerID string) (Resource, error)
+	GetResourceList(resourceServerID string, limit, offset int) ([]Resource, error)
+	GetResourceListByParent(resourceServerID string, parentID *string, limit, offset int) ([]Resource, error)
+	GetResourceListCount(resourceServerID string) (int, error)
+	GetResourceListCountByParent(resourceServerID string, parentID *string) (int, error)
+	UpdateResource(id, resourceServerID string, res Resource) error
+	DeleteResource(id, resourceServerID string) error
+	IsResourceExist(id, resourceServerID string) (bool, error)
+	CheckResourceHandleExistsUnderParent(
```
ThaminduDilshan: Should fit into a single line

backend/internal/resource/store.go
```diff
@@ -0,0 +1,1043 @@
...
+	resourceServerID string, parentID *string, limit, offset int,
+) ([]Resource, error) {
+	dbClient, err := s.getIdentityDBClient()
+	if err != nil {
+		return nil, err
+	}
+
+	var results []map[string]interface{}
+	// Treat empty string parent ID same as nil (top-level resources)
+	if parentID == nil || *parentID == "" {
```
ThaminduDilshan: Should be able to use a single query as well for this with a nullable parameter. Can utilize two helper methods to switch the parameter

backend/internal/resource/store.go
```diff
@@ -0,0 +1,1043 @@
...
+		return false, fmt.Errorf("failed to check action handle: %w", err)
+	}
+
+	return parseBoolFromCount(results)
+}
+
+// Helper methods
+
+// getIdentityDBClient retrieves the identity database client.
+func (s *resourceStore) getIdentityDBClient() (provider.DBClientInterface, error) {
```
ThaminduDilshan:
```suggestion
func (s *resourceStore) getConfigDBClient() (provider.DBClientInterface, error) {
```

backend/internal/resource/store.go
```diff
@@ -0,0 +1,1043 @@
...
+		return false, err
+	}
+	return count > 0, nil
+}
+
+// buildResourceServerFromResultRow builds a ResourceServer from a database result row.
+func buildResourceServerFromResultRow(row map[string]interface{}) (ResourceServer, error) {
+	rs := ResourceServer{}
+
+	if id, ok := row["resource_server_id"].(string); ok {
```
ThaminduDilshan, line 771: Shall we define constants for these?

backend/internal/resource/store.go
```diff
@@ -0,0 +1,1043 @@
...
+		res.Handle = handle
+	} else {
+		return res, fmt.Errorf("handle field is missing or invalid")
+	}
+
+	if desc, ok := row["description"].(string); ok {
+		res.Description = desc
+	}
+
+	// PROPERTIES column exists in DB but not mapped to model (store as empty JSON)
```
ThaminduDilshan, line 826: Shall we have a TODO comment instead?

backend/internal/resource/store_test.go
```diff
@@ -0,0 +1,1621 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+//nolint:lll // Test file with long mock setup calls
```
ThaminduDilshan: This should be added to ignore the warning for 120 character line length. Rather than ignoring, shall we fix this?

Also wonder if this adds a package level exclusion which could lead to excluding line length for all files in the package. Suspecting as I spot few such occurrences

backend/dbscripts/thunderdb/postgres.sql
senthalan: Advantage of this would be 
1. Smaller and faster indexes
- INT = 4 or 8 bytes vs UUID = 16 bytes
- Indexes on foreign keys, joins, and clustered primary keys become smaller, denser, and cache-friendlier.
- Especially beneficial for write-heavy and join-heavy workloads.
2. Faster joins
- Foreign keys referencing INT columns are faster to compare and hash.
- You avoid the CPU overhead of comparing 16-byte UUIDs during joins or lookups.

So here for the primary key and foreign key references, I am using the INT.

senthalan: Advantage of this would be 
1. Smaller and faster indexes
- INT = 4 or 8 bytes vs UUID = 16 bytes
- Indexes on foreign keys, joins, and clustered primary keys become smaller, denser, and cache-friendlier.
- Especially beneficial for write-heavy and join-heavy workloads.
2. Faster joins
- Foreign keys referencing INT columns are faster to compare and hash.
- You avoid the CPU overhead of comparing 16-byte UUIDs during joins or lookups.

So here for the primary key and foreign key references, I am using the INT.

senthalan: > This should also help get rid of some methods like CheckResourceServerExistAndGetInternalID, CheckResourceExistAndGetInternalID which do two things

But they execute only one query

backend/internal/resource/handler.go
senthalan: The service layer will not return an error for the non existing cases. Idempotent delete is handled within the service layer itself. https://github.com/senthalan/thunder/blob/resource-mgt/backend/internal/resource/service.go#L295

backend/internal/resource/store.go
senthalan: This is added to cater any future requirements to store any properties. As we don't have any such property now, adding it as empty JSON. I don't think having TODO would make sense here.

backend/dbscripts/thunderdb/postgres.sql
ThumulaPerera: In that case, we should come to an agreement and update our existing schemas to use INT ids for foreign keys. WDYT ?

backend/internal/resource/handler.go
senthalan: updated

backend/internal/resource/service_test.go
senthalan: referred this and improved the tests

backend/internal/resource/service.go
```diff
@@ -0,0 +1,980 @@
...
+	GetResourceList(
+		resourceServerID string, parentID *string, limit, offset int,
+	) (*ResourceList, *serviceerror.ServiceError)
+	UpdateResource(resourceServerID, id string, res Resource) (*Resource, *serviceerror.ServiceError)
+	DeleteResource(resourceServerID, id string) *serviceerror.ServiceError
+
+	// Action operations
+	CreateAction(resourceServerID string, resourceID *string, action Action) (*Action, *serviceerror.ServiceError)
+	GetAction(resourceServerID string, resourceID *string, id string) (*Action, *serviceerror.ServiceError)
+	GetActionList(resourceServerID string, resourceID *string, limit, offset int) (*ActionList, *serviceerror.ServiceError)
```
ThaminduDilshan, line 56: Check for line length

backend/internal/resource/service.go
```diff
@@ -0,0 +1,980 @@
...
+	}
+}
+
+// Resource Server Methods
+
+// CreateResourceServer creates a new resource server.
+func (rs *resourceService) CreateResourceServer(
+	resourceServer ResourceServer,
+) (*ResourceServer, *serviceerror.ServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
```
ThaminduDilshan: Shall we define this logger inside resourceService struct so that we don't need to initiate it in every method? 
We've been following this pattern in some new components.

backend/internal/resource/service.go
senthalan: The issue here is the tab character.

>The lll (line length) linter, when used as part of golangci-lint, counts a tab character (\t) as a single character by default, not four. 

That's why it's not detected by linter checks. We can override the tab width. (We can set to 4, as it's the VS codes default tab width)
```
lll:
  line-length: 120
  tab-width: 4 # This changes the default from 1 to 4
```

I tried to change this config, it's breaking in many existing files. I will send another PR to change this configuration

PR #673, opened 2025-11-04
Refactor authentication services to use di pattern

ThaminduDilshan:
## Purpose

$subject


backend/internal/authn/oidc/service_test.go
```diff
@@ -37,8 +40,12 @@ const (
 type OIDCAuthnServiceTestSuite struct {
 	suite.Suite
 	mockOAuthService *oauthmock.OAuthAuthnServiceInterfaceMock
+	mockHTTPClient   *httpmock.HTTPClientInterfaceMock
+	mockIdpService   *idpmock.IDPServiceInterfaceMock
+	mockUserService  *usermock.UserServiceInterfaceMock
 	mockJWTService   *jwtmock.JWTServiceInterfaceMock
-	service          OIDCAuthnServiceInterface
+	endpoints        oauth.OAuthEndpoints
+	service          oidcAuthnService
```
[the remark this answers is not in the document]
ThaminduDilshan: This is to allow injecting the internal oauth service. Otherwise will have to mock lot of internal dependencies of oauth authentication service

backend/internal/authn/service_test.go
```diff
@@ -111,17 +112,18 @@ func (suite *AuthenticationServiceTestSuite) SetupSuite() {
...
-	suite.mockOIDCService = oidcmock.NewOIDCAuthnServiceInterfaceMock(suite.T())
-	suite.mockGoogleService = googlemock.NewGoogleOIDCAuthnServiceInterfaceMock(suite.T())
-	suite.mockGithubService = githubmock.NewGithubOAuthAuthnServiceInterfaceMock(suite.T())
+	suite.mockAssertGenerator = &assertmock.AuthAssertGeneratorInterfaceMock{}
+	suite.mockCredentialsService = &credentialsmock.CredentialsAuthnServiceInterfaceMock{}
+	suite.mockOTPService = &otpmock.OTPAuthnServiceInterfaceMock{}
+	suite.mockOAuthService = &oauthmock.OAuthAuthnServiceInterfaceMock{}
+	suite.mockOIDCService = &oidcmock.OIDCAuthnServiceInterfaceMock{}
+	suite.mockGoogleService = &googlemock.GoogleOIDCAuthnServiceInterfaceMock{}
+	suite.mockGithubService = &githubmock.GithubOAuthAuthnServiceInterfaceMock{}
```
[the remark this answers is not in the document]
ThaminduDilshan: It's done intentionally as we'll be removing the deprecated public constructors. Since this uses a sub package, it should be mocked this way

backend/internal/authn/oidc/service_test.go
```diff
@@ -47,8 +54,23 @@ func TestOIDCAuthnServiceTestSuite(t *testing.T) {
...
+		UserInfoEndpoint:      "https://localhost:8090/oauth/userinfo",
+	}
+
+	service := newOIDCAuthnService(suite.mockHTTPClient, suite.mockIdpService,
+		suite.mockUserService, suite.mockJWTService, suite.endpoints)
+
+	cast, ok := service.(*oidcAuthnService)
+	suite.True(ok, "service is not of type *oidcAuthnService")
+	suite.service = *cast
+	suite.service.internal = suite.mockOAuthService
```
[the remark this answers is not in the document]
ThaminduDilshan: This is done intentionally as oidc authn service shouldn't expect outside components to inject it's internal service (This is inheritance. Not injected dependencies)

backend/cmd/server/servicemanager.go
```diff
@@ -71,12 +80,24 @@ func registerServices(mux *http.ServeMux) {
...
+	oAuthAuthnService := authnoauth.Initialize(idpService, userService)
+	oidcAuthnService := authnoidc.Initialize(idpService, userService, jwtService)
+	githubAuthnService := authngithub.Initialize(idpService, userService)
+	googleAuthnService := authngoogle.Initialize(idpService, userService, jwtService)
+	_ = authn.Initialize(
+		mux,
+		idpService, jwtService,
+		assertGenerator, credentialsAuthnService, otpAuthnService,
+		oAuthAuthnService, oidcAuthnService, googleAuthnService, githubAuthnService,
+	)
```
senthalan: Rather than creating all these Service in the servicemanager level, can't we do this 

In the servicemanager.go, we will Initialize the authentication service with all the needed services
```
# servicemanager.go
authnService, executorProvider = authn.Initialize(
		mux,, userService
		idpService, jwtService
	)

....

_ := flowExec.Initialize(mux, executorProvider)
```

And we will have a authn/init.go, we will Initialize each of these authenticaiton package which will return  authenticaiton service and corresponding executor. Using that we can then create authnService and executorProvider(I am not sure whether this name) which will keep the map of executors which will be used for the flowexec
```
# authn/init.go
   ....
	oauthAuthnSvc, oAuthExectuor := authnoauth.Initialize(idpService, userService)
	oidcAuthnSvc, oidcExectuor := authnoidc.Initialize(idpService, userService, jwtService)
   ......
    authnService := newAuthenticationService(
		...
		oauthAuthnSvc,
		oidcAuthnSvc,
		....
	)
   ...
   
   executorProvider := ...

  return authnService, executorProvider
```

In the authn/oauth/init.go, we can Initialize each of authenticaiton service and corresponding executor as well and return both. 
```
# authn/oauth/init.go
 oauthAuthnSvc := newOAuthAuthnService(...)
 oauthAuthExceutor := newOAuthAuthExceutor(oauthAuthnSvc)
return oauthAuthnSvc, oauthAuthExceutor
```


So with this change there won't be need for the executor package. 

I like this approach because,
1. It keeps the main servicemanager clean, it doesn't need to know about individual authentication service
2. It's better if we can keep the authentication service and corresponding executor under the same package
3. This will give us add recoveryservice and it's executors in another package and feed into the flowExce.

ThaminduDilshan: Then isn't this introducing a cyclic dependency when we try to construct executor from the internal/authn/oauth package?

senthalan: The directory structure I am suggesting will be like this
```
authn/
    oauth2/ 
        service.go
        oauthexecutor.go
        init.go
    github/ 
        service.go
        githubexecutor.go
        init.go
   service.go
   handler.go
   init.go
```

ThaminduDilshan: Let's track this as an improvement as discussed offline. This will require refactoring the executors which will conflict with my other pending changes

ThaminduDilshan: Created an issue to track this: https://github.com/asgardeo/thunder/issues/674.

Will push the other suggestion we discussed offline

ThaminduDilshan: > Will push the other suggestion we discussed offline

Addressed with https://github.com/asgardeo/thunder/compare/22de2d58dac5aaff9a803b66f09aa5c55ca0718c..86e5674d38cb3863e03a48245ea754a10fd469cc

darshanasbg (approved):

PR #676, opened 2025-11-05
Refactor crypto package and add sign verify methods

hwupathum:
- Moved crypto package into crypto/encrypt
- Refactor crypto package to introduce CryptoAlgorithm type and improve encryption handling
- Created new crypto/sign package to include sign and verify util methods supporting following algorithms
    - RSA-SHA256 
    - RSA-SHA512 
    - ECDSA-SHA256 
    - ECDSASHA512 
    - ED25519 
- Moved hash package into crypto/hash


hwupathum:
Will send a different PR to improve code coverage

backend/internal/system/crypto/service.go
```diff
@@ -31,13 +31,19 @@ import (
 	"sync"
 
 	"github.com/asgardeo/thunder/internal/system/config"
-	"github.com/asgardeo/thunder/internal/system/hash"
+	"github.com/asgardeo/thunder/internal/system/crypto/hash"
 	"github.com/asgardeo/thunder/internal/system/log"
 )
 
+// CryptoAlgorithm represents the supported encryption algorithms
+type CryptoAlgorithm string
```
darshanasbg: Shall we move this to model.go?

darshanasbg: Along with the const values

backend/internal/system/crypto/hash/utils.go
```diff
@@ -30,21 +30,29 @@ import (
 	"github.com/asgardeo/thunder/internal/system/log"
 )
 
+// CredAlgorithm represents the supported credential hashing algorithms
+type CredAlgorithm string
```
darshanasbg: Shall we move this to the hash/model.go? Along with the constants?

backend/internal/system/crypto/encrypt/service.go
```diff
@@ -31,13 +31,19 @@ import (
 	"sync"
```
darshanasbg, line 31: Would it be better to move the file in the `crypto` package to `crypto/encrypt`, so that crypto package would have three packages: `sign`, `encrypt` & `hash`.

backend/internal/system/crypto/encrypt/model.go
```diff
@@ -16,11 +16,20 @@
  * under the License.
  */
 
-package crypto
+// Package encrypt provides cryptographic functionality with algorithm agility.
```
ThaminduDilshan: We don't need to have the package comment in all the files. Just having it in a single file is sufficient. Usually we keep it in the service file

backend/internal/system/crypto/sign/constants.go
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2025, WSO2 LLC. (http://www.wso2.com).
```
ThaminduDilshan, line 1: Shall we rename this to constants.go?

backend/internal/system/crypto/sign/utils.go
```diff
@@ -0,0 +1,190 @@
...
+var (
+	ErrUnsupportedAlgorithm = errors.New("unsupported signature algorithm")
+	ErrInvalidPrivateKey    = errors.New("invalid private key type for algorithm")
+	ErrInvalidPublicKey     = errors.New("invalid public key type for algorithm")
+	ErrInvalidSignature     = errors.New("signature verification failed")
+)
+
+// Generate takes a byte array, hashes it according to the specified algorithm,
+// and returns the digital signature using the provided private key.
+func Generate(data []byte, alg SignAlgorithm, privateKey crypto.PrivateKey) ([]byte, error) {
```
ThaminduDilshan, line 44: We may need to have a service here to enable mocking hash functionalities from the external packages.
Can be addressed with a separate effort

backend/internal/system/crypto/hash/utils.go
```diff
@@ -31,20 +31,18 @@ import (
...
-
+	defaultSaltSize         = 16
 	defaultPBKDF2Iterations = 600000
 	defaultPBKDF2KeyLength  = 32
 )
 
-// NewCredential creates a credential using the configured hash provider.
-func NewCredential(credentialValue []byte) Credential {
+// Generate creates a credential using the configured hash provider.
+func Generate(credentialValue []byte) Credential {
```
ThaminduDilshan, line 40: We may need this as a service to allow mocking these functionalities in the external packages.
Can be addressed with a separate effort

backend/internal/system/crypto/encrypt/service.go
```diff
@@ -16,8 +16,8 @@
  * under the License.
  */
 
-// Package crypto provides cryptographic functionality with algorithm agility.
-package crypto
+// Package encrypt provides cryptographic functionality with algorithm agility.
+package encrypt
```
ThaminduDilshan, line 20: With the package restructuring, should this `CryptoService` renamed to `EncryptionService`. Suggesting this because it only handles encryption/ decryption. 
Can be addressed with a separate effort

PR #717, opened 2025-11-10
Improvement to the observability package

sahandilshan:
## Purpose
With this PR, we have done the below improvements to the observability component
- Refactor the observability component
Remove redundant codes and followed the YAGNI pattern to remove unwanted features (Breaking Change)
Remove unwanted sub packages inside the observability component (Breaking Change)
Followed dependency injection for service initialization (Breaking Change)
Followed factory pattern to register subscribers (Breaking Change)
- Read the config from thunder runtime
Earlier we had dedicated configuration management for observability package. With this improvement we will prioritise only the thunder runtime configurations.
Refactored the way we define the configs. With new configs, all subscribers are defined under a key called *output*
```
observability:
  enabled: true
  output:
    console:
      enabled: true
      format: json
      categories: ["observability.all"]

    # OpenTelemetry configuration for distributed tracing
    opentelemetry:
      enabled: true
      exporter_type: "otlp"           # "otlp" or "stdout"
      otlp_endpoint: "localhost:4317" # Jaeger/Tempo collector endpoint
      service_name: "thunder-iam"
      service_version: "1.0.0"
      environment: "development"      # "production", "staging", "development"
      sample_rate: 1.0                # 1.0 = 100% sampling (recommended for dev)
      insecure: true                 # Set to true for non-TLS connections
 ```
- Publish flow related events only from engine
Without publishing events from each executor, we will only use the engine.go to publish flow related events
- Add support to publish observability to OpenTelemetry collectors

***Note: With this PR, we have changed the way that we configure a subscriber for observability

### Related Issue(s)
- https://github.com/asgardeo/thunder/issues/695

### CODE REVIEW

File: backend/internal/observability/event/category.go [hunk 1]
```diff
@@ -116,67 +78,38 @@ var eventTypeToCategory = map[EventType]EventCategory{
...
 	}
-
-	// Log warning for unmapped event type to aid debugging
-	// This should not happen in production if all event types are properly mapped
-	// Using a simple approach to avoid circular dependencies with the log package
-	// The caller (Event.GetCategory) can log this if needed
-	return CategoryUnknown
+	// Panic to catch unmapped event types during development/testing
+	panic("event type not mapped to category: " + string(eventType) + " - all event types must be explicitly mapped")
 }
```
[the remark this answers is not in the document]
sahandilshan: It's okay to have this breaking change since we haven't put this to production level yet

backend/internal/observability/event/datakeys.go
```diff
@@ -32,36 +32,24 @@ var DataKey = struct {
...
-	IPAddress      string
-	UserAgent      string
-	SessionID      string
-	SessionDataKey string
-
 	// Flow Execution Keys
-	FlowID     string
-	NodeID     string
-	RedirectTo string
+	FlowID        string
```
senthalan, line 36: Do we need to keep this `DataKey`? it won't be scalable with the number of events we support grow. Can't each package keep this as constant in that side and event datakey will only support the main set of keys

backend/go.mod
```diff
@@ -11,16 +11,35 @@ require (
 )
 
 require (
+	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
```
senthalan, line 19: why this thirdparty dependency is included ?

sahandilshan: This is being used by the opentelemetry libraries
```
internal/observability/otel
      └─> go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc
          └─> otlptracegrpc/internal/retry
              └─> github.com/cenkalti/backoff/v4
```

ThaminduDilshan: What is the requirement to use the opentelemetry library? If it's for publishing Can't we use http requests to publish data?

backend/cmd/server/repository/conf/deployment.yaml
```diff
@@ -33,8 +33,10 @@ immutable_resources:
 observability:
   enabled: false
   output:
-    type: "console"
-    format: "json"
-  metrics:
-    enabled: false
+    console:
```
ThaminduDilshan: Let's add these configs to default.json too. 
Since this is by default disabled, do we really need to keep observability configs in deployment.yaml?
We only need to keep the main/ must changed configs here.

backend/cmd/server/main.go
```diff
@@ -142,33 +145,25 @@ func initCacheManager(logger *log.Logger) {
...
 	if err != nil {
 		logger.Error("Failed to initialize observability service", log.Error(err))
 		return
 	}
 
 	if svc.IsEnabled() {
-		logger.Debug("Observability service initialized successfully with console adapter and JSON format")
+		logger.Debug("Observability service initialized successfully")
+
+		// Register subscribers explicitly after service initialization
```
ThaminduDilshan: It would be better if we can follow a similar approach to the authn service or executor registration here too. Feels we're going to handle lot of observability specific logic in the main package. Rather we can delegate this to some service/ some component in observability package and only initialize that from main. Then it can initialize whatever internal components of the observability. 

Just think about it. No need to do it in this PR

backend/internal/flow/flowexec/engine.go
```diff
@@ -454,3 +487,234 @@ func createExecutionAttempt(nodeRecord *model.NodeExecutionRecord, nodeResp *mod
 
 	return attempt
 }
+
+// publishNodeExecutionStartedEvent publishes an observability event when node execution starts.
+func publishNodeExecutionStartedEvent(ctx *model.EngineContext, node model.NodeInterface) {
+	obsSvc := observability.GetService()
+	if !obsSvc.IsEnabled() {
```
ThaminduDilshan: Hope `PublishEvent(evt)` is also handling this check and here it is only to prevent additional processing when observability is disabled.

backend/internal/flow/flowexec/engine.go
```diff
@@ -454,3 +487,234 @@ func createExecutionAttempt(nodeRecord *model.NodeExecutionRecord, nodeResp *mod
...
+	executionStartTime int64, executionEndTime int64) {
+	obsSvc := observability.GetService()
+	if !obsSvc.IsEnabled() {
+		return
+	}
+
+	// Get node execution record to determine step number and attempt
+	record := ctx.ExecutionHistory[node.GetID()]
+	if record == nil {
+		return // Should not happen, but guard against nil
```
ThaminduDilshan: Shall we remove this inline comment?

backend/internal/flow/flowexec/engine.go
```diff
@@ -454,3 +487,234 @@ func createExecutionAttempt(nodeRecord *model.NodeExecutionRecord, nodeResp *mod
...
+			eventType = event.EventTypeFlowNodeExecutionFailed
+			status = event.StatusFailure
+			nodeStatus = string(common.FlowStatusError)
+		default:
+			eventType = event.EventTypeFlowNodeExecutionCompleted
+			status = event.StatusSuccess
+			nodeStatus = string(nodeResp.Status)
+		}
+	} else {
+		// Should not happen, but handle gracefully
```
ThaminduDilshan: Let's remove this comment

backend/internal/observability/event/category.go
```diff
@@ -116,67 +78,38 @@ var eventTypeToCategory = map[EventType]EventCategory{
...
 		return category
 	}
-
-	// Log warning for unmapped event type to aid debugging
-	// This should not happen in production if all event types are properly mapped
-	// Using a simple approach to avoid circular dependencies with the log package
-	// The caller (Event.GetCategory) can log this if needed
-	return CategoryUnknown
+	// Panic to catch unmapped event types during development/testing
+	panic("event type not mapped to category: " + string(eventType) + " - all event types must be explicitly mapped")
```
ThaminduDilshan: Where do we use this function? If it's not getting called during server startup, we shouldn't use panic. This causes server to restart

backend/internal/observability/event/category_test.go
```diff
@@ -264,46 +130,73 @@ func TestEvent_GetCategory(t *testing.T) {
 	}
 }
 
+func TestGetCategory_PanicsOnUnmappedEventType(t *testing.T) {
```
ThaminduDilshan: Shall we stick to test suite pattern for consistency? Can be addressed in a separate PR

backend/internal/observability/event/constants.go
```diff
@@ -18,27 +18,18 @@
...
-	// EventTypeAuthorizationValidated is triggered after authorization request validation.
-	EventTypeAuthorizationValidated EventType = "AUTHORIZATION_VALIDATED"
-
-	// EventTypeAuthorizationRedirect is triggered when redirecting to authentication.
-	EventTypeAuthorizationRedirect EventType = "AUTHORIZATION_REDIRECT"
-
-	// EventTypeAuthorizationCodeGenerated is triggered when authorization code is created.
-	EventTypeAuthorizationCodeGenerated EventType = "AUTHORIZATION_CODE_GENERATED"
+	// ComponentFlowEngine identifies events from the flow execution engine.
+	ComponentFlowEngine = "FlowEngine"
```
ThaminduDilshan, line 25: Can't we use the same logger component name defined in respective packages?

backend/internal/observability/publisher/category_publisher_test.go
```diff
@@ -36,6 +35,14 @@ func (m *mockSubscriberV2) Close() error {
 	return nil
 }
 
+func (m *mockSubscriberV2) IsEnabled() bool {
```
ThaminduDilshan, line 38: Can't we use mockery generated mocks for this rather than defining custom mocks in tests?

backend/internal/observability/subscriber/console_subscriber.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/asgardeo/thunder/internal/observability/subscriber"
+	"github.com/asgardeo/thunder/internal/system/log"
+	"github.com/asgardeo/thunder/internal/system/utils"
+)
+
+const loggerComponentName = "ConsoleSubscriber"
+
+// ConsoleSubscriber writes observability events to the console.
+// It supports category-based filtering and JSON formatting.
+type ConsoleSubscriber struct {
```
ThaminduDilshan, line 34: Let's define these structs private and provides a private constructor method. It will use di pattern to inject it to other services.

If these are internal, we don't need to return them to outside to the package. And let's try to use a flat hierarchy to prevent exposing unnecessary details/ services/ structs/ etc to outside packages.

I think we can have all subscribers in a flat level. No need to have sub packages

backend/internal/observability/subscriber/subscriber.go
```diff
@@ -24,6 +24,7 @@ import (
 )
 
 // Subscriber is the interface that all event subscribers must implement.
+// Subscribers are now responsible for their own activation and configuration.
```
ThaminduDilshan, line 27: Shall we rename this to SubscriberInterface for consistency?

backend/internal/observability/service.go
```diff
@@ -68,19 +63,18 @@ var (
 // Uses default configuration. For custom configuration, use InitializeWithConfig first.
 func GetService() *Service {
 	serviceOnce.Do(func() {
-		serviceInstance = newService(nil)
+		serviceInstance = newService()
```
ThaminduDilshan: Constructor method should be private and should follow di pattern to inject dependencies. Also we need to have a interface defined for the services and constructor method should return a type interface

backend/go.mod
```diff
@@ -11,16 +11,35 @@ require (
...
+	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/google/uuid v1.6.0 // indirect
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.19.0 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/ncruces/go-strftime v0.1.9 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
+	go.opentelemetry.io/otel v1.24.0 // indirect
+	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.24.0 // indirect
```
ThaminduDilshan, line 32: Shall we run `go mod tidy` and see whether these gets added as direct dependencies?

backend/internal/flow/flowexec/engine.go
sahandilshan: Yes, PublishEvent is also checking this underneath inside that method

backend/cmd/server/main.go
```diff
@@ -73,8 +70,8 @@ func main() {
 		logger.Fatal("Failed to load private key", log.Error(err))
 	}
 
-	// Register the services.
-	registerServices(mux, jwtService)
+	// Register the services and get observability service for shutdown
+	observabilitySvc := registerServices(mux, jwtService)
```
senthalan: Rather than getting it back to the main.go just to shutdown. Can't we have an `unRegisterServices()` function in the servicemanager.go and call it from the main.go before calling the `gracefulShutdown`

In servicemanager.go, we can keep the observabilitySvc outside and Shutdown

backend/cmd/server/servicemanager.go
```diff
@@ -43,10 +44,18 @@ import (
 	"github.com/asgardeo/thunder/internal/userschema"
 )
 
+var ObservabilitySvc observability.ObservabilityServiceInterface
```
ThaminduDilshan: We're not assigning any value/ service to this variable right?
If this is not needed, can remove code in main.go as well

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+		eventType = event.EventTypeFlowNodeExecutionCompleted
+		status = event.StatusSuccess
+		nodeStatus = string(common.FlowStatusComplete)
+	}
+
+	// Calculate duration in milliseconds
+	durationMs := (executionEndTime - executionStartTime) * 1000
+
+	evt := event.NewEvent(
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+		WithData(event.DataKey.AttemptNumber, fmt.Sprintf("%d", attemptNumber)).
+		WithData(event.DataKey.DurationMs, fmt.Sprintf("%d", durationMs)).
+		WithData(event.DataKey.AppID, ctx.AppID)
+
+	// Set executor details if applicable (only for executor-backed nodes)
+	if node.GetType() == common.NodeTypeTaskExecution {
+		if executableNode, ok := node.(core.ExecutorBackedNodeInterface); ok {
+			executor := executableNode.GetExecutor()
+			if executor != nil {
+				record.ExecutorName = executor.GetName()
```
ThaminduDilshan: This will modify execution record right? Do we need to do this when publishing events. Or should we add this info to the event data?

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+		evt.WithData(event.DataKey.Error, nodeErr.Error).
+			WithData(event.DataKey.ErrorCode, nodeErr.Code)
+		if nodeErr.ErrorDescription != "" {
+			evt.WithData(event.DataKey.Message, nodeErr.ErrorDescription)
+		}
+	} else if nodeResp != nil && nodeResp.FailureReason != "" {
+		evt.WithData(event.DataKey.FailureReason, nodeResp.FailureReason)
+	}
+
+	// Add user ID if authenticated
```
ThaminduDilshan, line 736: Don't we need IsAuthenticated field as a data?

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+}
+
+// publishFlowStartedEvent publishes an observability event when flow execution starts.
+func publishFlowStartedEvent(ctx *EngineContext, obsSvc observability.ObservabilityServiceInterface) {
+	if obsSvc == nil || !obsSvc.IsEnabled() {
+		return
+	}
+
+	evt := event.NewEvent(
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+) {
+	if obsSvc == nil || !obsSvc.IsEnabled() {
+		return
+	}
+
+	// Calculate duration in milliseconds
+	durationMs := (flowEndTime - flowStartTime) * 1000
+
+	evt := event.NewEvent(
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/flowexec/engine.go
```diff
@@ -494,3 +535,250 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+	flowStartTime int64, flowEndTime int64, obsSvc observability.ObservabilityServiceInterface) {
+	if obsSvc == nil || !obsSvc.IsEnabled() {
+		return
+	}
+
+	// Calculate duration in milliseconds
+	durationMs := (flowEndTime - flowStartTime) * 1000
+
+	evt := event.NewEvent(
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/flowexec/engine_event_enabled_test.go
```diff
@@ -0,0 +1,492 @@
...
+	authncm "github.com/asgardeo/thunder/internal/authn/common"
+	"github.com/asgardeo/thunder/internal/flow/common"
+	"github.com/asgardeo/thunder/internal/flow/core"
+	"github.com/asgardeo/thunder/internal/system/config"
+	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+	"github.com/asgardeo/thunder/tests/mocks/observabilitymock"
+)
+
+// mockNode implements core.NodeInterface for testing
+type mockNodeForEvents struct {
```
ThaminduDilshan: Can't we use the mockery generated mocks in `tests/mocks/flow/coremock`?

backend/internal/flow/flowexec/engine_event_enabled_test.go
```diff
@@ -0,0 +1,492 @@
...
+	// Setup common expectations - allow any number of calls
+	mockObs.On("IsEnabled").Return(true).Maybe()
+	mockObs.On("PublishEvent", mock.Anything).Return().Maybe()
+	mockObs.On("Shutdown").Return().Maybe()
+
+	return mockObs
+}
+
+// TestPublishFlowStartedEvent tests the flow started event publishing
+func TestPublishFlowStartedEvent(t *testing.T) {
```
ThaminduDilshan, line 69: Shall we follow data suite pattern?

backend/internal/observability/adapter/console_adapter.go
```diff
@@ -16,15 +16,12 @@
  * under the License.
  */
 
-// Package console provides a console-based output adapter for analytics events.
-package console
+package adapter
```
ThaminduDilshan, line 19: Shall we make ConsoleAdapter struct and `NewConsoleAdapter()` private?
Same applies to the FileAdapter.

If this is called from another package in observability/ it's sub packages, we can follow a approach similar to authn package.

Sub packages have a init.go with `Initialize()` function [1], that gets called from the root `init.go` [2].

[1] https://github.com/asgardeo/thunder/blob/9990a18b51b1c23d55e5ac2aa414ee35def6888c/backend/internal/authn/credentials/init.go#L24
[2] https://github.com/asgardeo/thunder/blob/9990a18b51b1c23d55e5ac2aa414ee35def6888c/backend/internal/authn/init.go#L84

backend/internal/observability/adapter/console_adapter.go
```diff
@@ -33,7 +30,7 @@ type ConsoleAdapter struct {
 	closed bool
 }
 
-var _ adapter.OutputAdapter = (*ConsoleAdapter)(nil)
+var _ OutputAdapter = (*ConsoleAdapter)(nil)
```
ThaminduDilshan: Also for consistency, shall we rename `OutputAdapter` to `OutputAdapterInterface`?

backend/internal/observability/formatter/json_formatter.go
```diff
@@ -16,20 +16,19 @@
...
 
 import (
 	"encoding/json"
 
 	"github.com/asgardeo/thunder/internal/observability/event"
-	"github.com/asgardeo/thunder/internal/observability/formatter"
 )
 
 // JSONFormatter formats events as JSON.
 type JSONFormatter struct{}
```
ThaminduDilshan, line 29: Same as comment added to the adapter. Shall we validate it for other sub packages as well?

sahandilshan: As mentioned in that comment, I have removed all redundant sub packages

backend/cmd/server/servicemanager.go
sahandilshan: This is getting initialize inside the register service and will be used in the main.go when the server is shutdown

backend/internal/flow/flowexec/engine.go
sahandilshan: At the end of the node, I'm adding user id if the user is authenticated. I thought of that entry itself will say if the use is authenticated or not. If needed we can add whether the user is authenticated or not in a separate event key

backend/internal/flow/flowexec/engine_event_enabled_test.go
sahandilshan: Can you point to an example if possible

backend/internal/observability/adapter/console_adapter.go
sahandilshan: This is done

backend/internal/flow/flowexec/engine_event_enabled_test.go
sahandilshan: This is done

backend/internal/observability/subscriber/registry.go
```diff
@@ -0,0 +1,113 @@
...
+			log.String("subscriberType", name))
+	}
+
+	factoryRegistry[name] = factory
+}
+
+// GetAllFactories returns a copy of all registered subscriber factories.
+// This is called during service initialization to instantiate subscribers.
+// Returns a map of subscriber name -> factory function.
+func GetAllFactories() map[string]SubscriberFactory {
```
rajithacharith: Does this needs to be exported?

backend/cmd/server/main.go
```diff
@@ -288,9 +254,8 @@ func gracefulShutdown(logger *log.Logger, server *http.Server) {
 	}
 
 	// Shutdown observability service
-	observabilitySvc := observability.GetService()
-	if observabilitySvc != nil {
-		observabilitySvc.Shutdown()
+	if ObservabilitySvc != nil {
```
senthalan: without accessing ObservabilitySvc in main.go, shall we create an unregister function in servicemanager and have this logic there?

backend/internal/flow/flowexec/engine.go
```diff
@@ -39,12 +41,18 @@ type flowEngineInterface interface {
 // FlowEngine is the main engine implementation for orchestrating flow executions.
 type flowEngine struct {
 	executorRegistry executor.ExecutorRegistryInterface
+	observabilitySvc observability.ObservabilityServiceInterface
 }
 
-// GetFlowEngine returns a singleton instance of FlowEngine.
-func newFlowEngine(executorRegistry executor.ExecutorRegistryInterface) flowEngineInterface {
+// newFlowEngine creates a new flow engine with the given dependencies.
+// The observabilitySvc parameter is optional (can be nil) - if nil, observability events won't be published.
```
senthalan: do we need this comment 
```suggestion
```

backend/internal/observability/event/category.go
senthalan: This need to be fixed. we can't have panic inside code

backend/internal/observability/formatter/init.go
```diff
@@ -0,0 +1,47 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package formatter
+
+// Initialize initializes and returns a formatter based on the specified format type.
+// This is the single entry point for creating formatters, following the initialization
+// pattern used throughout Thunder (similar to authn/credentials/init.go).
```
senthalan: do we need these comments

backend/internal/observability/opentelemetry/init.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+)
+
+// Initialize initializes and returns an OpenTelemetry TracerProvider.
+// This is the single entry point for creating tracer providers, following the
+// initialization pattern used throughout Thunder (similar to authn/credentials/init.go).
```
senthalan: do we need these comments

backend/internal/observability/subscriber/registry.go
sahandilshan: This can be package private. Will do the change

backend/cmd/server/main.go
sahandilshan: Will do the changes

backend/cmd/server/servicemanager.go
```diff
@@ -45,12 +46,20 @@ import (
...
+
 // registerServices registers all the services with the provided HTTP multiplexer.
-func registerServices(mux *http.ServeMux, jwtService jwt.JWTServiceInterface) {
+func registerServices(
+	mux *http.ServeMux,
+	jwtService jwt.JWTServiceInterface,
+) {
 	logger := log.GetLogger()
 
+	observabilitySvc := observability.Initialize()
```
rajithacharith: This will not work as intended 

The := operator creates a new local variable inside the function, which shadows the package-level variable.
So the global observabilitySvc is never assigned, and remains nil.


```suggestion
	observabilitySvc = observability.Initialize()
```

backend/internal/flow/flowexec/engine.go
```diff
@@ -526,3 +566,235 @@ func createExecutionAttempt(nodeRecord *common.NodeExecutionRecord, nodeResp *co
...
+		WithData(event.DataKey.AttemptNumber, fmt.Sprintf("%d", attemptNumber)).
+		WithData(event.DataKey.AppID, ctx.AppID)
+
+	// Set executor details if applicable (only for executor-backed nodes)
+	if node.GetType() == common.NodeTypeTaskExecution {
+		if executableNode, ok := node.(core.ExecutorBackedNodeInterface); ok {
+			executor := executableNode.GetExecutor()
+			if executor != nil && record != nil {
+				record.ExecutorName = executor.GetName()
+				record.ExecutorType = executor.GetType()
```
rajithacharith, line 647: Have we used these variables?

backend/internal/observability/formatter/json_formatter.go
```diff
@@ -16,20 +16,19 @@
...
 
-var _ formatter.Formatter = (*JSONFormatter)(nil)
+var _ FormatterInterface = (*JSONFormatter)(nil)
 
 // NewJSONFormatter creates a new JSON formatter.
 func NewJSONFormatter() *JSONFormatter {
```
rajithacharith: Do we need this fn exported?

backend/cmd/server/servicemanager.go
rajithacharith: This will not work as intended. 

Reason: 
The := operator creates a new local variable inside the function, which shadows the package-level variable.
So the global observabilitySvc is never assigned, and remains nil.
```suggestion
	observabilitySvc = observability.Initialize()
```

backend/internal/flow/flowexec/engine.go
sahandilshan: We are not using these variables. These are just string representation. The real value will be showing in the event. Introduce these keys to avoid duplicate string representations

backend/internal/observability/formatter/json_formatter.go
sahandilshan: This is fixed

backend/internal/observability/event/datakeys.go
senthalan: What is our stand on this?

Are we planning to keep all the datakeys in one place?

backend/internal/observability/subscriber/console_subscriber.go
senthalan: it's better if we can fix this here and in other subscribers. 

Let's have public functions only in needed cases

backend/cmd/server/repository/conf/deployment.yaml
sahandilshan: This is fixed now

backend/internal/observability/subscriber/console_subscriber.go
sahandilshan: Hi @senthalan @ThaminduDilshan,

This is fixed now.
1. Moved everything to a flat level and keep only mandatory sub packages
2. There are no public constructors. This will get initialse from init.go files
3. Followed DI pattern to initiate observability service

backend/internal/observability/publisher/category_publisher_test.go
sahandilshan: Used mockery to generate service and other required test cases

backend/internal/observability/event/constants.go
sahandilshan: Hi @ThaminduDilshan,

Didn't get your requirement. Do you want to use the same name that have been used in logger component, yes, I have used the same name

ThaminduDilshan: NVM. I was going to ask to use the same constant defined in respective packages to avoid defining duplicate constants. If we're to do that, we need to make logger component constants public.

backend/internal/observability/event/datakeys.go
sahandilshan: Hi @senthalan,
we can follow your approaches as well where we keep the keys in each component level. The reason why moved those inside the observability package is that 
1. Mange everything from a single place
2. There may be keys that will be commonly used accross multiple components (ex: applicationId, userId) , in that case we have to define the same key on multiple components.

If needed we can follow your approaches as well

backend/internal/observability/event/category.go
sahandilshan: This is fixed with latest changes

backend/internal/observability/subscriber/init.go
```diff
@@ -0,0 +1,126 @@
...
+//   - observabilityConfig: The observability configuration from deployment.yaml
+//
+// Returns:
+//   - []SubscriberInterface: List of successfully initialized and enabled subscribers
+//   - error: Error if initialization fails in strict mode, nil otherwise
+//
+// Error Handling:
+//   - In "strict" failure mode: Returns error on first initialization failure
+//   - In "lenient" failure mode: Logs warning and continues with remaining subscribers
+func Initialize(observabilityConfig config.ObservabilityConfig) ([]SubscriberInterface, error) {
```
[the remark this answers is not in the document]
sahandilshan: Fixed now

backend/internal/observability/subscriber/otel_subscriber.go
```diff
@@ -0,0 +1,274 @@
...
+		}
+		span.SetStatus(codes.Error, errorMsg)
+		span.RecordError(fmt.Errorf("%s", errorMsg),
+			trace.WithTimestamp(evt.Timestamp))
+	} else {
+		span.SetStatus(codes.Ok, evt.Status)
+	}
+
+	// End span (instant span - starts and ends at same time)
+	span.End(trace.WithTimestamp(evt.Timestamp))
```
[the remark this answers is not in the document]
sahandilshan: This is fixed now

PR #785, opened 2025-11-17
Support password hashing with customizable params

hwupathum:
### Purpose

Currently, the parameter values for password hashing algorithms are hardcoded. This PR adds configuring capability to password hashing algorithm parameters.

---
### ⚠️ Breaking Changes

This PR introduces breaking changes that may require updates to your existing error code validations.

**1. ✂️ Changes UserCredential object**

With this changes, user credential object structure is changed by adding credential salt inside parameters.

**2. ✂️ Changes Hash Configuration**

With this changes, configuration format of hashing is changed.

---

### Approach
Following configuration will be used to modify the password hashing con figurations

```yaml
crypto:
  password_hashing:
    algorithm: PBKDF2
    parameters:
      iterations: 600000
      key_size: 32
      salt_size: 16
```

### Related Issues
- https://github.com/asgardeo/thunder/discussions/754

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/system/crypto/hash/model.go
```diff
@@ -26,11 +26,26 @@ const (
...
+
 // Credential represents the credentials of a hashed value.
 type Credential struct {
-	Algorithm CredAlgorithm
-	Hash      string
-	Salt      string
+	Algorithm  CredAlgorithm
+	Parameters CredParameters
+	Hash       string
+	Salt       string
```
darshanasbg: Ain't this can also be considered as a parameter?

If this is a parameter that all the credential hashing methods will have, we can have keep this as a first class attribute, but if this is going to be a optional one, we need to move it to the Parameters.

backend/internal/system/crypto/hash/utils.go
```diff
@@ -97,47 +107,61 @@ func verifySHA256Credential(credentialValueToVerify []byte, referenceCredential
...
+	keySize := config.GetThunderRuntime().Config.Crypto.PasswordHashing.Parameters.KeySize
+	if saltSize == 0 {
+		saltSize = defaultSaltSize
+	}
+	if iterations == 0 {
+		iterations = defaultPBKDF2Iterations
+	}
+	if keySize == 0 {
+		keySize = defaultPBKDF2KeySize
+	}
```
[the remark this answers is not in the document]
hwupathum: Negative values are handled by the following methods

backend/internal/user/model.go
```diff
@@ -34,11 +34,11 @@ type User struct {
...
-	StorageType    string             `json:"storageType"`
-	StorageAlgo    hash.CredAlgorithm `json:"storageAlgo"`
-	Value          string             `json:"value"`
-	Salt           string             `json:"salt"`
+	CredentialType    string              `json:"credentialType"`
+	StorageType       string              `json:"storageType"`
+	StorageAlgo       hash.CredAlgorithm  `json:"storageAlgo"`
+	StorageAlgoParams hash.CredParameters `json:"storageAlgoParams"`
+	Value             string              `json:"value"`
 }
```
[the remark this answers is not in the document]
hwupathum: We don't need to think about backward compatibility at this moment

backend/internal/system/crypto/hash/utils.go
```diff
@@ -97,47 +107,61 @@ func verifySHA256Credential(credentialValueToVerify []byte, referenceCredential
...
+	credSalt := generateSalt(saltSize)
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "PBKDF2HashProvider"))
-	hash, err := pbkdf2.Key(sha256.New, string(credentialValue), credSalt, defaultPBKDF2Iterations, defaultPBKDF2KeyLength)
+	hash, err := pbkdf2.Key(sha256.New, string(credentialValue), credSalt, iterations, keySize)
```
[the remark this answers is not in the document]
hwupathum: generateSalt will throw a panic if error happens. so no need to handle the error

backend/internal/system/crypto/hash/utils.go
```diff
@@ -33,21 +33,25 @@ import (
...
 	case SHA256:
 		logger.Debug("Using SHA256 hash provider as per configuration.")
 		return newSHA256Credential(credentialValue)
-	default:
+	case PBKDF2:
 		logger.Debug("Using PBKDF2 hash provider as per configuration.")
 		return newPBKDF2Credential(credentialValue)
+	default:
+		logger.Error("Unsupported hash algorithm in configuration",
+			log.String("algorithm", string(algorithm)))
```
senthalan: Shouldn't the server startup fails if there is an issue with the configuration ?

hwupathum: Added a panic and, call the Generate method with a dummy input during server starup to check for the config issues. Will need to create a proper way to check for configs in the future.

backend/internal/system/crypto/hash/service.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+type pbkdf2HashProvider struct {
+	SaltSize   int
+	Iterations int
+	KeySize    int
+}
+
+// GetHashService returns the singleton instance of the HashService
+func GetHashService() HashService {
```
senthalan: Shall we follow the Dependency Injection pattern we follow in other modules,

We have to have an init.go with Initialize which will create the service and return. And constructor (newService()) have to be private. So that it won't be called by other services directly. 

servicemanager.go will Initialize the crypto and pass the instance of the service to the user package to Initialize.
 
```
// internal/system/crypto/init.go
package application

func Initialize(mux *http.ServeMux) *Service {
    service := newService()
    return service
}


// cmd/server/servicemanager.go
package main

// registerServices registers all the services with the provided HTTP multiplexer.
func registerServices(
	mux *http.ServeMux,
	jwtService jwt.JWTServiceInterface,
) {
	...
     cryptoService := crypto.Initialize()
	userService := user.Initialize(mux, ouService, userSchemaService, cryptoService)
     ...
}

// internal/user/init.go
package user

// Initialize initializes the user service and registers its routes.
func Initialize(
	mux *http.ServeMux,
	ouService oupkg.OrganizationUnitServiceInterface,
	userSchemaService userschema.UserSchemaServiceInterface,
     cryptoService crypto.HashServiceInterface
) UserServiceInterface {
	userService := newUserService(ouService, userSchemaService, cryptoService)
	setUserService(userService) // Set the provider for backward compatibility
	userHandler := newUserHandler(userService)
	registerRoutes(mux, userHandler)
	return userService
}
```

backend/internal/system/crypto/hash/service.go
```diff
@@ -0,0 +1,196 @@
...
+)
+
+var (
+	instance HashService
+	once     sync.Once
+	logger   = log.GetLogger().With(log.String(log.LoggerKeyComponentName, "HashService"))
+)
+
+// HashService defines the interface for hashing services.
+type HashService interface {
```
senthalan:
```suggestion
type HashServiceInterface interface {
```
To align with existing naming conversions

backend/internal/system/crypto/hash/service.go
```diff
@@ -0,0 +1,194 @@
...
+		saltSize = defaultSaltSize
+	}
+	return &sha256HashProvider{
+		SaltSize: saltSize,
+	}
+}
+
+// Generate SHA256Credential generates a SHA256 hash
+func (a *sha256HashProvider) Generate(credentialValue []byte) Credential {
+	credSalt := generateSalt(a.SaltSize)
```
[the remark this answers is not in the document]
senthalan: better if we can improve this. without returning Credential{}, we should return an error from here. We can change the signature of this function. So that proper error message can be returned to the caller.

```
func (a *sha256HashProvider) Generate(credentialValue []byte) (Credential, error) {
	credSalt := generateSalt(a.SaltSize)
    if credSalt == nil {
         return nil, fmt.Errorf("Failed to generate salt")
   } 
```

As this is the already existing logic, I will approve this PR, shall we create another PR for this improvement.

backend/internal/system/crypto/hash/service.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+	}
+}
+
+// Verify SHA256Credential checks if the SHA256 hash of the input data and salt matches the expected hash.
+func (a *sha256HashProvider) Verify(credentialValueToVerify []byte, referenceCredential Credential) bool {
+	saltBytes, err := hex.DecodeString(referenceCredential.Parameters.Salt)
+	if err != nil {
+		return false
+	}
```
senthalan, line 108: Here also, it's better if we can return the error from here.

backend/internal/system/crypto/hash/service.go
```diff
@@ -0,0 +1,185 @@
...
+		KeySize:    keySize,
+	}
+}
+
+// Generate PBKDF2Credential generates a PBKDF2 hash of the input data using the provided salt.
+func (a *pbkdf2HashProvider) Generate(credentialValue []byte) Credential {
+	credSalt := generateSalt(a.SaltSize)
+	hash, err := pbkdf2.Key(sha256.New, string(credentialValue), credSalt, a.Iterations, a.KeySize)
+	if err != nil {
+		logger.Error("Error hashing data with PBKDF2: %v", log.Error(err))
```
senthalan, line 138: Better to return the error

PR #886, opened 2025-12-04
Support hybrid datastore for Organization Units

rajithacharith:
### Purpose
[Support hybrid datastore for Organization Units](https://github.com/asgardeo/thunder/commit/3b27f9d1ffb015d53152793a771ffb622187a5b2)

### DIFF
--- a/backend/internal/ou/hybrid_store.go
+++ b/backend/internal/ou/hybrid_store.go
@@ -1,0 +1,465 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package ou
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/asgardeo/thunder/internal/system/log"
+)
+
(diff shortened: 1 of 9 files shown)


backend/internal/ou/hybrid_store.go
```diff
@@ -0,0 +1,465 @@
...
+			uniqueDBCount++
+		}
+	}
+
+	return fileCount + uniqueDBCount, nil
+}
+
+// GetOrganizationUnitList retrieves root organization units from both stores with pagination.
+// File-based OUs are returned first, followed by database OUs.
+func (h *ouHybridStore) GetOrganizationUnitList(limit, offset int) ([]OrganizationUnitBasic, error) {
```
senthalan, line 90: Handling pagination is going to be complex

backend/internal/ou/hybrid_store.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	// Check name conflict in file store
+	hasConflict, err = h.fileStore.CheckOrganizationUnitNameConflict(ou.Name, ou.Parent)
+	if err != nil {
+		return fmt.Errorf("error checking name conflict in file store: %w", err)
+	}
+	if hasConflict {
+		return errors.New("cannot create organization unit: name conflicts with immutable file-based OU")
+	}
```
senthalan, line 159: shouldn't this part of the service layer?

Service layer will call `hybridStore.CheckOrganizationUnitNameConflict` and hybridStore will check from both stores

backend/internal/ou/hybrid_store.go
```diff
@@ -0,0 +1,465 @@
...
+	return h.dbStore.CheckOrganizationUnitHandleConflict(handle, parent)
+}
+
+// UpdateOrganizationUnit updates an organization unit in the database store only.
+// Returns error if the OU exists in the file store (immutable).
+func (h *ouHybridStore) UpdateOrganizationUnit(ou OrganizationUnit) error {
+	// Check if exists in file store (immutable)
+	_, err := h.fileStore.GetOrganizationUnit(ou.ID)
+	if err == nil {
+		return errors.New("cannot update organization unit: exists in immutable file-based store")
```
senthalan, line 318: this means when listing the resource or getting the resource, a flag needs to be sent to indicate whether the resource is readOnly. Else handling the UX in developer portal is going to be problem. 

In highlevel, we need think though this hybrid pattern a bit more.

Issue #919, opened 2025-12-08
Improve flow definitions to support UI translation flexibility


ThaminduDilshan:
### Current Limitation

Thunder supports executing server orchestrated flows with a defined graph structure in json format. Currently it's supported only for authentication and registration flows, however can be extended for different flows in future.
These flow definitions have the following limitations.
- Doesn't support UI layouts/ provide flexibility
- Doesn't support defining a node with both inputs and multiple options
- Only supports moving forward in the flow graphs

The current flow definitions require some changes to accommodate UI translations proposed in https://github.com/asgardeo/thunder/discussions/812

### Suggested Improvement

Following improvements can be done to the current flow definition syntax, flow execution layer and the APIs.

1. Replace `PROMPT_ONLY` node with new `PROMPT` node which allows prompting for user inputs and actions
2. Modify `PROMPT` nodes to have user selectable actions instead of generic "next" nodes
3. Introduce `onSuccess` and `onFailure` navigations to the `TASK_EXECUTION` nodes
4. Remove "next" node concept from the node definitions. Prompt nodes define it via the actions and task_execution nodes will have onSuccess/ onFailure edges
5. Improve condition evaluation for the nodes. When nodes can have multiple outgoing edges, condition evaluation needs to know which node to move next upon condition dissatisfaction. This can be supported with a `onSkip` edge defined in the condition block
6. Rename `actionId` in flow/execute API requests to `action` so that it will be aligned with inputs
7. Introduce `START` node for all the flow graphs. Currently only the registration flows have a start node

PR #920, opened 2025-12-08
Improve flow definitions and execution API request/responses

ThaminduDilshan:
### Purpose
Current server orchestrated flows have the following limitations.
- Doesn't support UI layouts/ provide flexibility
- Doesn't support defining a node with both inputs and multiple options
- Only supports moving forward in the flow graphs

This pull request improves the flow definition structure and execution API request/ response body to accommodate the UI layouts.

---
### ⚠️ Breaking Changes

This PR introduces breaking changes that require updates to the existing flow definitions and execution API usages.

**1. ✂️ Removal of `PROMPT_ONLY` node type and addition of new `PROMPT` node type**

The `PROMPT_ONLY` node type has been removed. A new `PROMPT` node type has been introduced to handle user prompts. Existing flow definitions using `PROMPT_ONLY` must be updated to use the new `PROMPT` node type.

The new `PROMPT` node type requires an `actions` field to define the next steps after user input. Here is an example of how to update a `PROMPT_ONLY` node to the new `PROMPT` format:

```json
{
    "id": "prompt_mobile",
    "type": "PROMPT",
    "inputs": [
        {
            "ref": "input_001",
            "identifier": "mobileNumber",
            "type": "string",
            "required": true
        }
    ],
    "actions": [
        {
            "ref": "action_001",
            "nextNode": "sms_otp_auth"
        }
    ]
},
```

**2. ✂️ Removal of `DECISION` node type**

The `DECISION` node type has been removed. Flow definitions that use `DECISION` nodes must be refactored to use `PROMPT` nodes with appropriate actions to handle branching logic.

Here is an example of how to refactor a `DECISION` node into a `PROMPT` node:

```json
{
  "id": "choose_auth",
  "type": "PROMPT",
  "actions": [
      {
          "ref": "basic_auth",
          "nextNode": "basic_auth"
      },
      {
          "ref": "prompt_mobile",
          "nextNode": "prompt_mobile"
      }
  ]
}
```

**3. 🔄 Changes to the node navigation**

The way nodes navigate to the next node has changed. Instead of using a single `next` field, `PROMPT` nodes now use an `actions` array to define possible transitions. Each action specifies the next node to navigate to.
`TASK_EXECUTION` nodes now use an `onSuccess` and `onFailure` field to define the next nodes based on the task execution result. Here is an example of the updated navigation structure:

```json
{
  "id": "basic_auth",
  "type": "TASK_EXECUTION",
  "inputs": [
      {
          "ref": "input_001",
          "identifier": "username",
          "type": "string",
          "required": true
      },
      {
          "ref": "input_002",
          "identifier": "password",
          "type": "string",
          "required": true
      }
  ],
  "executor": {
      "name": "BasicAuthExecutor"
  },
  "onSuccess": "attr_collector",
  "onFailure": "auth_failure"
}
```

Note: "onFailure" is optional and can be omitted if not needed. However if defining, it has to point to a "PROMPT" node.

**4. 🧩 Introduction of `identifier` and `ref` for the inputs**

All input definitions within nodes now require both an `identifier` and a `ref` field. The `identifier` is used to uniquely identify the input within the flow (previous `name`), while the `ref` is used for referencing the input in layouts.

**5. ⚙️ Enhancement of condition definitions**

Conditions now require defining a `onSkip` field to specify the next node when a condition is not met. Here is an example of the updated condition structure:

```json
{
  "id": "conditional_ou_creation",
  "type": "TASK_EXECUTION",
  "condition": {
      "key": "{{ context.userEligibleForProvisioning }}",
      "value": "true",
      "onSkip": "authenticated"
  },
  "executor": {
      "name": "OUExecutor"
  },
  "onSuccess": "provisioning"
}
```

**6. 🔄 Changes to the Flow Execution API**

The flow execution API now expects the selected action to be sent as "action", instead of the previously used "actionId".

Additionally, the flow execution API now require specifying "action" when progressing from a `PROMPT` node. The action input should correspond to the `ref` defined in the `actions` array of the `PROMPT` node.

Here is an example of the updated API request format:

```sh
curl --location 'https://localhost:8090/flow/execute' \
--header 'Content-Type: application/json' \
--data-raw '{
    "flowId": "0fe0f363-2af5-486f-8f23-cf10b9946a02",
    "action": "action_001",
    "inputs": {
        "username": "thor",
        "password": "thor@123"
    }
}
'
```

---

### Approach
This PR introduces the following changes.

- Replace `PROMPT_ONLY` node with a new `PROMPT` node which allows prompting for user inputs and actions
- Modify `PROMPT` nodes to have user selectable actions instead of generic "next" nodes
- Remove `DECISION` node that will get represented with `PROMPT` nodes with actions
- Introduce `onSuccess` and `onFailure` navigations to the `TASK_EXECUTION` nodes
- Remove "next" node concept from the node definitions. Prompt nodes define it via the actions and task_execution nodes will have onSuccess/ onFailure edges
- Improve condition evaluation for the nodes. When nodes can have multiple outgoing edges, condition evaluation needs to know which node to move next upon condition dissatisfaction. This is supported with a `onSkip` edge defined in the condition block
- Rename `actionId` in flow/execute API requests to `action` so that it will be aligned with inputs

In addition to the above listed changes, this PR does some minor refactoring to the existing implementation.

### Related Issues
- https://github.com/asgardeo/thunder/issues/919

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks


backend/internal/flow/core/prompt_node.go
```diff
@@ -24,97 +24,170 @@ import (
...
 		Actions:        make([]common.Action, 0),
 		RuntimeData:    make(map[string]string),
 	}
 
-	if n.checkInputData(ctx, nodeResp) {
-		logger.Debug("Required input data is not available in the context, returning incomplete response",
-			log.Any("requiredData", nodeResp.RequiredData))
-		nodeResp.Status = common.NodeStatusIncomplete
-		nodeResp.Type = common.NodeResponseTypeView
+	// Check if this prompt is handling a failure
```
senthalan, line 67: In which user flow this will get involved ?

Can't we detect prompt failure in any other form than reading an value from RuntimeData? I hope this failureReason should be set by another node

backend/internal/flow/flowexec/engine.go
```diff
@@ -430,33 +421,62 @@ func (fe *flowEngine) handleIncompleteResponse(nodeResp *common.NodeResponse, fl
 	// TODO: Handle retry scenarios with nodeResp.Type == common.NodeResponseTypeRetry
 }
 
-// resolveToNextNode resolves the next node to execute based on the current node.
-func (fe *flowEngine) resolveToNextNode(graph core.GraphInterface, currentNode core.NodeInterface,
-	nodeResp *common.NodeResponse) (core.NodeInterface, error) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "FlowEngine"))
+// handleForwardResponse handles forwarding to the next node (e.g., onFailure handler)
```
senthalan, line 424: Why we need to forward to the next node during onFailure ?

Will this be another configuration for the node? Something like this?
```
"nodes": [
        {
            "id": "basic_auth",
            "type": "TASK_EXECUTION",
            "inputs": [
                {
                    "ref": "input_001",
                    "type": "TEXT_INPUT",
                    "identifier": "username",
                    "required": true
                },
                {
                    "ref": "input_002",
                    "type": "PASSWORD_INPUT",
                    "identifier": "password",
                    "required": true
                }
            ],
            "executor": {
                "name": "BasicAuthExecutor"
            },
            "onSuccess": "authorization_check",
            "onFailure": "socialmedia_login",
        },
```

ThaminduDilshan: It's when a `onFailure` path is defined for a task execution node. For now, we're restricting onFailures to be a PROMPT node. The requirement is to display a custom error page during flow execution errors.

In that case, during failure, we need to execute that PROMPT node and return it's content along with the same failureReason

backend/internal/flow/core/prompt_node.go
ThaminduDilshan: > In which user flow this will get involved ?

It's when this PROMPT node is a redirection from a previous TASK_EXECUTION node onFailure callback

> Can't we detect prompt failure in any other form than reading an value from RuntimeData? I hope this failureReason should be set by another node

It has to be with something preserved in runtime data. We can introduce different runtime data key (boolean) to indicate this. However still have to pass failureReason from previous node to the current node. The approach followed is to retrieve it from runtime data

senthalan (approved):

PR #928, opened 2025-12-10
Refactor immutable resources code

rajithacharith:
### Purpose
This PR contains the changes to refactoring the existing file based runtime and export packages. 
The main objective of this refactoring is to organize the domain specific codes to separate domain packages and keep the immutable resource and export package domain logic independent. 
This will allow developers to onboard immutable configuration support to any resource easily


backend/internal/system/export/exporters.go
```diff
@@ -0,0 +1,291 @@
...
+
+const (
+	paramTypApplication        = "Application"
+	paramTypIdentityProvider   = "IdentityProvider"
+	paramTypNotificationSender = "NotificationSender"
+	paramTypUserSchema         = "UserSchema"
+)
+
+// ApplicationExporter implements ResourceExporter for applications.
+type ApplicationExporter struct {
```
senthalan: What if we move all the ApplicationExporters to the individual package. We can create a new go file with the name `export.go` ?

backend/internal/application/init.go
```diff
@@ -55,32 +54,51 @@ func Initialize(
 	appService := newApplicationService(appStore, certService, flowMgtService, brandingService, userSchemaService)
 
 	if config.GetThunderRuntime().Config.ImmutableResources.Enabled {
-		configs, err := filebasedruntime.GetConfigs("applications")
-		if err != nil {
-			logger.Fatal("Failed to read application configs from file-based runtime", log.Error(err))
+		// Type assert to access Storer interface for resource loading
```
senthalan: what if we move this to another go file and keep the init.go simple?

senthalan:
I have proposed to introduce two files, but it's not a good approach. It's better if we can merge them to a single file.

backend/internal/system/export/service.go
```diff
@@ -69,12 +70,20 @@ func newExportService(appService application.ApplicationServiceInterface,
 	notificationSenderService notification.NotificationSenderMgtSvcInterface,
 	userSchemaService userschema.UserSchemaServiceInterface,
 	param ParameterizerInterface) ExportServiceInterface {
+	// Create registry and register all exporters
+	registry := NewResourceExporterRegistry()
+	registry.Register(application.NewApplicationExporter(appService))
+	registry.Register(idp.NewIDPExporter(idpService))
+	registry.Register(notification.NewNotificationSenderExporter(notificationSenderService))
+	registry.Register(userschema.NewUserSchemaExporter(userSchemaService))
+
```
senthalan: Rather than passing the services and creating the exporter in the export service, what if we can pass the exporter from the servicemanager.go?

This way, we can keep the exporter constructors private.

backend/internal/system/export/registry.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/asgardeo/thunder/internal/system/immutableresource"
+)
+
+// ResourceExporterRegistry holds all registered resource exporters.
+type ResourceExporterRegistry struct {
+	exporters map[string]immutableresource.ResourceExporter
+}
+
+// NewResourceExporterRegistry creates a new registry for resource exporters.
+func NewResourceExporterRegistry() *ResourceExporterRegistry {
```
senthalan: can't we make it private

backend/internal/system/export/service.go
senthalan: We can keep a list of immutableresource.ResourceExporter in the servicemanager.go. And from each package init, we can get the created exporter and keep in this list. Then we can pass to here as a list.

I hope, with this no need to import individual package(app, idp) to the export package

backend/internal/system/immutableresource/parameterizer.go
```diff
@@ -48,8 +48,8 @@ type parameterizer struct {
 	rules templatingRules
 }
 
-// newParameterizer creates a new Parameterizer instance with the given templating rules
-func newParameterizer(rules templatingRules) *parameterizer {
+// NewParameterizer creates a new Parameterizer instance with the given templating rules
+func NewParameterizer(rules templatingRules) *parameterizer {
```
senthalan: What is the reason for this?

rajithacharith: Moved back to export package

backend/internal/system/export/template_rules.go
senthalan, line 1: Can't this be also moved to individual packages ?  it can be another method in the ResourceExporter Interface. 

With this the export and immutableresource package won't have anything specific to app or idp or etc

backend/internal/system/immutable_resource/config.go
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package immutableresource provides a generic framework for managing immutable resources.
+package immutableresource
```
senthalan, line 20: what is our standard on the package name?

File names we use `_`. Do we need to follow the same here  `immutable_resource` ?

Let's not change it with this PR as it's existing implementation, we can decide on the format and address this concern in a later PR(if needed)

backend/internal/system/immutableresource/loader.go
```diff
@@ -0,0 +1,103 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package immutableresource
+
+import (
+	"fmt"
+
+	filebasedruntime "github.com/asgardeo/thunder/internal/system/file_based_runtime"
```
senthalan: why we need to have this as a separate package ? can't we merge it to `immutableresource` package?

backend/internal/system/export/registry.go
rajithacharith: Fixed

backend/internal/system/export/service.go
rajithacharith: Fixed

backend/internal/system/immutableresource/loader.go
rajithacharith: Fixed, immutable_resource was added with this PR. merged into immutable_resource

backend/internal/system/immutable_resource/config.go
rajithacharith: Fixed

backend/internal/system/export/template_rules.go
rajithacharith: Fixed

senthalan (approved):

Issue #876, opened 2025-12-02
Internationalization Support (i18n)


ThaminduDilshan:
### Problem

Thunder currently has no structured i18n infrastructure. All existing backend and UI components (management APIs, gate client, login portal, error responses, executor messages, etc.) lack a consistent way to externalize and retrieve localized strings.

Without a unified system, maintaining localization across core flows and extensions becomes inconsistent and error-prone.

### Proposed Solution

Design and implement a i18n service that provides consistent internationalization capabilities across Thunder. This includes;

1. Backend i18n Server
    - Introduce a dedicated module responsible for translation key/value management
    - Support locales (e.g., en-US, fr-FR, si-LK)
    - Include storage design (DB tables for keys, locales, values, metadata)
    - Provide versioning and caching

2. i18n Management APIs
    - APIs for CRUD operations

3. i18n Retrieval API
    - Provide a lightweight open endpoint for Login portal / gate client (I.e. GET /i18n?locale=en-US)
    - Fast-loading format optimized for frontends (minified payload or grouped namespace approach) [Open for discussion]
    - Evaluate option to return i18n translations within the branding resolver response reducing two network calls

4. Error Message Localization
    - Introduce i18n keys for all existing error messages (service errors, API errors, validation messages) and other applicable places
    - Provide fallback locale behavior when translation is missing. I.e. keep current error messages, etc as the default
    - (Will have to remove current `serviceerror.CustomServiceError()` implementation as it cannot support translations for the error description)

5. Configuration & Developer Experience
    - Provide Develop app UI for viewing available i18n keys, edit translations per locale and searching and filtering

6. Governance Model
    Define whether Thunder:
    - Should allow free-form keys, OR
    - Restrict to predefined keys/ schemas/ per product area (auth flows, branding, errors, login pages…)
    - Hybrid model is also a suggestion in case custom keys are needed.
        - Core keys predefined
        - Custom keys allowed in a specific namespace (I.e. custom.*)

PR #940, opened 2025-12-14
Add initial i18n framework

ThumulaPerera:
### Purpose

The main purpose of this PR is to add the initial i18n (internationalization) framework.

That includes:
  1. Adding a management API to manage and retrieve translations.
  2. Adding a way for developers to define system API responses with i18n.
  3. Providing a reference Implementation on using i18n in system API responses.

### Approach

- Automated Extraction Tooling (tools/i18n-extractor): 

    - A standalone Go tool was built to parse the Go AST of the backend/internal directory and extract all core.Message{} struct literals. 
    - This tool generates the defaults.go file, ensuring the source code is the single source of truth for all translatable keys. 
    - A make generate_i18n target was added, with verification during linting to ensure the generated file is kept up-to-date.

- Core I18n Library (system/i18n/core): This new package provides the foundational primitives:
   
    - I18nMessage: A struct for developers to define a translation key and its default English message directly in code.

- Management API & Persistence (system/i18n/mgt): A new management module was added for CRUD operations on translations:

    - Store: A database-backed repository for persisting translation overrides and custom translations.
    - Service: A service layer encapsulating the business logic for managing translations.
    - Handler: HTTP handlers exposing a REST API to retrieve and update translation dynamically at runtime.

### Related Issues
- https://github.com/asgardeo/thunder/issues/876

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


api/i18n.yaml
```diff
@@ -0,0 +1,575 @@
...
+                    locales:
+                      - en
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
+  /i18n/locales/{locale}/translations/resolve:
+    parameters:
+      - $ref: '#/components/parameters/localePathParam'
+
+    post:
```
senthalan: Any specific reason to keep this as POST ?

api/i18n.yaml
```diff
@@ -0,0 +1,575 @@
...
+      tags:
+        - bulk-translations
+      summary: Resolve all translations for a locale
+      description: Resolves all translations for a locale by merging custom and default values; returns 200 even when only defaults exist (404 only if neither custom nor default exists for the requested scope).
+      operationId: resolveTranslationsByLocale
+      parameters:
+        - name: namespace
```
senthalan, line 74: won't we need pagination for this

api/i18n.yaml
```diff
@@ -0,0 +1,575 @@
...
+                      auth:
+                        login.button: Iniciar sesión
+                        login.title: Bienvenido
+                        logout.button: Cerrar sesión
+        '400':
+          $ref: '#/components/responses/BadRequest'
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
+  /i18n/locales/{locale}/translations/set-override:
```
senthalan: do we need to use `override` here ?
```
POST  /i18n/locales/{locale}/translations // to set translations for a locale (The provide keys will be set, an non-provided keys will have the defaults in the resolving time)
DELETE /i18n/locales/{locale}/translations // to delete the translations for a locale
```

Here we are setting the translations for locales right?
The override is only happening for `locale=en`, so I am thinking will this `override` term make sense is general perspective.

We can follow the same pattern for the single-translation
```
POST  /i18n/locales/{locale}/translations/ns/{namespace}/keys/{key}
DELETE /i18n/locales/{locale}/translations/ns/{namespace}/keys/{key}
```

api/i18n.yaml
```diff
@@ -0,0 +1,575 @@
...
+                      - fr-CA
+                default-only:
+                  summary: Default locale only
+                  value:
+                    locales:
+                      - en
+        '500':
+          $ref: '#/components/responses/InternalServerError'
+
+  /i18n/locales/{locale}/translations/resolve:
```
senthalan: I would like to group the two resolving APIs (bulk and individual) grouped as different tag (resolve)

To group the management APIs vs runtime API

api/i18n.yaml
ThumulaPerera: I think all 3 of "resolve", "set-overrides", "clear-overrides" behave more like processing functions. Hence the use of POST. But in this case, it might be okay to go with GET for resolve given that there still won't be any side effects

ThumulaPerera: Updated the Spec and Implementation as discussed offline

api/i18n.yaml
ThumulaPerera: Updated the Spec and Implementation as discussed offline

api/i18n.yaml
ThumulaPerera: Updated

backend/internal/system/error/serviceerror/error.go
```diff
@@ -51,6 +55,16 @@ func CustomServiceError(svcError ServiceError, errorDesc string) *ServiceError {
 	return err
 }
 
+// I18nServiceError defines a service error structure with i18n support.
+// This is the new error type that should be used for services being migrated to i18n.
```
senthalan, line 59: I hope the expectation here is to remove the existing CustomServiceError, once we migrate all the errors to this format.

backend/internal/system/i18n/core/defaults.go
```diff
@@ -0,0 +1,46 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package core
+
+// DefaultMessages contains all i18n keys extracted from source code.
+// This map is populated by the i18n extraction script at build time.
+// Helper functions for accessing these messages are in helpers.go.
+var DefaultMessages = map[string]string{
```
senthalan: Shall we keep this map package private,?

backend/internal/system/i18n/core/message.go
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package core provides internationalization support for the Thunder server.
```
senthalan:
```suggestion
// Package core provides internationalization support.
```

backend/internal/system/i18n/mgt/model.go
```diff
@@ -0,0 +1,61 @@
...
+	TotalResults int                          `json:"totalResults,omitempty"`
+	Translations map[string]map[string]string `json:"translations"`
+}
+
+// SetTranslationsRequest represents the request body for setting custom translations for a language.
+type SetTranslationsRequest struct {
+	Translations map[string]map[string]string `json:"translations"`
+}
+
+// --- Internal Service Models ---
```
senthalan:
```suggestion
// --- Service Models ---
```

backend/internal/system/i18n/mgt/service.go
```diff
@@ -0,0 +1,421 @@
...
+func newI18nService(store i18nStoreInterface) I18nServiceInterface {
+	return &i18nService{
+		store: store,
+	}
+}
+
+// ListLanguages retrieves all locale codes that have translations in the system.
+// The default locale is always included in the response, even if it has no translations in the DB.
+func (s *i18nService) ListLanguages() ([]string, *serviceerror.I18nServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
```
senthalan: Shall we move the logger to the i18nService struct ?

backend/internal/system/i18n/mgt/service.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+	if !ValidateLanguage(language) {
+		return nil, &ErrorInvalidLanguage
+	}
+	if !ValidateNamespace(namespace) {
+		return nil, &ErrorInvalidNamespace
+	}
+	if !ValidateKey(key) {
+		return nil, &ErrorInvalidKey
+	}
```
senthalan: Shall we move it to a common function as `validate(language string, namespace string, key string) (error)` 

because this seems to be duplicated

api/i18n.yaml
ThumulaPerera: It's bit complex to paginate this as we are returning a single large json object rather than a list of json objects. If the response grows in the future, a better approach would be to get the values filtered by namespace which is already supported.

backend/internal/system/error/serviceerror/error.go
ThumulaPerera: Yes

backend/internal/system/i18n/core/defaults.go
ThumulaPerera: Already updated

backend/internal/system/i18n/mgt/service.go
ThumulaPerera: done

backend/internal/system/i18n/mgt/service.go
ThumulaPerera: done

backend/internal/system/i18n/mgt/model.go
ThumulaPerera: done

backend/internal/system/i18n/core/message.go
ThumulaPerera: fixed

backend/internal/system/i18n/mgt/store.go
```diff
@@ -0,0 +1,279 @@
...
+type i18nStoreInterface interface {
+	GetDistinctLanguages() ([]string, error)
+	GetTranslations() (map[string]map[string]Translation, error)
+	GetTranslationsByNamespace(namespace string) (map[string]map[string]Translation, error)
+	GetTranslationsByKey(key string, namespace string) (map[string]Translation, error)
+	UpsertTranslationsByLanguage(language string, translations []Translation) error
+	UpsertTranslation(trans Translation) error
+	DeleteTranslationsByLanguage(language string) error
+	DeleteTranslation(language string, key string, namespace string) error
+}
```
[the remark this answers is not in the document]
ThumulaPerera: store interface is already private

backend/dbscripts/thunderdb/postgres.sql
```diff
@@ -339,3 +339,31 @@ CREATE INDEX idx_flow_version_deployment_id ON FLOW_VERSION (DEPLOYMENT_ID);
...
+    LANGUAGE_CODE   VARCHAR(10) NOT NULL,
+    NAMESPACE       VARCHAR(50) NOT NULL DEFAULT 'default',
+    VALUE           TEXT NOT NULL,
+    CREATED_AT      TIMESTAMPTZ DEFAULT NOW(),
+    UPDATED_AT      TIMESTAMPTZ DEFAULT NOW(),
+    CONSTRAINT uq_translation_key_lang_ns UNIQUE (MESSAGE_KEY, LANGUAGE_CODE, NAMESPACE, DEPLOYMENT_ID)
+);
+
+-- Index for deployment isolation on TRANSLATION
+CREATE INDEX idx_translation_deployment_id ON TRANSLATION (DEPLOYMENT_ID);
```
senthalan: we can remove this index right?

ThumulaPerera: In a case where DB is used by large number of deployments, this will be needed to
1. Get the supported languages of a given deployment
2. Get the translations for a given deployment (as we have a fallback mechanism, even when `fr` translations are requested, we pull all translations as we fallback for best matching language when a `fr` translation is not available for a given key)

backend/internal/system/i18n/mgt/storeconstants.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	// queryDeleteTranslation deletes a translation by language, key, and namespace.
+	queryDeleteTranslation = dbmodel.DBQuery{
+		ID: "I18N-07",
+		Query: `DELETE FROM TRANSLATION ` +
+			`WHERE LANGUAGE_CODE = $1 AND MESSAGE_KEY = $2 AND NAMESPACE = $3 AND DEPLOYMENT_ID = $4`,
+	}
+
+	// queryDeleteTranslation deletes a translation by language, key, and namespace.
```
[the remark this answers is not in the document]
ThumulaPerera: Updated

backend/internal/system/i18n/mgt/store.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	_, err = dbClient.Execute(queryDeleteTranslation, language, key, namespace, s.deploymentID)
+	if err != nil {
+		return fmt.Errorf("failed to delete translation: %w", err)
+	}
+	return nil
+}
+
+// DeleteTranslation deletes a translation by language, key, and namespace.
```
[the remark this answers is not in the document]
ThumulaPerera: Updated

backend/dbscripts/thunderdb/postgres.sql
ThumulaPerera: Removed the index and updated the composite index to use DEPLOYMENT_ID as the leftmost field

Issue #950, opened 2025-12-15
PS1 scripts compatibility with default Windows PowerShell 5.1 (Windows 10/11)


warnakulasuriya-fds-e23:
### Current Limitation

The `setup.ps1` and `start.ps1` scripts currently encounter compatibility issues when run on the default Windows PowerShell 5.1 (pre-installed on Windows 10/11).

### Observations

1. Syntax Errors: The scripts use modern syntax (like the null-coalescing operator ??) supported only in PowerShell 7+.

2. Encoding Issues: PowerShell 7 handles standard `UTF-8` files correctly. However, PowerShell 5.1 misinterprets the encoding (reading as ANSI/Windows-1252), causing "mojibake" (corrupted characters) in logs and regex parsing errors unless the file encoding format is manually changed to `UTF-8 with BOM.`

### Importance
As per [official documentation](https://learn.microsoft.com/en-us/powershell/scripting/what-is-windows-powershell), "Microsoft is no longer updating Windows PowerShell with new features". Windows PowerShell 5.1 is the final version of the legacy shell and is no longer receiving new features. Consequently, users on standard Windows installs cannot run these scripts without separately installing PowerShell 7+.

### Observed Environment
OS: Windows 11
Shells: Windows PowerShell 5.1 & PowerShell 7.5.4

![Screenshot of failure to run scripts](https://raw.githubusercontent.com/warnakulasuriya-fds-e23/reference-images/refs/heads/main/thunder/issue%23950/Screenshot%202025-12-15%20153135.png)

![Screenshot with powershell version table](https://raw.githubusercontent.com/warnakulasuriya-fds-e23/reference-images/refs/heads/main/thunder/issue%23950/Screenshot%202025-12-15%20154020.png)

### Suggested Improvement

Since the scripts are designed for modern PowerShell (pwsh), add a version check at the top of the scripts to detect if the user is running the older Windows PowerShell (5.1). If detected, gracefully exit and prompt the user to install PowerShell 7+.

### Version

v0.15.0

PR #951, opened 2025-12-15
Address Windows PowerShell Compatibility issues

warnakulasuriya-fds-e23:
### Purpose
This PR addresses compatibility issues preventing setup.ps1 and start.ps1 from running on the default Windows PowerShell 5.1 environment (standard on Windows 10 and 11).

Currently, these scripts fail immediately on Windows due to two main factors:
1. **Parser Crashes**: The scripts contain non-ASCII characters (emojis) and modern syntax (like ??) saved in UTF-8 (No BOM). Windows PowerShell 5.1 reads these as ANSI, causing immediate ParserError crashes (e.g., "The string is missing the terminator") before execution even begins.
2. **Alias Conflicts**: The usage of curl conflicts with the built-in PowerShell alias for Invoke-WebRequest, causing health checks to fail.

This change ensures the scripts are "parser-safe" for legacy shells, allowing them to gracefully notify the user of the requirement for PowerShell 7+ instead of crashing with version related errors.

### Approach
1. **Parser Safety (Emoji Removal)**: Replaced all Unicode emojis with standard ASCII text tags (e.g., [INFO], [ERROR], [WARN]).
      - Reason: Since PowerShell 5.1 attempts to parse the entire file before execution, the presence of UTF-8 emojis caused the parser to crash on load. Removing them allows the script to successfully parse on legacy shells so the logic below can actually run.

2. **Version Guard Clause**: Added a compatibility check at the very top of both scripts.
      - Reason: Checks $PSVersionTable.PSVersion.Major. If the version is less than 7, the script prints a clear error message directing the user to install PowerShell Core (pwsh) and exits cleanly.

3. **Command Disambiguation**: Replaced calls to curl with explicit calls to curl.exe.
      - Reason: This bypasses the PowerShell Invoke-WebRequest alias, ensuring the scripts use the actual cURL tool required for the API health checks.

4. **Syntax Standardization**: Updated regex strings to use standard quoting, preventing parsing errors on older interpreter versions.

### Related Issues
- #950

### DIFF
--- a/samples/apps/react-sdk-sample/start.ps1
+++ b/samples/apps/react-sdk-sample/start.ps1
@@ -17,6 +17,23 @@
 # under the License.
 # ----------------------------------------------------------------------------
 
+# Check for PowerShell Version Compatibility
+if ($PSVersionTable.PSVersion.Major -lt 7) {
+    Write-Host ""
+    Write-Host "================================================================" -ForegroundColor Red
+    Write-Host " [ERROR] UNSUPPORTED POWERSHELL VERSION" -ForegroundColor Red
+    Write-Host "================================================================" -ForegroundColor Red
+    Write-Host ""
+    Write-Host " You are currently running PowerShell $($PSVersionTable.PSVersion.ToString())" -ForegroundColor Yellow
+    Write-Host " Thunder requires PowerShell 7 (Core) or later." -ForegroundColor Yellow
+    Write-Host ""
+    Write-Host " Please install the latest version from:"
+    Write-Host " https://github.com/PowerShell/PowerShell" -ForegroundColor Cyan
+    Write-Host ""
+    exit 1
+}
+
+
 $SERVER_PORT = 3000
 $scriptDir = Split-Path -Parent $MyInvocation.MyCommand.Definition
 
@@ -46,7 +63,7 @@ KillPort -Port $SERVER_PORT
 # Check if npx is available
 $npx = Get-Command npx -ErrorAction SilentlyContinue
 if (-not $npx) {
-    Write-Host "❌ Error: npx is not installed. Please install Node.js and npm."
+    Write-Host "[ERROR] Error: npx is not installed. Please install Node.js and npm."
     exit 1
 }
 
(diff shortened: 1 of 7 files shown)


samples/apps/react-sdk-sample/start.ps1
```diff
@@ -46,7 +63,7 @@ KillPort -Port $SERVER_PORT
 # Check if npx is available
 $npx = Get-Command npx -ErrorAction SilentlyContinue
 if (-not $npx) {
-    Write-Host "❌ Error: npx is not installed. Please install Node.js and npm."
+    Write-Host "[ERROR] Error: npx is not installed. Please install Node.js and npm."
```
senthalan, line 66: Was these emoji characters are causing issue in PowerShell 7 (Core) or later versions ?

warnakulasuriya-fds-e23: No, not in the later versions. They had to be removed because they prevent the execution of the ps1 script at all for the Windows Powershell (5.1 version). Like a compile error. 
Suggested fix to show an error message asking user to install latest Powershell wont be displayed at all because of the emojis, instead error messages pointing to syntax errors would be seen.

senthalan: Understood. Thank you for details.

setup.ps1
```diff
@@ -17,6 +17,22 @@
...
 
+# Check for PowerShell Version Compatibility
+if ($PSVersionTable.PSVersion.Major -lt 7) {
+    Write-Host ""
+    Write-Host "================================================================" -ForegroundColor Red
+    Write-Host " [ERROR] UNSUPPORTED POWERSHELL VERSION" -ForegroundColor Red
+    Write-Host "================================================================" -ForegroundColor Red
+    Write-Host ""
+    Write-Host " You are currently running PowerShell $($PSVersionTable.PSVersion.ToString())" -ForegroundColor Yellow
+    Write-Host " This project requires PowerShell 7 (Core) or later." -ForegroundColor Yellow
```
senthalan:
```suggestion
    Write-Host " Thunder requires PowerShell 7 (Core) or later." -ForegroundColor Yellow
```

Shall we change `This project` -> `Thunder` ? in other places. We can keep the ps1 files within the sample directory as `This project`

warnakulasuriya-fds-e23: Message improved, changed `This project` to `Thunder`

warnakulasuriya-fds-e23:
Found some issues when running the **build.ps1**, currently looking into them. The `build `option works for build.ps1 under the latest changes, but there are some errors for the `test_unit` and `test_integration` and `run_frontend` options of the build.ps1. @senthalan would it be okay to send fixes for these issues along with this single PR? 

For example there is a 32,766 character limit imposed by the underlying Win32 API on commands run via powershell in windows, this affects the `test_unit `option and cause the `test_unit `option to fail when run in the latest powershell (Windows specific issue). PR builder doesnt show a fail because this particular option of build.ps1 doesnt seem to be tested in it. Fixing this might require refactoring  the test_unit function in build.ps1.

Issue #610, opened 2025-10-29
Add Passkey-Based Authentication


KaveeshaPiumini:
### Problem

When end users want to log in using Thunder securely and conveniently, the only available passwordless option is SMS OTP. This results in a poor user experience, higher number of password reset requests, and a higher risk of account compromise. With the growing browser and platform support for passkeys, introducing passkey-based authentication would significantly enhance security, improve usability, and allow users to log in seamlessly without needing to remember passwords.

### Proposed Solution

Add passkey-based authentication support using WebAuthn APIs for registration and login.

Implementation tasks:

- [x] - Add endpoints for passkey registration and authentication.
- [x] - Store credential metadata (credential ID, public key, user handle, sign count).
- [x] - Support multiple passkey registrations per user.
- [x] - Enforce challenge–response verification and replay protection.
- [x] - Support Atomic API based Passkey registration and authentication.
- [x] - Support Flow Execution based Passkey registration and authentication.
- [x] - Support on-the-fly passkey registration per user.
- [x] - Support username-less passkey support.
- [ ] - Ensure compatibility with major browsers, mobile devices and platforms.
- [x] - Add UI support to register and authenticate passkeys.
- [ ] - Add admin support to manage registered passkeys.

PR #958, opened 2025-12-17
Refactor: change credential storage from array to Map

KaveeshaPiumini:
### Purpose

This PR refactors the user credential management by introducing explicit credential type handling. The change enables the system to manage multiple credential types independently, enforce validation at the boundary, and ensure that sensitive credentials are always stored in a hashed form.

## ⚠️ Breaking Change:
This PR introduces breaking changes to the credential storage model, the user service interface, and the Credential data structure. Existing consumers, service implementations, and stored data handling must be updated to conform to the new design.

#### This PR introduces breaking changes to:

- The credential storage structure in the user table
- The user service interface for updating credentials
- The Credential model

### Implementation:

1. A dedicated constants file was introduced to define supported credential types such as password, pin, secret, and passkey, along with utility methods to validate credential types and determine whether a given type requires hashing. Centralizing this logic ensures consistent validation and security rules across the system.

2. (**Breaking Change**) Previously, credentials were stored as an array of Credential objects, each embedding its own credential type. With this PR, credentials are now stored as a map where the key is the credential type and the value is an array of credentials for that type. This removes duplication of credential type information, simplifies retrieval by type, and provides a clearer separation between different credential categories.

3. The user handler was updated to reflect this new model and request format. Incoming credential updates are now parsed as a map of credential types to values. The handler validates each credential type early to fail fast on invalid input. For each entry, it supports both an array of credential objects and a single string value, automatically normalizing string values into credential arrays. Before invoking the service layer, the handler hashes credentials for types that require hashing, specifically passwords, PINs, and secrets, while leaving passkey credentials unmodified. This ensures that the service layer only receives pre-processed, correctly structured credential data.

4. To support above behavior, a new hashCredentialsIfNeeded helper was introduced in the handler. This method determines whether hashing is required based on the credential type and, when applicable, hashes each credential value using the configured hash service. The resulting credentials include the hash value along with the algorithm, parameters, and salt, ensuring that all necessary metadata is preserved for future verification. Any hashing failures are logged and handled as internal server errors to avoid partial or inconsistent updates.

5. (**Breaking Change**) The credentialType field was removed from the struct, as the type is now derived from the surrounding map key. This change enforces structural correctness and prevents mismatches between a credential’s type and its stored location.

6. (**Breaking Change**) The user service interface for updating credentials was updated from accepting a generic attributes payload to explicitly accepting a user ID, a credential type, and a list of credentials. This change makes credential updates more explicit, type-safe, and aligned with the new storage structure, but requires all existing service implementations and mocks to be updated accordingly.

7. A new service method, GetUserCredentialsByType, was introduced to retrieve credentials of a specific type for a given user. This method validates the input parameters, retrieves all credentials for the user from the store, and returns only the credentials associated with the requested type. If the user does not have credentials of the specified type, the method returns an empty array rather than an error, allowing callers to distinguish between “no credentials” and failure scenarios. Proper error handling and logging are included for invalid input, missing users, and unexpected store failures.

### Related Issue
- https://github.com/asgardeo/thunder/issues/610


backend/internal/user/store.go
```diff
@@ -147,9 +147,10 @@ func (us *userStore) CreateUser(user User, credentials []Credential) error {
 	}
 
 	// Convert credentials array to JSON string
```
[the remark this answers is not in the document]
KaveeshaPiumini: done

backend/internal/user/model.go
```diff
@@ -41,6 +41,11 @@ type Credential struct {
 	Value             string              `json:"value"`
 }
 
+// CredentialMap represents the credential storage structure where credentials are organized by type.
+// Key: Credential type (e.g., "password", "pin", "secret", "passkey")
+// Value: Array of credentials of that type
+type CredentialMap map[string][]Credential
```
senthalan:
```suggestion
type Credentials map[string][]Credential
```

Do we need to indicate the obj type in name ?

backend/internal/user/service.go
```diff
@@ -61,6 +70,13 @@ type UserServiceInterface interface {
 	VerifyUser(userID string, credentials map[string]interface{}) (*User, *serviceerror.ServiceError)
 	AuthenticateUser(request AuthenticateUserRequest) (*AuthenticateUserResponse, *serviceerror.ServiceError)
 	ValidateUserIDs(userIDs []string) ([]string, *serviceerror.ServiceError)
+	AddPasskeyCredential(userID string, credential Credential) *serviceerror.ServiceError
```
senthalan: Do we need interface level method for a specific credential type ?

Shouldn't we have generic methods here ?
```
// Returns specific credentialType of specific user
GetUserCredential(userID string, credentialType string) ([]Credential, *serviceerror.ServiceError)
// Create a specific Credential of specific user
AddUserCredentials(userID string, credentialType string, credential Credential) (Credential, *serviceerror.ServiceError)
// Update a specific Credential of specific user
UpdateUserCredentials(userID string, credentialType string, updatedCredentials []Credential) (*serviceerror.ServiceError)
// Delete a specific Credentials of specific user
DeleteUserCredentials(userID string, credentialType string) (*serviceerror.ServiceError)
```
(I haven't checked the complete implementation, I might be missing some requirements here. The idea is not to have interface methods with specific credential type).

backend/internal/user/model.go
senthalan: I am not sure my suggestion is correct or not after checking implementation. Please check the go naming convention regarding this

backend/internal/user/service.go
```diff
@@ -782,6 +790,293 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Successfully added passkey credential", log.String("userID", log.MaskString(userID)))
+	return nil
+}
+
+// GetUserCredentials retrieves credentials of a specific type for a user.
+// If credentialType is empty, returns all credentials.
```
senthalan: let's have another method for `GetAllUserCredentials(userID string)` to return all the credentials, rather than missing it

backend/internal/user/service.go
```diff
@@ -782,6 +790,285 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			log.String("credentialType", credentialType))
+		return []Credential{}, nil
+	}
+
+	logger.Debug("Retrieved credentials for type",
+		log.String("userID", log.MaskString(userID)),
+		log.String("credentialType", credentialType),
+		log.Int("count", len(credentials)))
+	return credentials, nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: added test coverage

backend/internal/user/service.go
```diff
@@ -782,6 +790,285 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Successfully updated credentials in map",
+		log.String("userID", log.MaskString(userID)),
+		log.String("credentialType", credentialType))
+	return nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: added test coverage

backend/internal/user/service.go
```diff
@@ -37,11 +37,20 @@ import (
 
 const loggerComponentName = "UserService"
 
+// Credential type constants
+const (
+	CredentialTypePassword = "password"
```
ThaminduDilshan: Shall we define a type for these?
Also let's move them to a constant file

backend/internal/user/service.go
ThaminduDilshan: IMO we should use generic methods in a service like these

backend/internal/user/handler.go
```diff
@@ -387,11 +388,26 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
+		return
+	}
+
+	// Update each credential type (replace mode by default for user-initiated updates)
+	for credentialType, credentialValue := range attrs {
+		if svcErr := uh.userService.UpdateUserCredentials(userID, credentialType, credentialValue, false); svcErr != nil {
+			handleError(w, svcErr)
+			return
+		}
+	}
```
senthalan: I don't think performing a for loop here is not the best approach. Can't we have the interface method to accepts credentials list ?

backend/internal/user/service.go
KaveeshaPiumini: Since retrieving all the user credentials doesn't seem to have a valid usecase, I have removed this capability now

backend/internal/user/service.go
```diff
@@ -782,6 +790,285 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+	}
+
+	// Append the new passkey credential
+	existingCredentials[CredentialTypePasskey] = append(
+		existingCredentials[CredentialTypePasskey],
+		credential,
+	)
+
+	// Update credentials in the database
+	err = us.userStore.UpdateUserCredentials(userID, existingCredentials)
```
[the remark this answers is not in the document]
KaveeshaPiumini: removed code block

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			logger,
+			"Failed to update credential map",
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Successfully updated credential map", log.String("userID", log.MaskString(userID)))
+	return nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Successfully updated credentials in map",
+		log.String("userID", log.MaskString(userID)),
+		log.String("credentialType", credentialType))
+	return nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			logger,
+			"Failed to update user credentials",
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Successfully added passkey credential", log.String("userID", log.MaskString(userID)))
+	return nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	logger.Debug("Retrieved user with credentials",
+		log.String("userID", log.MaskString(userID)),
+		log.Int("credentialTypeCount", len(credentialMap)))
+	return &user, credentialMap, nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/service.go
```diff
@@ -782,6 +790,285 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			err,
+			log.String("userID", userID),
+		)
+	}
+
+	// Update the credentials for the specified type
+	existingCredMap[credentialType] = updatedCredentials
+
+	// Save the updated credential map
+	err = us.userStore.UpdateUserCredentials(userID, existingCredMap)
```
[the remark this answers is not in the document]
KaveeshaPiumini: this code block is now removed

backend/internal/user/model.go
KaveeshaPiumini: updated to Credentials

backend/internal/user/service.go
KaveeshaPiumini: Modified to have a generic function. 

Updating the credentials use the already existing method - `UpdateUserCredentials`
For retrieving the credentials a new method added - `GetUserCredentialsByType`

backend/internal/user/service.go
KaveeshaPiumini: moved to constant file

backend/internal/user/service.go
```diff
@@ -552,39 +621,63 @@ func (us *userService) UpdateUserCredentials(userID string, attributes json.RawM
...
+	if credList, exists := extractedCreds[credentialType]; exists {
+		credentials[credentialType] = credList
+		logger.Debug("Replaced credentials",
+			log.String("credentialType", credentialType),
+			log.Int("count", len(credList)))
 	}
 
-	return merged
+	return nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/handler.go
```diff
@@ -387,15 +391,95 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
+		if svcErr != nil {
+			handleError(w, svcErr)
+			return
+		}
+
+		if svcErr := uh.userService.UpdateUserCredentials(userID, credentialType, hashedCredentials); svcErr != nil {
+			handleError(w, svcErr)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/user/service.go
```diff
@@ -464,127 +462,137 @@ func (us *userService) containsCredentialFields(attributes json.RawMessage) (boo
...
-		}
-		return logErrorAndReturnServerError(logger, "Failed to update user credentials", err, log.String("id", userID))
+// validateCredential validates a single credential.
+func (us *userService) validateCredential(credential *Credential) error {
+	if credential == nil {
+		return errors.New("credential is nil")
+	}
+	if strings.TrimSpace(credential.Value) == "" {
+		return errors.New("credential value is empty")
 	}
```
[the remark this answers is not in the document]
KaveeshaPiumini: for passkey, these values can be empty

backend/internal/user/service.go
```diff
@@ -464,127 +462,137 @@ func (us *userService) containsCredentialFields(attributes json.RawMessage) (boo
...
+func (us *userService) prepareUpdatedCredentials(
+	existingCredentials Credentials,
+	credentialType string,
+	newCredentials []Credential,
+	logger *log.Logger,
+) (Credentials, error) {
+	// Copy existing credentials to preserve other types
+	updatedCredentials := make(Credentials)
+	for credType, credList := range existingCredentials {
+		updatedCredentials[credType] = append([]Credential{}, credList...)
```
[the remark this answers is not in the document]
KaveeshaPiumini: this is not necessary to be deep copied

backend/internal/user/handler.go
```diff
@@ -387,15 +391,104 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
```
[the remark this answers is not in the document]
KaveeshaPiumini: the ability to update just one credential type should be there.

backend/internal/user/service.go
```diff
@@ -1021,7 +1179,7 @@ func buildPaginationLinks(path string, limit, offset, totalResults int) []Link {
 	return links
 }
 
-// buildTreePaginationLinks builds pagination links for user responses.
+// buildTreePaginationLinks builds pagination linksapp for user responses.
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+		}
+		for _, cred := range credentials {
+			if cred.CredentialType != credType {
+				logger.Debug("Credential type mismatch in map",
+					log.String("mapKey", credType),
+					log.String("credentialType", cred.CredentialType))
+				return &ErrorInvalidRequestFormat
+			}
+		}
+	}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this code block is now removed

backend/internal/user/store.go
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/user/service.go
```diff
@@ -782,6 +788,281 @@ func (us *userService) ValidateUserIDs(userIDs []string) ([]string, *serviceerro
...
+			log.String("credentialType", credentialType))
+		return []Credential{}, nil
+	}
+
+	logger.Debug("Retrieved credentials for type",
+		log.String("userID", log.MaskString(userID)),
+		log.String("credentialType", credentialType),
+		log.Int("count", len(credentials)))
+	return credentials, nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: this method is now removed

backend/internal/user/model.go
```diff
@@ -34,13 +34,17 @@ type User struct {
 
 // Credential represents the credentials of a user.
 type Credential struct {
-	CredentialType    string              `json:"credentialType"`
 	StorageType       string              `json:"storageType"`
 	StorageAlgo       hash.CredAlgorithm  `json:"storageAlgo"`
 	StorageAlgoParams hash.CredParameters `json:"storageAlgoParams"`
 	Value             string              `json:"value"`
 }
```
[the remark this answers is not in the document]
KaveeshaPiumini: mentioned in the PR

backend/internal/user/constants.go
```diff
@@ -0,0 +1,54 @@
...
+// Credential type constants define the supported credential types.
+const (
+	CredentialTypePassword CredentialType = "password"
+	CredentialTypePin      CredentialType = "pin"
+	CredentialTypeSecret   CredentialType = "secret"
+	CredentialTypePasskey  CredentialType = "passkey"
+)
+
+// SupportedCredentialTypes defines the set of credential types that are supported.
+var SupportedCredentialTypes = map[CredentialType]struct{}{
```
ThaminduDilshan: Shall we use an array instead? You should be able to use `slices.Contains()` to check for the value existence

backend/internal/user/constants.go
```diff
@@ -0,0 +1,54 @@
...
+
+// IsValid checks if the credential type is supported.
+func (ct CredentialType) IsValid() bool {
+	_, ok := SupportedCredentialTypes[ct]
+	return ok
+}
+
+// RequiresHashing checks if this credential type requires hashing
+func (ct CredentialType) RequiresHashing() bool {
+	return ct == CredentialTypePassword || ct == CredentialTypePin || ct == CredentialTypeSecret
```
ThaminduDilshan: Rather, shall we define another array for hashing required credentials? This will allow to easily extend this for future credential types

backend/internal/user/model.go
```diff
@@ -34,13 +34,17 @@ type User struct {
...
 
+// Credentials represents the credential storage structure where credentials are organized by type.
+// Key: Credential type (e.g., "password", "pin", "secret", "passkey")
+// Value: Array of credentials of that type
+type Credentials map[string][]Credential
```
ThaminduDilshan: Shall we use the `CredentialType` defined in constant for the map key?

backend/internal/user/handler.go
```diff
@@ -387,15 +391,104 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
+
+		// Hash credentials that require hashing (password, pin, secret)
+		// Note: The service expects pre-hashed credentials
+		hashedCredentials, svcErr := uh.hashCredentialsIfNeeded(credentialType, credentials)
+		if svcErr != nil {
+			handleError(w, svcErr)
+			return
+		}
+
+		if svcErr := uh.userService.UpdateUserCredentials(userID, credentialType, hashedCredentials); svcErr != nil {
```
ThaminduDilshan: Rather than updating user profile in a loop for each credential, shall we modify service layer to accept a credential map and update database in a single DB call?

backend/internal/user/handler.go
```diff
@@ -387,15 +391,104 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
+		return
+	}
+
+	if len(attrs) == 0 {
+		handleError(w, &ErrorMissingCredentials)
 		return
 	}
 
+	// Update each credential type
+	for credentialType, credentialValue := range attrs {
```
ThaminduDilshan: Shouldn't we move this to service layer? IMO validation for credential types, parsing and hashing should be performed at the service layer.
In future, there maybe a requirement to update credentials from a executor or some other service and it directly calls the user service. With the current implementation executor or some external service has to duplicate the validation/ hashing/ etc in their side.
Should be able to accept `json.RawMessage` in the user service method similar to `UpdateUserAttributes` method

backend/internal/user/service.go
```diff
@@ -249,20 +243,21 @@ func (us *userService) CreateUserByPath(
 	return us.CreateUser(user)
 }
 
-// extractCredentials extracts the credentials from the user attributes and returns a Credentials array.
-func (us *userService) extractCredentials(user *User) ([]Credential, error) {
+// extractCredentials extracts the credentials from the user attributes and returns Credentials.
```
ThaminduDilshan:
```suggestion
// extractCredentials extracts the credentials from the user attributes and returns Credentials struct.
```

backend/internal/user/constants.go
ThaminduDilshan: Alternatively we can define a struct for the credential that will store meta data along with the type.

For example;
```go
type CredentialType struct {
    name string
    requireHashing bool
    multiValued bool
}
```

backend/internal/user/constants.go
KaveeshaPiumini: done

backend/internal/user/constants.go
KaveeshaPiumini: Introduced a new array named `HashedCredentialTypes` defining the credential types that require hashing. Using a `CredentialType struct` introduces unintentional complexities to the code. Hence the first approach was implemented.

backend/internal/user/model.go
KaveeshaPiumini: done

backend/internal/user/handler.go
KaveeshaPiumini: done

backend/internal/user/handler.go
KaveeshaPiumini: updated to json.RawMessage. 

`UpdateUserCredentials(userID string, credentialType CredentialType, credentials json.RawMessage) *serviceerror.ServiceError`

backend/internal/user/service.go
```diff
@@ -1021,7 +1220,7 @@ func buildPaginationLinks(path string, limit, offset, totalResults int) []Link {
 	return links
 }
 
-// buildTreePaginationLinks builds pagination links for user responses.
+// buildTreePaginationLinks builds pagination linksapp for user responses.
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/user/service.go
```diff
@@ -464,129 +462,273 @@ func (us *userService) containsCredentialFields(attributes json.RawMessage) (boo
...
+		}
+
+		processedCredentials[ct] = hashedCredentials
 	}
 
-	for _, credType := range providedOrder {
-		if cred, found := providedMap[credType]; found {
-			merged = append(merged, cred)
+	return processedCredentials, nil
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: removed

backend/internal/user/constants.go
```diff
@@ -0,0 +1,68 @@
...
+// IsValid checks if the credential type is supported.
+func (ct CredentialType) IsValid() bool {
+	return slices.Contains(SupportedCredentialTypes, ct)
+}
+
+// ParseCredentialType converts a string to a CredentialType.
+func ParseCredentialType(name string) (CredentialType, bool) {
+	ct := CredentialType(name)
+	return ct, ct.IsValid()
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: removed

backend/internal/user/service.go
KaveeshaPiumini: done

backend/internal/user/handler.go
```diff
@@ -382,16 +383,41 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
...
+			handleError(w, &ErrorInvalidCredential)
+			return
+		}
+
+		// Call service to update this credential type
+		if svcErr := uh.userService.UpdateUserCredentials(userID, credType, credValue); svcErr != nil {
+			handleError(w, svcErr)
+			return
+		}
+	}
```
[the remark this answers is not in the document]
KaveeshaPiumini: refactored to remove the credentialType parameter from UpdateUserCredentials to solve the race condition issue.

backend/internal/user/store.go
```diff
@@ -300,7 +301,7 @@ func (us *userStore) UpdateUser(user *User) error {
 }
 
 // UpdateUserCredentials updates the credentials for a given user.
-func (us *userStore) UpdateUserCredentials(userID string, credentials []Credential) error {
+func (us *userStore) UpdateUserCredentials(userID string, credentials CredentialMap) error {
 	dbClient, err := provider.GetDBProvider().GetUserDBClient()
 	if err != nil {
 		return fmt.Errorf("failed to get database client: %w", err)
```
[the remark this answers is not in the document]
KaveeshaPiumini: this code block is now removed

backend/internal/user/handler.go
```diff
@@ -382,11 +383,12 @@ func (uh *userHandler) HandleSelfUserCredentialUpdateRequest(w http.ResponseWrit
 		return
 	}
 
-	if updateRequest == nil || len(updateRequest.Attributes) == 0 {
+	if updateRequest == nil || len(updateRequest.Attributes) == 0 || string(updateRequest.Attributes) == "{}" {
 		handleError(w, &ErrorMissingCredentials)
 		return
 	}
 
+	// Call service to update all credential types atomically
```
ThaminduDilshan: Shall we remove these unnecessary comments?

backend/internal/user/service.go
```diff
@@ -464,129 +462,253 @@ func (us *userService) containsCredentialFields(attributes json.RawMessage) (boo
...
 
 	return false, nil
 }
 
-// UpdateUserCredentials updates the credentials of a user.
-func (us *userService) UpdateUserCredentials(userID string, attributes json.RawMessage) *serviceerror.ServiceError {
+// UpdateUserCredentials updates one or more credential types for a user.
+// Accepts a JSON object where keys are credential types (e.g., "password", "pin")
+// and values are the credential data (string or array of Credential objects).
+// All credential types in the request are processed atomically in a single database transaction.
```
ThaminduDilshan: Let's remove this part

backend/internal/user/handler.go
ThaminduDilshan: Check other places as well

ThaminduDilshan (approved):

backend/internal/user/handler.go
KaveeshaPiumini: done

backend/internal/user/service.go
KaveeshaPiumini: removed here and other places as well

PR #998, opened 2025-12-22
Improve event publishing

sahandilshan:
### Purpose
- Remove events getting published from executor level
- Publish thunder events as otel traces tags

### Related PR(s)
- the dashboard changes that sent with https://github.com/asgardeo/thunder/pull/999 should be merged after merging this


backend/internal/flow/executor/basic_auth_executor.go
```diff
@@ -94,20 +93,6 @@ func (b *basicAuthExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResp
 	logger := b.logger.With(log.String(log.LoggerKeyFlowID, ctx.FlowID))
 	logger.Debug("Executing basic authentication executor")
 
-	if b.observabilitySvc.IsEnabled() {
```
rajithacharith, line 97: With this removal observablitiySvc have no usages in the executor, but the residual observablity Dependency has not removed

Issue #1276, opened 2026-02-03
Allow configuring response type for OAuth userinfo in application


ThaminduDilshan:
### Current Limitation

Currently OAuth user info response is returned as a plain JSON. However spec allows the server to sign and/or encrypt the user info response. This behaviour should be configurable via the application.

### Suggested Improvement

Allow configuring response type of the OAuth user info via an application config

Issue #1220, opened 2026-01-28
Sign OAuth user info response


ThaminduDilshan:
### Current Limitation

Currently OAuth user info response is returned as a plain JSON. However the OIDC spec allows to sign the user info response based on a client configuration.

> The UserInfo Claims MUST be returned as the members of a JSON object unless a signed or encrypted response was requested during Client Registration.
> 
> If the UserInfo Response is signed and/or encrypted, then the Claims are returned in a JWT and the content-type MUST be application/jwt. The response MAY be encrypted without also being signed. If both signing and encryption are requested, the response MUST be signed then encrypted, with the result being a Nested JWT, as defined in [[JWT]](https://openid.net/specs/openid-connect-core-1_0.html#JWT).
> 
> If signed, the UserInfo Response MUST contain the Claims iss (issuer) and aud (audience) as members. The iss value MUST be the OP's Issuer Identifier URL. The aud value MUST be or include the RP's Client ID value.

### Suggested Improvement

Allow to sign the user info response. Application developers should be able to configure whether to sign or not the user info response with a application configuration

nandhu-kumar:
@ThaminduDilshan can you please assign it to me?

PR #1362, opened 2026-02-11
Support configurable OAuth UserInfo response_type (JSON/JWS) with JWS signing implementation

nandhu-kumar:
## Purpose

Currently, the OAuth UserInfo response is always returned as plain JSON. However, as per the OIDC specification, the server should support returning signed (JWS) or encrypted (JWE) UserInfo responses when configured by the client.

This PR introduces support to:

1. Configure the `userinfo.response_type` flag per application.
2. Generate and return a signed UserInfo response (JWS) when configured.

Allowed values:

- JSON (default fallback)
- JWS
- JWE (configuration support only; encryption logic not implemented yet)

When `response_type = JWS`, the UserInfo response is returned as a signed JWT with `Content-Type: application/jwt`, in compliance with the OIDC specification.

The configured value is persisted in the database (OAuth config JSON) and returned via the Application APIs.

---

## No Breaking Changes

This change is fully backward compatible.

- If `response_type` is not provided, it defaults to `JSON`.
- If `response_type` is invalid, it falls back to `JSON`.
- Existing applications continue to receive JSON responses without any changes.

---

### Application Configuration (Issue #1276)

- Added `response_type` field inside `UserInfoConfig`.
- Added typed constants for supported response types:
  - JSON
  - JWS
  - JWE
- Added validation to allow only supported values.
- Implemented default fallback to JSON when the flag is not provided or invalid.
- Updated DTOs and processed models to support `user_info.response_type`.

### Store Layer Updates

- Persisted `response_type` inside `oauth_config_json`.
- Updated marshalling and unmarshalling logic.
- Ensured value is properly stored and returned in:
  - Create Application
  - Update Application
  - Get Application

---

### UserInfo Endpoint Enhancement (Issue #1220)

- Refactored the UserInfo service to return a structured response model.
- Introduced support for signed (JWS) UserInfo responses.
- When `response_type = JWS`:
  - Claims are signed using the existing JWT service.
  - `Content-Type` is set to `application/jwt`.
  - Issuer and audience values are populated appropriately.
- Added error handling for JWT generation failures.
- Maintained thin handler architecture and clean service-layer separation.

---

### Unit Tests

- JSON response flow validation.
- JWS response flow validation.
- JWT generation failure handling.
- Grant type validation (client_credentials rejected).
- Scope validation (openid required).
- Claim filtering and scope-to-claims mapping.
- Response type fallback behavior.

### Integration Tests

- Application API tests for:
  - JWS configuration
  - JWE configuration
  - Invalid response_type fallback
- End-to-end UserInfo JWS validation:
  - Authorization Code flow
  - JWT format validation
  - `Content-Type: application/jwt` verification

---

## Note

This PR implements configuration support and JWS signing support for the UserInfo endpoint.

JWE configuration is supported at the application level, but encryption logic will be implemented separately.

---

## Related Issues

- Allow configuring response type for OAuth userinfo in application #1276
- Sign OAuth user info response #1220

---

## Related PRs

N/A

---

## Checklist

- [x] Documentation provided (if applicable).
- [x] Tests provided.
  - [x] Unit Tests
  - [x] Integration Tests
- [ ] Breaking changes (Fill if applicable)
- [ ] Breaking change label added.

---

## Security checks

### DIFF
--- a/backend/internal/application/model/constants.go
+++ b/backend/internal/application/model/constants.go
@@ -33,13 +33,10 @@ type UserInfoResponseType string
 
 const (
 	// UserInfoResponseTypeJSON represents the JSON userinfo response type.
-	UserInfoResponseTypeJSON = "JSON"
+	UserInfoResponseTypeJSON UserInfoResponseType = "JSON"
 
 	// UserInfoResponseTypeJWS represents the JWS userinfo response type.
-	UserInfoResponseTypeJWS = "JWS"
-
-	// UserInfoResponseTypeJWE represents the JWE userinfo response type.
-	UserInfoResponseTypeJWE = "JWE"
+	UserInfoResponseTypeJWS UserInfoResponseType = "JWS"
 )
 
 // ApplicationNotFoundError is the error returned when an application is not found.
--- a/backend/internal/application/service.go
+++ b/backend/internal/application/service.go
@@ -1378,8 +1378,7 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
 		responseType := model.UserInfoResponseType(strings.ToUpper(string(userInfoConfigInput.ResponseType)))
 
 		switch responseType {
-		case model.UserInfoResponseTypeJWS,
-			model.UserInfoResponseTypeJWE:
+		case model.UserInfoResponseTypeJWS:
 			oauthUserInfo.ResponseType = responseType
 		default:
 			oauthUserInfo.ResponseType = model.UserInfoResponseTypeJSON
--- a/backend/internal/oauth/oauth2/userinfo/handler.go
+++ b/backend/internal/oauth/oauth2/userinfo/handler.go
@@ -21,6 +21,7 @@ package userinfo
 import (
 	"net/http"
 
+	appmodel "github.com/asgardeo/thunder/internal/application/model"
 	"github.com/asgardeo/thunder/internal/oauth/oauth2/constants"
 	serverconst "github.com/asgardeo/thunder/internal/system/constants"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
@@ -55,7 +56,7 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	userInfo, svcErr := h.service.GetUserInfo(accessToken)
+	result, svcErr := h.service.GetUserInfo(accessToken)
 	if svcErr != nil {
 		h.writeServiceErrorResponse(w, svcErr)
 		return
@@ -64,7 +65,13 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
 	w.Header().Set(serverconst.CacheControlHeaderName, serverconst.CacheControlNoStore)
 	w.Header().Set(serverconst.PragmaHeaderName, serverconst.PragmaNoCache)
 
-	utils.WriteSuccessResponse(w, http.StatusOK, userInfo)
+	if result.Type == appmodel.UserInfoResponseTypeJWS {
+		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJWT)
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(result.JWTBody))
+	} else {
+		utils.WriteSuccessResponse(w, http.StatusOK, result.JSONBody)
+	}
 
 	h.logger.Debug("UserInfo response sent successfully")
 }
--- a/backend/internal/oauth/oauth2/userinfo/model.go
+++ b/backend/internal/oauth/oauth2/userinfo/model.go
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package userinfo
+
+import appmodel "github.com/asgardeo/thunder/internal/application/model"
+
+// UserInfoResponse represents the structured response returned by the
+// UserInfo service. It supports both JSON and JWS response types.
+// Only one of JSONBody or JWTBody will be populated based on Type.
+type UserInfoResponse struct {
+	Type     appmodel.UserInfoResponseType
+	JSONBody map[string]interface{}
+	JWTBody  string
+}
--- a/backend/internal/oauth/oauth2/userinfo/service.go
+++ b/backend/internal/oauth/oauth2/userinfo/service.go
@@ -29,6 +29,7 @@ import (
 	"github.com/asgardeo/thunder/internal/oauth/oauth2/tokenservice"
 	oauth2utils "github.com/asgardeo/thunder/internal/oauth/oauth2/utils"
 	"github.com/asgardeo/thunder/internal/ou"
+	"github.com/asgardeo/thunder/internal/system/config"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/jose/jwt"
 	"github.com/asgardeo/thunder/internal/system/log"
@@ -39,7 +40,7 @@ const serviceLoggerComponentName = "UserInfoService"
 
 // userInfoServiceInterface defines the interface for OIDC UserInfo endpoint.
 type userInfoServiceInterface interface {
-	GetUserInfo(accessToken string) (map[string]interface{}, *serviceerror.ServiceError)
+	GetUserInfo(accessToken string) (*UserInfoResponse, *serviceerror.ServiceError)
 }
 
 // userInfoService implements the userInfoServiceInterface.
@@ -68,7 +69,7 @@ func newUserInfoService(
 }
 
 // GetUserInfo validates the access token and returns user information based on authorized scopes.
-func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{}, *serviceerror.ServiceError) {
+func (s *userInfoService) GetUserInfo(accessToken string) (*UserInfoResponse, *serviceerror.ServiceError) {
 	if accessToken == "" {
 		return nil, &errorInvalidAccessToken
 	}
@@ -115,7 +116,55 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
 		return nil, svcErr
 	}
 
-	return response, nil
+	// Decide response type
+	responseType := appmodel.UserInfoResponseTypeJSON
+	if oauthApp != nil && oauthApp.UserInfo != nil {
+		responseType = oauthApp.UserInfo.ResponseType
+	}
+
+	if responseType == appmodel.UserInfoResponseTypeJWS {
+		return s.generateJWSUserInfo(sub, tokenClaims, response)
+	}
+
+	return &UserInfoResponse{
+		Type:     appmodel.UserInfoResponseTypeJSON,
+		JSONBody: response,
+	}, nil
+}
+
+// generateJWSUserInfo creates a signed JWT UserInfo response
+// based on the application configuration.
+func (s *userInfoService) generateJWSUserInfo(
+	sub string,
+	tokenClaims map[string]interface{},
+	response map[string]interface{},
+) (*UserInfoResponse, *serviceerror.ServiceError) {
+	clientID := ""
+	if cid, ok := tokenClaims["client_id"].(string); ok {
+		clientID = cid
+	}
+
+	runtime := config.GetThunderRuntime()
+
+	issuer := runtime.Config.JWT.Issuer
+	validity := runtime.Config.JWT.ValidityPeriod
+
+	signedJWT, _, err := s.jwtService.GenerateJWT(
+		sub,
+		clientID,
+		issuer,
+		validity,
+		response,
+	)
+	if err != nil {
+		s.logger.Error("Failed to generate signed UserInfo JWT")
+		return nil, &serviceerror.InternalServerError
+	}
+
+	return &UserInfoResponse{
+		Type:    appmodel.UserInfoResponseTypeJWS,
+		JWTBody: signedJWT,
+	}, nil
 }
 
 // validateAndDecodeToken validates the JWT signature and decodes the payload.
--- a/backend/internal/oauth/oauth2/userinfo/handler_test.go
+++ b/backend/internal/oauth/oauth2/userinfo/handler_test.go
@@ -145,7 +145,7 @@ func (s *UserInfoHandlerTestSuite) TestHandleUserInfo_Success() {
 		"email": "john@example.com",
 	}
 
-	s.mockService.On("GetUserInfo", "valid-token").Return(userInfo, nil)
+	s.mockService.On("GetUserInfo", "valid-token").Return(jsonResponse(userInfo), nil)
 
 	s.handler.HandleUserInfo(rr, req)
 
@@ -169,7 +169,7 @@ func (s *UserInfoHandlerTestSuite) TestHandleUserInfo_Success_POST() {
 		"sub": "user123",
 	}
 
-	s.mockService.On("GetUserInfo", "valid-token").Return(userInfo, nil)
+	s.mockService.On("GetUserInfo", "valid-token").Return(jsonResponse(userInfo), nil)
 
 	s.handler.HandleUserInfo(rr, req)
 
@@ -190,7 +190,7 @@ func (s *UserInfoHandlerTestSuite) TestHandleUserInfo_Success_WithGroups() {
 		"groups": []interface{}{"admin", "users"},
 	}
 
-	s.mockService.On("GetUserInfo", "valid-token").Return(userInfo, nil)
+	s.mockService.On("GetUserInfo", "valid-token").Return(jsonResponse(userInfo), nil)
 
 	s.handler.HandleUserInfo(rr, req)
 
@@ -211,7 +211,7 @@ func (s *UserInfoHandlerTestSuite) TestHandleUserInfo_CaseInsensitiveBearer() {
 		"sub": "user123",
 	}
 
-	s.mockService.On("GetUserInfo", "valid-token").Return(userInfo, nil)
+	s.mockService.On("GetUserInfo", "valid-token").Return(jsonResponse(userInfo), nil)
 
 	s.handler.HandleUserInfo(rr, req)
 
@@ -229,7 +229,7 @@ func (s *UserInfoHandlerTestSuite) TestHandleUserInfo_BEARERUpperCase() {
 		"sub": "user123",
 	}
 
-	s.mockService.On("GetUserInfo", "valid-token").Return(userInfo, nil)
+	s.mockService.On("GetUserInfo", "valid-token").Return(jsonResponse(userInfo), nil)
 
 	s.handler.HandleUserInfo(rr, req)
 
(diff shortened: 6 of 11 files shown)


backend/internal/application/service.go
```diff
@@ -1334,11 +1335,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
 
 	if len(app.InboundAuthConfig) > 0 && app.InboundAuthConfig[0].OAuthAppConfig != nil &&
 		app.InboundAuthConfig[0].OAuthAppConfig.UserInfo != nil {
-		oauthUserInfo.UserAttributes = app.InboundAuthConfig[0].OAuthAppConfig.UserInfo.UserAttributes
+		input := app.InboundAuthConfig[0].OAuthAppConfig.UserInfo
+		oauthUserInfo.UserAttributes = input.UserAttributes
+		responseType := strings.ToUpper(input.ResponseType)
+
+		switch responseType {
+		case "", "JSON":
```
thiva-k: Can't we simplify this as?

```
case "JWS", "JWE":
    oauthUserInfo.ResponseType = responseType
default:
    oauthUserInfo.ResponseType = "JSON"
```

nandhu-kumar: @thiva-k Updated the logic.

thiva-k:
We should add the `response_type` to the `userInfoConfig`  model in the store layer too and add it to relevant store functions as well. Seems like currently it is being returned from the application cache, which stores the application struct during the create request at the service layer.

thiva-k:
Shall we also squash the commits?

backend/internal/application/model/constants.go
```diff
@@ -28,6 +28,17 @@ const (
 	OAuthInboundAuthType InboundAuthType = "oauth2"
 )
 
+const (
```
ThaminduDilshan, line 34: Shall we define a go type for this also similar to ` type InboundAuthType string`?

backend/internal/application/service.go
```diff
@@ -1367,11 +1368,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
 
 	if len(app.InboundAuthConfig) > 0 && app.InboundAuthConfig[0].OAuthAppConfig != nil &&
 		app.InboundAuthConfig[0].OAuthAppConfig.UserInfo != nil {
-		oauthUserInfo.UserAttributes = app.InboundAuthConfig[0].OAuthAppConfig.UserInfo.UserAttributes
+		input := app.InboundAuthConfig[0].OAuthAppConfig.UserInfo
```
ThaminduDilshan:
```suggestion
		userInfoConfigInput := app.InboundAuthConfig[0].OAuthAppConfig.UserInfo
```

backend/internal/application/service.go
```diff
@@ -1367,11 +1368,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
...
+		oauthUserInfo.UserAttributes = input.UserAttributes
+		responseType := strings.ToUpper(input.ResponseType)
+
+		switch responseType {
+		case model.UserInfoResponseTypeJWS,
+			model.UserInfoResponseTypeJWE:
+			oauthUserInfo.ResponseType = responseType
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/application/service.go
```diff
@@ -1367,11 +1368,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
...
+		default:
+			oauthUserInfo.ResponseType = model.UserInfoResponseTypeJSON
+		}
 	}
 	if oauthUserInfo.UserAttributes == nil {
 		oauthUserInfo.UserAttributes = idTokenConfig.UserAttributes
 	}
+	if oauthUserInfo.ResponseType == "" {
```
ThaminduDilshan, line 1390: We don't need to check response type is empty here as switch-case defaults to JSON right ?

nandhu-kumar: The switch defaults to JSON only when UserInfo is present.
If UserInfo is not configured at all, we never enter the block and ResponseType remains empty.
Therefore, the final empty check is required to enforce the default behavior.

backend/internal/application/model/constants.go
nandhu-kumar: updated it

backend/internal/application/service.go
nandhu-kumar: changed it

backend/internal/application/service.go
nandhu-kumar: removed the line

tests/integration/application/model.go
```diff
@@ -79,6 +79,7 @@ type OAuthTokenConfig struct {
 
 // UserInfoConfig represents the UserInfo endpoint configuration.
 type UserInfoConfig struct {
+	ResponseType   string   `json:"response_type,omitempty"`
 	UserAttributes []string `json:"user_attributes,omitempty"`
 }
```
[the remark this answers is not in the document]
nandhu-kumar: Included ResponseType in UserInfo equality checks as suggested.

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,48 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+		)
+		if err != nil {
+			s.logger.Error("Failed to generate signed UserInfo JWT")
+			return nil, &serviceerror.InternalServerError
+		}
+
+		return &UserInfoResponse{
+			ResponseType: appmodel.UserInfoResponseTypeJWS,
+			JWTBody:      signedJWT,
+			ContentType:  "application/jwt",
```
thiva-k: Can we define this as a constant?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,48 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+		}, nil
+	}
+
+	return &UserInfoResponse{
+		ResponseType: appmodel.UserInfoResponseTypeJSON,
+		JSONBody:     response,
+		ContentType:  "application/json",
```
thiva-k: Same as above

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,48 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+		issuer := ""
+		if oauthApp != nil && oauthApp.Token != nil {
+			issuer = oauthApp.Token.Issuer
+		}
+
+		signedJWT, _, err := s.jwtService.GenerateJWT(
+			sub,
+			clientID,
+			issuer,
+			0,
```
thiva-k: Can we use the default validity period from our configurations, rather than having no validity period

backend/internal/application/model/constants.go
```diff
@@ -28,6 +28,20 @@ const (
...
+
+const (
+	// UserInfoResponseTypeJSON represents the JSON userinfo response type.
+	UserInfoResponseTypeJSON UserInfoResponseType = "JSON"
+
+	// UserInfoResponseTypeJWS represents the JWS userinfo response type.
+	UserInfoResponseTypeJWS UserInfoResponseType = "JWS"
+
+	// UserInfoResponseTypeJWE represents the JWE userinfo response type.
+	UserInfoResponseTypeJWE UserInfoResponseType = "JWE"
```
thiva-k: There is a case in the spec where the response can be signed and encrypted. Can we have a type for that also ? (e.g. SignedJWE)

tests/integration/application/model.go
```diff
@@ -79,6 +79,7 @@ type OAuthTokenConfig struct {
 
 // UserInfoConfig represents the UserInfo endpoint configuration.
 type UserInfoConfig struct {
```
thiva-k, line 80: Shouldn't we also have configs to set the signing, encryption algorithms? May be support existing algorithms as of now

backend/internal/oauth/oauth2/userinfo/model.go
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package userinfo
+
+import appmodel "github.com/asgardeo/thunder/internal/application/model"
+
+// UserInfoResponse represents the structured response returned by the
+// UserInfo service. It supports both JSON and JWS response types.
+type UserInfoResponse struct {
+	ResponseType appmodel.UserInfoResponseType
```
thiva-k: Do we need both ResponseType and ContentType? Both carry the same information right?

tests/integration/application/model.go
nandhu-kumar: @thiva-k -: For this MVP implementation, the signing algorithm is derived from the server-level JWT configuration and not exposed as a per-application setting.

We can extend UserInfoConfig in a future enhancement to support algorithm-level configurability together with full JWE support.

backend/internal/application/model/constants.go
nandhu-kumar: For this MVP, we are limiting implementation to JSON and JWS response types.
Nested JWT (sign-then-encrypt) support can be introduced together with full JWE implementation in a follow-up enhancement.

@ThaminduDilshan Correct me if i am wrong

thiva-k: We can then remove the JWE type too, if we are not planning to support it anytime soon. WDYT?

tests/integration/application/model.go
thiva-k: Sure. 

Also we have onboarded the basic JWE utilities recently, which can be used for JWE implementation in later efforts

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: updated it

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: updated it

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: updated it. pls review once

backend/internal/application/model/constants.go
nandhu-kumar: removed JWE.

backend/internal/oauth/oauth2/userinfo/model.go
nandhu-kumar: removed responsetype and handled it using contenttype.

backend/internal/oauth/oauth2/userinfo/model.go
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package userinfo
+
+// UserInfoResponse represents the structured response returned by the
+// UserInfo service. It supports both JSON and JWS response types.
+type UserInfoResponse struct {
+	JSONBody    map[string]interface{}
+	JWTBody     string
+	ContentType string
```
ThaminduDilshan: Instead of adding ContentType here and handle in the service layer, shall we handle content type in the handler based on whether JWTBody or JSONBody is present?

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -176,6 +176,12 @@ const (
 	TokenTypeJWT    = "JWT"
 )
 
+// OAuth2 content types.
+const (
+	ContentTypeApplicationJWT  = "application/jwt"
```
ThaminduDilshan: Shall we define this in `internal/system/constants/server_constants.go` file? There's already a constant for "application/json" there

backend/internal/oauth/oauth2/userinfo/handler.go
```diff
@@ -64,7 +64,13 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
 	w.Header().Set(serverconst.CacheControlHeaderName, serverconst.CacheControlNoStore)
 	w.Header().Set(serverconst.PragmaHeaderName, serverconst.PragmaNoCache)
 
-	utils.WriteSuccessResponse(w, http.StatusOK, userInfo)
+	if result.ContentType == constants.ContentTypeApplicationJWT {
+		w.Header().Set("Content-Type", result.ContentType)
```
ThaminduDilshan: Let's use ContentTypeHeaderName from internal/system/constants/server_constants.go rather than hard coding it here

backend/internal/oauth/oauth2/userinfo/model.go
ThaminduDilshan: Alternatively you can keep the previous `ResponseType` and use it to determine applicable type

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,55 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
 		return nil, svcErr
 	}
 
-	return response, nil
+	// Decide response type
+	var responseType appmodel.UserInfoResponseType = appmodel.UserInfoResponseTypeJSON
```
ThaminduDilshan:
```suggestion
	responseType := appmodel.UserInfoResponseTypeJSON
```

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,55 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+	if oauthApp != nil && oauthApp.UserInfo != nil {
+		responseType = oauthApp.UserInfo.ResponseType
+	}
+
+	// JWS Case
+	if responseType == appmodel.UserInfoResponseTypeJWS {
+		clientID := ""
```
ThaminduDilshan: Shall we extract out the implementation inside this if condition into a private helper method?

Something like;
```go
func (s *userInfoService) signUserInfoResponse(...
```

or 
```go
func (s *userInfoService) generateJWSUserInfo(...
```

backend/internal/oauth/oauth2/userinfo/model.go
nandhu-kumar: removed contenttype and handled it with response body.

backend/internal/oauth/oauth2/constants/constants.go
nandhu-kumar: updated it.

backend/internal/oauth/oauth2/userinfo/handler.go
nandhu-kumar: modified it.

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: updated

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: created the helper and handled the signing logic inside it.

backend/internal/oauth/oauth2/userinfo/handler.go
```diff
@@ -64,7 +64,14 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
 	w.Header().Set(serverconst.CacheControlHeaderName, serverconst.CacheControlNoStore)
 	w.Header().Set(serverconst.PragmaHeaderName, serverconst.PragmaNoCache)
 
-	utils.WriteSuccessResponse(w, http.StatusOK, userInfo)
+	if result.JWTBody != "" {
+		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJWT)
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(result.JWTBody))
+		return
```
ThaminduDilshan: Shall we remove this return statement? Otherwise `h.logger.Debug("UserInfo response sent successfully")` will not be hit.

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +115,60 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+	if responseType == appmodel.UserInfoResponseTypeJWS {
+		return s.generateJWSUserInfo(sub, tokenClaims, oauthApp, response)
+	}
+
+	return &UserInfoResponse{
+		JSONBody: response,
+	}, nil
+}
+
+func (s *userInfoService) generateJWSUserInfo(
```
ThaminduDilshan, line 137: Let's add a method comment

nandhu-kumar: added method comment.

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -68,7 +68,7 @@ func newUserInfoService(
 }
 
 // GetUserInfo validates the access token and returns user information based on authorized scopes.
-func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{}, *serviceerror.ServiceError) {
+func (s *userInfoService) GetUserInfo(accessToken string) (*UserInfoResponse, *serviceerror.ServiceError) {
```
sacrana0, line 72: I think the design of this method could be better. Instead of returning UserInfoResponse which has json and jws field, we can do one of the following:
1. UserInfoResponse has two fields - type and value(string)
2. We directly return  responseType, userinfo(string), error

json, jws or jwe all can be returned as a string.

ThaminduDilshan: But the value cannot be a string when it's a json response. In that case one option is to use the generic interface{} type return value which is unsafe. Or else we need to return a JSON string. 
(Each option has it's own trade offs including the currently implemented one)

sacrana0: @ThaminduDilshan, I was talking json string only. I don't think there is any issue with that. Currently implemented way does not establish the constraint that both jws and json will not be present in the struct. At first I thought looking at code that both json and signed values are returned from the function.

ThaminduDilshan: Agreed with the concern on current approach. But IMO returning JSON string is not nice for a service level method and may raise issues with white spaces, escape characters,  etc

ThaminduDilshan: @coderabbitai what's your opinion on this?

ThaminduDilshan: If we add that type back, it will reduce the confusion. Then handler can rely on this type to read the correct field and write Response Type header to the response 

```go
type UserInfoResponse struct {
    Type UserInfoType
	JSONBody map[string]interface{}
	JWTBody  string
}
```

@sacrana0 WDYT?

sacrana0: Lets leave it as it is, in that case.

nandhu-kumar: as discussed, updated struct with Type

type UserInfoResponse struct {
    Type UserInfoType
	JSONBody map[string]interface{}
	JWTBody  string
}

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -115,7 +116,61 @@ func (s *userInfoService) GetUserInfo(accessToken string) (map[string]interface{
...
+	}
+
+	issuer := config.GetThunderRuntime().Config.JWT.Issuer
+
+	var validity int64
+	if oauthApp != nil &&
+		oauthApp.Token != nil &&
+		oauthApp.Token.IDToken != nil &&
+		oauthApp.Token.IDToken.ValidityPeriod > 0 {
+		validity = oauthApp.Token.IDToken.ValidityPeriod
```
ThaminduDilshan: Is it correct/ usual practice to use the id token validity period for the userinfo JWT?

nandhu-kumar: Good point. I reused the ID token validity for now, but I agree that UserInfo JWS is separate from the ID token.

Would it be better to just pass 0 here and let GenerateJWT use the default JWT validity from the runtime config (same way it already handles it)?

nandhu-kumar: @thiva-k Earlier you suggested using the config, but I mistakenly used the ID token validity here — that’s my mistake.
If we are using config, should this rely on Config.JWT.ValidityPeriod?
Or would you prefer passing 0 and letting GenerateJWT fall back to the default runtime JWT validity?

ThaminduDilshan: I prefer to use `Config.JWT.ValidityPeriod` rather than passing 0. Logic makes clear this way.

nandhu-kumar: @ThaminduDilshan so for now shall i go ahead with "Config.JWT.ValidityPeriod"?

ThaminduDilshan: +1

nandhu-kumar: updated as per the above discussion.

hwupathum:
⚠️ **Note on JWE:** No encryption logic is implemented in this PR. The UserInfo endpoint does not produce JWE tokens. JWE encryption support will be added in a follow-up PR.

PR #1611, opened 2026-02-28
Fix: Replace xxd with openssl in build.sh

### LINKED ISSUE #1578: XXD needed to be installed manually on Debian 12
### Description

`XXD` which creates a hex dump of a given file or standard input, is used in the project but not listed in the pre-requisite list.
When **building** the project on a Debian 12 system, the script fails giving an error.

I would suggest to have `xxd` on the prerequisites list.

### Steps to Reproduce

Cloning the repository and running 

> make all 

on Debian 12

### Version

bc0ebe0

Debian 12

RandithaK:
Removes the xxd dependency by using openssl in build.sh as openssl is already a guaranteed dependency. Fixes #1578

## Summary by CodeRabbit

* **Chores**
  * Switched build-time cryptographic key generation to an industry-standard tool for more consistent key creation.
  * Updated build logging and error messages to reflect the new generation method.
  * Preserves existing behavior: if no key exists, one is generated and persisted; no user-facing workflow changes.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -910,12 +910,10 @@ function ensure_crypto_file() {
     else
         echo "Default crypto key file not found. Generating new key at $KEY_FILE..."
         
-        # Generate 32-byte key (64 hex characters) using /dev/urandom
+        # Generate 32-byte key (64 hex characters) using openssl
         local NEW_KEY
-        NEW_KEY=$(head -c 32 /dev/urandom | xxd -p -c 32)
-
-        if [[ -z "$NEW_KEY" ]]; then
-            echo "ERROR: Failed to generate crypto key from /dev/urandom."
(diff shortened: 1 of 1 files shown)


ThaminduDilshan:
Also let's squash to a single commit

darshanasbg:
@hwupathum Thunder `build.sh` used to use openssl for the key generation, but i has been changed with [1]. That PR doesn't indicate any specific reason for that change.. Was there any specific reason to make that change? Do you see any issue moving back to openssl?

A side note, the powershell script use openssl.

[1] https://github.com/asgardeo/thunder/pull/937/changes#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823

PR #1645, opened 2026-03-03
Fix OAuth2 redirect and CORS for local dev mode

### LINKED ISSUE #1644: OAuth2 flow fails in dev mode due to missing gate_client config and CORS origins
## Description

When running Thunder in local development mode (`make run`), the OAuth2 authorization flow fails because:

1. **Missing `gate_client` configuration** — The default `gate_client` config points to port 8090 (the backend), but in dev mode the Gate frontend runs as a separate Vite dev server on port 5190. This causes the OAuth2 flow to redirect to `https://localhost:8090/gate/signin` instead of `https://localhost:5190/gate/signin`, resulting in a 404 since the backend does not serve the Gate frontend in dev mode.

2. **Missing CORS origins** — The frontend dev servers on ports 5190 and 5191 are not included in the default CORS allowed origins, which can cause cross-origin request failures.

## Steps to Reproduce

1. Run `make run` to start Thunder in dev mode
2. Navigate to `https://localhost:5191/develop` in a browser
3. The OAuth2 flow redirects to `https://localhost:8090/gate/signin` → 404

## Expected Behavior

The OAuth2 flow should redirect to `https://localhost:5190/gate/signin` (the Gate Vite dev server), where the sign-in form loads correctly.

## Fix

Add `gate_client` configuration in `deployment.yaml` pointing to port 5190, and add CORS allowed origins for ports 5190 and 5191.

RandithaK:
Fixes #1644

## Summary

When running Thunder in local dev mode (`make run`), the OAuth2 flow redirects to `https://localhost:8090/gate/signin` (the backend) instead of `https://localhost:5190/gate/signin` (the Gate Vite dev server), resulting in a 404.

## Changes

- **`deployment.yaml`**: Add `gate_client` config pointing to port 5190 so the OAuth2 flow redirects to the Gate frontend dev server
- **`deployment.yaml`**: Add CORS allowed origins for frontend dev server ports 5190 and 5191

## Summary by CodeRabbit

* **New Features**
  * Added a development configuration override system allowing local developers to customize settings without touching production configs.

* **Documentation**
  * Added detailed README and contributing guide instructions for creating and using the local development override and run workflows.

* **Chores**
  * Updated build/packaging and run scripts to use and exclude the development override from production artifacts.
  * Added the development override to ignore rules.

* **Tests**
  * Added extensive tests covering override loading, path derivation, merging, and error cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -113,3 +113,6 @@ backend/coverage_unit.html
 # IDE settings
 .cursor/rules/nx-rules.mdc
 .github/instructions/nx.instructions.md
+
+# Dev configuration override (local developer settings)
+deployment.dev.yaml
--- a/Makefile
+++ b/Makefile
@@ -92,10 +92,10 @@ build_with_coverage_only:
 	@echo "================================================================"
 
 run:
-	./build.sh run $(OS) $(ARCH)
+	THUNDER_PROFILE=dev ./build.sh run $(OS) $(ARCH)
 
 run_backend:
-	./build.sh run_backend $(OS) $(ARCH)
+	THUNDER_PROFILE=dev ./build.sh run_backend $(OS) $(ARCH)
 
 run_frontend:
 	./build.sh run_frontend $(OS) $(ARCH)
--- a/README.md
+++ b/README.md
@@ -400,6 +400,52 @@ The product will now use the PostgreSQL database for its operations.
 
 </details>
 
+<details>
+<summary><h3>Development Configuration Override</h3></summary>
+
+Thunder supports a `deployment.dev.yaml` file that lets you override values from `deployment.yaml` for local development without changing the main configuration.
+
+#### Setup
+
+1. Navigate to the configuration directory:
+
+    ```bash
+    cd backend/cmd/server/repository/conf/
+    ```
+
+2. Copy the example file:
+
+    ```bash
+    cp deployment.dev.yaml.example deployment.dev.yaml
+    ```
+
+3. Edit `deployment.dev.yaml` with only the values you want to override:
+
+    ```yaml
+    # Example: Override the server port for local development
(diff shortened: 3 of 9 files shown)


backend/cmd/server/repository/conf/deployment.yaml
```diff
@@ -2,6 +2,12 @@ server:
   hostname: "localhost"
   port: 8090
 
+gate_client:
```
ThaminduDilshan: We shouldn't add this to server deployment config file. This is only for development time and is not applicable for release distributions. Whatever configs added here should be only for distributions.

Developers have to follow the [contribution guideline](https://asgardeo.github.io/thunder/docs/community/contributing/contributing-code/configure-and-run) and add the respective configs manually

senthalan: Agreed with Thamindu, we can't have this added to the default deployment.yaml

@RandithaK, Can you look into the possibility of having another deployment.yaml (something like - dev.deployment.yaml) which can be used for the dev mode ? So it will be easier for the developers. But this new file shouldn't be packed within the thunder distribution. This is something you can research on if you have time.

darshanasbg: IMO, we should work on this in a discussion mode first.. The problem is clear, and we already have way of solving that. As we are trying to explore more convenient solution, we have first discuss and evaluate and select an approach.

@RandithaK The updated approach you are taking now, Thunder production/release distributions are capable of work in the dev mode, even its not activated by default. Unless its intentional to provide such capability to our released distributions, we should avoid having development time capabilities in the released distributions.

So, lets discuss the constraints, solutions etc in the git discussion thread and come to agreement of the solution and then work on the code.

PR #1604, opened 2026-02-27
1551 | Add nonce support for OIDC authorization code flow

nandhu-kumar:
### Purpose

This PR adds full OpenID Connect (OIDC) `nonce` support to the OAuth2 Authorization Code flow.

Previously, the `nonce` parameter sent in the `/oauth2/authorize` request was not persisted through the authorization code flow and was not included in the issued ID token.

This change ensures:

- The `nonce` parameter is accepted at the authorization endpoint
- The `nonce` is persisted within the authorization code
- The `nonce` is propagated to the ID token during token issuance
- The ID token includes the `nonce` claim when provided

This aligns the implementation with OpenID Connect Core specification requirements for replay protection and ID token validation.

---

### Approach

The implementation follows the standard OIDC flow design:

1. **Authorization Endpoint**
   - Capture the `nonce` parameter from the authorization request.
   - Store it inside the authorization request context.

2. **Authorization Code Persistence**
   - Extend the authorization code model to persist `nonce`.
   - Ensure JSON serialization/deserialization includes the new field.
   - Add unit test coverage for authorization code persistence.

3. **Token Service (ID Token Builder)**
   - Extend `IDTokenBuildContext` to include `Nonce`.
   - Populate the nonce from the authorization code during token exchange.
   - Add the `nonce` claim to the ID token only if it was originally requested.

4. **Testing**
   - Added unit tests for:
     - Authorization code JSON persistence
     - ID token building logic including nonce
   - Added integration test:
     - End-to-end authorization code flow with `nonce`
     - Validates that the issued ID token contains the expected `nonce` claim

No breaking changes were introduced.

---

### Related Issues
- https://github.com/asgardeo/thunder/issues/1551

### Related PRs
- N/A

### Checklist
- [x] Followed the contribution guidelines.
- [x] Manual test round performed and verified.
- [ ] Documentation provided. (Add links if there are any)
    - [ ] Ran Vale and fixed all errors and warnings
- [x] Tests provided.
    - [x] Unit Tests
    - [x] Integration Tests
- [ ] Breaking changes.
    - [ ] Breaking changes section filled.
    - [ ] `breaking change` label added.

### Security checks
- [x] Followed secure coding standards in WSO2 Secure Coding Guidelines.
- [x] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.

## Summary by CodeRabbit

* **New Features**
[...]

PR #1664, opened 2026-03-05
Add strict display attribute validation with dot-notation support in user schema

### LINKED ISSUE #1513: Add display attribute validation in user schema service
Parent issue: #1511

## Scope

Add validation logic for the `displayAttribute` field during user schema create and update operations.

## Tasks

- [ ] Add error constants:
  - `ErrorInvalidDisplayAttribute` (USRS-1008): "Display attribute must reference a top-level string attribute that is not a credential type"
  - `ErrorMissingDisplayAttribute` (USRS-1009): "A display attribute must be specified when the schema has multiple non-credential string attributes"
- [ ] Add validation in `CreateUserSchema` and `UpdateUserSchema` (after `model.CompileUserSchema` succeeds):
  1. Parse schema JSON → collect top-level string attributes
  2. Filter out credential types (`password`, `pin`, `secret`, `passkey`)
  3. **Single eligible** → auto-set if empty; validate if provided
  4. **Multiple eligible** → require explicit `displayAttribute`; validate in eligible set
  5. **Zero eligible** → allow empty (fallback to user ID at display time)

## Files

- `backend/internal/userschema/service.go`
- `backend/internal/userschema/error_constants.go`

## Design Discussion

#1510

DonOmalVindula:
## Summary
- Validate that `systemAttributes.display`, when provided in create/update requests, references a valid attribute in the schema definition
- Support dot-notation paths for nested attributes (e.g., `address.city`, `profile.name.first`)
- Three-layer validation: existence (USRS-1011) → displayability (USRS-1012) → non-credential (USRS-1013)
- Only `string` and `number` types are considered displayable (boolean, object, array are rejected)
- The field remains fully optional — omitting it raises no error
- Add `resolvePropertyPath()` on compiled `Schema` to walk nested object trees via dot-notation
- Rewrite `HasAttribute()`, `IsDisplayableAttribute()`, `IsCredentialAttribute()` to delegate to the path resolver
- Refactor `validateUserSchemaDefinition` to return the compiled schema (avoids double compilation)
- Add `validateDisplayAttribute()` helper integrated into `CreateUserSchema`, `UpdateUserSchema`, and declarative resource loading
- Add error constants: `ErrorInvalidDisplayAttribute` (USRS-1011), `ErrorNonDisplayableAttribute` (USRS-1012), `ErrorCredentialDisplayAttribute` (USRS-1013)

## Test plan
- [x] Unit tests for `HasAttribute()` across all property types and dot-notation paths (nested objects, deeply nested, non-existent leaf, intermediate non-object, intermediate array)
- [x] Unit tests for `IsDisplayableAttribute()` — flat and nested paths, string/number displayable, boolean/object/array not displayable
- [x] Unit tests for `IsCredentialAttribute()` — flat and nested paths, credential vs non-credential
- [x] Unit tests for `validateDisplayAttribute()` — nil systemAttributes, empty display, valid refs, non-existent attribute, non-displayable type, credential attribute, dot-notation paths
- [x] Integration test through `CreateUserSchema` and `UpdateUserSchema` with invalid display attributes
- [x] Declarative resource validation covers display attribute checks during init
- [x] All existing tests pass (`go test ./internal/userschema/...`)
- [x] Build passes (`go build ./...`)
- [x] Vet passes (`go vet ./...`)

Closes #1513

## Summary by CodeRabbit

* **New Features**
  * Display-attribute validation: display attribute must exist, be a displayable primitive (string/number), and not be a credential; three new user-facing error codes cover invalid cases.
  * Schema now provides attribute existence, displayability, and credential-status checks (including nested/dotted paths).

* **Tests**
  * Expanded test coverage for display-attribute validation and attribute checks across types, nesting, and edge cases.

### DIFF
--- a/backend/internal/userschema/declarative_resource.go
+++ b/backend/internal/userschema/declarative_resource.go
@@ -30,6 +30,7 @@ import (
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/log"
 	"github.com/asgardeo/thunder/internal/system/security"
+	"github.com/asgardeo/thunder/internal/userschema/model"
 
 	"gopkg.in/yaml.v3"
 )
@@ -227,12 +228,19 @@ func validateUserSchema(schemaDTO *UserSchema, ouService oupkg.OrganizationUnitS
 			schemaDTO.OrganizationUnitID, schemaDTO.Name)
 	}
 
-	// Validate schema is valid JSON
-	if len(schemaDTO.Schema) > 0 {
-		var testSchema map[string]interface{}
-		if err := json.Unmarshal(schemaDTO.Schema, &testSchema); err != nil {
-			return fmt.Errorf("invalid schema JSON for user schema '%s': %w", schemaDTO.Name, err)
-		}
+	// Validate schema definition is present and valid.
+	if len(schemaDTO.Schema) == 0 {
+		return fmt.Errorf("schema definition is required for user schema '%s'", schemaDTO.Name)
+	}
+
+	compiledSchema, compileErr := model.CompileUserSchema(schemaDTO.Schema)
+	if compileErr != nil {
+		return fmt.Errorf("invalid schema for user schema '%s': %w", schemaDTO.Name, compileErr)
+	}
+
+	if svcErr := validateSystemAttributes(compiledSchema, schemaDTO.SystemAttributes); svcErr != nil {
+		return fmt.Errorf("invalid system attributes for user schema '%s': %s",
+			schemaDTO.Name, svcErr.ErrorDescription)
 	}
 
 	return nil
--- a/backend/internal/userschema/error_constants.go
+++ b/backend/internal/userschema/error_constants.go
@@ -99,6 +99,31 @@ var (
 		Error:            "Consent synchronization failed",
 		ErrorDescription: "Failed to synchronize consent configurations for the user schema",
 	}
+	// ErrorInvalidDisplayAttribute is the error returned when the display attribute
+	// does not reference a valid top-level attribute in the schema.
+	ErrorInvalidDisplayAttribute = serviceerror.ServiceError{
+		Type:  serviceerror.ClientErrorType,
+		Code:  "USRS-1011",
+		Error: "Invalid display attribute",
+		ErrorDescription: "Display attribute must reference an attribute defined in the schema " +
+			"(use dot notation for nested attributes, e.g. 'address.city')",
+	}
+	// ErrorNonDisplayableAttribute is the error returned when the display attribute
+	// references an attribute with a non-displayable type (e.g. object or array).
+	ErrorNonDisplayableAttribute = serviceerror.ServiceError{
+		Type:             serviceerror.ClientErrorType,
+		Code:             "USRS-1012",
+		Error:            "Non-displayable attribute type",
+		ErrorDescription: "Display attribute must reference a string or number type",
+	}
+	// ErrorCredentialDisplayAttribute is the error returned when the display attribute
+	// references an attribute marked as a credential.
+	ErrorCredentialDisplayAttribute = serviceerror.ServiceError{
+		Type:             serviceerror.ClientErrorType,
+		Code:             "USRS-1013",
+		Error:            "Credential attribute not allowed as display",
+		ErrorDescription: "Display attribute must not reference a credential attribute",
+	}
 )
 
 // Server errors for user schema management operations.
--- a/backend/internal/userschema/model/boolean.go
+++ b/backend/internal/userschema/model/boolean.go
@@ -38,6 +38,10 @@ func (p *boolean) isCredential() bool {
 	return false
 }
 
+func (p *boolean) isDisplayable() bool {
+	return false
+}
+
 func (p *boolean) validateValue(value interface{}, path string, logger *log.Logger) (bool, error) {
 	_, ok := value.(bool)
 	if !ok {
--- a/backend/internal/userschema/model/schema.go
+++ b/backend/internal/userschema/model/schema.go
@@ -21,6 +21,7 @@ package model
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/asgardeo/thunder/internal/system/log"
 )
@@ -42,6 +43,7 @@ const (
 type property interface {
 	isRequired() bool
 	isCredential() bool
+	isDisplayable() bool
 	validateValue(value interface{}, path string, logger *log.Logger) (bool, error)
 	validateUniqueness(value interface{}, path string,
 		identifyUser func(map[string]interface{}) (*string, error), logger *log.Logger) (bool, error)
@@ -50,6 +52,63 @@ type property interface {
 // Schema represents a user schema with a set of properties.
 type Schema struct {
 	properties map[string]property
+}
+
+// getPropertyByPath returns the property at the given dot-notation path
+// (e.g. "address.city") by walking through nested object types. For a simple
+// (non-dotted) name, it returns the top-level property directly.
+func (cs *Schema) getPropertyByPath(path string) (property, bool) {
+	segments := strings.Split(path, ".")
+	currentProps := cs.properties
+
+	for i, segment := range segments {
+		prop, exists := currentProps[segment]
+		if !exists {
+			return nil, false
+		}
+
+		if i == len(segments)-1 {
+			return prop, true
+		}
+
+		obj, ok := prop.(*object)
+		if !ok {
+			return nil, false
+		}
+		currentProps = obj.properties
+	}
+
+	return nil, false
+}
+
+// DisplayAttributeStatus represents the result of validating an attribute as a display attribute.
+type DisplayAttributeStatus int
+
+const (
+	// DisplayAttributeValid indicates the attribute is valid for use as a display attribute.
+	DisplayAttributeValid DisplayAttributeStatus = iota
+	// DisplayAttributeNotFound indicates the attribute does not exist in the schema.
+	DisplayAttributeNotFound
+	// DisplayAttributeNotDisplayable indicates the attribute type is not displayable.
+	DisplayAttributeNotDisplayable
+	// DisplayAttributeIsCredential indicates the attribute is marked as a credential.
+	DisplayAttributeIsCredential
+)
+
+// ValidateAsDisplayAttribute resolves the path once and checks existence, displayability,
+// and credential status in a single pass.
+func (cs *Schema) ValidateAsDisplayAttribute(name string) DisplayAttributeStatus {
+	prop, exists := cs.getPropertyByPath(name)
+	if !exists {
+		return DisplayAttributeNotFound
+	}
+	if !prop.isDisplayable() {
+		return DisplayAttributeNotDisplayable
+	}
+	if prop.isCredential() {
+		return DisplayAttributeIsCredential
+	}
+	return DisplayAttributeValid
 }
 
 // GetCredentialAttributes returns the names of top-level properties marked as credentials.
--- a/backend/internal/userschema/service.go
+++ b/backend/internal/userschema/service.go
@@ -188,6 +188,7 @@ func (us *userSchemaService) CreateUserSchema(
 	schemaToValidate := UserSchema{
 		Name:               request.Name,
 		OrganizationUnitID: request.OrganizationUnitID,
+		SystemAttributes:   request.SystemAttributes,
 		Schema:             request.Schema,
 	}
 	if validationErr := validateUserSchemaDefinition(schemaToValidate); validationErr != nil {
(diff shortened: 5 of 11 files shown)


backend/internal/userschema/service_test.go
```diff
@@ -779,3 +780,127 @@ func (s *GetCredentialAttributesTestSuite) TestStoreError_ReturnsInternalError()
...
+func TestValidateDisplayAttribute_ValidBooleanAttribute(t *testing.T) {
+	compiled, err := model.CompileUserSchema(json.RawMessage(`{
+		"email":{"type":"string"},
+		"active":{"type":"boolean"}
+	}`))
+	require.NoError(t, err)
+
+	svcErr := validateDisplayAttribute(compiled, &SystemAttributes{Display: "active"})
+	require.Nil(t, svcErr)
+}
```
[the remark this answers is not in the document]
DonOmalVindula: This is intentional. Our design decision is to allow all primitive types (string, number, and boolean) as valid display attributes — not just string. The display attribute is meant to show a human-readable identifier in the UI, and numbers (e.g., employee ID) and booleans (e.g., active status) are valid use cases.

The validation enforces three checks in `validateDisplayAttribute()`:
1. **Existence** — attribute must exist in the schema (USRS-1010)
2. **Displayability** — must be a primitive type: string, number, or boolean; rejects object/array (USRS-1011)
3. **Non-credential** — must not be marked as `credential: true`, since credentials are not included in user responses (USRS-1012)

backend/internal/userschema/error_constants.go
```diff
@@ -82,6 +82,14 @@ var (
...
 		ErrorDescription: "The user schema is declarative and cannot be modified or deleted",
 	}
+	// ErrorInvalidDisplayAttribute is the error returned when the display attribute
+	// does not reference a valid top-level attribute in the schema.
+	ErrorInvalidDisplayAttribute = serviceerror.ServiceError{
+		Type:             serviceerror.ClientErrorType,
+		Code:             "USRS-1010",
+		Error:            "Invalid display attribute",
+		ErrorDescription: "Display attribute must reference a top-level attribute defined in the schema",
+	}
```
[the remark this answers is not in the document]
DonOmalVindula: This is intentional. The `ErrorInvalidDisplayAttribute` error constant only covers the **existence** check (step 1 of 3). The displayability and non-credential checks have their own separate error constants (`ErrorNonDisplayableAttribute` USRS-1012 and `ErrorCredentialDisplayAttribute` USRS-1013). Each error description accurately reflects its specific check. Also, our design intentionally allows all primitive types (string, number, boolean) as display attributes — not just string (see discussion on comment thread below).

backend/internal/userschema/service.go
```diff
@@ -655,37 +665,51 @@ func logAndReturnServerError(
...
+func validateDisplayAttribute(
+	compiledSchema *model.Schema, systemAttrs *SystemAttributes,
+) *serviceerror.ServiceError {
+	if systemAttrs == nil || systemAttrs.Display == "" {
+		return nil
+	}
+	if !compiledSchema.HasAttribute(systemAttrs.Display) {
+		return &ErrorInvalidDisplayAttribute
+	}
 	return nil
```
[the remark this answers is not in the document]
DonOmalVindula: This is not applicable. The display attribute is fully **optional** — when omitted, the UI falls back to showing user ID. There is no auto-population or required-when-multiple behavior in our design. The validation has three intentional layers when a display value IS provided:

1. **Existence** — attribute must exist in schema (USRS-1011)
2. **Displayability** — must be a primitive type: string, number, or boolean (USRS-1012)
3. **Non-credential** — must not be marked `credential: true` (USRS-1013)

All three checks are implemented. The empty/nil display case correctly returns nil (no error) because it's a valid state.

backend/internal/userschema/model/schema.go
```diff
@@ -52,6 +53,22 @@ type Schema struct {
...
+
+// IsDisplayableAttribute checks if the named attribute exists and has a displayable type
+// (string, number, or boolean). Complex types like object and array are not displayable.
+func (cs *Schema) IsDisplayableAttribute(name string) bool {
+	prop, exists := cs.properties[name]
+	if !exists {
+		return false
+	}
+	return prop.isDisplayable()
+}
```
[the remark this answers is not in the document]
DonOmalVindula: This is intentional. `IsDisplayableAttribute` is deliberately separated from the credential check — these are two independent concerns:

- **`IsDisplayableAttribute`** answers: "Is this attribute type suitable for display?" → Primitive types (string, number, boolean) return true; complex types (object, array) return false.
- **`IsCredentialAttribute`** answers: "Is this attribute marked as a credential?"

The caller (`validateDisplayAttribute`) chains these as separate checks with distinct error codes (USRS-1012 for non-displayable type, USRS-1013 for credential). Merging credential logic into `IsDisplayableAttribute` would conflate two orthogonal concerns and prevent returning the correct specific error to the API consumer.

backend/internal/userschema/error_constants.go
```diff
@@ -99,6 +99,31 @@ var (
...
+	// ErrorCredentialDisplayAttribute is the error returned when the display attribute
+	// references an attribute marked as a credential.
+	ErrorCredentialDisplayAttribute = serviceerror.ServiceError{
+		Type:  serviceerror.ClientErrorType,
+		Code:  "USRS-1013",
+		Error: "Credential attribute not allowed as display",
+		ErrorDescription: "Display attribute must not reference a credential attribute, " +
+			"as credentials are not included in user responses",
+	}
```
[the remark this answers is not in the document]
DonOmalVindula: The error codes are correct as-is. `USRS-1010` is already taken by `ErrorConsentSyncFailed`, which was added by a separate upstream PR (#1604) that merged before this one. Our three display validation errors correctly use the next available codes:

- `USRS-1011` — `ErrorInvalidDisplayAttribute` (existence check)
- `USRS-1012` — `ErrorNonDisplayableAttribute` (displayability check)
- `USRS-1013` — `ErrorCredentialDisplayAttribute` (non-credential check)

The "agreed mapping" CodeRabbit references (USRS-1010/1011/1012) was from its own earlier learning that was recorded before the consent sync error was merged. The learning itself used the wrong base codes.

backend/internal/userschema/model/boolean.go
```diff
@@ -38,6 +38,10 @@ func (p *boolean) isCredential() bool {
 	return false
 }
 
+func (p *boolean) isDisplayable() bool {
+	return true
```
ThumulaPerera: This might not make sense. Shall we keep boolean as non displayable?

DonOmalVindula: Fixed

backend/internal/userschema/model/schema.go
```diff
@@ -52,6 +54,59 @@ type Schema struct {
 	properties map[string]property
 }
 
+// resolvePropertyPath resolves a dot-notation path (e.g. "address.city") to the
+// leaf property by walking through nested object types. For a simple (non-dotted)
+// name, it returns the top-level property directly.
+func (cs *Schema) resolvePropertyPath(path string) (property, bool) {
```
ThumulaPerera: To me, `resolvePropertyPath` gives the impression this outputs a path. Shall we rename this to `getPropertyByPath` or something similar?

ThumulaPerera: For path string, I think we should support JSONPath syntax [1]. We might not need to do it right now, but shall we track this as an improvement ?

[1] https://www.rfc-editor.org/rfc/rfc9535.html

backend/internal/userschema/service.go
```diff
@@ -710,37 +720,59 @@ func logAndReturnServerError(
...
+func validateDisplayAttribute(
+	compiledSchema *model.Schema, systemAttrs *SystemAttributes,
+) *serviceerror.ServiceError {
+	if systemAttrs == nil || systemAttrs.Display == "" {
+		return nil
+	}
+	if !compiledSchema.HasAttribute(systemAttrs.Display) {
+		return &ErrorInvalidDisplayAttribute
+	}
+	if !compiledSchema.IsDisplayableAttribute(systemAttrs.Display) {
```
ThumulaPerera: HasAttribute(...), IsDisplayableAttribute(...), IsCredentialAttribute(...) all internally call resolvePropertyPath(...). Is there some way we can optimise this?

backend/internal/userschema/error_constants.go
```diff
@@ -99,6 +99,32 @@ var (
...
+		ErrorDescription: "Display attribute must reference a string or number type",
+	}
+	// ErrorCredentialDisplayAttribute is the error returned when the display attribute
+	// references an attribute marked as a credential.
+	ErrorCredentialDisplayAttribute = serviceerror.ServiceError{
+		Type:  serviceerror.ClientErrorType,
+		Code:  "USRS-1013",
+		Error: "Credential attribute not allowed as display",
+		ErrorDescription: "Display attribute must not reference a credential attribute, " +
+			"as credentials are not included in user responses",
```
ThumulaPerera: Do we need to explicitly say `as credentials are not included in user responses`. 

IMO `Display attribute must not reference a credential attribute` is self explanatory. Even if we are giving a justification, I think the better justification would be to say, `as credentials are secrets and should not be displayed`

backend/internal/userschema/model/schema.go
DonOmalVindula: Yes.. I agree. Created an issue to track this - https://github.com/asgardeo/thunder/issues/1674

DonOmalVindula: Fixed

backend/internal/userschema/service.go
```diff
@@ -332,11 +337,16 @@ func (us *userSchemaService) UpdateUserSchema(ctx context.Context, schemaID stri
...
 		Schema:             request.Schema,
 	}
-	if validationErr := validateUserSchemaDefinition(schemaToValidate); validationErr != nil {
+	compiledSchema, validationErr := validateUserSchemaDefinition(schemaToValidate)
+	if validationErr != nil {
 		logger.Debug("User schema validation failed", log.String("id", schemaID))
 		return nil, validationErr
 	}
 
+	if svcErr := validateDisplayAttribute(compiledSchema, request.SystemAttributes); svcErr != nil {
```
ThumulaPerera: Shall we move this logic into validateUserSchemaDefinition(...)
It already validates other fields in the request (like name and ouID). System attributes is also another such field IMO.

Also, that removes the requirement to return the `compiledSchema` from validateUserSchemaDefinition(...) method

ThumulaPerera: To clarify, I'm thinking something like

```
validateUserSchemaDefinition(schema Schema) {
       // validate schema.Name
       // validate schema.ou
       // compile schema.Schema (this is basically the validation for schema)
       validateSystemAttributes(compiledSchema, schema.SystemAttributes)
}

validateSystemAttributes(compiledSchema Schema, systemAttrs schema.SystemAttributes) {
       validateDisplayAttribute(compiledSchema, systemAttrs.Display)
}
```

backend/internal/userschema/service.go
DonOmalVindula: Yes.. added the following approach. Hope it is fine

```
func validateDisplayAttribute(
	compiledSchema *model.Schema, systemAttrs *SystemAttributes,
) *serviceerror.ServiceError {
	if systemAttrs == nil || systemAttrs.Display == "" {
		return nil
	}

	switch compiledSchema.ValidateAsDisplayAttribute(systemAttrs.Display) {
	case model.DisplayAttributeNotFound:
		return &ErrorInvalidDisplayAttribute
	case model.DisplayAttributeNotDisplayable:
		return &ErrorNonDisplayableAttribute
	case model.DisplayAttributeIsCredential:
		return &ErrorCredentialDisplayAttribute
	default:
		return nil
	}
}

```

backend/internal/userschema/error_constants.go
DonOmalVindula: Fixed

backend/internal/userschema/service.go
DonOmalVindula: Fixed

backend/internal/userschema/declarative_resource.go
```diff
@@ -227,11 +228,16 @@ func validateUserSchema(schemaDTO *UserSchema, ouService oupkg.OrganizationUnitS
 			schemaDTO.OrganizationUnitID, schemaDTO.Name)
 	}
 
-	// Validate schema is valid JSON
+	// Validate schema is valid JSON and compile it for further checks.
 	if len(schemaDTO.Schema) > 0 {
```
ThumulaPerera: shall we change the validation to behave the same as service

backend/internal/userschema/model/schema.go
```diff
@@ -52,6 +54,89 @@ type Schema struct {
...
+			return nil, false
+		}
+		currentProps = obj.properties
+	}
+
+	return nil, false
+}
+
+// HasAttribute checks if an attribute with the given name or dot-notation path exists in the schema.
+func (cs *Schema) HasAttribute(name string) bool {
```
ThumulaPerera: Shall we remove these methods if there are no usages

backend/internal/userschema/declarative_resource.go
DonOmalVindula: Fixed

PR #1690, opened 2026-03-06
Add batch fetch support for users and groups and fix display attribute resolution

### LINKED ISSUE #1518: Fix role assignment display to use actual display attribute value
Parent issue: #1511

## Scope

Update the role assignment `include=display` feature to resolve the actual display attribute value for user-type assignments instead of returning the user ID.

## Tasks

- [x] Replace `getDisplayNameForAssignment` with batch `populateDisplayNames` in `role/service.go`:
  - Batch fetch users via `userService.GetUsersByIDs` (from #1524)
  - Resolve display attribute paths via `userSchemaService.GetDisplayAttributesByNames` (reads `systemAttributes.display` from cached schema)
  - Extract display value from user attributes JSON using dot-notation path traversal
  - Fall back to user ID if display attribute is empty or not configured
- [x] Add `userSchemaService userschema.UserSchemaServiceInterface` to `roleService` struct
- [x] Update `role/init.go` `Initialize` to accept and pass `userSchemaService`
- [x] Update `servicemanager.go` to pass `userSchemaService` to `role.Initialize`
- [x] Update `api/role.yaml` docs to reflect that user display now uses actual display attribute value

## Files

- `backend/internal/role/service.go` — `populateDisplayNames`, `resolveDisplayAttributePaths`, `resolveUserDisplay`, `extractDisplayValue`
- `backend/internal/role/init.go`
- `backend/internal/system/servicemanager/servicemanager.go`
- `api/role.yaml`

## Implementation

Resolved as part of PR #1690 which also implements batch fetch (#1524). The display name resolution uses the schema-configured `systemAttributes.display` path to extract the correct attribute value from user attributes JSON, falling back to user ID when not configured.

## Design Discussion

#1510

### LINKED ISSUE #1524: Add batch fetch support (GetUsersByIDs / GetGroupsByIDs) for user and group services

Several endpoints that reference users or groups by ID currently fetch them **one at a time** in a loop, resulting in N+1 database queries. For example, the role assignments endpoint calls `userService.GetUser()` per assignment to resolve display names — if a role has 50 user assignments, that's 50 individual queries to the user DB.

This pattern appears in:
- **Role assignments** (`role/service.go` → `getDisplayNameForAssignment`): fetches each user/group individually to resolve display names
- **Any future endpoint** that needs to resolve a batch of user/group IDs into objects

**API consumers and administrators** who work with endpoints returning references to multiple users/groups. As the number of assignments, members, or references grows, the N+1 pattern creates linear query growth that degrades response times.

### Proposed Solution

Add batch fetch methods to the user and group service interfaces:

#### User Service

```go
// GetUsersByIDs retrieves multiple users by their IDs in a single query.
GetUsersByIDs(ctx context.Context, userIDs []string) (map[string]*User, *serviceerror.ServiceError)
```

Backed by a store-level query using `WHERE USER_ID IN (...)`:

```sql
SELECT USER_ID, OU_ID, TYPE, ATTRIBUTES
FROM "USER"
WHERE USER_ID IN ($1, $2, ..., $N) AND DEPLOYMENT_ID = $M
```

#### Group Service

```go
// GetGroupsByIDs retrieves multiple groups by their IDs in a single query.
GetGroupsByIDs(ctx context.Context, groupIDs []string) (map[string]*Group, *serviceerror.ServiceError)
```

Backed by:

```sql
SELECT GROUP_ID, NAME, DESCRIPTION, OU_ID
FROM "GROUP"
WHERE GROUP_ID IN ($1, $2, ..., $N) AND DEPLOYMENT_ID = $M
```

#### Usage example (role assignments)

Before (N+1):
```go
for _, assignment := range assignments {
    user, _ := rs.userService.GetUser(ctx, assignment.ID)  // 1 query per user
    display = resolveDisplay(user)
}
```

After (batch):
```go
userIDs := collectUserIDs(assignments)
users, _ := rs.userService.GetUsersByIDs(ctx, userIDs)  // 1 query for all users
for _, assignment := range assignments {
    display = resolveDisplay(users[assignment.ID])
}
```

### Implementation Notes

- Return a `map[string]*User` / `map[string]*Group` keyed by ID for O(1) lookups
- Handle large batches by chunking (e.g., 100 IDs per query) if needed
- Missing IDs should be absent from the map (not error), allowing callers to handle gracefully
- Update `role/service.go` `getDisplayNameForAssignment` as the first consumer

### Related

- Part of the display attribute feature discussion: #1510
- Complements the cache-backed user schema store (#1514) — batch fetch resolves **user data** from user DB, cache-backed store resolves **schema metadata** from config DB

DonOmalVindula:
## Summary

- Adds `GetUsersByIDs` and `GetGroupsByIDs` methods to both store and service layers for batch retrieval of users/groups by ID lists
- Refactors role service `GetRoleAssignments` to use batch fetch for display name resolution, eliminating N+1 queries (50 assignments = 2 queries instead of 50)
- Resolves user display names using schema-configured display attribute (via `systemAttributes.display`) instead of returning the user ID
- Extracts shared `buildUserINClauseQuery` / `buildGroupINClauseQuery` helpers to reduce duplication in query builders

Closes #1524
Closes #1518

## Changes

| File | Change |
|------|--------|
| `internal/user/store_constants.go` | Add `buildUserINClauseQuery` helper, `buildGetUsersByIDsQuery` (ASQ-USER_MGT-21) |
| `internal/user/store.go` | Add `GetUsersByIDs` to `userStoreInterface` + DB implementation |
| `internal/user/file_based_store.go` | Add `GetUsersByIDs` file-based store implementation |
| `internal/user/composite_store.go` | Add `GetUsersByIDs` composite store implementation |
| `internal/user/service.go` | Add `GetUsersByIDs` to `UserServiceInterface` + implementation (returns `map[string]*User`) |
| `internal/group/store_constants.go` | Add `buildGroupINClauseQuery` helper, `buildGetGroupsByIDsQuery` (GRQ-GROUP_MGT-17) |
| `internal/group/store.go` | Add `GetGroupsByIDs` to `groupStoreInterface` + DB implementation |
| `internal/group/service.go` | Add `GetGroupsByIDs` to `GroupServiceInterface` + implementation (returns `map[string]*Group`) |
| `internal/role/service.go` | Replace N+1 `getDisplayNameForAssignment` with batch `populateDisplayNames`; resolve user display via schema display attribute path |
| `internal/role/service_test.go` | Update display name tests to use batch mock expectations |
| `internal/user/store_constants_test.go` | Add tests for `buildGetUsersByIDsQuery` |
| `internal/group/store_constants_test.go` | **New** — tests for `buildGetGroupsByIDsQuery` |
| Mock files (6) | Add `GetUsersByIDs` / `GetGroupsByIDs` mock methods |

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./...` — all tests pass
- [x] `golangci-lint run ./...` — 0 issues
- [x] Manual: Verify role assignments endpoint returns correct display names with batch fetch
- [x] Manual: Verify with mixed user/group assignments (>10 assignments)

## Summary by CodeRabbit

* **New Features**
  * Batch retrieval for users and groups by IDs; role APIs expose bulk lookup capability.
  * Role service can resolve display names via schema-driven attribute paths.

* **Performance Improvements**
  * Role display-name population now uses bulk lookups for faster resolution.

* **Bug Fixes**
  * Batch operations tolerate partial failures (log and continue).

* **Tests**
  * Expanded unit tests for query builders, batch retrievals, display-extraction, stores, and mocks.

### DIFF
--- a/backend/internal/group/service.go
+++ b/backend/internal/group/service.go
@@ -52,6 +52,7 @@ type GroupServiceInterface interface {
 	GetGroupMembers(ctx context.Context, groupID string, limit, offset int) (
 		*MemberListResponse, *serviceerror.ServiceError)
 	ValidateGroupIDs(ctx context.Context, groupIDs []string) *serviceerror.ServiceError
+	GetGroupsByIDs(ctx context.Context, groupIDs []string) (map[string]*Group, *serviceerror.ServiceError)
 	AddGroupMembers(ctx context.Context, groupID string, members []Member) (*Group, *serviceerror.ServiceError)
 	RemoveGroupMembers(ctx context.Context, groupID string, members []Member) (*Group, *serviceerror.ServiceError)
 }
@@ -924,6 +925,46 @@ func (gs *groupService) ValidateGroupIDs(ctx context.Context, groupIDs []string)
 	}
 
 	return nil
+}
+
+// GetGroupsByIDs retrieves groups by a list of IDs.
+func (gs *groupService) GetGroupsByIDs(
+	ctx context.Context, groupIDs []string,
+) (map[string]*Group, *serviceerror.ServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
+
+	if len(groupIDs) == 0 {
+		return map[string]*Group{}, nil
+	}
+
+	// Deduplicate IDs before passing to store.
+	seen := make(map[string]struct{}, len(groupIDs))
+	uniqueIDs := make([]string, 0, len(groupIDs))
+	for _, id := range groupIDs {
+		if _, ok := seen[id]; !ok {
+			seen[id] = struct{}{}
+			uniqueIDs = append(uniqueIDs, id)
+		}
+	}
+
+	groupDAOs, err := gs.groupStore.GetGroupsByIDs(ctx, uniqueIDs)
+	if err != nil {
+		logger.Error("Failed to get groups by IDs", log.Error(err))
+		return nil, &ErrorInternalServerError
+	}
+
+	result := make(map[string]*Group, len(groupDAOs))
+	for _, dao := range groupDAOs {
+		group := convertGroupDAOToGroup(GroupDAO{
+			ID:                 dao.ID,
+			Name:               dao.Name,
+			Description:        dao.Description,
+			OrganizationUnitID: dao.OrganizationUnitID,
+		})
+		result[dao.ID] = &group
+	}
+
+	return result, nil
 }
 
 // convertGroupDAOToGroup constructs a Group from a GroupDAO.
--- a/backend/internal/group/store.go
+++ b/backend/internal/group/store.go
@@ -51,6 +51,7 @@ type groupStoreInterface interface {
 		ctx context.Context, organizationUnitID string, limit, offset int) ([]GroupBasicDAO, error)
 	AddGroupMembers(ctx context.Context, groupID string, members []Member) error
 	RemoveGroupMembers(ctx context.Context, groupID string, members []Member) error
+	GetGroupsByIDs(ctx context.Context, groupIDs []string) ([]GroupBasicDAO, error)
 }
 
 // groupStore is the default implementation of groupStoreInterface.
@@ -495,6 +496,55 @@ func (s *groupStore) RemoveGroupMembers(ctx context.Context, groupID string, mem
 	}
 
 	return nil
+}
+
+// GetGroupsByIDs retrieves groups by a list of IDs.
+func (s *groupStore) GetGroupsByIDs(ctx context.Context, groupIDs []string) ([]GroupBasicDAO, error) {
+	const batchSize = 100
+
+	if len(groupIDs) == 0 {
+		return []GroupBasicDAO{}, nil
+	}
+
+	dbClient, err := s.dbProvider.GetUserDBClient()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get database client: %w", err)
+	}
+
+	groups := make([]GroupBasicDAO, 0, len(groupIDs))
+
+	for start := 0; start < len(groupIDs); start += batchSize {
+		end := start + batchSize
+		if end > len(groupIDs) {
+			end = len(groupIDs)
+		}
+		chunk := groupIDs[start:end]
+
+		query, args, err := buildGetGroupsByIDsQuery(chunk, s.deploymentID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build get groups by IDs query: %w", err)
+		}
+
+		results, err := dbClient.QueryContext(ctx, query, args...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to execute query: %w", err)
+		}
+
+		for _, row := range results {
+			group, err := buildGroupFromResultRow(row)
+			if err != nil {
+				return nil, fmt.Errorf("failed to build group from result row: %w", err)
+			}
+			groups = append(groups, GroupBasicDAO{
+				ID:                 group.ID,
+				Name:               group.Name,
+				Description:        group.Description,
+				OrganizationUnitID: group.OrganizationUnitID,
+			})
+		}
+	}
+
+	return groups, nil
 }
 
 // buildGroupFromResultRow constructs a GroupDAO from a database result row.
--- a/backend/internal/group/store_constants.go
+++ b/backend/internal/group/store_constants.go
@@ -215,33 +215,54 @@ var (
 	}
 )
 
-// buildBulkGroupExistsQuery constructs a query to check which group IDs exist from a list.
-func buildBulkGroupExistsQuery(groupIDs []string, deploymentID string) (dbmodel.DBQuery, []interface{}, error) {
+// buildGroupINClauseQuery constructs a query with an IN clause for group IDs.
+func buildGroupINClauseQuery(
+	queryID, baseQuery string, groupIDs []string, deploymentID string,
+) (dbmodel.DBQuery, []interface{}, error) {
 	if len(groupIDs) == 0 {
 		return dbmodel.DBQuery{}, nil, fmt.Errorf("groupIDs list cannot be empty")
 	}
+
 	args := make([]interface{}, len(groupIDs)+1)
-	args[0] = deploymentID
 
 	postgresPlaceholders := make([]string, len(groupIDs))
 	sqlitePlaceholders := make([]string, len(groupIDs))
 
 	for i, groupID := range groupIDs {
-		postgresPlaceholders[i] = fmt.Sprintf("$%d", i+2)
+		postgresPlaceholders[i] = fmt.Sprintf("$%d", i+1)
 		sqlitePlaceholders[i] = "?"
-		args[i+1] = groupID
+		args[i] = groupID
 	}
+	args[len(groupIDs)] = deploymentID
 
-	baseQuery := "SELECT ID FROM \"GROUP\" WHERE DEPLOYMENT_ID = %s AND ID IN (%s)"
-	postgresQuery := fmt.Sprintf(baseQuery, "$1", strings.Join(postgresPlaceholders, ","))
-	sqliteQuery := fmt.Sprintf(baseQuery, "?", strings.Join(sqlitePlaceholders, ","))
+	deploymentPlaceholder := fmt.Sprintf("$%d", len(groupIDs)+1)
+	postgresQuery := fmt.Sprintf(baseQuery, strings.Join(postgresPlaceholders, ","), deploymentPlaceholder)
+	sqliteQuery := fmt.Sprintf(baseQuery, strings.Join(sqlitePlaceholders, ","), "?")
 
 	query := dbmodel.DBQuery{
-		ID:            "GRQ-GROUP_MGT-18",
+		ID:            queryID,
 		Query:         postgresQuery,
 		PostgresQuery: postgresQuery,
 		SQLiteQuery:   sqliteQuery,
 	}
 
 	return query, args, nil
+}
+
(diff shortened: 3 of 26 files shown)


backend/internal/group/service.go
```diff
@@ -926,6 +927,37 @@ func (gs *groupService) ValidateGroupIDs(ctx context.Context, groupIDs []string)
 	return nil
 }
 
+// GetGroupsByIDs retrieves groups by a list of IDs.
+// Returns a map of groupID -> *Group for O(1) lookups. Missing IDs are absent from the map.
```
senthalan: Do we need this comment

backend/internal/group/store.go
```diff
@@ -497,6 +498,65 @@ func (s *groupStore) RemoveGroupMembers(ctx context.Context, groupID string, mem
...
+
+	// Deduplicate IDs to avoid fetching the same group across batches.
+	seen := make(map[string]struct{}, len(groupIDs))
+	uniqueIDs := make([]string, 0, len(groupIDs))
+	for _, id := range groupIDs {
+		if _, ok := seen[id]; !ok {
+			seen[id] = struct{}{}
+			uniqueIDs = append(uniqueIDs, id)
+		}
+	}
```
senthalan: Shouldn't this part of the service layer?

I think the store logic shouldn't Deduplicate IDs

backend/internal/role/service.go
```diff
@@ -385,23 +391,13 @@ func (rs *roleService) GetRoleAssignments(ctx context.Context, id string, limit,
...
-			if err != nil {
-				logger.Warn("Failed to get display name for assignment",
-					log.String("assignmentID", assignments[i].ID),
-					log.String("assignmentType", string(assignments[i].Type)),
-					log.Error(err))
-				// Continue with empty display name rather than failing the entire request
-				displayName = ""
-			}
+	if includeDisplay {
+		rs.populateDisplayNames(ctx, assignments, serviceAssignments, logger)
```
senthalan: do we need to pass logger?

backend/internal/user/store.go
```diff
@@ -536,6 +537,60 @@ func (us *userStore) ValidateUserIDs(ctx context.Context, userIDs []string) ([]s
...
+
+	// Deduplicate IDs to avoid fetching the same user across batches.
+	seen := make(map[string]struct{}, len(userIDs))
+	uniqueIDs := make([]string, 0, len(userIDs))
+	for _, id := range userIDs {
+		if _, ok := seen[id]; !ok {
+			seen[id] = struct{}{}
+			uniqueIDs = append(uniqueIDs, id)
+		}
+	}
```
senthalan: Same as group service, can we move to service level

backend/internal/user/store_constants.go
```diff
@@ -242,30 +242,33 @@ func buildIdentifyQuery(filters map[string]interface{}, deploymentID string) (mo
 	return filterQuery, args, nil
 }
 
-// buildBulkUserExistsQuery constructs a query to check which user IDs exist from a list.
-func buildBulkUserExistsQuery(userIDs []string, deploymentID string) (model.DBQuery, []interface{}, error) {
+// buildUserINClauseQuery constructs a query with an IN clause for user IDs.
+// The base query must use %s placeholders in the order: IN clause first, DEPLOYMENT_ID second.
```
senthalan: Do we need this comment

backend/internal/group/store_constants.go
```diff
@@ -215,33 +215,55 @@ var (
 	}
 )
 
-// buildBulkGroupExistsQuery constructs a query to check which group IDs exist from a list.
-func buildBulkGroupExistsQuery(groupIDs []string, deploymentID string) (dbmodel.DBQuery, []interface{}, error) {
+// buildGroupINClauseQuery constructs a query with an IN clause for group IDs.
+// The base query must use %s placeholders in the order: IN clause first, DEPLOYMENT_ID second.
```
senthalan: Do we need this comment

backend/internal/group/service.go
DonOmalVindula: Fixed

backend/internal/group/store.go
DonOmalVindula: Fixed

backend/internal/role/service.go
DonOmalVindula: Fixed

backend/internal/user/store.go
DonOmalVindula: Fixed

backend/internal/user/store_constants.go
DonOmalVindula: Fixed

backend/internal/group/store_constants.go
DonOmalVindula: Fixed

backend/internal/role/service.go
```diff
@@ -711,28 +707,148 @@ func buildPaginationLinks(base string, limit, offset, totalCount int) []Link {
...
-	case AssigneeTypeGroup:
-		groupResp, svcErr := rs.groupService.GetGroup(ctx, assignment.ID)
+	if len(groupIDs) > 0 {
+		groups, svcErr := rs.groupService.GetGroupsByIDs(ctx, groupIDs)
 		if svcErr != nil {
-			return "", fmt.Errorf("failed to get group: %w", errors.New(svcErr.Error))
+			logger.Warn("Failed to batch fetch groups for display names", log.Any("error", svcErr))
+		} else {
+			groupsMap = groups
+		}
```
senthalan:
```suggestion
		groupsMap, svcErr = rs.groupService.GetGroupsByIDs(ctx, groupIDs)
        if svcErr != nil {
			logger.Warn("Failed to batch fetch groups for display names", log.Any("error", svcErr))
		}
```
Won't this work?

backend/internal/role/service.go
```diff
@@ -711,28 +707,148 @@ func buildPaginationLinks(base string, limit, offset, totalCount int) []Link {
...
+	var groupsMap map[string]*group.Group
+
+	if len(userIDs) > 0 {
+		users, svcErr := rs.userService.GetUsersByIDs(ctx, userIDs)
 		if svcErr != nil {
-			return "", fmt.Errorf("failed to get user: %w", errors.New(svcErr.Error))
+			logger.Warn("Failed to batch fetch users for display names", log.Any("error", svcErr))
+		} else {
+			usersMap = users
 		}
```
senthalan:
```suggestion
		usersMap, svcErr = rs.userService.GetUsersByIDs(ctx, userIDs)
if svcErr != nil {
			logger.Warn("Failed to batch fetch users for display names", log.Any("error", svcErr))
		} 
}
```

backend/internal/role/service.go
```diff
@@ -711,28 +707,148 @@ func buildPaginationLinks(base string, limit, offset, totalCount int) []Link {
...
+			logger.Warn("Failed to batch fetch groups for display names", log.Any("error", svcErr))
+		} else {
+			groupsMap = groups
+		}
+	}
+
+	// Resolve display attribute paths for user types
+	displayAttrPaths := rs.resolveDisplayAttributePaths(ctx, usersMap)
+
+	// Map lookup for each assignment (O(1) per assignment)
```
senthalan: Can we remove this comment

backend/internal/role/service.go
```diff
@@ -711,28 +707,148 @@ func buildPaginationLinks(base string, limit, offset, totalCount int) []Link {
...
-		return groupResp.Name, nil
+	}
+
+	if len(uniqueTypes) == 0 {
+		return nil
+	}
+
+	typeNames := make([]string, 0, len(uniqueTypes))
+	for t := range uniqueTypes {
+		typeNames = append(typeNames, t)
```
senthalan: Why we need to create a map and then convert to slice, we can 
```
var typeNames []string
for _, u := range usersMap {
		if u.Type != "" {
			typeNames = append(typeNames, u.Type)
		}
	}
```

backend/internal/role/service.go
DonOmalVindula: Fixed

backend/internal/role/service.go
DonOmalVindula: Fixed

backend/internal/role/service.go
DonOmalVindula: Fixed

backend/internal/role/service.go
DonOmalVindula: This approach would introduce duplicated entries in the array right? however, I have improved the logic as follows

```
	seen := make(map[string]struct{})
	var typeNames []string
	for _, u := range usersMap {
		if u.Type != "" {
			if _, ok := seen[u.Type]; !ok {
				seen[u.Type] = struct{}{}
				typeNames = append(typeNames, u.Type)
			}
		}
	}
```

backend/internal/role/service.go
```diff
@@ -711,30 +707,147 @@ func buildPaginationLinks(base string, limit, offset, totalCount int) []Link {
...
+		if current == nil {
+			return ""
+		}
+	}
+
+	if s, ok := current.(string); ok {
+		return s
+	}
+
+	return fmt.Sprintf("%v", current)
```
senthalan: when the current is not string, we are returning stringified current object here(it should be mostly JSON), is is correct?

Don't we need to return "" here

DonOmalVindula: Yes. agreed. I will make the change

senthalan (dismissed):

backend/internal/role/service.go
DonOmalVindula: As we are allowing both string and number type attributes as display values, i have updated the logic as follows.

```
	switch current.(type) {
	case string, float64:
		return utils.ConvertInterfaceValueToString(current)
	default:
		return "", fmt.Errorf("unknown assignment type: %s", assignment.Type)
		return ""
	}
```

senthalan (approved):

PR #1710, opened 2026-03-09
Add include=display support for user and group listing endpoints

### LINKED ISSUE #1515: Add `include=display` support to user listing endpoints
Parent issue: #1511

## Scope

Add `include=display` query parameter support to `GET /users` and `GET /users/tree/{path}` endpoints using a SQL JOIN approach.

## Tasks

- [ ] Add `Display string` field (with `json:"display,omitempty"`) to `user.User` model
- [ ] Add `extractDisplayValue(attributes json.RawMessage, displayAttrName string) string` utility
- [ ] Add JOIN query variant in `user/store_constants.go`:
  ```sql
  SELECT u.USER_ID, u.OU_ID, u.TYPE, u.ATTRIBUTES, us.DISPLAY_ATTRIBUTE
  FROM "USER" u
  LEFT JOIN USER_SCHEMAS us ON u.TYPE = us.NAME
      AND us.DEPLOYMENT_ID = u.DEPLOYMENT_ID
  WHERE u.DEPLOYMENT_ID = $3
  ORDER BY u.USER_ID LIMIT $1 OFFSET $2
  ```
- [ ] Add `GetUserListWithDisplay` to store interface and implement
- [ ] Add `includeDisplay bool` parameter to `GetUserList` and `GetUsersByPath` service methods
- [ ] Parse `include=display` query parameter in `HandleUserListRequest` and `HandleUserListByPathRequest`
- [ ] Update all existing callers to pass `false` for `includeDisplay`
- [ ] Update `api/user.yaml` — add `include` query param and `display` response field

## Files

- `backend/internal/user/model.go`
- `backend/internal/user/store_constants.go`
- `backend/internal/user/store.go`
- `backend/internal/user/service.go`
- `backend/internal/user/handler.go`
- `api/user.yaml`

## Design Discussion

#1510

### LINKED ISSUE #1517: Add `include=display` support to group member listing endpoint
Parent issue: #1511

## Scope

Add `include=display` query parameter support to `GET /groups/{id}/members` endpoint using **service-layer resolution**.

> **Note**: `USER_SCHEMAS` lives in the config DB while `USER` lives in the user DB — these are separate databases, so SQL JOINs are not possible. Display values are resolved in the service layer using a single batch query against the config DB.

## Tasks

- [ ] Add `Display string` field (with `json:"display,omitempty"`) to `group.Member` model
- [ ] Add `includeDisplay bool` to `GetGroupMembers` service method
- [ ] Implement service-layer display resolution when `includeDisplay` is true:
  - For **user members**:
    1. Collect distinct user types from the member list
    2. **Single batch call**: `userSchemaService.GetDisplayAttributesByNames(distinctTypes)` → `map[typeName]displayAttributeName` (reads `systemAttributes.display` from each schema)
    3. For each user member, extract the display value from their JSON attributes using the map
    4. Fall back to user ID if display attribute name is empty or extraction fails
  - For **group members**: use group name directly as display
- [ ] Add `userSchemaService userschema.UserSchemaServiceInterface` dependency to `groupService` struct
- [ ] Update `group/init.go` and `servicemanager.go` to wire `userSchemaService`
- [ ] Parse `include=display` in handler
- [ ] Update `api/group.yaml` — add `include` query param and `display` response field

## Performance

Only **2 DB queries** per request: 1 user DB query (list group members) + 1 config DB query (batch `WHERE NAME IN (...)` for `systemAttributes`).

## Files

- `backend/internal/group/model.go`
- `backend/internal/group/service.go`
- `backend/internal/group/handler.go`
- `backend/internal/group/init.go`
- `backend/internal/system/servicemanager/servicemanager.go`
- `api/group.yaml`

## Design Discussion

#1510

DonOmalVindula:
## Summary
- Adds `include=display` query parameter support to user, group member, and role assignment listing endpoints
- Resolves user display names from the schema-configured `systemAttributes.display` attribute path (falls back to user ID)
- Resolves group member display: user members use schema display attribute, group members use group name
- Role assignment listing (`GET /roles/{id}/assignments`) resolves display for user and group assignments
- Extracts shared utilities to `internal/system/utils`: `ResolveUserDisplay`, `ResolveDisplayAttributePaths`, `BuildPaginationLinks`
- Wires `userSchemaService` into group service for display resolution
- Pagination links carry `include=display` forward across pages
- Updates API specs (`user.yaml`, `group.yaml`, `role.yaml`) with `include` query param and `display` field documentation

Closes #1515, closes #1517

## Supported Flows

| # | Endpoint | Description |
|---|----------|-------------|
| 1 | `GET /users?include=display` | User listing with display names |
| 2 | `GET /users/tree/{path}?include=display` | User listing by OU path with display names |
| 3 | `GET /groups/{id}/members?include=display` | Group member listing with display names |
| 4 | `GET /roles/{id}/assignments?include=display` | Role assignment listing with display names |

## Test plan
- [x] Unit tests for shared display utilities (`ResolveUserDisplay`, `ResolveDisplayAttributePaths`) 
- [x] Unit tests for `BuildPaginationLinks` shared utility
- [x] Unit tests for `populateUserDisplayNames` in user service
- [x] Unit tests for `populateMemberDisplayNames` in group service
- [x] Pagination link tests verify `include=display` is carried forward
- [x] Handler tests verify `include=display` query param parsing (user, group)
- [x] All existing tests pass with no regressions

### DIFF
--- a/backend/internal/group/service.go
+++ b/backend/internal/group/service.go
@@ -34,6 +34,7 @@ import (
 	"github.com/asgardeo/thunder/internal/system/sysauthz"
 	"github.com/asgardeo/thunder/internal/system/utils"
 	"github.com/asgardeo/thunder/internal/user"
+	"github.com/asgardeo/thunder/internal/userschema"
 )
 
 const loggerComponentName = "GroupMgtService"
@@ -49,7 +50,7 @@ type GroupServiceInterface interface {
 	GetGroup(ctx context.Context, groupID string) (*Group, *serviceerror.ServiceError)
 	UpdateGroup(ctx context.Context, groupID string, request UpdateGroupRequest) (*Group, *serviceerror.ServiceError)
 	DeleteGroup(ctx context.Context, groupID string) *serviceerror.ServiceError
-	GetGroupMembers(ctx context.Context, groupID string, limit, offset int) (
+	GetGroupMembers(ctx context.Context, groupID string, limit, offset int, includeDisplay bool) (
 		*MemberListResponse, *serviceerror.ServiceError)
 	ValidateGroupIDs(ctx context.Context, groupIDs []string) *serviceerror.ServiceError
 	GetGroupsByIDs(ctx context.Context, groupIDs []string) (map[string]*Group, *serviceerror.ServiceError)
@@ -59,26 +60,29 @@ type GroupServiceInterface interface {
 
 // groupService is the default implementation of the GroupServiceInterface.
 type groupService struct {
-	groupStore    groupStoreInterface
-	ouService     oupkg.OrganizationUnitServiceInterface
-	userService   user.UserServiceInterface
-	transactioner transaction.Transactioner
-	authzService  sysauthz.SystemAuthorizationServiceInterface
+	groupStore        groupStoreInterface
+	ouService         oupkg.OrganizationUnitServiceInterface
+	userService       user.UserServiceInterface
+	userSchemaService userschema.UserSchemaServiceInterface
+	transactioner     transaction.Transactioner
+	authzService      sysauthz.SystemAuthorizationServiceInterface
 }
 
 // newGroupService creates a new instance of GroupService with injected dependencies.
 func newGroupService(
 	ouService oupkg.OrganizationUnitServiceInterface,
 	userService user.UserServiceInterface,
+	userSchemaService userschema.UserSchemaServiceInterface,
 	authzService sysauthz.SystemAuthorizationServiceInterface,
 	transactioner transaction.Transactioner,
 ) GroupServiceInterface {
 	return &groupService{
-		groupStore:    newGroupStore(),
-		ouService:     ouService,
-		userService:   userService,
-		authzService:  authzService,
-		transactioner: transactioner,
+		groupStore:        newGroupStore(),
+		ouService:         ouService,
+		userService:       userService,
+		userSchemaService: userSchemaService,
+		authzService:      authzService,
+		transactioner:     transactioner,
 	}
 }
 
@@ -127,7 +131,7 @@ func (gs *groupService) listAllGroups(ctx context.Context, limit, offset int) (
 		Groups:       groupBasics,
 		StartIndex:   offset + 1,
 		Count:        len(groupBasics),
-		Links:        buildPaginationLinks("/groups", limit, offset, totalCount),
+		Links:        utils.BuildPaginationLinks("/groups", limit, offset, totalCount, ""),
 	}
 
 	return response, nil
@@ -143,7 +147,7 @@ func (gs *groupService) listGroupsByOUIDs(ctx context.Context, ouIDs []string, l
 			Groups:       []GroupBasic{},
 			StartIndex:   offset + 1,
 			Count:        0,
-			Links:        []Link{},
+			Links:        []utils.Link{},
 		}, nil
 	}
 
@@ -159,7 +163,7 @@ func (gs *groupService) listGroupsByOUIDs(ctx context.Context, ouIDs []string, l
 			Groups:       []GroupBasic{},
 			StartIndex:   offset + 1,
 			Count:        0,
-			Links:        []Link{},
+			Links:        []utils.Link{},
 		}, nil
 	}
 
@@ -179,7 +183,7 @@ func (gs *groupService) listGroupsByOUIDs(ctx context.Context, ouIDs []string, l
 		Groups:       groupBasics,
 		StartIndex:   offset + 1,
 		Count:        len(groupBasics),
-		Links:        buildPaginationLinks("/groups", limit, offset, totalCount),
+		Links:        utils.BuildPaginationLinks("/groups", limit, offset, totalCount, ""),
 	}
 
 	return response, nil
@@ -236,7 +240,7 @@ func (gs *groupService) GetGroupsByPath(
 		Groups:       groupBasics,
 		StartIndex:   offset + 1,
 		Count:        len(groupBasics),
-		Links:        buildPaginationLinks("/groups", limit, offset, totalCount),
+		Links:        utils.BuildPaginationLinks("/groups", limit, offset, totalCount, ""),
 	}
 
 	return response, nil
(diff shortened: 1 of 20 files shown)


backend/internal/group/service.go
```diff
@@ -611,6 +620,118 @@ func (gs *groupService) GetGroupMembers(ctx context.Context, groupID string, lim
...
+			displayAttrPaths = gs.resolveDisplayAttributePaths(ctx, usersMap, logger)
+		}
+	}
+
+	// Batch-fetch groups for group member display names.
+	var groupsMap map[string]*Group
+	if len(groupIDs) > 0 {
+		var svcErr *serviceerror.ServiceError
+		groupsMap, svcErr = gs.GetGroupsByIDs(ctx, groupIDs)
+		if svcErr != nil {
```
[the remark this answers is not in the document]
DonOmalVindula: Acknowledged — this is a deliberate trade-off:

1. **Cross-table JOINs aren't straightforward** — `GetGroupMembers` queries the `group_member` table. Inlining display data would require JOINs to both `USER` (for Type + Attributes) and `GROUP` (for Name) with different columns per member type, significantly complicating the store query.
2. **Package boundaries** — The group store intentionally doesn't access user tables. Resolving user display fields at the store layer would break this separation.
3. **Batched, not N+1** — `GetUsersByIDs` and `GetGroupsByIDs` use `IN` clauses with chunking (100 items/batch), so each is a single DB round-trip per batch — not per member.
4. **Bounded cost** — For one page of members it's at most 3 DB calls (1 schema lookup + 1 batched user fetch + 1 batched group fetch), constant regardless of page size.
5. **Consistent pattern** — The role service's `populateDisplayNames` uses the same approach for role assignments.

We can revisit with a store-level JOIN optimisation if profiling shows this becomes a bottleneck.

backend/internal/ou/service.go
```diff
@@ -1083,21 +1115,61 @@ func (ous *organizationUnitService) getResourceListWithExistenceCheck(
...
+		typeNames := make([]string, 0, len(uniqueTypes))
+		for t := range uniqueTypes {
+			typeNames = append(typeNames, t)
+		}
+
+		paths, svcErr := ous.displayAttrResolver.GetDisplayAttributesByNames(ctx, typeNames)
+		if svcErr != nil {
+			logger.Warn("Failed to get display attribute paths", log.Any("error", svcErr))
+		} else {
+			displayAttrPaths = paths
```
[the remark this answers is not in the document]
DonOmalVindula: This is an intentional design decision, display names are optional enrichment on top of the core user list response. If the schema lookup fails, we prefer to degrade gracefully (fall back to user ID) rather than fail the entire request. The caller asked for a list of users and should still get it.

This matches the pattern used in both the user and group services (`populateUserDisplayNames`, `populateMemberDisplayNames`), which also log a warning and continue with ID fallback on schema resolution failure.

backend/internal/group/service.go
```diff
@@ -598,8 +602,14 @@ func (gs *groupService) GetGroupMembers(ctx context.Context, groupID string, lim
 		return nil, &ErrorInternalServerError
 	}
 
+	displayQuery := ""
+	if includeDisplay {
+		displayQuery = "&include=display"
+		gs.populateMemberDisplayNames(ctx, members, logger)
+	}
```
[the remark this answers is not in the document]
DonOmalVindula: Same rationale as the OU service, display enrichment is intentionally best-effort. If the schema or batch-fetch call fails, we degrade gracefully (fall back to user/group ID) rather than failing the entire member listing request. The caller asked for a list of members and should still receive it.

This is consistent across all three packages (user, group, OU) they all log a warning and continue with ID fallback on display resolution failure. Returning a 500 for a non-critical enrichment failure would be a worse user experience than returning members without display names.

backend/internal/ou/service.go
```diff
@@ -36,6 +36,22 @@ import (
 
 const loggerComponentNameService = "OrganizationUnitService"
 
+// DisplayAttributeResolver resolves display attribute paths for user types.
+// This interface is satisfied by userschema.UserSchemaServiceInterface and avoids
+// a circular import between the ou and userschema packages.
```
senthalan: Do we need this comment

backend/internal/ou/service.go
```diff
@@ -36,6 +36,22 @@ import (
...
+type DisplayAttributeResolver interface {
+	GetDisplayAttributesByNames(
+		ctx context.Context, names []string,
+	) (map[string]string, *serviceerror.ServiceError)
+}
+
+// DisplayAttributeConfigurable allows injecting a display attribute resolver
+// after initialization. This is separate from OrganizationUnitServiceInterface
+// because it is an initialization concern, not a domain operation.
```
senthalan:
```suggestion
// after initialization. 
```

backend/internal/user/displayutil.go
senthalan, line 1: shall we move this to the user package?

As this is related to resolving the display attribute of the user

backend/internal/system/utils/displayutil.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// DisplayAttributeLookupFunc is a function type for looking up display attribute paths by user type names.
+type DisplayAttributeLookupFunc func(ctx context.Context, typeNames []string) (map[string]string, error)
+
+// ResolveDisplayAttributePaths collects unique user types from the provided type-name pairs and
+// resolves their display attribute paths using the given lookup function.
+// Returns nil if there are no types to resolve or if the lookup fails.
+func ResolveDisplayAttributePaths(
+	ctx context.Context, userTypes []string, lookupFn DisplayAttributeLookupFunc,
```
senthalan: why `DisplayAttributeLookupFunc` this needs to be provided as callback function? 

display attribute paths for the user is always resolved from the user type

backend/internal/system/utils/displayutil.go
```diff
@@ -0,0 +1,114 @@
...
+) map[string]string {
+	if lookupFn == nil || len(userTypes) == 0 {
+		return nil
+	}
+
+	seen := make(map[string]struct{})
+	var uniqueTypes []string
+	for _, t := range userTypes {
```
rajithacharith:
```suggestion
	for _, userType := range userTypes {
```

backend/internal/system/utils/displayutil.go
```diff
@@ -0,0 +1,114 @@
...
+		if t != "" {
+			if _, ok := seen[t]; !ok {
+				seen[t] = struct{}{}
+				uniqueTypes = append(uniqueTypes, t)
+			}
+		}
+	}
```
rajithacharith: Shall we extract this to a separate function
uniqueTypes := uniqueStrings(userTypes)

backend/internal/system/utils/displayutil.go
```diff
@@ -0,0 +1,114 @@
...
+func ResolveUserDisplay(id, userType string, attributes json.RawMessage, displayAttrPaths map[string]string) string {
+	if displayAttrPaths != nil && userType != "" {
+		if path, ok := displayAttrPaths[userType]; ok && path != "" {
+			if val := ExtractDisplayValue(attributes, path); val != "" {
+				return val
+			}
+		}
+	}
+
+	return id
```
rajithacharith: Shall we follow the return first pattern. Seems it is more readable
```suggestion
    if displayAttrPaths == nil || userType == "" {
       return id
    }
    path, ok := displayAttrPaths[userType];
    if !ok || path == "" {
       return id
    }
    value :=  ExtractDisplayValue(attributes, path);
    if value == "" {
       return id
    }
    return val
```

PR #1739, opened 2026-03-10
Restrict scopes in client credentials grant

thiva-k:
This pull request updates the client credentials grant flow in the OAuth2 implementation to temporarily restrict scopes from being issued until proper client authorization is implemented. The changes ensure that access tokens issued via the client credentials grant do not include any scopes, regardless of what is requested. Corresponding tests are updated to reflect this new behavior.

---
The `client_credentials` grant type no longer returns requested scopes in the issued access token. Any scopes provided in the token request are silently ignored, and the `scope` claim is omitted from the JWT entirely.

#### 💥 Impact
Clients using the client credentials grant that rely on scopes being present in the access token will receive tokens without any scope claim. Resource servers validating scope-based access will deny such requests.

#### 🔄 Migration Guide
Remove scope-based access control assumptions for tokens issued via the client credentials grant until client authorization is implemented. Resource servers should not expect a `scope` claim in tokens issued through this grant type.

---

**Client credentials grant scope restriction:**

* In `client_credentials.go`, the handler now sets `scopes` to an empty list for client credentials grants, regardless of the requested scopes. This is a temporary measure until client authorization is implemented.

**Test updates to match new scope behavior:**

* In `client_credentials_test.go`, the test case "WithValidScope" is updated to expect no scopes in the JWT claims or scopes list, reflecting the new restriction.
* In the same test file, the expectation for the access token's scopes is changed to an empty string for the nil token attributes test, ensuring consistency with the new behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Client credentials grant now rejects any provided scope and issues access tokens without propagated scopes.
* **Tests**
  * Updated tests to expect an invalid_scope error when a scope is supplied to the client credentials flow and to assert tokens contain no scopes.

### DIFF
--- a/backend/internal/oauth/oauth2/granthandlers/client_credentials.go
+++ b/backend/internal/oauth/oauth2/granthandlers/client_credentials.go
@@ -57,6 +57,14 @@ func (h *clientCredentialsGrantHandler) ValidateGrant(tokenRequest *model.TokenR
 		}
 	}
 
+	// Scope is not supported for client credentials grant until client authorization is implemented.
+	if tokenRequest.Scope != "" {
+		return &model.ErrorResponse{
+			Error:            constants.ErrorInvalidScope,
+			ErrorDescription: "Scope is not supported for the client credentials grant type",
+		}
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


senthalan:
Since we have onboard the application RBAC, do we need this PR ? @thiva-k

thiva-k:
Closing this PR since app RBAC is implemented.

Issue #1397, opened 2026-02-13
Flow type is not validated during Application creation


rajithacharith:
### Description

The registration_flow_id can have a authentication flow as well

### Steps to Reproduce

Create a authentication flow. 
Create an application with registration flow id as the authentication flow

### Version

v0.22.0

Dilusha-Madushan:
Here I observe another issue happens if we send the registration flow id as auth flow id in application creation. So these both issues are related because both time use same validation and that validation missing the flow type check.

PR #1906, opened 2026-03-20
Validate application flow types during creation

Dilusha-Madushan:
### What does this PR do?

Fixes a validation bug where `registration_flow_id` and `auth_flow_id` in application creation/update endpoints accepted any valid flow ID regardless of the flow's type. Also refactors `IsValidFlow` to be type-aware with proper error discrimination, and removes the now-dead `IsFlowExists` store method.

### Why is this necessary?

The previous logic used `IsValidFlow()` which only verified a flow existed — no type enforcement. An `AUTHENTICATION` flow could be incorrectly assigned as `registration_flow_id` and vice versa, leading to misconfigured applications and potential runtime errors during user orchestration. Additionally, any store/DB failure was silently collapsed into a false return, causing transient server errors to be surfaced as 4xx client validation failures instead of 5xx internal errors.

### 1. `IsValidFlow` refactored
- Made `flowType common.FlowType` a required parameter
- Changed the return type from `bool` to `(bool, *serviceerror.ServiceError)`
- This distinguishes between:
  - Flow not found/type mismatch → `(false, nil)` → 4xx
  - Store failure → `(false, serviceError)` → 5xx  
- Now consistent with the rest of `FlowMgtServiceInterface`

#### 2. Validation updated
- `validateRegistrationFlowID` now explicitly requires `FlowTypeRegistration`
- `validateAuthFlowID` now requires `FlowTypeAuthentication`
- Both correctly propagate server errors

#### 3. `IsFlowExists` removed
- With `IsValidFlow` now using `GetFlowByID` internally, `IsFlowExists` became dead code
- Removed from:
  - Store interface
  - All store implementations:
    - `flowStore`
    - `compositeFlowStore`
    - `cacheBackedFlowStore`
    - `fileBasedStore`
  - SQL query constant
  - All related tests

#### 4. Mocks regenerated
- All mock files regenerated via `make mockery` after interface changes

### Updated tests
- Updated all mock call sites in `service_test.go` with:
  - New `IsValidFlow` signature
  - Includes `flowType` and `(bool, *serviceerror.ServiceError)` return

#### New test cases
- `TestValidateRegistrationFlowID_WithAuthenticationFlowType`
  - Confirms an `AUTHENTICATION` flow passed as `registration_flow_id` is rejected with `APP-1008`
- `TestValidateAuthFlowID_WithRegistrationFlowType`
  - Confirms a `REGISTRATION` flow passed as `auth_flow_id` is rejected
- Store error path covered
  - Verifies transient DB failures return a server error, not a client validation error

#### Removed tests
- Removed all `IsFlowExists` store-level tests across:
  - `store_test.go`
  - `cache_backed_store_test.go`
  - `composite_store_test.go`
  - `file_based_store_test.go`
  - `declarative_resource_test.go`

### Test Results

- ✅ All unit tests pass with **87.3% coverage** (`make test_unit`)
- ✅ Lint clean — **0 issues** (`make lint_backend`)

### **Related Issues**
 - https://github.com/asgardeo/thunder/issues/1397

## Summary by CodeRabbit

* **Bug Fixes**
  * Flow validation tightened: authentication and registration flows are now validated against explicit flow types and actual stored flow records, reducing acceptance of incorrect flow IDs and surface internal errors as appropriate.

* **Tests**
  * Expanded test coverage for typed flow validation (wrong-type IDs, empty IDs, not-found and store-error cases) and updated mocks to enforce type-specific checks.


backend/internal/flow/mgt/service.go
thiva-k: Will this function be used elsewhere after this change? If not shall we modify this to accept FlowType as an arg?

Dilusha-Madushan: No, I checked the code and only use in test file now. So we can merge these two functions where keep one function where keep type check as optional check.

thiva-k:
Shall we also create an issue and attach to this PR?

backend/internal/flow/mgt/service.go
```diff
@@ -458,18 +458,30 @@ func (s *flowMgtService) GetGraph(ctx context.Context, flowID string) (
 }
 
 // IsValidFlow checks if a valid flow exists for the given flow ID.
-func (s *flowMgtService) IsValidFlow(ctx context.Context, flowID string) bool {
+// An optional flowType argument can be provided to also validate that the flow
+// matches the expected type. If no flowType is provided, only the existence of
+// the flow is checked. Only the first flowType value is used if multiple are provided.
```
thiva-k: Is there a requirement to check flow existence without a flow type or pass multiple flow types? If not wouldn't it be better to always pass a single flow type as an arg to this function, instead of using variadic args and selecting the first one?

backend/internal/flow/mgt/service.go
```diff
@@ -458,18 +458,30 @@ func (s *flowMgtService) GetGraph(ctx context.Context, flowID string) (
...
+// Returns (false, nil) when the flow is not found or type mismatches (client error).
+// Returns (false, *serviceerror.ServiceError) when a store failure occurs (server error).
+func (s *flowMgtService) IsValidFlow(
+	ctx context.Context, flowID string, flowType ...common.FlowType) (bool, *serviceerror.ServiceError) {
 	if flowID == "" {
-		return false
+		return false, nil
 	}
 
-	exists, err := s.store.IsFlowExists(ctx, flowID)
```
thiva-k, line 466: Will there be any usage for this store function? If not shall we remove it?

backend/internal/flow/mgt/service.go
Dilusha-Madushan: No currently, I just arrange the function in dual focus manner keeping original focus also from function. If this existance check use case no longer needed then we can make the function single focus one.

Dilusha-Madushan: Had discussion with @rajithacharith and make the function single focus

backend/internal/flow/mgt/service.go
Dilusha-Madushan: Had discussion with @rajithacharith and removed this store function

Issue #1923, opened 2026-03-23
Support email confirmation during user self-registration


ThaminduDilshan:
### Current Limitation

Thunder's self-registration flow currently has no built-in mechanism to confirm the user's email address before activating their account. After completing self-registration, the user account is created without any email verification step, making it impossible to ensure the user owns the email they registered with.

Additionally, the template service currently only supports the \`USER_INVITE\` scenario type. There is no \`ACCOUNT_CONFIRMATION\` (or equivalent) template scenario to support sending a confirmation email as part of the registration flow.

### Suggested Improvement

Add support for an email confirmation step within self-registration flows by:

1. **New email template scenario** — Introduce a new \`ScenarioType\` (e.g., \`ACCOUNT_CONFIRMATION\`) in the template service alongside the existing \`USER_INVITE\` scenario. This allows administrators to define an email containing a confirmation link that is sent to the user during registration.

2. **Flow definition** — Provide a reference/sample flow definition that models the expected sequence:
   - User submits self-registration form.
   - System sends a confirmation email (using the new template) with a time-limited confirmation link.
   - User clicks the confirmation link.
   - System creates the account.
   - User continues with the remaining steps in the registration flow (e.g., profile completion, password setup).

3. **No new executor required** — The existing \`EmailExecutor\` can be reused with the new scenario/template type. The flow graph handles orchestration.

The change is primarily a template/flow-definition concern. No significant backend or frontend code changes are expected beyond registering the new template scenario type and providing the sample flow and default template."

PR #1970, opened 2026-03-26
Add Self Registration Invite Flow

Dilusha-Madushan:
### Purpose

This PR introduces self-registration via an invite link flow, allowing users to sign up for an application by providing their email address, receiving a registration link, and completing their profile setup through that link.

Key improvements included alongside the new flow:

- **Invite link security**: The invite link is no longer leaked to the browser in self-registration flows. It is stored server-side only and delivered exclusively via email, preventing bypass of the email verification step.
- **Email sent signalling**: The `emailSent` flag (`"true"` / `"false"`) is now always present in the flow response `additionalData`, giving frontend clients a reliable way to detect whether an email was dispatched.
- **Application logo fix**: All flow JSON templates were using `logo_url` (snake_case) to resolve the application logo, but the flow metadata response uses `logoUrl` (camelCase). This mismatch caused the application logo to never render in login and registration prompts. Fixed across all affected flows.
- **Prompt meta always included**: Flow prompt node meta (UI component definitions) was previously only included in the response when verbose mode was enabled. Since meta contains required UI data, it is now always included when present.

---

### Approach

**New self-registration flow (`registeration_flow_basic_invite.json`)**

The flow follows this sequence:
1. Resolve user type → prompt for email → validate email uniqueness
2. Generate invite token + link (stored server-side only)
3. Send registration email via `EmailExecutor` (with `SELF_REGISTRATION` template)
4. Await invite token verification (user clicks the link from the email)
5. Prompt for user details → validate uniqueness → prompt for password → provision user

**Invite link handling**

`InviteExecutor` now stores the invite link only in `RuntimeData` (server-side). `EmailExecutor` is responsible for deciding when to expose it in `AdditionalData`:
- `USER_ONBOARDING` flows: invite link exposed in `AdditionalData` as fallback for admin UIs (allows inline display when SMTP is not configured)
- `REGISTRATION` (self-registration) flows: invite link never exposed in `AdditionalData` regardless of SMTP status

**Email not configured behaviour**

When SMTP is not configured, `EmailExecutor` always completes as a no-op (`ExecComplete`) with `emailSent: "false"` for all flow types. The self-registration flow relies on `emailSent: "true"` in the frontend to show the "check your email" state.

**Frontend**

- `SignUpBox`: added `emailSent` state — shows a "Check your email" alert when `emailSent === "true"` in `additionalData`
- `AcceptInviteBox`: personalises the completion message with the application name when available; removed the hardcoded "Go to Sign In" navigation button
- `ImageAdapter`: added null guard for unresolvable `src` and resolves `alt` text through the template literal resolver

### Related Issues
- https://github.com/asgardeo/thunder/issues/1923
 

---

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added a self-service multi-step registration flow (email, user details, invite, password, provisioning).
  * Added personalized self-registration email invites with app name and a “Complete Registration” link.
  * Signup and invite UIs now show application name in completion messages when available.

* **Localization**
  * Added signup localization keys for email, user-details, and credential screens.

* **Bug Fixes**
  * Email-sent indicator now set explicitly; invite link exposure limited to onboarding flows.

* **Tests**
  * Expanded tests for templates, email sending, invite generation, and UI flows.

### DIFF
--- a/backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
+++ b/backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
@@ -1,0 +1,350 @@
+{
+  "name": "Default Self-Invite Registration Flow",
+  "handle": "default-self-invite-flow",
+  "flowType": "REGISTRATION",
+  "nodes": [
+    {
+      "id": "start",
+      "type": "START",
+      "onSuccess": "user_type_resolver"
+    },
+    {
+      "id": "user_type_resolver",
+      "type": "TASK_EXECUTION",
+      "executor": {
+        "name": "UserTypeResolver"
+      },
+      "onSuccess": "prompt_email",
+      "onIncomplete": "prompt_usertype"
+    },
+    {
+      "id": "prompt_usertype",
+      "type": "PROMPT",
+      "meta": {
+        "components": [
+          {
+            "alt": "{{ t(signup:images.app_logo.alt) }}",
+            "category": "DISPLAY",
+            "height": "60",
+            "id": "image",
+            "resourceType": "ELEMENT",
+            "src": "{{ meta(application.logoUrl) }}",
+            "type": "IMAGE",
+            "width": ""
+          },
+          {
+            "align": "center",
+            "type": "TEXT",
+            "id": "heading_usertype",
+            "label": "{{ t(signup:forms.user_type.title) }}",
+            "variant": "HEADING_1"
+          },
+          {
+            "type": "BLOCK",
+            "id": "block_usertype",
+            "components": [
+              {
+                "type": "SELECT",
+                "id": "usertype_input",
+                "ref": "userType",
+                "label": "{{ t(signup:forms.user_type.fields.user_type.label) }}",
+                "placeholder": "{{ t(signup:forms.user_type.fields.user_type.placeholder) }}",
+                "required": true,
+                "options": []
+              },
+              {
+                "type": "ACTION",
+                "id": "action_usertype",
+                "label": "{{ t(signup:forms.user_type.actions.continue.label) }}",
+                "variant": "PRIMARY",
+                "eventType": "SUBMIT"
+              }
+            ]
+          }
+        ]
+      },
+      "prompts": [
+        {
+          "inputs": [
+            {
+              "ref": "usertype_input",
+              "identifier": "userType",
+              "type": "SELECT",
+              "required": true
+            }
+          ],
+          "action": {
+            "ref": "action_usertype",
+            "nextNode": "user_type_resolver"
+          }
+        }
+      ]
+    },
+    {
+      "id": "prompt_email",
+      "type": "PROMPT",
+      "meta": {
+        "components": [
+          {
+            "alt": "{{ t(signup:images.app_logo.alt) }}",
+            "category": "DISPLAY",
+            "height": "60",
+            "id": "image",
+            "resourceType": "ELEMENT",
+            "src": "{{ meta(application.logoUrl) }}",
+            "type": "IMAGE",
+            "width": ""
+          },
+          {
+            "align": "center",
+            "type": "TEXT",
+            "id": "text_header_email",
+            "label": "{{ t(signup:forms.email.title) }}",
+            "variant": "HEADING_1"
+          },
+          {
+            "type": "BLOCK",
+            "id": "block_email",
+            "components": [
+              {
+                "id": "input_prompt_email",
+                "ref": "email",
+                "type": "EMAIL_INPUT",
+                "label": "{{ t(signup:forms.email.fields.email.label) }}",
+                "required": true,
+                "placeholder": "{{ t(signup:forms.email.fields.email.placeholder) }}"
+              },
+              {
+                "type": "ACTION",
+                "id": "action_submit_email",
+                "label": "{{ t(signup:forms.email.actions.next.label) }}",
+                "variant": "PRIMARY",
+                "eventType": "SUBMIT"
+              }
+            ]
+          }
+        ]
+      },
+      "prompts": [
+        {
+          "inputs": [
+            {
+              "ref": "input_prompt_email",
+              "identifier": "email",
+              "type": "EMAIL_INPUT",
+              "required": true
+            }
+          ],
+          "action": {
+            "ref": "action_submit_email",
+            "nextNode": "check_email_uniqueness"
+          }
+        }
+      ]
+    },
+    {
+      "id": "check_email_uniqueness",
+      "type": "TASK_EXECUTION",
+      "executor": {
+        "name": "AttributeUniquenessValidator"
+      },
+      "onSuccess": "invite_generate",
+      "onIncomplete": "prompt_email"
+    },
+    {
+      "id": "invite_generate",
+      "type": "TASK_EXECUTION",
+      "executor": {
+        "name": "InviteExecutor",
+        "mode": "generate"
+      },
+      "onSuccess": "send_registration_email"
+    },
+    {
+      "id": "send_registration_email",
+      "type": "TASK_EXECUTION",
+      "properties": {
+        "emailTemplate": "SELF_REGISTRATION"
+      },
+      "executor": {
+        "name": "EmailExecutor",
+        "mode": "send"
+      },
+      "onSuccess": "invite_verify"
+    },
+    {
+      "id": "invite_verify",
+      "type": "TASK_EXECUTION",
+      "inputs": [
+        {
+          "ref": "input_003",
+          "identifier": "inviteToken",
+          "type": "HIDDEN",
+          "required": true
+        }
+      ],
+      "executor": {
+        "name": "InviteExecutor",
+        "mode": "verify"
+      },
+      "onSuccess": "prompt_user_details"
+    },
+    {
+      "id": "prompt_user_details",
+      "type": "PROMPT",
+      "meta": {
+        "components": [
+          {
+            "alt": "{{ t(signup:images.app_logo.alt) }}",
+            "category": "DISPLAY",
+            "height": "60",
+            "id": "image",
+            "resourceType": "ELEMENT",
+            "src": "{{ meta(application.logoUrl) }}",
+            "type": "IMAGE",
+            "width": ""
+          },
+          {
+            "align": "center",
+            "type": "TEXT",
+            "id": "text_header_details",
+            "label": "{{ t(signup:forms.user_details.title) }}",
+            "variant": "HEADING_1"
+          },
+          {
+            "type": "BLOCK",
+            "id": "block_user_details",
+            "components": [
+              {
+                "id": "input_prompt_username",
+                "ref": "username",
+                "type": "TEXT_INPUT",
+                "label": "{{ t(signup:forms.user_details.fields.username.label) }}",
+                "required": true,
+                "placeholder": "{{ t(signup:forms.user_details.fields.username.placeholder) }}"
+              },
+              {
+                "id": "input_prompt_given_name",
+                "ref": "given_name",
+                "type": "TEXT_INPUT",
+                "label": "{{ t(signup:forms.user_details.fields.first_name.label) }}",
+                "required": false,
+                "placeholder": "{{ t(signup:forms.user_details.fields.first_name.placeholder) }}"
+              },
+              {
+                "id": "input_prompt_family_name",
+                "ref": "family_name",
+                "type": "TEXT_INPUT",
+                "label": "{{ t(signup:forms.user_details.fields.last_name.label) }}",
+                "required": false,
+                "placeholder": "{{ t(signup:forms.user_details.fields.last_name.placeholder) }}"
+              },
+              {
+                "id": "input_prompt_password",
+                "ref": "password",
+                "type": "PASSWORD_INPUT",
+                "label": "{{ t(signup:forms.credential.fields.password.label) }}",
+                "required": true,
+                "placeholder": "{{ t(signup:forms.credential.fields.password.placeholder) }}"
+              },
+              {
+                "type": "ACTION",
+                "id": "action_submit_details",
+                "label": "{{ t(signup:forms.credential.actions.submit.label) }}",
+                "variant": "PRIMARY",
+                "eventType": "SUBMIT"
+              }
+            ]
+          }
+        ]
+      },
+      "prompts": [
+        {
+          "inputs": [
+            {
+              "ref": "input_prompt_username",
+              "identifier": "username",
+              "type": "TEXT_INPUT",
+              "required": true
+            },
+            {
+              "ref": "input_prompt_given_name",
+              "identifier": "given_name",
+              "type": "TEXT_INPUT",
+              "required": false
+            },
+            {
+              "ref": "input_prompt_family_name",
+              "identifier": "family_name",
+              "type": "TEXT_INPUT",
+              "required": false
+            },
+            {
+              "ref": "input_prompt_password",
+              "identifier": "password",
+              "type": "PASSWORD_INPUT",
+              "required": true
+            }
+          ],
+          "action": {
+            "ref": "action_submit_details",
+            "nextNode": "check_user_details_uniqueness"
+          }
+        }
+      ]
+    },
+    {
+      "id": "check_user_details_uniqueness",
+      "type": "TASK_EXECUTION",
+      "executor": {
+        "name": "AttributeUniquenessValidator"
+      },
+      "onSuccess": "provisioning",
+      "onIncomplete": "prompt_user_details"
+    },
+    {
+      "id": "provisioning",
+      "type": "TASK_EXECUTION",
+      "inputs": [
+        {
+          "ref": "input_email",
+          "identifier": "email",
+          "type": "EMAIL_INPUT",
+          "required": true
+        },
+        {
+          "ref": "input_username",
+          "identifier": "username",
+          "type": "TEXT_INPUT",
+          "required": true
+        },
+        {
+          "ref": "input_password",
+          "identifier": "password",
+          "type": "PASSWORD_INPUT",
+          "required": true
+        },
+        {
+          "ref": "input_given_name",
+          "identifier": "given_name",
+          "type": "TEXT_INPUT",
+          "required": false
+        },
+        {
+          "ref": "input_family_name",
+          "identifier": "family_name",
+          "type": "TEXT_INPUT",
+          "required": false
+        }
+      ],
+      "executor": {
+        "name": "ProvisioningExecutor"
+      },
+      "onSuccess": "end"
+    },
+    {
+      "id": "end",
+      "type": "END"
+    }
+  ]
+}
--- a/backend/internal/flow/executor/email_executor.go
+++ b/backend/internal/flow/executor/email_executor.go
@@ -19,7 +19,6 @@
 package executor
 
 import (
-	"context"
 	"errors"
 	"fmt"
 
@@ -31,8 +30,7 @@ import (
 )
 
 // emailExecutor sends emails based on the configured email template and runtime context data.
-// When email is not configured (emailClient is nil), the executor completes as a no-op
-// without setting the emailSent flag, allowing the SDK to fall back to displaying the invite link.
+// When email is not configured (emailClient is nil), it completes as a no-op with emailSent=false.
 type emailExecutor struct {
 	core.ExecutorInterface
 	logger          *log.Logger
@@ -41,7 +39,7 @@ type emailExecutor struct {
 }
 
 // newEmailExecutor creates a new instance of the email executor.
-// emailClient may be nil if SMTP is not configured; the executor will act as a no-op in that case.
+// emailClient may be nil if SMTP is not configured; the executor completes as a no-op in that case.
 func newEmailExecutor(flowFactory core.FlowFactoryInterface, emailClient email.EmailClientInterface,
 	templateService template.TemplateServiceInterface) *emailExecutor {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "EmailExecutor"))
@@ -83,8 +81,8 @@ func (e *emailExecutor) executeSend(ctx *core.NodeContext) (*common.ExecutorResp
 	}
 
 	// If email client is not configured, complete as a no-op.
-	// The emailSent flag will NOT be set, so the SDK can fall back to showing the invite link.
 	if e.emailClient == nil {
+		execResp.AdditionalData[common.DataEmailSent] = dataValueFalse
 		logger.Debug("Email client not configured, skipping email send")
 		execResp.Status = common.ExecComplete
 		return execResp, nil
@@ -120,15 +118,16 @@ func (e *emailExecutor) executeSend(ctx *core.NodeContext) (*common.ExecutorResp
 	}
 
 	inviteLink := ctx.RuntimeData[common.RuntimeKeyInviteLink]
-	if scenario == template.ScenarioUserInvite && inviteLink == "" {
+	if (scenario == template.ScenarioUserInvite || scenario == template.ScenarioSelfRegistration) && inviteLink == "" {
 		return nil, errors.New("invite link not found in runtime data")
 	}
 
 	templateData := template.TemplateData{
 		"inviteLink": inviteLink,
+		"appName":    ctx.Application.Name,
 	}
 
-	rendered, svcErr := e.templateService.Render(context.Background(), scenario, templateData)
+	rendered, svcErr := e.templateService.Render(ctx.Context, scenario, templateData)
 	if svcErr != nil {
 		return nil, fmt.Errorf("failed to render email template: %s", svcErr.Code)
 	}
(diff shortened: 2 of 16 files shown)


backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
```diff
@@ -0,0 +1,398 @@
...
+    {
+      "id": "start",
+      "type": "START",
+      "onSuccess": "user_type_resolver"
+    },
+    {
+      "id": "user_type_resolver",
+      "type": "TASK_EXECUTION",
+      "executor": {
+        "name": "UserTypeResolver"
```
darshanasbg, line 15: In this flow, `UserTypeResolver` executor should only list the user-types who have enabled self sign up.. Is that the case?

darshanasbg: cc: @thiva-k @ThaminduDilshan 

If not we should improve the UserTypeResolver. For admin invite, it should list any type, but for registration type flow, it should only allow to pick users who have self registration enabled.

backend/cmd/server/bootstrap/flows/registration/registeration_flow_basic_invite.json
```diff
@@ -0,0 +1,398 @@
...
+          ],
+          "action": {
+            "ref": "action_usertype",
+            "nextNode": "user_type_resolver"
+          }
+        }
+      ]
+    },
+    {
+      "id": "prompt_email",
```
darshanasbg, line 84: Considering the common experience, i would like have the default flow to request all the user attributes in one page and have the email confirmation pushed to the end just before the provisioning node.

So lets merge this with the prompt_user_details

backend/cmd/server/repository/resources/templates/registeration_invite.yaml
```diff
@@ -0,0 +1,20 @@
...
+  	<h2>Complete Your Registration for {{ctx(appName)}}</h2>
+  	<p>You recently started the registration process for <strong>{{ctx(appName)}}</strong>. Click the link below to complete your account setup:</p>
+  	<p><a href="{{ctx(inviteLink)}}" style="display: inline-block; padding: 10px 20px;
+  		background-color: #ff7300; color: #fff; text-decoration: none;
+  		border-radius: 4px;">Complete Registration</a></p>
+  	<p>If the button above doesn't work, copy and paste the following link into your browser:</p>
+  	<p><a href="{{ctx(inviteLink)}}">{{ctx(inviteLink)}}</a></p>
+  	<p>If you did not initiate this request, you can safely ignore this email.</p>
+  </body>
+  </html>
```
darshanasbg: Add a new line

backend/internal/flow/core/prompt_node.go
```diff
@@ -109,8 +109,8 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 	logger.Debug("Required inputs or action not available, prompting user",
 		log.Any("inputs", nodeResp.Inputs), log.Any("actions", nodeResp.Actions))
 
-	// Include meta in the response if verbose mode is enabled
-	if ctx.Verbose && n.GetMeta() != nil {
+	// Always include meta if the node has it defined — meta.components is required UI data, not optional debug info
+	if n.GetMeta() != nil {
```
darshanasbg: I don't think we can remove `ctx.Verbose &&` part..
cc: @ThaminduDilshan

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
Dilusha-Madushan: > In this flow, `UserTypeResolver` executor should only list the user-types who have enabled self sign up.. Is that the case?

Yes, that is why I added that but in admin invite any user type is listed for invite.

thiva-k: Yes this is the current behavior. Only for admin invite all types are listed. For registration flow both allowed user types from app config as well as schema's self registration enabled is checked.

backend/internal/flow/executor/invite_executor.go
```diff
@@ -151,6 +152,14 @@ func (e *inviteExecutor) generateInviteLink(ctx *core.NodeContext, inviteToken s
...
+	queryParams := url.Values{
+		"flowId":      []string{ctx.FlowID},
+		"inviteToken": []string{inviteToken},
+	}
+
+	if ctx.AppID != "" {
+		queryParams.Set(oauth2const.AppID, ctx.AppID)
+	}
 
-	return fmt.Sprintf("%s/invite?flowId=%s&inviteToken=%s", gateAppURL, ctx.FlowID, inviteToken)
```
darshanasbg, line 155: (Not related to this PR)
@ThaminduDilshan, @rajithacharith : looking at the existing code, having the url constructed as simple text, is bit alarming. Lets think on someway we can avoid such changes getting in.. And we have to look other places that have done similar way, and may be we can reuse the above approach to catch the existing places too..

backend/internal/flow/executor/email_executor.go
```diff
@@ -83,8 +82,16 @@ func (e *emailExecutor) executeSend(ctx *core.NodeContext) (*common.ExecutorResp
 	}
 
 	// If email client is not configured, complete as a no-op.
-	// The emailSent flag will NOT be set, so the SDK can fall back to showing the invite link.
+	// For USER_ONBOARDING flows, expose the invite link in AdditionalData so admin UIs can
+	// display it inline for manual sharing. For self-registration flows, do NOT expose the
+	// invite link — leaking it back to the browser bypasses the email verification step.
 	if e.emailClient == nil {
+		if ctx.FlowType == common.FlowTypeUserOnboarding {
```
thiva-k: Let's keep the email executor generic. Instead, let's not add the invite link in the additional data for registration flows in the invite executor

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
```diff
@@ -0,0 +1,398 @@
+{
```
thiva-k, line 1: Let's fix the typo in the file name

backend/internal/flow/core/prompt_node_test.go
```diff
@@ -189,58 +189,22 @@ func (s *PromptOnlyNodeTestSuite) TestExecuteWithVerboseModeEnabled() {
 		},
 	})
 
-	// Test with verbose mode enabled
```
thiva-k, line 192: Shall we revert these changes?

backend/internal/flow/flowexec/engine.go
```diff
@@ -182,6 +182,15 @@ func (fe *flowEngine) Execute(ctx *EngineContext) (FlowStep, *serviceerror.Servi
 		flowStep.Assertion = ctx.Assertion
 	}
 
+	// Include accumulated additional data and app name for the frontend
+	flowStep.Data.AdditionalData = ctx.AdditionalData
+	if flowStep.Data.AdditionalData == nil {
+		flowStep.Data.AdditionalData = make(map[string]string)
+	}
+	if ctx.Application.Name != "" {
```
thiva-k: Let's not add application name to AdditionalData in all completed flows at the engine-level

backend/cmd/server/bootstrap/flows/registration/registeration_flow_basic_invite.json
Dilusha-Madushan: - Removed prompt_email and check_email_uniqueness nodes
- Merged email input into prompt_user_details — now one page collects email, username, first name, last name, check_user_details_uniqueness now validates all attributes including email

backend/cmd/server/repository/resources/templates/registeration_invite.yaml
Dilusha-Madushan: Added the new line

backend/internal/flow/core/prompt_node.go
Dilusha-Madushan: When a user clicks an invite link (which has a pre-existing flowId), the engine calls loadPrevContext to resume the flow. This function was NOT accepting the verbose parameter from the request, so engineCtx.Verbose was always loaded from the DB as false (since the invite was created by the admin without verbose=true). Even though the AcceptInvite SDK component correctly sends verbose: true, it was being silently ignored — so meta (containing the components the frontend needs to render the form) was never returned.

Fix in service.go: Pass verbose to loadPrevContext and set engineCtx.Verbose = verbose so the current request's verbose flag is honored when resuming existing flows.

backend/internal/flow/executor/email_executor.go
Dilusha-Madushan: Moved the invite link AdditionalData exposure logic from EmailExecutor to InviteExecutor, keeping the email executor generic.

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
Dilusha-Madushan: Fixed that

backend/internal/flow/core/prompt_node_test.go
Dilusha-Madushan: loadPrevContext was loading Verbose: false from the DB (set when the invite was created), so even though AcceptInvite SDK sends verbose: true, it was ignored — causing meta to never be returned and the invite page to render blank; fixed by passing verbose from the request into loadPrevContext so it correctly updates engineCtx.Verbose.

backend/internal/flow/flowexec/engine.go
Dilusha-Madushan: Moved appName population into AdditionalData from the engine level down to the provisioning executor, so it's only set when a user is actually provisioned

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
```diff
@@ -0,0 +1,341 @@
...
+      "type": "TASK_EXECUTION",
+      "properties": {
+        "emailTemplate": "SELF_REGISTRATION"
+      },
+      "executor": {
+        "name": "EmailExecutor",
+        "mode": "send"
+      },
+      "onSuccess": "invite_verify"
+    },
```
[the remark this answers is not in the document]
Dilusha-Madushan: here we cannot expose the inviteLink anyway for the user. So if email not sent, user see the default failure message from SignUpBox.

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
```diff
@@ -57,12 +57,62 @@ const StyledPaper = styled(Paper)(({theme}) => ({
...
+      );
+    }
+    if (!error) {
+      return (
+        <Alert severity="error" sx={{mb: 2}}>
+          <AlertTitle>{t("Oops, that didn't work")}</AlertTitle>
+          {t("We're sorry, we ran into a problem. Please try again!")}
+        </Alert>
+      );
+    }
```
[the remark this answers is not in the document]
Dilusha-Madushan: This is not a bug — the condition is intentional. Look at the full context of renderFlowContent:


components.length > 0  → render form
emailSent              → show email sent message
!error                 → show "Oops" fallback
return null            → (error already shown above the component)
The if (!error) case handles an unexpected empty state — when the flow returns no components, no email was sent, and there's also no error. It's a fallback for when something went wrong but no error was reported.

If there IS an error (error is truthy), it returns null here because the error is already rendered above at lines 151–156:


{error && (
  <Alert severity="error" sx={{mb: 2}}>
    <AlertTitle>{t('signup:errors.signup.failed.message')}</AlertTitle>
    ...
  </Alert>
)}

backend/internal/flow/executor/invite_executor.go
```diff
@@ -89,7 +91,14 @@ func (e *inviteExecutor) executeGenerate(ctx *core.NodeContext) (*common.Executo
 
 	execResp.RuntimeData[common.RuntimeKeyStoredInviteToken] = inviteToken
 	execResp.RuntimeData[common.RuntimeKeyInviteLink] = inviteLink
-	execResp.AdditionalData[common.DataInviteLink] = inviteLink
+
+	// For USER_ONBOARDING flows, expose the invite link in AdditionalData so admin UIs
+	// can display it inline for manual sharing when SMTP is not configured.
+	// For self-registration flows, do NOT expose it to prevent leaking the link back to
+	// the browser, which would bypass the email verification step.
```
thiva-k: Let's remove this verbose comment

backend/internal/flow/core/prompt_node.go
ThaminduDilshan: We can't remove `ctx.Verbose`. Engine supposed to work in two modes; with UI meta components and in pure backend mode. This will break the backend scenario

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite.json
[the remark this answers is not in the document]
Dilusha-Madushan: emailSent in SignUpBox is now boolean | null (was boolean):

null — initial state, emailSent not yet known from the flow
true — email sent successfully → show "Check your inbox" info alert
false — email send skipped (no SMTP) → show "Email service unavailable" warning alert
onFlowChange now captures both 'true' and 'false' values from additionalData.emailSent. The engine accumulates AdditionalData across steps, so emailSent: "false" from the email executor is present in the response even though invite_verify runs after it.

backend/internal/flow/executor/invite_executor.go
Dilusha-Madushan: removed that

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -183,6 +183,13 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
 	execResp.AuthenticatedUser = authenticatedUser
 	execResp.Status = common.ExecComplete
 
+	if ctx.Application.Name != "" {
```
ThaminduDilshan: Why this is handled in the provisioning executor? Adding these application specific data doesn't seem to be a responsibility of the user provisioning executor

backend/internal/flow/flowexec/engine.go
```diff
@@ -182,6 +182,12 @@ func (fe *flowEngine) Execute(ctx *EngineContext) (FlowStep, *serviceerror.Servi
 		flowStep.Assertion = ctx.Assertion
 	}
 
+	// Include accumulated additional data for the frontend
+	flowStep.Data.AdditionalData = ctx.AdditionalData
```
ThaminduDilshan: Reaching here means the flow execution has been completed and flow context will be removed next by the flow exec service.
Why is this change required and why is it added here?

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
```diff
@@ -22,23 +22,93 @@
...
+              component={component}
+              index={index}
+              values={values ?? {}}
+              touched={touched}
+              fieldErrors={fieldErrors}
+              isLoading={isLoading}
+              resolve={resolve}
+              onInputChange={handleInputChange}
+              onSubmit={(action, inputs) => {
+                // Tracker: https://github.com/asgardeo/javascript/issues/222
```
brionmario: We can remove this
```suggestion
```

backend/internal/flow/common/constants.go
```diff
@@ -137,6 +137,10 @@ const (
 	DataEmailSent = "emailSent"
 	// DataRootOUID is the key used to pass the root OU ID to the frontend for the OU tree picker.
 	DataRootOUID = "rootOuId"
+	// DataAppName is the key used for the application name in the flow response additional data.
+	DataAppName = "appName"
+	// DataAppURL is the key used for the application access URL in the flow response additional data.
+	DataAppURL = "appUrl"
```
ThaminduDilshan: Why do we need these two parameters?

I guess this is to display application name and application url in the email sent confirmation page. If that's the case, why do we need to display that in the email sent confirmation page? User is already logging into a application and I don't see a valid requirement to show app name/ url here.

Even if we need to display them, there is a different way to obtain these parameters using `flow/meta`. In the flow graph we can configure placeholders for these variables, and in runtime gate will resolve them by retrieving values through the `flow/meta` API call

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
```diff
@@ -22,23 +22,93 @@
...
+    touched: any,
+    fieldErrors: any,
+    handleInputChange: any,
+    handleSubmit: any,
+  ): JSX.Element | null => {
+    if (components.length > 0) {
+      return (
+        <Box sx={{display: 'flex', flexDirection: 'column', gap: 2}}>
+          {isLoading && (
+            <Typography sx={{textAlign: 'center'}}>{t('signup:creating', 'Creating account...')}</Typography>
```
brionmario: Is this added to the `thunder-i18n` bundle?
```suggestion
            <Typography sx={{textAlign: 'center'}}>{t('signup:create_account.loading', 'Creating account...')}</Typography>
```

backend/internal/flow/common/constants.go
ThaminduDilshan: cc: @brionmario

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
```diff
@@ -66,62 +146,46 @@ export default function SignUpBox(): JSX.Element {
...
-                    textDecoration: 'underline',
-                    '&:hover': {
+            {emailSent !== true && (
+              <Typography sx={{textAlign: 'center', mt: 3}}>
+                <Trans i18nKey="signup:redirect.to.signin">
+                  Already have an account?
+                  <Button
+                    variant="text"
+                    onClick={() => {
+                      // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
brionmario: try void and see rather than suppressing.

backend/internal/flow/executor/provisioning_executor.go
thiva-k: We do not need to send the application name via flow response at all right? This should be resolved from the `flow/meta` api since we are having the app id in the url?

backend/internal/flow/common/constants.go
brionmario: +1. Incase we need to show, we can utilize the flow/meta if we have the app UUID.

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
Dilusha-Madushan: I missed that, now added.

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
Dilusha-Madushan: updated this part

backend/internal/flow/common/constants.go
Dilusha-Madushan: I removed those values and now using app/meta to get application data. Test the flow and it's working.

frontend/apps/thunder-gate/src/components/SignUp/SignUpBox.tsx
Dilusha-Madushan: Addressed

backend/internal/flow/flowexec/engine.go
Dilusha-Madushan: I added that to get app name and access url values into response. Now I changed the code and now use flow/meta to fetch app data

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: Fixed that

backend/internal/flow/flowexec/service.go
```diff
@@ -241,12 +241,13 @@ func (s *flowExecService) getFlowExpirySeconds(flowType common.FlowType) int64 {
 
 // loadPrevContext retrieves the flow context from the store based on the given details.
 func (s *flowExecService) loadPrevContext(ctx context.Context, flowID, action string,
-	inputs map[string]string, logger *log.Logger) (*EngineContext, *serviceerror.ServiceError) {
+	inputs map[string]string, verbose bool, logger *log.Logger) (*EngineContext, *serviceerror.ServiceError) {
```
thiva-k: Do we need to pass verbose as an arg here?

Dilusha-Madushan: The reason verbose is passed to loadPrevContext:

InitiateFlow (line 526) always initializes a flow context with verbose = true:


engineCtx, err := s.initContext(ctx, initContext.ApplicationID, flowType, true, logger)
The comment there explains it: "This uses verbose true to ensure step layouts are returned during execution."

This is used for the invite/self-registration flow — the flow is pre-initialized before the user actually starts executing steps. The stored context already has verbose=true.

Now when the user subsequently calls Execute, they might pass verbose=false (or the client doesn't explicitly request verbose). Without the OR logic, that would overwrite the stored verbose=true and the frontend would stop receiving the step layout (meta/components) needed to render the UI forms.

ThaminduDilshan: Where do we use `InitiateFlow()` for the self registration path? Idea of this function is for a different path. This is only called from the oauth service. If someone is using oauth and flow intiated there, verbose will always be enabled through this method.

But a general flow/execution API request will always go through the Execute() method. It accepts verbose as a parameter during the initial flow execute request. There after we're not accepting this as a request parameter.

1. If the invite user registration flow is started with verbose=true in the very first request, it will contain meta.
2. If the invite user registration flow is started without passing verbose flag, that means they're not using our UIs either though SDK/ gate. In that case, they don't need meta even after user clicking on the invite

ThaminduDilshan: If you check [1], we're treating a request as an existing flow, only if the flowId is present in request. If the flowId is present, we're not even reading verbose from the request params [2]. So in your case, email invite will contain the flowId, and therefore it will go through [2]. It doesn't overide verbose with the request param, rather take the value from context

[1] https://github.com/asgardeo/thunder/blob/57a6902cb939cb5d4ab4a27caf290c1cbe767340/backend/internal/flow/flowexec/service.go#L91
[2] https://github.com/asgardeo/thunder/blob/57a6902cb939cb5d4ab4a27caf290c1cbe767340/backend/internal/flow/flowexec/service.go#L120

ThaminduDilshan: @thiva-k I think the problem lies in a different place (what we discussed offline). Here we're always redirecting user to the gate. Gate require verbose to be true by design. But in a native flow, verbose can be false and ideally the redirection should happen to the client app. But in the current implementation, it always goes to the gate and UI will fail if the flow is started in non verbose mode

thiva-k: The redirection part here is just the way the invite link is created right? We can maybe have a node property to determine whether the invite link should be created based on thunder gate app or a custom client url or maybe just send the invite token only in additional data(which can be used for the app-native flow), so the client has full control on the invite link creation, experience, etc.

Dilusha-Madushan: After reverting the verbose param and || from loadPrevContext, both user invite and registration invite flows still work correctly. This confirms the change is redundant — InitiateFlow already stores verbose=true in DB, and loadPrevContext already restores it via ToEngineContext. I'm removing this unnecessary change.

backend/internal/flow/core/prompt_node.go
Dilusha-Madushan: This one is handled and rollback the change

frontend/apps/thunder-gate/src/components/AcceptInvite/AcceptInviteBox.tsx
```diff
@@ -107,8 +110,25 @@ export default function AcceptInviteBox(): JSX.Element {
             return (
               <Box sx={{textAlign: 'center', py: 2}}>
                 <Alert severity="success">
-                  {t('invite:complete.description', 'Your account has been successfully set up.')}
+                  {appName
+                    ? t(
+                        'invite:complete.description.withApp',
```
brionmario, line 118: We already have a namespace for `onboarding` right?
`onboarding:xxx`? BTW, to which lang bundle did we add this? Runtime bundle (`en-US.json`) or the local i18n bundle (`en-US.ts`).

cc: / @thiva-k

thiva-k: It is in the runtime bundle `en-US.json`

Dilusha-Madushan: I added new namespace for invite  in en-us.json and move the en-US.ts invite keys into there.

PR #1995, opened 2026-03-27
Fix schema validation failing on user update when credentials are absent

### LINKED ISSUE #1991: Schema validation fails on user update when credential fields are required
## Description

When updating a user via the UI (or API), the request fails with:

```json
{
    "code": "USR-1019",
    "message": "Schema validation failed",
    "description": "User attributes do not conform to the required schema"
}
```

## Root Cause

In the `UpdateUser` flow, credential attributes (e.g. `password`) are extracted from `user.Attributes` before schema validation runs. However, `Schema.Validate()` checks **all** schema properties including credential ones. If a credential field is marked `required: true` in the schema, validation fails because the credential has already been removed from the attributes map.

**Flow:**
1. `extractCredentials()` removes credential fields from `user.Attributes` → `backend/internal/user/service.go:629`
2. `validateUserAndUniqueness()` validates the now-stripped attributes against the full schema → `backend/internal/user/service.go:644`
3. `Schema.Validate()` sees the required credential field is missing → returns `false` → `backend/internal/userschema/model/schema.go:155-160`

## Fix

Skip credential properties during schema attribute validation since they are extracted and validated separately in the user service layer.

**PR:** https://github.com/asgardeo/thunder/pull/1990

DonOmalVindula:
## Summary
- Schema validation was rejecting user updates because required credential fields (e.g. `password`) were missing from the payload — the UI doesn't send credentials during updates
- Added `skipCredentialRequired` parameter to `ValidateUser` so update flows skip the required check for credential properties while still enforcing it during creation
- Reordered `UpdateUser` to validate attributes **before** extracting credentials, so credential values (regex, type) are still validated when present

### Single credential schema (`password`: required, credential, regex `^[A-Za-z0-9@#$%^&+=!]{8,}$`)

**Create:**
- [x] Both required fields + valid password → success
- [x] Missing password → USR-1019
- [x] Empty string password → USR-1019 (fails regex)
- [x] Too-short password → USR-1019 (fails regex)
- [x] Password with invalid chars (spaces) → USR-1019
- [x] Password as number (wrong type) → USR-1019
- [x] Password as boolean (wrong type) → USR-1019
- [x] Password as null → USR-1019
- [x] Extra attribute not in schema → USR-1019

**Update (PUT):**
- [x] Without credentials → success
- [x] With valid password → success
- [x] Too-short password → USR-1019
- [x] Password with spaces → USR-1019
- [x] Password as number → USR-1019
- [x] Password as boolean → USR-1019
- [x] Password as null → USR-1019
- [x] Empty string password → USR-1019
- [x] Extra attribute not in schema → USR-1019
- [x] Missing required non-credential field → USR-1019
- [x] Empty attributes (missing all required) → USR-1019

### Multi-credential schema (`password` + `pin`: both required, credential, with regex)

**Create:**
- [x] Both credentials valid → success
- [x] Both missing → USR-1019
- [x] Only password, missing pin → USR-1019
- [x] Only pin, missing password → USR-1019
- [x] Valid password, invalid pin (too long) → USR-1019
- [x] Invalid password (short), valid pin → USR-1019
- [x] Both credentials invalid → USR-1019
- [x] Pin with letters (fails regex) → USR-1019

**Update (PUT):**
- [x] No credentials at all → success
- [x] Only valid password, no pin → success
- [x] Only valid pin, no password → success
- [x] Both credentials valid → success
- [x] Valid password, invalid pin → USR-1019
- [x] Invalid password, valid pin → USR-1019
- [x] Both credentials invalid → USR-1019
- [x] Only invalid password, no pin → USR-1019
- [x] Only invalid pin, no password → USR-1019

Closes #1991

### DIFF
--- a/backend/internal/user/service.go
+++ b/backend/internal/user/service.go
@@ -324,7 +324,7 @@ func (us *userService) CreateUser(ctx context.Context, user *User) (*User, *serv
 		return nil, svcErr
 	}
 
-	if svcErr := us.validateUserAndUniqueness(ctx, user.Type, user.Attributes, logger, ""); svcErr != nil {
+	if svcErr := us.validateUserAndUniqueness(ctx, user.Type, user.Attributes, logger, "", false); svcErr != nil {
 		return nil, svcErr
 	}
 
@@ -626,11 +626,6 @@ func (us *userService) UpdateUser(ctx context.Context, userID string, user *User
 			fmt.Errorf("schema service error: %s", svcErr.ErrorDescription), log.String("id", userID))
 	}
 
-	credentials, err := us.extractCredentials(user, schemaCredentialAttributes)
-	if err != nil {
-		return nil, logErrorAndReturnServerError(logger, "Failed to extract credentials", err, log.String("id", userID))
-	}
-
 	var capturedSvcErr *serviceerror.ServiceError
 
 	err = us.transactioner.Transact(ctx, func(txCtx context.Context) error {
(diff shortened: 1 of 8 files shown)


backend/internal/user/service.go
```diff
@@ -641,12 +636,22 @@ func (us *userService) UpdateUser(ctx context.Context, userID string, user *User
...
+			txCtx, user.Type, user.Attributes, logger, user.ID, true,
+		); svcErr != nil {
 			capturedSvcErr = svcErr
 			return errors.New("rollback for validation error")
 		}
 
-		err := us.userStore.UpdateUser(txCtx, user)
+		credentials, err := us.extractCredentials(user, schemaCredentialAttributes)
+		if err != nil {
+			return err
```
ThaminduDilshan:
```suggestion
			return fmt.Errorf("Failed to extract credentials: %w", err)
```
Otherwise we'll lose the failing credentials validation part right?

DonOmalVindula: Yes fixed

PR #1996, opened 2026-03-27
Add allowed user schema support to node poperty 

thiva-k:
This pull request adds support for specifying an `allowedUserTypes` property at the node level in user registration and onboarding flows. This allows flows to further restrict which user types are available for selection or auto-selection, beyond the application-level allowed user types. The implementation includes filtering logic, validation, and comprehensive test coverage for these new behaviors.

**Feature: Node-level Allowed User Types**

* Added a new node property key, `allowedUserTypes`, to the constants in `constants.go` to support node-level user type restrictions.
* Implemented logic in `user_type_resolver.go` to:
  - Filter the list of allowed user types at the node level during registration and onboarding flows.
  - Validate user input against the node-level allowed list and return appropriate failure reasons if no valid user types are available.
  - Auto-select the user type if only one valid option remains after filtering. [[1]](diffhunk://#diff-d8781035d0c0c7f33924e03093b5921306623cdaaab0ed5709dc17836de00cb7R141-R161) [[2]](diffhunk://#diff-d8781035d0c0c7f33924e03093b5921306623cdaaab0ed5709dc17836de00cb7R185-R198) [[3]](diffhunk://#diff-d8781035d0c0c7f33924e03093b5921306623cdaaab0ed5709dc17836de00cb7L197-R312)

**Testing: Node-level Allowed User Types**

* Added comprehensive unit tests in `user_type_resolver_test.go` to cover:
  - Filtering of application-allowed user types by node-level restrictions.
  - Auto-selection when only one user type is allowed at the node level.
  - Prompting the user when multiple valid options exist.
  - Handling of invalid or mismatched allowed user types at the node level.
  - Validation of user input against the node-level allowed list for both registration and onboarding flows. [[1]](diffhunk://#diff-45787b2d1d1c67f5c397ac0d7e6e003982e616230047eb3c4031969df2598502R738-R846) [[2]](diffhunk://#diff-45787b2d1d1c67f5c397ac0d7e6e003982e616230047eb3c4031969df2598502R1023-R1050) [[3]](diffhunk://#diff-45787b2d1d1c67f5c397ac0d7e6e003982e616230047eb3c4031969df2598502R1084-R1229)

## Summary by CodeRabbit

* **New Features**
  * Added support for restricting allowed user types at the flow node level in registration and user onboarding workflows
  * Automatic filtering of available schemas and options based on configured user type restrictions
  * Enhanced validation with early error detection when no valid user types are available

### DIFF
--- a/backend/internal/flow/executor/user_type_resolver.go
+++ b/backend/internal/flow/executor/user_type_resolver.go
@@ -138,6 +138,27 @@ func (u *userTypeResolver) handleRegistrationFlows(ctx *core.NodeContext, execRe
 		return execResp, nil
 	}
 
+	// If allowedUserTypes is configured in node properties, filter the application-level allowed list
+	nodeAllowedUserTypes := u.getAllowedUserTypesFromProperties(ctx)
+	if len(nodeAllowedUserTypes) > 0 {
+		filtered := make([]string, 0, len(allowed))
+		for _, userType := range allowed {
+			if slices.Contains(nodeAllowedUserTypes, userType) {
+				filtered = append(filtered, userType)
+			}
+		}
+
+		if len(filtered) == 0 {
+			logger.Debug("No valid user types after filtering with node allowedUserTypes",
+				log.Any("applicationAllowed", allowed), log.Any("nodeAllowed", nodeAllowedUserTypes))
+			execResp.Status = common.ExecFailure
+			execResp.FailureReason = "No valid user types available for this flow"
+			return execResp, nil
+		}
+
+		allowed = filtered
+	}
+
 	// Check if userType is provided in inputs
 	if u.HasRequiredInputs(ctx, execResp) {
 		err := u.resolveUserTypeFromInput(reqCtx, execResp, ctx.UserInputs[userTypeKey], allowed)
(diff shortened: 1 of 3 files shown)


backend/internal/flow/executor/user_type_resolver.go
```diff
@@ -194,15 +227,89 @@ func (u *userTypeResolver) handleUserOnboardingFlows(ctx *core.NodeContext,
...
 
+// getAllowedUserTypesFromProperties reads the optional allowedUserTypes property from node properties.
+func (u *userTypeResolver) getAllowedUserTypesFromProperties(ctx *core.NodeContext) []string {
+	if ctx.NodeProperties == nil {
+		return nil
+	}
+
+	val, exists := ctx.NodeProperties[propertyKeyAllowedUserTypes]
+	if !exists {
+		return nil
```
ThaminduDilshan, line 270: If `propertyKeyAllowedUserTypes` is not defined, we're returning nil here. If defines, but empty, I think the function will return an empty list.

Can we use the same validation in the usages of this method? Idea is if someone doesn't want to enable invite/ registration for any user type, they can configure an empty list.
Not sure this validation can work when it's passed from top layers

darshanasbg: > if someone doesn't want to enable invite/ registration for any user type, they can configure an empty list.

IMO, in this case it should be disabled in more direct form, rather using the empty array.

thiva-k: Will address in follow-up PR if needed. Merging now for the release

PR #2094, opened 2026-03-31
Add coderabbit.yaml with specific instructions

sahandilshan:
### Purpose
$subject

## Summary by CodeRabbit

* **Chores**
  * Added configuration for automated code reviews (en-US) with language-specific checks for Go, documentation-validation rules, and explicit guidance when user-facing changes lack docs.
  * Enabled auto-review with title filters (e.g., WIP / DO NOT MERGE) and set main as the base branch.
  * Excluded common test/build paths and backup docs from review scope.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,53 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: en-US
+
+reviews:
+  request_changes_workflow: false
+  high_level_summary: true
+  poem: false
+
+  path_instructions:
+    - path: "**/*.go"
+      instructions: |
+        Check whether this change introduces or modifies any of the following:
+        - Public REST/OAuth2/OIDC API endpoints (new routes, changed request/response schemas, new query/path/header parameters)
+        - Configuration options (YAML/TOML keys, environment variables, deployment settings)
+        - Authentication or authorization flows (new grant types, new authenticators, changed flow steps)
+        - User-facing behavior changes (new features, changed defaults, removed capabilities)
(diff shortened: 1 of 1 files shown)


.coderabbit.yaml
```diff
@@ -0,0 +1,52 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: en-US
+
+reviews:
+  request_changes_workflow: true
```
pavinduLakshan: This allows Devs to merge PRs without any further approval from any human committer, once coderabbit has approved the pr. It was decided to keep this disabled. Will share the discussion offline

.coderabbit.yaml
```diff
@@ -0,0 +1,52 @@
...
+  high_level_summary: true
```
pavinduLakshan, line 6: We can disable sequence diagrams and poem as well if needed :)

.coderabbit.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  path_instructions:
+    - path: "**/*.go"
```
pavinduLakshan, line 10: We could add specific instructions for each folder path for coderabbit to better understand and enforce these instructions.

sahandilshan: So, currently all .go files are there with the backend changes. This instructions is generic for all those files. That's why added it as this. But when the specific requirement comes for each sub level, will folow this

.coderabbit.yaml
sahandilshan: Let's disable the poem

.coderabbit.yaml
sahandilshan: Mark this as false. Let the coderabbit to only to put comments

Issue #1877, opened 2026-03-19
Redis support for Runtime Database


rajithacharith:
### What problem are we solving?

Currently runtime db is SQL based and there is room for improvement in performance by moving this to Redis. 

### Who are we solving this for?

All the users- Needs to improve performance

### Why should we solve this now?

Runtime DB is a frequently used DB. Better if we can use Redis for this as Redis tends to give better perf compared to SQL 

### Proposed Solution

Use Redis for Runtime DB

KaveeshaPiumini:
Thunder Runtime Database Performance - https://docs.google.com/document/d/1GorZce0PV4eh1soRLBo83nZ73AQmDEDjVflM2P2YklI/edit?tab=t.tgmvlsigp5ij

PR #2120, opened 2026-04-01
Refactor runtime DB usages to fit redis

KaveeshaPiumini:
### Purpose
This pull request refactors the storage and retrieval of passkey session data and simplifies the related database schema, especially for flow context and user data. The main themes are database schema simplification and streamlining the session management logic and interfaces.

**Database Schema Simplification:**

* The `FLOW_CONTEXT` table is simplified: multiple columns related to flow state are replaced with a single `CONTEXT` field, and the primary key is now `(ID, DEPLOYMENT_ID)` instead of `(FLOW_ID, DEPLOYMENT_ID)` (`backend/dbscripts/runtimedb/postgres.sql`, `backend/dbscripts/runtimedb/sqlite.sql`) [[1]](diffhunk://#diff-7b38a144a3b6839a8d68351a45ba0035a72760787ff899b9556bc9479e87c076L32-R40) [[2]](diffhunk://#diff-e7ce2a8d87ad5ab4722ef69e4accc5fb5c63e9cca3261ed4567c6e302d3e8a5bL32-R40).
* The `FLOW_USER_DATA` table is removed entirely from both Postgres and SQLite schemas, reducing complexity and eliminating the need for cascade deletes (`backend/dbscripts/runtimedb/postgres.sql`, `backend/dbscripts/runtimedb/sqlite.sql`, `backend/dbscripts/runtimedb/postgres-cleanup.sql`) [[1]](diffhunk://#diff-7b38a144a3b6839a8d68351a45ba0035a72760787ff899b9556bc9479e87c076L55-L78) [[2]](diffhunk://#diff-e7ce2a8d87ad5ab4722ef69e4accc5fb5c63e9cca3261ed4567c6e302d3e8a5bL55-L78) [[3]](diffhunk://#diff-efa435ee158f72589a47ef9ad1e2cc4f041b4a9cfafbe0e6d2faa6dbf1de91deL52).

**Session Management Refactoring:**

* The `storeSessionData` and `storeSession` methods are refactored to take only the session data and expiry seconds, removing the need to pass `userID` and `relyingPartyID` as separate arguments. This change is reflected in both the implementation and the mocks/tests (`backend/internal/authn/passkey/service.go`, `backend/internal/authn/passkey/session.go`, `backend/internal/authn/passkey/sessionStoreInterface_mock_test.go`, `backend/internal/authn/passkey/session_test.go`) [[1]](diffhunk://#diff-531907e19c493cf9df2e91b807b7a43156929afd54f2d1afe336234d011503caL147-R147) [[2]](diffhunk://#diff-531907e19c493cf9df2e91b807b7a43156929afd54f2d1afe336234d011503caL370-R369) [[3]](diffhunk://#diff-39e8d5cc227671bad1e2aa0e63cf91cc4259bcff5a3bfdd493018e7423b9a16dL47-R55) [[4]](diffhunk://#diff-9fb05941365966026620e07db52ecf1b0f840737b0edd91e9bc5149fd9ca56cdL155-R161) [[5]](diffhunk://#diff-66abd9c9309c07ec0bf054bcfd6534bb121ef81025d371a46e67e87b384c3b82L95-R98).
* The `retrieveSessionData` method now decodes `userID` and `relyingPartyID` from the session data payload, simplifying the retrieval interface and ensuring all session-related data is encapsulated in the session object (`backend/internal/authn/passkey/session.go`).
* All relevant mocks and tests are updated to match the new method signatures and logic, ensuring test coverage remains accurate and up-to-date (`backend/internal/authn/passkey/sessionStoreInterface_mock_test.go`, `backend/internal/authn/passkey/session_test.go`) [[1]](diffhunk://#diff-9fb05941365966026620e07db52ecf1b0f840737b0edd91e9bc5149fd9ca56cdL92-R99) [[2]](diffhunk://#diff-66abd9c9309c07ec0bf054bcfd6534bb121ef81025d371a46e67e87b384c3b82R148-R158).

These changes collectively make the codebase easier to maintain and extend, while reducing the risk of inconsistencies between session data and its storage/retrieval logic.

---

The database schema and passkey session management logic have been significantly simplified.

- The FLOW_CONTEXT table structure has changed, replacing multiple columns with a single CONTEXT field and updating the primary key to (ID, DEPLOYMENT_ID).
- The FLOW_USER_DATA table has been removed entirely.
- Passkey session storage and retrieval APIs have been refactored:
  - storeSessionData and storeSession no longer accept userID and relyingPartyID as parameters
  - retrieveSessionData now extracts these values from the stored session payload

#### 💥 Impact

- Existing databases using the old schema will be incompatible without migration.
- Any code relying on:
  - Old FLOW_CONTEXT schema
  - FLOW_USER_DATA table
  - Previous session method signatures
  - will break and require updates.
- Custom integrations or extensions depending on explicit userID / relyingPartyID parameters in session APIs will need to be refactored.
#### 🔄 Migration Guide
- Update database schema:
  - Apply new FLOW_CONTEXT structure
  - Drop FLOW_USER_DATA table
- Refactor code:
  - Update calls to storeSessionData / storeSession to pass only session data and expiry
  - Remove dependencies on separate userID and relyingPartyID arguments
  - Ensure session payload includes all required fields for retrieval
- Review any direct DB queries or integrations that depend on removed tables or old schema

---

### Related Issues
- https://github.com/asgardeo/thunder/issues/1877

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * Flow context now stored as a single JSON payload; separate per-flow user-data storage removed.
  * WebAuthn/passkey sessions simplified to store full session payloads with TTLs, omitting separate user/relying-party identifiers.
  * Authorization code consumption now returns full code details on success.

* **Chores**
  * Cleanup script comments updated to reflect the new storage model.

* **Tests**
  * Unit tests and mocks updated to match the refactored storage, session, and auth-code behaviors.


backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -90,38 +86,11 @@ func newAuthorizeService(
...
-			return errors.New("authorization code already used")
-		}
-
-		return errors.New("invalid authorization code")
-	})
-
+	authCode, err := as.authCodeStore.ConsumeAuthorizationCode(ctx, clientID, code)
 	if err != nil {
 		as.logger.Error("Failed to get authorization code details", log.Error(err))
 		return nil, err
```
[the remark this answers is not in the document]
KaveeshaPiumini: this is not relevant to the changes of this PR, @coderabbitai, create an issue to track this separately

backend/internal/flow/flowexec/store.go
```diff
@@ -150,112 +130,54 @@ func withRuntimeDBClientContext(_ context.Context, dbProvider provider.DBProvide
...
-		UserAttributes:      userAttributes,
-		Token:               token,
-		AvailableAttributes: availableAttributes,
-		ExecutionHistory:    executionHistory,
-		ExpiryTime:          expiryTime,
+	return &FlowContextDB{
+		ID:         id,
+		Context:    *contextStr,
+		ExpiryTime: expiryTime,
 	}, nil
```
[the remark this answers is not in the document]
KaveeshaPiumini: These values doesn't have any operational usages, hence those has been dropped, this doesn't fall under the scope of this PR either

backend/internal/oauth/oauth2/authz/auth_code_store.go
```diff
@@ -93,23 +104,42 @@ func (acs *authorizationCodeStore) InsertAuthorizationCode(ctx context.Context,
...
+		if rowsAffected == 0 {
+			return errAuthorizationCodeNotFound
+		}
+
+		results, err := dbClient.QueryContext(ctx, queryGetAuthorizationCode, clientID, authCode, acs.deploymentID)
+		if err != nil {
+			return fmt.Errorf("error retrieving authorization code after consume: %w", err)
+		}
+		if len(results) == 0 {
+			return errAuthorizationCodeNotFound
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/oauth/oauth2/authz/auth_code_store.go
```diff
@@ -83,8 +93,9 @@ func (acs *authorizationCodeStore) InsertAuthorizationCode(ctx context.Context,
 		return err
 	}
 
+	expiryTime := time.Now().UTC().Add(time.Duration(expirySeconds) * time.Second)
 	_, err = dbClient.ExecuteContext(ctx, queryInsertAuthorizationCode, authzCode.CodeID, authzCode.Code,
-		authzCode.ClientID, authzCode.State, jsonDataBytes, authzCode.TimeCreated, authzCode.ExpiryTime,
+		authzCode.ClientID, authzCode.State, jsonDataBytes, authzCode.TimeCreated, expiryTime,
 		acs.deploymentID)
```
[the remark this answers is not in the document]
KaveeshaPiumini: reverted

backend/internal/flow/flowexec/store_test.go
```diff
@@ -71,16 +72,9 @@ func (s *StoreTestSuite) TestStoreFlowContext_WithToken() {
...
-		mock.Anything, mock.Anything, "test-graph-id",
-		mock.Anything, mock.Anything, mock.Anything, "test-deployment").Return(int64(0), nil)
-
-	// Token encryption/decryption is tested in model_test.go, so we just use mock.Anything here
-	mockDBClient.EXPECT().ExecuteContext(mock.Anything, QueryCreateFlowUserData, "test-flow-id", true, mock.Anything,
-		mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, mock.Anything, "test-deployment").
-		Return(int64(0), nil)
+	// Expect a single ExecuteContext call for FLOW_CONTEXT with the context JSON
+	mockDBClient.EXPECT().ExecuteContext(mock.Anything, QueryCreateFlowContext,
+		"test-flow-id", "test-deployment", mock.Anything, mock.Anything).Return(int64(0), nil)
```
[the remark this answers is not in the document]
KaveeshaPiumini: updated

backend/internal/authn/passkey/session.go
```diff
@@ -64,20 +61,25 @@ func (w *passkeyService) storeSessionData(
...
 	// Retrieve session data from database
-	sessionData, userID, relyingPartyID, err := w.sessionStore.retrieveSession(sessionKey)
+	sd, err := w.sessionStore.retrieveSession(sessionKey)
 	if err != nil {
 		return nil, "", "", &ErrorSessionExpired
 	}
 
-	if sessionData == nil {
+	if sd == nil {
 		return nil, "", "", &ErrorSessionExpired
```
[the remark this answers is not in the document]
KaveeshaPiumini: updated

backend/internal/flow/flowexec/model.go
```diff
@@ -106,35 +106,54 @@ type FlowInitContext struct {
...
+	ExecutionHistory    *string `json:"executionHistory,omitempty"`
+	IsAuthenticated     bool    `json:"isAuthenticated"`
+	UserID              *string `json:"userId,omitempty"`
+	OUID                *string `json:"ouId,omitempty"`
+	UserType            *string `json:"userType,omitempty"`
+	UserInputs          *string `json:"userInputs,omitempty"`
+	UserAttributes      *string `json:"userAttributes,omitempty"`
+	Token               *string `json:"token,omitempty"`
+	AvailableAttributes *string `json:"availableAttributes,omitempty"`
+}
```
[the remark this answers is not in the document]
KaveeshaPiumini: I think it’s not necessary to persist. The use cases are:

* **Additional data**: return any extra parameters needed in the flow execution response
* **Assertion**: authentication assertion returned in the final flow/execute response
* **Forwarded data**: pass data from an executor to another node

backend/internal/oauth/oauth2/authz/auth_code_store.go
```diff
@@ -93,23 +102,42 @@ func (acs *authorizationCodeStore) InsertAuthorizationCode(ctx context.Context,
...
+	var consumed *AuthorizationCode
+	err = acs.transactioner.Transact(ctx, func(ctx context.Context) error {
+		rowsAffected, err := dbClient.ExecuteContext(ctx, queryConsumeAuthorizationCode,
+			AuthCodeStateInactive, clientID, authCode, AuthCodeStateActive, acs.deploymentID)
+		if err != nil {
+			return fmt.Errorf("error consuming authorization code: %w", err)
+		}
+		if rowsAffected == 0 {
+			return errAuthorizationCodeNotFound
+		}
```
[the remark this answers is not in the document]
KaveeshaPiumini: updated

backend/internal/flow/flowexec/model.go
```diff
@@ -106,35 +106,54 @@ type FlowInitContext struct {
...
+	}
+	return content.GraphID, nil
 }
 
 // ToEngineContext converts the database model to the flow engine context.
-func (f *FlowContextWithUserDataDB) ToEngineContext(graph core.GraphInterface) (EngineContext, error) {
+func (f *FlowContextDB) ToEngineContext(graph core.GraphInterface) (EngineContext, error) {
+	var content flowContextContent
+	if err := json.Unmarshal([]byte(f.Context), &content); err != nil {
+		return EngineContext{}, err
```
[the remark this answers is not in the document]
KaveeshaPiumini: updated

backend/internal/flow/flowexec/model.go
```diff
@@ -106,35 +107,62 @@ type FlowInitContext struct {
...
+	}
+
+	var content flowContextContent
+	if err := json.Unmarshal([]byte(f.Context), &content); err != nil {
+		return EngineContext{}, err
+	}
+
+	if content.GraphID == "" {
+		return EngineContext{}, fmt.Errorf("stored context has empty graph ID")
+	}
```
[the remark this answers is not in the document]
KaveeshaPiumini: not necessary to handle in model.go

backend/internal/authn/passkey/store.go
```diff
@@ -218,28 +214,25 @@ func (s *sessionStore) serializeSessionData(sessionData *sessionData) ([]byte, e
...
 	var jsonData map[string]interface{}
-	if err := json.Unmarshal([]byte(sessionDataJSON), &jsonData); err != nil {
-		s.logger.Error("Failed to unmarshal session data JSON",
+	if err := json.Unmarshal([]byte(payloadJSON), &jsonData); err != nil {
+		s.logger.Error("Failed to unmarshal session payload JSON",
 			log.Error(err),
-			log.String("jsonData", sessionDataJSON))
-		return nil, "", "", err
+			log.String("payloadJSON", payloadJSON))
+		return nil, err
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/authn/passkey/store.go
```diff
@@ -77,48 +76,46 @@ func (s *sessionStore) storeSession(sessionKey, userID, relyingPartyID string,
...
 	dbClient, err := s.dbProvider.GetRuntimeDBClient()
 	if err != nil {
 		s.logger.Error("Failed to get database client", log.Error(err))
-		return nil, "", "", err
+		return nil, err
 	}
 
 	// Check expiry by comparing with current time
 	now := time.Now()
 	results, err := dbClient.Query(queryGetSession, sessionKey, now, s.deploymentID)
```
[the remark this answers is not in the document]
KaveeshaPiumini: fixed

backend/internal/authn/passkey/session.go
```diff
@@ -64,20 +62,22 @@ func (w *passkeyService) storeSessionData(
 }
 
 // retrieveSessionData retrieves the session data from the database using the session key.
+// userID and relyingPartyID are decoded from the embedded payload.
```
senthalan: Do we need this comment

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -90,38 +86,12 @@ func newAuthorizeService(
...
 	}
-
+	// TODO: Revoke all access tokens already granted for this authorization code.
```
senthalan: this revoke should happen when the auth code state is inactive right ?

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -90,38 +86,12 @@ func newAuthorizeService(
 func (as *authorizeService) GetAuthorizationCodeDetails(
 	ctx context.Context, clientID string, code string,
 ) (*AuthorizationCode, error) {
-	var authCode *AuthorizationCode
-	err := as.transactioner.Transact(ctx, func(ctx context.Context) error {
```
senthalan, line 94: any specific reason to remove the transactioner here?

backend/internal/flow/flowexec/store.go
```diff
@@ -150,112 +130,54 @@ func withRuntimeDBClientContext(_ context.Context, dbProvider provider.DBProvide
...
-func (s *flowStore) parseOptionalString(value interface{}) *string {
+// parseRequiredString parses a required string field from the database row.
+func (s *flowStore) parseRequiredString(value interface{}) *string {
 	if value == nil {
 		return nil
 	}
 	if str, ok := value.(string); ok {
 		return &str
 	}
-	// Handle []byte type (PostgreSQL may return TEXT/JSON as []byte)
```
ThaminduDilshan, line 222: Better to keep this comment for future references

backend/internal/flow/flowexec/store.go
```diff
@@ -150,112 +130,54 @@ func withRuntimeDBClientContext(_ context.Context, dbProvider provider.DBProvide
...
-}
-
 // parseTimeField safely parses a time field from the database row handling multiple formats.
-// This follows the pattern used in other stores for consistency.
 func (s *flowStore) parseTimeField(field interface{}, fieldName string) (time.Time, error) {
 	const customTimeFormat = "2006-01-02 15:04:05.999999999"
 
 	switch v := field.(type) {
 	case string:
-		// Handle SQLite datetime strings
```
ThaminduDilshan, line 254: Better to keep this comment for future references

backend/internal/flow/flowexec/store.go
```diff
@@ -150,112 +130,54 @@ func withRuntimeDBClientContext(_ context.Context, dbProvider provider.DBProvide
...
 		trimmedTime := s.trimTimeString(v)
 		parsedTime, err := time.Parse(customTimeFormat, trimmedTime)
 		if err != nil {
-			// Try alternative ISO 8601 format as fallback
```
ThaminduDilshan, line 258: Better to keep this comment for future references

backend/internal/oauth/oauth2/authz/service.go
ThaminduDilshan: In a way this change aligns with the service-store separation pattern. Store is the one responsible for transection handling, not service right?

KaveeshaPiumini: The transactioner was removed from `service.go` mainly to move transaction ownership to the appropriate layer and to better align with different storage implementations (including Redis).

Previously, the service handled transactions to keep ConsumeAuthorizationCode (UPDATE) and GetAuthorizationCode (SELECT) atomic. This exposed database concerns at the service layer. Now, ConsumeAuthorizationCode encapsulates both operations within the store itself (using a transaction internally for SQL), and returns the full record. This keeps the service layer clean and makes the interface backend-agnostic.

The previous pattern is actually incompatible with Redis:

- ConsumeAuthorizationCode would map to GETDEL (atomic delete + return).
- A subsequent GetAuthorizationCode would always fail since the key is already deleted.
- Service-level transactions cannot solve this, as Redis does not support that kind of transactional visibility.

backend/internal/flow/flowexec/store.go
KaveeshaPiumini: reverted

backend/internal/flow/flowexec/store.go
KaveeshaPiumini: reverted

backend/internal/flow/flowexec/store.go
KaveeshaPiumini: reverted

backend/internal/oauth/oauth2/authz/service.go
KaveeshaPiumini: updated the comment placement

backend/internal/authn/passkey/session.go
KaveeshaPiumini: removed

backend/internal/authn/passkey/session.go
```diff
@@ -68,16 +66,17 @@ func (w *passkeyService) retrieveSessionData(
 	sessionKey string,
 ) (*sessionData, string, string, *serviceerror.ServiceError) {
```
rajithacharith, line 67: Shall we try to get this.. 
Treat userId and relyingPartyId as something in session data.
```suggestion
func (w *passkeyService) retrieveSessionData(
sessionKey string,
) (*sessionData, *serviceerror.ServiceError) {
```
WDYT?

backend/internal/oauth/oauth2/authz/auth_code_store.go
```diff
@@ -53,26 +54,34 @@ const (
 // AuthorizationCodeStoreInterface defines the interface for managing authorization codes.
 type AuthorizationCodeStoreInterface interface {
 	InsertAuthorizationCode(ctx context.Context, authzCode AuthorizationCode) error
-	ConsumeAuthorizationCode(ctx context.Context, clientID, authCode string) (bool, error)
+	ConsumeAuthorizationCode(ctx context.Context, clientID, authCode string) (*AuthorizationCode, error)
```
senthalan: Do we need to pass the clientID as an argument here? If the refactoring effort is to align the store interface function definition for Redis plugability, then we should keep this as 
```
GetAuthorizationCode(ctx context.Context, authCode string) (*AuthorizationCode, error)
ConsumeAuthorizationCode(ctx context.Context, authCode string) (bool, error)
```

And in the service layer, we can,
1. Get the auth code using GetAuthorizationCode 
2. Perform the client ID match check
3. Consume the auth code using ConsumeAuthorizationCode

We can perform all these in a single transaction managed in the service layer

backend/internal/oauth/oauth2/authz/service.go
rajithacharith: Hi @KaveeshaPiumini 
As I understood
So, to support GETDEL, the store layer should support single method. So having both deleting and retrieving both on the store "ConsumeAuthorizationCode" is okay. 
Shall we check whether is there a way to keep the transaction code in the service level. 
Having the `as.transactioner.Transact(ctx, func(ctx context.Context) error {` here or store level does not make a difference  IMO, therefore that should be possible.

backend/internal/authn/passkey/session.go
KaveeshaPiumini: UserID and relyingPartyID are currently used directly in the service layer, not within the session itself. We would still need to extract them from the session data regardless, so it might be cleaner to handle that in the service layer rather than making session.go responsible for resolving them.

That said, my intention in keeping it here was to avoid duplication. There are a few places in the service layer using retrieveSessionData, and extracting userID and relyingPartyID in each of those places would lead to repeated logic. Keeping it here helps centralize that extraction while still keeping the flow simple.

Open to adjusting this based on what we feel is the right separation of responsibilities

backend/internal/oauth/oauth2/authz/auth_code_store.go
KaveeshaPiumini: I think, if I am not wrong, the specification requires that the `client_id` in the token request must match the one the code was issued to and enforces this at the DB query level (WHERE CLIENT_ID = ? AND AUTHORIZATION_CODE = ?) 

CC: @thiva-k

backend/internal/oauth/oauth2/authz/service.go
KaveeshaPiumini: You’re right that for SQL, having the Transact call in the service or store would behave the same.

The main reason for moving it to the store is to keep the service layer backend-agnostic, especially with Redis in mind. The transactioner is tied to the SQL runtime DB, and for Redis we don’t have an equivalent — it would end up being a no-op, which can be misleading.

By keeping transaction handling inside the store:

- SQL uses real transactions internally
- Redis can rely on native atomic operations (e.g., GETDEL)

The service simply calls ConsumeAuthorizationCode and doesn’t need to be aware of how atomicity is achieved. This keeps responsibilities clearer and avoids leaking backend-specific behavior into the service layer.

backend/internal/oauth/oauth2/authz/auth_code_store.go
senthalan: Yes, this check have to be done by the server according to the specification. But the place of this check can be in the DB query level or in the service level. What I am suggesting is to move it to the service level. So that the store level functions will be simple. So we can plug different store implementation easily

thiva-k: Yes. But if the check is moved to the service layer, this should be redundant.

KaveeshaPiumini: Moving the client ID check to the service level may introduce an additional check to get the client ID from the auth code. 
If not we can validate the client ID using GetAuthorizationCode and call ConsumeAuthorizationCode to update the status, this approach make it difficult to use the DB transactioner which was there to preserve atomicity.

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -81,47 +78,32 @@ func newAuthorizeService(
...
+	if err != nil {
+		as.logger.Error("Failed to get authorization code details", log.Error(err))
+		return nil, err
+	}
 
-		return errors.New("invalid authorization code")
-	})
+	if record.ClientID != clientID {
+		as.logger.Error("Client ID mismatch for authorization code")
+		return nil, errors.New("client ID mismatch for authorization code")
```
senthalan: we don't need to log it as error or return error object. It's client error and should return to the caller as 400.

backend/internal/oauth/oauth2/authz/auth_code_store.go
```diff
@@ -53,26 +54,34 @@ const (
 // AuthorizationCodeStoreInterface defines the interface for managing authorization codes.
 type AuthorizationCodeStoreInterface interface {
 	InsertAuthorizationCode(ctx context.Context, authzCode AuthorizationCode) error
-	ConsumeAuthorizationCode(ctx context.Context, clientID, authCode string) (bool, error)
-	GetAuthorizationCode(ctx context.Context, clientID, authCode string) (*AuthorizationCode, error)
+	ConsumeAuthorizationCode(ctx context.Context, authCode string) (*AuthorizationCode, error)
```
senthalan: why we should return the AuthorizationCode here? the service implementation already have the AuthorizationCode retrieved from the GetAuthorizationCode.

senthalan: As previously, 
- If the code is not used, return mark it as used and return true
- If the code is already used, return false
- If there are any errors, then return error

KaveeshaPiumini: Updated the implementation to reflect above behaviour

backend/internal/oauth/oauth2/authz/service.go
KaveeshaPiumini: error log is removed

Issue #1926, opened 2026-03-23
Implement a new SMS executor for generic SMS delivery


ThaminduDilshan:
### What problem are we solving?

When flow designers want to send SMS messages for use cases beyond OTP (such as notifications, confirmations, or multi-step flows), they are forced to use the tightly coupled SMSOTPAuthExecutor, which only supports OTP scenarios. This prevents flexible SMS delivery and limits extensibility for new flows.

### Who are we solving this for?

Primary: app developers

- Goal: Design flexible, modular flows that can send SMS for any scenario (not just OTP)
- Pain: Cannot reuse SMS delivery logic for non-OTP use cases; must duplicate or workaround existing executor limitations

### Why should we solve this now?

- Value to Persona: Enables new onboarding, notification, and confirmation flows using SMS, improving user experience and security
- Value to Org: Increases product flexibility, and unlocks new customer requirements

### Proposed Solution

- Implement a new sms executor responsible solely for sending SMS messages similar to the existing email executor.
- The executor should accept recipient and message content as inputs, and support a send mode.
- Integrate with the existing notification/message sender infrastructure.
- Allow the executor to be plugged into any flow node, independent of OTP logic.

### Alternatives

- Continue extending the current SMSOTPAuthExecutor for each new use case (increases complexity and coupling).
- Use HTTP request executors to call external SMS APIs directly (loses central management and template support).

Issue #1928, opened 2026-03-23
SMS Template Support


ThaminduDilshan:
### What problem are we solving?

When administrators or developers want to customize SMS content for different scenarios (OTP, confirmation, notifications), they cannot use templates as with email. SMS content is hardcoded, making it impossible to manage or localize messages centrally.

### Who are we solving this for?

Primary: Administrators, developers

- Goal: Define and manage SMS content for various flows and scenarios
- Pain: Cannot customize SMS messages for scenario-specific requirements

### Why should we solve this now?

- Value to Persona: Enables scenario-specific SMS content, improving communication and compliance
- Value to Org: Increases product flexibility, supports enterprise requirements, and reduces support burden

### Proposed Solution

- Extend the template service to support SMS templates, similar to email templates.
- Allow administrators to define SMS templates for scenarios such as OTP and confirmation.
- Update the new SMSExecutor to use the template service for rendering SMS content.
- Ensure templates can be selected and rendered based on flow scenario and runtime context.

### Alternatives

-

PR #2128, opened 2026-04-02
Add generic SMS executor for flow-based SMS sending

Dilusha-Madushan:
### Purpose
Adds a new generic SMSExecutor utility executor that enables flow designers to send arbitrary SMS messages at any point within a flow. This is distinct from the existing SMSOTPAuthExecutor, which is tightly coupled to OTP-based authentication. The new executor enables use cases such as sending a welcome SMS after registration, delivery notifications, or any other flow-triggered SMS communication.

### Approach

**SMSSenderService** — a new service layer (SMSSenderServiceInterface / smsSenderService) introduced in the notification package. It abstracts SMS delivery by looking up the configured sender by ID, resolving the appropriate message client (Twilio, Vonage, custom webhook, etc.), and dispatching the message. This service is wired through notification.Initialize and injected into the executor at startup.

**SMSExecutor** — a new utility-type flow executor registered as SMSExecutor. It operates in send mode and reads two required properties from the node configuration:

**senderId** — the notification sender to use
**messageBody** — the SMS content to deliver
The recipient (mobileNumber) is resolved from UserInputs first, falling back to RuntimeData, allowing flexibility in where the phone number is collected within the flow. If no SMS sender service is configured, the executor completes gracefully as a no-op (**smsSent=false**) rather than failing the flow.

Error handling follows the existing executor pattern:

Client errors (invalid sender, bad recipient) → ExecFailure with a failure reason, allowing the flow to handle it gracefully
Server errors → propagated as a Go error to halt the flow

### Related Issues
- https://github.com/asgardeo/thunder/issues/1926

## Summary by CodeRabbit

* **New Features**
  * SMS messaging added to the notification system; workflows can send SMS using a configurable sender ID and default message. Execution responses include an smsSent flag and clear failure reasons when recipient or configuration is missing.
* **Tests**
  * Added unit tests and autogenerated mocks covering the SMS sender service and the workflow SMS executor across success and failure scenarios.


backend/internal/flow/executor/constants.go
```diff
@@ -88,8 +89,13 @@ const (
 	propertyKeyRequiredScopes   = "requiredScopes"
 	propertyKeyEmailTemplate    = "emailTemplate"
 	propertyKeyAllowedUserTypes = "allowedUserTypes"
+	propertyKeySMSSenderID = "senderId"
 )
 
+// smsDefaultMessage is a temporary hardcoded SMS body used until template-based message support is implemented.
+// TODO: Replace with a proper message template engine in a future PR.
+const smsDefaultMessage = "You have a pending notification from the system."
```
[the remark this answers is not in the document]
Dilusha-Madushan: This is intentional for this PR's scope. The SMSExecutor introduced here is a foundational utility executor — its primary purpose is to wire up the SMS sending infrastructure (sender lookup, client dispatch, error classification) into the flow engine. Template-based messageBody support — reading from node properties, variable interpolation, and localization — is a non-trivial feature tracked as follow-up work and will be implemented in a dedicated PR.

Shipping with a hardcoded default message is a deliberate trade-off: it keeps this PR focused and reviewable, avoids a half-finished template system, and the TODO comment in constants.go makes the intent explicit. Flow authors won't be able to configure the message body yet, which is acceptable since this executor is backend-only and not exposed in the UI palette until the template feature lands.

backend/internal/notification/sms_sender_service.go
```diff
@@ -0,0 +1,69 @@
...
+		return &ErrorRequestedSenderIsNotOfExpectedType
+	}
+
+	client, svcErr := s.clientProvider.GetMessageClient(*sender)
+	if svcErr != nil {
+		return svcErr
+	}
+
+	if err := client.SendSMS(common.SMSData{To: recipient, Body: body}); err != nil {
+		return &ErrorInternalServerError
```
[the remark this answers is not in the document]
Dilusha-Madushan: This is a pre-existing limitation of MessageClientInterface — it accepts no context.Context and returns a plain error, so context propagation and typed error classification at the provider layer are not possible without changing the interface. The same gap exists in OTPService.SendSMS (otp_service.go:271). Fixing it requires updating the interface, all three client implementations (Twilio, Vonage, custom), the OTP service, and their tests. That refactor is out of scope for this PR and should be tracked as a separate issue.

backend/internal/flow/executor/constants.go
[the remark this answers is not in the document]
Dilusha-Madushan: This git issue tracking that issue - https://github.com/asgardeo/thunder/issues/1928

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, fmt.Errorf("senderId is not configured in node properties: %w", err)
+	}
+
+	// TODO: Replace smsDefaultMessage with a proper template-based message body in a future PR.
+	svcErr := e.smsSenderService.SendSMS(ctx.Context, senderID, recipient, smsDefaultMessage)
+	if svcErr != nil {
+		if svcErr.Type == serviceerror.ClientErrorType {
+			execResp.Status = common.ExecFailure
+			execResp.FailureReason = svcErr.Error
+			return execResp, nil
```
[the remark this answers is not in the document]
Dilusha-Madushan: Addressed

backend/internal/notification/sms_sender_service.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+
+	"github.com/asgardeo/thunder/internal/notification/common"
+	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+)
+
+// SMSSenderServiceInterface defines the interface for sending SMS messages.
+type SMSSenderServiceInterface interface {
```
ThaminduDilshan: Rather than making this specific to SMS, can we write this as a generic notification sender service?

`SMSSenderServiceInterface` -> `NotificationSendServiceInterface`
`SendSMS()` -> `Send()` and let's pass channel as a method param (channel=sms)

Dilusha-Madushan: Done! Renamed SMSSenderServiceInterface → NotificationSenderServiceInterface and SendSMS() → Send() with channel common.ChannelType as a method param. The implementation dispatches internally via a channel switch (case ChannelTypeSMS → sendSMS()), so adding a new channel later only requires a new case and a private method — no interface changes needed.

Also renamed the file from sms_sender_service.go → notification_sender_service.go to match the generic naming.

backend/internal/flow/executor/constants.go
```diff
@@ -88,8 +89,13 @@ const (
 	propertyKeyRequiredScopes   = "requiredScopes"
 	propertyKeyEmailTemplate    = "emailTemplate"
 	propertyKeyAllowedUserTypes = "allowedUserTypes"
+	propertyKeySMSSenderID      = "senderId"
```
ThaminduDilshan:
```suggestion
	propertyKeyNotificationSenderID      = "senderId"
```

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/asgardeo/thunder/internal/flow/common"
+	"github.com/asgardeo/thunder/internal/flow/core"
+	"github.com/asgardeo/thunder/internal/notification"
+	notificationcommon "github.com/asgardeo/thunder/internal/notification/common"
+	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+	"github.com/asgardeo/thunder/internal/system/log"
+)
+
+// smsExecutor sends an SMS message using the configured sender and message body from node properties.
```
ThaminduDilshan: We don't take message body as a node parameter right? Shall we update the comment?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+// When notificationSenderService is nil, it completes as a no-op with smsSent=false.
+type smsExecutor struct {
+	core.ExecutorInterface
+	logger                    *log.Logger
+	notificationSenderService notification.NotificationSenderServiceInterface
+}
+
+// newSMSExecutor creates a new instance of smsExecutor.
+// notificationSenderService may be nil if no SMS provider is configured; executor completes as a no-op in that case.
```
ThaminduDilshan: This second sentence is incorrect right? Even if there's no sms provider, notificationSenderService will not be nil. Shall we remove this?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+		ExecutorInterface:         base,
+		logger:                    logger,
+		notificationSenderService: notificationSenderService,
+	}
+}
+
+// Execute sends an SMS using data from the runtime context.
+func (e *smsExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse, error) {
+	switch ctx.ExecutorMode {
+	case ExecutorModeSend:
```
ThaminduDilshan: Is there a case we need to support multiple modes for this executor? If not shall we remove mode based switch case?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+		return e.executeSend(ctx)
+	default:
+		return nil, fmt.Errorf("invalid executor mode for SMSExecutor: %s", ctx.ExecutorMode)
+	}
+}
+
+// executeSend resolves the recipient, sender ID, and message body from node properties, then sends the SMS.
```
ThaminduDilshan: Same as previous comment. We don't take message body as a node param right?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+// If the notification sender service is not configured, it completes without sending (no-op).
```
ThaminduDilshan: Continuing flow execution as a no-op may not be appropriate for all the cases. Consider this being used for sms otp registration/ authentication case. If the sms cannot be send, there's no way to verify the user. So flow should fail. Shall we modify this to return an error?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+	"github.com/asgardeo/thunder/internal/system/log"
+)
+
+// smsExecutor sends an SMS message using the configured sender and message body from node properties.
+// When notificationSenderService is nil, it completes as a no-op with smsSent=false.
+type smsExecutor struct {
+	core.ExecutorInterface
+	logger                    *log.Logger
+	notificationSenderService notification.NotificationSenderServiceInterface
```
ThaminduDilshan:
```suggestion
	notifSenderSvc notification.NotificationSenderServiceInterface
```

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+}
+
+// newSMSExecutor creates a new instance of smsExecutor.
+// notificationSenderService may be nil if no SMS provider is configured; executor completes as a no-op in that case.
+func newSMSExecutor(flowFactory core.FlowFactoryInterface,
+	notificationSenderService notification.NotificationSenderServiceInterface) *smsExecutor {
```
ThaminduDilshan:
```suggestion
	notifSenderSvc notification.NotificationSenderServiceInterface) *smsExecutor {
```

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+package executor
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/asgardeo/thunder/internal/flow/common"
+	"github.com/asgardeo/thunder/internal/flow/core"
+	"github.com/asgardeo/thunder/internal/notification"
+	notificationcommon "github.com/asgardeo/thunder/internal/notification/common"
```
ThaminduDilshan:
```suggestion
	notifcm "github.com/asgardeo/thunder/internal/notification/common"
```

Let's try to use short, but meaningful words for package names, variables, etc if that's going to be too lengthy. Otherwise code will be hard to read

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,146 @@
...
+	logger.Debug("SMS sent successfully", log.String("recipient", log.MaskString(recipient)))
+
+	execResp.AdditionalData[common.DataSMSSent] = dataValueTrue
+	execResp.Status = common.ExecComplete
+	return execResp, nil
+}
+
+// resolveRecipientMobile retrieves the recipient mobile number from user inputs or runtime data.
+func resolveRecipientMobile(ctx *core.NodeContext) string {
+	if mobile, ok := ctx.UserInputs[userAttributeMobileNumber]; ok && mobile != "" {
```
ThaminduDilshan: Shall we try to take this input from node inputs? Otherwise this executor will be hard code to mobile number. If the schema doesn't have a mobile number defined, but uses a different phone attribute, executor will fail

ThaminduDilshan: @coderabbitai check if the same issue exists for the email executor with email attribute. If so create an issue

backend/internal/notification/notification_sender_service.go
```diff
@@ -0,0 +1,83 @@
...
+		return &ErrorRequestedSenderIsNotOfExpectedType
+	}
+
+	client, svcErr := s.clientProvider.GetMessageClient(*sender)
+	if svcErr != nil {
+		return svcErr
+	}
+
+	if err := client.SendSMS(common.SMSData{To: recipient, Body: body}); err != nil {
+		return &ErrorInternalServerError
```
ThaminduDilshan, line 75: We should log the error here. Otherwise there's no way to figure out the exact root cause as e're only returning a generic server error

backend/internal/flow/executor/sms_executor.go
ThaminduDilshan: We should revisit this behaviour for email executor too. Tracked in https://github.com/asgardeo/thunder/issues/2195

backend/internal/flow/executor/sms_executor.go
ThaminduDilshan: Same should be addressed for email executor too. Tracked with https://github.com/asgardeo/thunder/issues/2196

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: This executor is intended for generic informational SMS notifications within a flow (e.g., welcome messages, invite messages) — not for OTP-based authentication. OTP send/verify is handled separately by smsOTPAuthExecutor which uses OTPServiceInterface.

The no-op behaviour here intentionally mirrors emailExecutor, which follows the same pattern — if the client/service is not configured, it completes with sent=false rather than failing the flow.

For authentication-critical SMS (OTP), smsOTPAuthExecutor is the right executor and it handles failures appropriately.

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/constants.go
Dilusha-Madushan: Addressed

backend/internal/notification/notification_sender_service.go
Dilusha-Madushan: Addressed

backend/cmd/server/servicemanager.go
```diff
@@ -202,8 +202,8 @@ func registerServices(mux *http.ServeMux) jwt.JWTServiceInterface {
 		logger.Fatal("Failed to initialize template service", log.Error(err))
 	}
 	execRegistry := executor.Initialize(flowFactory, ouService,
-		idpService, otpService, jwtService, authSvcRegistry, authZService, userSchemaService, observabilitySvc,
-		groupService, roleService, userProvider, attributeCacheService, emailClient, templateService)
+		idpService, otpService, smsSenderService, jwtService, authSvcRegistry, authZService, userSchemaService,
+		observabilitySvc, groupService, roleService, userProvider, attributeCacheService, emailClient, templateService)
```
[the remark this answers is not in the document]
ThaminduDilshan: This is a pending documentation area. Hence cannot document only this particular executor

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,156 @@
...
+		[]common.Input{
+			{Identifier: userAttributeMobileNumber, Type: common.InputTypePhone, Required: true},
+		},
+	)
+	return &smsExecutor{
+		ExecutorInterface: base,
+		logger:            logger,
+		smsSenderService:  smsSenderService,
+	}
+}
```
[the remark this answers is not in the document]
ThaminduDilshan: This is a pending documentation area. Hence cannot document only this particular executor

backend/internal/flow/executor/sms_executor.go
```diff
@@ -0,0 +1,134 @@
...
+	logger := e.logger.With(log.String(log.LoggerKeyFlowID, ctx.FlowID))
+	logger.Debug("Executing SMS executor")
+
+	execResp := &common.ExecutorResponse{
+		AdditionalData: make(map[string]string),
+		RuntimeData:    make(map[string]string),
+	}
+
+	if e.notifSenderSvc == nil {
+		execResp.AdditionalData[common.DataSMSSent] = dataValueFalse
```
ThaminduDilshan: This is a server error right? Let's return a go error

backend/internal/flow/executor/sms_executor.go
ThaminduDilshan: Let's address this too

backend/internal/notification/notification_sender_service.go
```diff
@@ -0,0 +1,79 @@
...
+// NotificationSenderServiceInterface defines the interface for sending notification messages.
+type NotificationSenderServiceInterface interface {
+	Send(ctx context.Context, channel common.ChannelType, senderID string,
+		data common.NotificationData) *serviceerror.ServiceError
+}
+
+// notificationSenderService implements NotificationSenderServiceInterface.
+type notificationSenderService struct {
+	senderMgtService NotificationSenderMgtSvcInterface
+	clientProvider   notificationClientProviderInterface
```
ThaminduDilshan, line 38: Shall we initialize the logger inside the struct and via the construct method rather than initializing in each method?

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Added a phoneAttribute node property that specifies which user attribute holds the phone number. Defaults to mobileNumber when not set, so existing flows that don't configure this property continue to work without changes.

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed and now return Go error - **_notification sender service is not configured_**

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Addressed this by sending Go error if the notification sender service is null.

backend/internal/notification/notification_sender_service.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/constants.go
```diff
@@ -83,13 +84,19 @@ const (
...
-	propertyKeyRequiredScopes   = "requiredScopes"
-	propertyKeyEmailTemplate    = "emailTemplate"
-	propertyKeyAllowedUserTypes = "allowedUserTypes"
+	propertyKeyAssignGroup          = "assignGroup"
+	propertyKeyAssignRole           = "assignRole"
+	propertyKeyRequiredScopes       = "requiredScopes"
+	propertyKeyEmailTemplate        = "emailTemplate"
+	propertyKeyAllowedUserTypes     = "allowedUserTypes"
+	propertyKeyNotificationSenderID = "senderId"
+	propertyKeyPhoneAttribute       = "phoneAttribute"
```
ThaminduDilshan: This is not the approach. Let's have a chat

Dilusha-Madushan: Addressed. Now use input from node instead of node properties.

Issue #1931, opened 2026-03-23
Support returning user roles in tokens and user info responses


ThaminduDilshan:
### Current Limitation

Currently, Thunder allows applications to configure which user attributes and special attributes (such as groups) are included in issued tokens (assertion, access token, ID token, and user info response). However, there is no support for including user roles in these tokens in a configurable way.

### Suggested Improvement

- Allow applications to configure whether user roles should be included in the assertion, access token, ID token, and user info response, similar to how user groups are handled.
- Extend the application token configuration UI and backend models to support "roles" as a selectable attribute.
- Ensure roles are only included if explicitly configured for the relevant token type.
- Update the token and user info claim-building logic to fetch and embed user roles when requested.

PR #2137, opened 2026-04-02
Provide support to return the roles with tokens

sahandilshan:
### Purpose
The role will be get as user attribute and returns with the
- Access token
- ID Token
- User info endpoint

### Related Issue(s)
- https://github.com/asgardeo/thunder/issues/1931

## Summary by CodeRabbit

* **New Features**
  * Added "roles" as a standard OIDC scope and user attribute; roles are now computed (including from group-derived assignments) and included in tokens when requested.
  * Resolved transitive group membership when building user attributes, improving derived-claim accuracy.
  * Console UI updated to expose "roles" and "groups" in application token configuration.

* **Tests**
  * Added and updated tests covering transitive group retrieval, role resolution, and related token/claim behaviors.

### DIFF
--- a/backend/internal/flow/executor/auth_assert_executor.go
+++ b/backend/internal/flow/executor/auth_assert_executor.go
@@ -388,17 +388,24 @@ func (a *authAssertExecutor) resolveUserAttributes(ctx *core.NodeContext, reques
 // appendComputedAttributes appends computed/derived attributes (groups, roles, userType, OU details) to the claims.
 func (a *authAssertExecutor) appendComputedAttributes(
 	ctx *core.NodeContext, requestedAttributes []string, attributes map[string]interface{}) error {
-	// Handle groups if configured in user attributes
-	if slices.Contains(requestedAttributes, oauth2const.UserAttributeGroups) && ctx.AuthenticatedUser.UserID != "" {
-		if err := a.appendGroupsToClaims(ctx.AuthenticatedUser.UserID, attributes); err != nil {
+	groupsRequested := slices.Contains(requestedAttributes, oauth2const.UserAttributeGroups)
+	rolesRequested := slices.Contains(requestedAttributes, oauth2const.UserAttributeRoles)
+
+	// Fetch all user groups once if either groups or roles are requested.
+	if (groupsRequested || rolesRequested) && ctx.AuthenticatedUser.UserID != "" {
+		allGroups, err := a.fetchAllUserGroups(ctx.AuthenticatedUser.UserID)
+		if err != nil {
 			return err
 		}
-	}
 
-	// Handle roles if configured in user attributes
-	if slices.Contains(requestedAttributes, oauth2const.UserAttributeRoles) && ctx.AuthenticatedUser.UserID != "" {
-		if err := a.appendRolesToClaims(ctx, attributes); err != nil {
-			return err
+		if groupsRequested {
+			a.appendGroupsToClaims(allGroups, attributes)
+		}
+
+		if rolesRequested {
+			if err := a.appendRolesToClaims(ctx, allGroups, attributes); err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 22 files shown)


backend/internal/role/file_based_store.go
```diff
@@ -375,6 +375,42 @@ func (f *fileBasedStore) GetAuthorizedPermissions(
 	return result, nil
 }
 
+// GetUserRoles retrieves the names of roles assigned to a user directly and/or through group membership.
+func (f *fileBasedStore) GetUserRoles(
```
sahandilshan, line 379: Group ID should be handled inside this function, but there's a sam approach done when getting permissions with`GetAuthorizedPermissions`. This is becuase this data is in the config db and the group data is in the user db. So to keep the separation of concern, getting the roles assoiciate to the user along with the groups like this. But later we may have to consider on how we can access these values in the two DB to avoid inconsistencies like this

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -514,6 +537,55 @@ func (a *authAssertExecutor) appendGroupsToClaims(
...
+	logger := a.logger.With(log.String("userID", ctx.AuthenticatedUser.UserID))
+
+	// Fetch user's group IDs for role resolution (roles can be inherited through groups)
+	groupIDs, err := a.getGroupIDsForRoleResolution(ctx.AuthenticatedUser.UserID)
+	if err != nil {
+		logger.Error("Failed to fetch group IDs for role resolution", log.Any("error", err))
+		return errors.New("something went wrong while fetching groups for role resolution")
+	}
+
+	roles, svcErr := a.roleService.GetUserRoles(ctx.Context, ctx.AuthenticatedUser.UserID, groupIDs)
```
senthalan, line 563: will this return the parant groups of user assigned group?

user is member of `group-a` and `group-a` is member of `group-b`. It should return both `group-a` and `group-b`. Then only the roles assigned to the `group-b` can be added to the token here

backend/internal/role/composite_store.go
```diff
@@ -338,6 +338,23 @@ func (c *compositeRoleStore) GetAuthorizedPermissions(
...
+	if err != nil {
+		return nil, err
+	}
+
+	fileRoleNames, err := c.fileStore.GetUserRoles(ctx, userID, groupIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	return mergePermissions(dbRoleNames, fileRoleNames), nil
```
thiva-k, line 355: Shall we have a separate function or make this function name generic instead?

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -514,6 +537,55 @@ func (a *authAssertExecutor) appendGroupsToClaims(
...
+	if svcErr != nil {
+		logger.Error("Failed to fetch user roles",
+			log.String("userID", ctx.AuthenticatedUser.UserID),
+			log.Any("error", svcErr))
+		return errors.New("something went wrong while fetching user roles: " + svcErr.ErrorDescription)
+	}
+
+	if len(roles) > 0 {
+		jwtClaims[oauth2const.UserAttributeRoles] = roles
+	}
```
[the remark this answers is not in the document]
sahandilshan: The apis.mdx is not a hand written one. It get auto generated

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -372,10 +377,28 @@ func (a *authAssertExecutor) resolveUserAttributes(ctx *core.NodeContext, reques
...
 	if slices.Contains(requestedAttributes, oauth2const.UserAttributeGroups) && ctx.AuthenticatedUser.UserID != "" {
 		if err := a.appendGroupsToClaims(ctx.AuthenticatedUser.UserID, attributes); err != nil {
-			return nil, err
+			return err
+		}
+	}
+
+	// Handle roles if configured in user attributes
+	if slices.Contains(requestedAttributes, oauth2const.UserAttributeRoles) && ctx.AuthenticatedUser.UserID != "" {
+		if err := a.appendRolesToClaims(ctx, attributes); err != nil {
```
thiva-k: Can't we pass the group IDs from the `appendGroupsToClaims` here to prevent redundant DB call if group claim is added?

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -372,10 +377,28 @@ func (a *authAssertExecutor) resolveUserAttributes(ctx *core.NodeContext, reques
...
+}
+
+// appendComputedAttributes appends computed/derived attributes (groups, roles, userType, OU details) to the claims.
+func (a *authAssertExecutor) appendComputedAttributes(
+	ctx *core.NodeContext, requestedAttributes []string, attributes map[string]interface{}) error {
 	// Handle groups if configured in user attributes
 	if slices.Contains(requestedAttributes, oauth2const.UserAttributeGroups) && ctx.AuthenticatedUser.UserID != "" {
 		if err := a.appendGroupsToClaims(ctx.AuthenticatedUser.UserID, attributes); err != nil {
-			return nil, err
+			return err
```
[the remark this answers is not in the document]
sahandilshan: This is fixed

tests/integration/oauth/authz/authz_scope_test.go
```diff
@@ -617,6 +616,21 @@ func (ts *OAuthAuthzScopeTestSuite) TestOAuthAuthzFlow_WithRequiredAttributes()
 		_ = testutils.DeleteUser(userID)
 	}()
 
+	// Create a role and assign it to the user so that the computed "roles" claim is populated
+	role := testutils.Role{
```
senthalan: shall we assign another role to the user through a group assignment, So that we can test both flows in the same tests. We have to assert the present of both roles

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -372,10 +377,35 @@ func (a *authAssertExecutor) resolveUserAttributes(ctx *core.NodeContext, reques
...
+	// Fetch all user groups once if either groups or roles are requested.
+	if (groupsRequested || rolesRequested) && ctx.AuthenticatedUser.UserID != "" {
+		allGroups, err := a.fetchAllUserGroups(ctx.AuthenticatedUser.UserID)
+		if err != nil {
+			return err
+		}
+
+		if groupsRequested {
+			a.appendGroupsToClaims(allGroups, attributes)
+		}
```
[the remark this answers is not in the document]
sahandilshan: No, the change that did here should reflect for groups as well where it shows transitive memberships

backend/internal/user/store.go
```diff
@@ -685,6 +686,32 @@ func (us *userStore) GetUserGroups(ctx context.Context, userID string, limit, of
...
+		group, err := buildGroupFromResultRow(row)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build group from result row: %w", err)
+		}
+
+		groups = append(groups, group)
+	}
+
+	return groups, nil
+}
```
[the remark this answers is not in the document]
sahandilshan: file-based (declarative) users are read only configuration entries that don't participate in group membership. Group assignments are a database only concept in Thunder's architecture. Merging two calls that return the same data (DB results + always-empty file results) would add unnecessary complexity with zero behavioral change.

backend/internal/flow/executor/auth_assert_executor.go
sahandilshan: This is fixed now. Added a new method that return all groups of the user including the transitive ones

tests/integration/oauth/authz/authz_scope_test.go
sahandilshan: Update the test case

senthalan (approved):

tests/integration/oauth/authz/authz_scope_test.go
```diff
@@ -698,8 +742,18 @@ func (ts *OAuthAuthzScopeTestSuite) TestOAuthAuthzFlow_WithRequiredAttributes()
...
+	ts.Require().NotNil(atClaims["roles"], "roles should be present in access token (required access_token attribute)")
+	rolesRaw, ok := atClaims["roles"].([]interface{})
+	ts.Require().True(ok, "roles claim should be an array")
+	roleNames := make([]string, len(rolesRaw))
+	for i, r := range rolesRaw {
+		roleNames[i], ok = r.(string)
+		ts.Require().True(ok, "each role should be a string")
+	}
+	ts.Assert().Contains(roleNames, "developer", "roles should include directly assigned role")
+	ts.Assert().Contains(roleNames, "reviewer", "roles should include group-inherited role")
```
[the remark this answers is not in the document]
sahandilshan: This is test file

Issue #2085, opened 2026-03-31
Implement a validator for flow definitions


ThaminduDilshan:
## Current limitation
Currently thunder doesn’t validate the flow definition when creating and updating flows. A flow definition validator should be implemented to validate the flow definition. This includes several layers of validations. Validate for node connectivity Validate for inputs, structures, etc, Validate for minimum requirements on nodes and executors, Validate inputs against schema (need to think on how to determine the applicable schema/ schemas). Prevent storing non valid flows. Additionally it doesn’t validate for references when deleting the created flows.

## Suggested improvement
Implement a validator for flow definitions. The validator should check for node connectivity, input and structure validity, minimum requirements on nodes and executors, and validate inputs against the schema. It should prevent storing invalid flows and validate references when deleting flows.

NutharaNR:
Discussion: https://github.com/thunder-id/thunderid/discussions/3537

NutharaNR:
Validation on reference at flow deletion is tracked under a separate issue: https://github.com/thunder-id/thunderid/issues/3554

PR #2203, opened 2026-04-06
ACR values support

### LINKED ISSUE #2102: ACR Values Support in Thunder

There is no standards-defined mechanism that lets relying parties express the required level of assurance at which an end-user must be authenticated.

Relying parties who use standards "acr_values" to enforce specific authentication methods (e.g., OTP-only, password-only, or a preference ordering of methods).

- Complete adherence with OIDC specification.
- Reduces the frequent changes in the flow graphs

### Proposed Solution

Using Authentication Context Class Reference and Authentication Method Reference concept. An Authentication Context Class specifies a set of business rules that authentications are being requested to satisfy. These rules can often be satisfied by using a number of different specific Authentication Methods, either singly or in combination.

For example:
ACR -> urn:thunder:silver maps to AMR -> OTP
ACR -> urn:thunder:gold maps to AMR -> BIO &&  OTP

For more details on the approach refer - https://github.com/anushasunkada/thunder/blob/main/feature-docs/acr-values.md

References: 
https://github.com/mosip/esignet/blob/master/esignet-service/src/main/resources/amr_acr_mapping.json
https://self-issued.info/?p=2726

### Alternatives

_No response_

sacrana0:
### Purpose
ACR values support for Thunder.

### Approach
Thunder currently ignores the acr_values OIDC parameter entirely. This feature adds full support for it, letting relying parties specify required authentication assurance levels (e.g., OTP-only vs. password-only).
The work covers eight areas: ACR-AMR mapping in server config, per-application ACR allowlists at registration, parsing and validating acr_values on the authorize endpoint, a new acr_options flow node variant for declarative ACR selection, filtering/reordering/auto-skipping that node at execution time, AMR validation in the auth executor to prevent method substitution attacks, including the acr claim in issued ID tokens, and publishing acr_values_supported in the OIDC discovery document.

### Related Issues
- [ISSUE 2102](https://github.com/asgardeo/thunder/issues/2102)

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added Authentication Class Reference (ACR) values support for OAuth2 applications.
  * Introduced login options prompt allowing users to select authentication methods during authorization.
  * ACR claim now included in OpenID Connect ID tokens.
  * OIDC discovery endpoint now exposes supported ACR values.

### DIFF
--- a/api/application.yaml
+++ b/api/application.yaml
@@ -158,6 +158,7 @@ paths:
                     scopeClaims:
                       profile: ["name", "family_name", "given_name", "picture"]
                       employee: ["emp_id", "department"]
+                    acrValues: ["urn:thunder:silver", "urn:thunder:gold"]
       responses:
         "201":
           description: Application created
@@ -227,6 +228,7 @@ paths:
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
                         employee: ["emp_id", "department"]
+                      acrValues: ["urn:thunder:silver", "urn:thunder:gold"]
         "400":
           description: Bad request
           content:
@@ -362,6 +364,7 @@ paths:
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
                         employee: ["emp_id", "department"]
+                      acrValues: ["urn:thunder:silver", "urn:thunder:gold"]
         "400":
           description: Bad request
           content:
@@ -485,6 +488,7 @@ paths:
                     scopeClaims:
                       profile: ["name", "family_name", "given_name", "picture"]
                       employee: ["emp_id", "department"]
+                    acrValues: ["urn:thunder:silver", "urn:thunder:gold"]
       responses:
         "200":
           description: Application updated
@@ -556,6 +560,7 @@ paths:
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
                         employee: ["emp_id", "department"]
+                      acrValues: ["urn:thunder:silver", "urn:thunder:gold"]
         "400":
           description: Bad request
           content:
@@ -1278,6 +1283,18 @@ components:
           example:
             profile: ["name", "family_name", "given_name", "picture"]
             employee: ["emp_id", "department"]
+        acrValues:
+          type: array
+          items:
+            type: string
+          description: |
+            ACR (Authentication Context Class Reference) values for the application.
+            Defines the allowed set of ACR values for authorization requests. When acr_values is
+            provided in an authorization request, only values present in this list are accepted;
+            unrecognised values are silently ignored. If filtering removes all requested ACRs,
+            the full list is used as a fallback. When acr_values is omitted from the request,
+            this configured list is used as the effective ACR set.
+          example: ["urn:thunder:silver", "urn:thunder:gold"]
 
     OAuthAppConfigComplete:
       type: object
@@ -1356,6 +1373,18 @@ components:
           example:
             profile: ["name", "family_name", "given_name", "picture"]
             employee: ["emp_id", "department"]
+        acrValues:
+          type: array
+          items:
+            type: string
+          description: |
+            ACR (Authentication Context Class Reference) values for the application.
+            Defines the allowed set of ACR values for authorization requests. When acr_values is
+            provided in an authorization request, only values present in this list are accepted;
+            unrecognised values are silently ignored. If filtering removes all requested ACRs,
+            the full list is used as a fallback. When acr_values is omitted from the request,
+            this configured list is used as the effective ACR set.
+          example: ["urn:thunder:silver", "urn:thunder:gold"]
 
     Error:
       type: object
--- a/backend/internal/application/service.go
+++ b/backend/internal/application/service.go
@@ -34,6 +34,7 @@ import (
 	oauth2const "github.com/asgardeo/thunder/internal/oauth/oauth2/constants"
 	oauthutils "github.com/asgardeo/thunder/internal/oauth/oauth2/utils"
 	oupkg "github.com/asgardeo/thunder/internal/ou"
+	"github.com/asgardeo/thunder/internal/system/config"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/i18n/core"
 	i18nmgt "github.com/asgardeo/thunder/internal/system/i18n/mgt"
@@ -709,6 +710,7 @@ func buildOAuthConfigData(inboundAuth model.InboundAuthConfigProcessedDTO) *inbo
 		Token:                              oa.Token,
 		UserInfo:                           oa.UserInfo,
 		Certificate:                        oa.Certificate,
+		AcrValues:                          oa.AcrValues,
 	}
 }
 
@@ -899,7 +901,31 @@ func validateOAuthParamsForCreateAndUpdate(app *model.ApplicationDTO) (*model.In
 		oauthAppConfig.TokenEndpointAuthMethod = oauth2const.TokenEndpointAuthMethodClientSecretBasic
 	}
 
+	if err := validateAcrValues(oauthAppConfig.AcrValues); err != nil {
+		return nil, err
+	}
+
 	return inboundAuthConfig, nil
+}
+
+// isValidACR reports whether acr is present in the deployment config ACR-AMR mapping.
+func isValidACR(acr string) bool {
+	mapping := config.GetServerRuntime().Config.OAuth.AuthClass
+	_, ok := mapping.AcrAMR[acr]
+	return ok
+}
+
+// validateAcrValues rejects acr values not registered in the ACR-AMR mapping.
+func validateAcrValues(acrValues []string) *serviceerror.ServiceError {
+	for _, acr := range acrValues {
+		if !isValidACR(acr) {
+			return serviceerror.CustomServiceError(ErrorInvalidAcrValues, core.I18nMessage{
+				Key:          "error.applicationservice.invalid_acr_values_unrecognized",
+				DefaultValue: fmt.Sprintf("ACR value %q is not recognized by the system", acr),
+			})
+		}
+	}
+	return nil
 }
 
 func translateOAuthValidationError(err error) *serviceerror.ServiceError {
@@ -1340,6 +1366,7 @@ func buildApplicationResponse(dto *model.ApplicationProcessedDTO) *model.Applica
 					Scopes:                             oauthAppConfig.Scopes,
 					UserInfo:                           oauthAppConfig.UserInfo,
 					ScopeClaims:                        oauthAppConfig.ScopeClaims,
+					AcrValues:                          oauthAppConfig.AcrValues,
 				},
 			})
 		}
@@ -1457,6 +1484,7 @@ func buildOAuthInboundAuthConfigProcessedDTO(
 			UserInfo:                           userInfo,
 			ScopeClaims:                        scopeClaims,
 			Certificate:                        certificate,
+			AcrValues:                          inboundAuthConfig.OAuthAppConfig.AcrValues,
 		},
 	}
 }
@@ -1512,6 +1540,7 @@ func buildReturnApplicationDTO(
 				UserInfo:                           userInfo,
 				ScopeClaims:                        scopeClaims,
 				Certificate:                        oauthCert,
+				AcrValues:                          inboundAuthConfig.OAuthAppConfig.AcrValues,
 			},
 		}
 		returnApp.InboundAuthConfig = []model.InboundAuthConfigDTO{returnInboundAuthConfig}
--- a/backend/internal/application/service_test.go
+++ b/backend/internal/application/service_test.go
@@ -2647,5 +2647,114 @@ func (suite *ServiceTestSuite) TestValidateApplication_ErrorFromProcessInboundAu
 	assert.Equal(suite.T(), &ErrorInvalidInboundAuthConfig, svcErr)
 }
 
-// TestValidateApplication_ErrorFromValidateAuthFlowID tests error from validateAuthFlowID
-// when an invalid auth flow ID is provided.
+var validAcrMapping = config.AuthClassConfig{
+	Amrs: []string{"PWD", "OTP"},
+	AcrAMR: map[string][]string{
+		"urn:thunder:acr:password":       {"PWD"},
+		"urn:thunder:acr:generated-code": {"OTP"},
+	},
+}
+
+type AcrValidationTestSuite struct {
+	suite.Suite
+}
+
+func TestAcrValidationTestSuite(t *testing.T) {
+	suite.Run(t, new(AcrValidationTestSuite))
+}
+
+func (s *AcrValidationTestSuite) initRegistry(mapping config.AuthClassConfig) {
+	config.ResetServerRuntime()
+	s.Require().NoError(config.InitializeServerRuntime("", &config.Config{
+		OAuth: config.OAuthConfig{
+			AuthClass: mapping,
+		},
+	}))
+	s.T().Cleanup(config.ResetServerRuntime)
+}
+
+func (s *AcrValidationTestSuite) TestValidateAcrValues_EmptyList() {
+	err := validateAcrValues(nil)
+	s.Nil(err)
+
+	err = validateAcrValues([]string{})
+	s.Nil(err)
+}
+
+func (s *AcrValidationTestSuite) TestValidateAcrValues_AllValid() {
+	s.initRegistry(validAcrMapping)
+
+	err := validateAcrValues([]string{
+		"urn:thunder:acr:password",
+		"urn:thunder:acr:generated-code",
+	})
+
+	s.Nil(err)
+}
+
+func (s *AcrValidationTestSuite) TestValidateAcrValues_SingleValid() {
+	s.initRegistry(validAcrMapping)
+
+	err := validateAcrValues([]string{"urn:thunder:acr:password"})
+
+	s.Nil(err)
+}
+
+func (s *AcrValidationTestSuite) TestValidateAcrValues_UnknownACR() {
+	s.initRegistry(validAcrMapping)
+
+	svcErr := validateAcrValues([]string{
+		"urn:thunder:acr:password",
+		"urn:thunder:acr:unknown-method",
+	})
+
+	s.NotNil(svcErr)
+	s.Equal("APP-1033", svcErr.Code)
+	s.Contains(svcErr.ErrorDescription.DefaultValue, "urn:thunder:acr:unknown-method")
+}
+
+func (s *AcrValidationTestSuite) TestValidateAcrValues_FirstEntryInvalid() {
+	s.initRegistry(validAcrMapping)
+
+	svcErr := validateAcrValues([]string{"totally-invalid-acr"})
+
+	s.NotNil(svcErr)
+	s.Equal("APP-1033", svcErr.Code)
+	s.Contains(svcErr.ErrorDescription.DefaultValue, "totally-invalid-acr")
+}
+
+func (s *AcrValidationTestSuite) TestIsValidACR_KnownACR() {
+	s.initRegistry(validAcrMapping)
+
+	s.True(isValidACR("urn:thunder:acr:password"))
+}
+
+func (s *AcrValidationTestSuite) TestIsValidACR_UnknownACR() {
+	s.initRegistry(validAcrMapping)
+
+	s.False(isValidACR("urn:thunder:acr:unknown"))
+}
+
+func (s *AcrValidationTestSuite) TestIsValidACR_EmptyString() {
+	s.initRegistry(validAcrMapping)
+
+	s.False(isValidACR(""))
+}
+
+func (s *AcrValidationTestSuite) TestIsValidACR_AllMappedACRs() {
+	s.initRegistry(validAcrMapping)
+
+	knownACRs := []string{
+		"urn:thunder:acr:password",
+		"urn:thunder:acr:generated-code",
+	}
+	for _, acr := range knownACRs {
+		s.True(isValidACR(acr), "expected ACR %q to be valid", acr)
+	}
+}
+
+func (s *AcrValidationTestSuite) TestIsValidACR_EmptyMapping() {
+	s.initRegistry(config.AuthClassConfig{})
+
+	s.False(isValidACR("urn:thunder:acr:password"))
+}
--- a/backend/internal/flow/common/constants.go
+++ b/backend/internal/flow/common/constants.go
@@ -146,6 +146,14 @@ const (
 // DefaultHTTPTimeout defines the default timeout duration for HTTP requests.
 const DefaultHTTPTimeout = 5 * time.Second
 
+// NodeVariant identifies a PROMPT node sub-type that activates a variant-specific code path.
+type NodeVariant string
+
+const (
+	// NodeVariantLoginOptions identifies a PROMPT node that presents login method choices to the user.
+	NodeVariantLoginOptions NodeVariant = "LOGIN_OPTIONS"
+)
+
 const (
 	// NodePropertyAllowAuthenticationWithoutLocalUser indicates whether authentication is allowed without a local user
 	NodePropertyAllowAuthenticationWithoutLocalUser = "allowAuthenticationWithoutLocalUser"
@@ -160,6 +168,8 @@ const (
 	NodePropertyOUResolveFrom = "resolveFrom"
 	// NodePropertyRecipientAttribute specifies the destination parameter to use for verification (e.g., email, phone).
 	NodePropertyRecipientAttribute = "destinationAttribute"
+	// NodePropertyAuthMethodMapping maps authentication classes to action refs on login_options PROMPT nodes.
+	NodePropertyAuthMethodMapping = "authMethodMapping"
 )
 
 const (
@@ -223,6 +233,12 @@ const (
 	InvalidMagicLinkToken = "Invalid magic link token"
 	// RuntimeKeyOAuthState holds the generated OAuth state parameter for CSRF validation.
 	RuntimeKeyOAuthState = "oauthState"
+	// RuntimeKeyRequestedAuthClasses holds the space-separated ACR values from acr_values.
+	RuntimeKeyRequestedAuthClasses = "requested_auth_classes"
+	// RuntimeKeySelectedAuthClass holds the ACR value of the chosen authentication method.
+	RuntimeKeySelectedAuthClass = "selected_auth_class"
+	// RuntimeKeyAllowedLoginOptions holds the space-separated action refs allowed on a LOGIN_OPTIONS node.
+	RuntimeKeyAllowedLoginOptions = "allowed_login_options"
 )
 
 // TODO: Define a go type for InputType when formalizing input types
--- a/backend/internal/flow/core/prompt_node.go
+++ b/backend/internal/flow/core/prompt_node.go
@@ -19,10 +19,15 @@
 package core
 
 import (
+	"slices"
+	"strings"
+
 	"github.com/asgardeo/thunder/internal/flow/common"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/log"
 )
+
+const failureReasonInvalidAction = "Invalid action selected"
 
 // PromptNodeInterface extends NodeInterface for nodes that require user interaction.
 type PromptNodeInterface interface {
@@ -36,6 +41,8 @@ type PromptNodeInterface interface {
 	GetMessage() string
 	SetMessage(message string)
 	IsDisplayOnly() bool
+	GetVariant() common.NodeVariant
+	SetVariant(variant common.NodeVariant)
 }
 
 // promptNode represents a node that prompts for user input/ action in the flow execution.
@@ -45,6 +52,7 @@ type promptNode struct {
 	meta     interface{}
 	nextNode string
 	message  string
+	variant  common.NodeVariant
 	logger   *log.Logger
 }
 
@@ -113,16 +121,20 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 		return nodeResp, nil
 	}
 
+	if n.variant == common.NodeVariantLoginOptions {
+		return n.executeLoginOptions(ctx, nodeResp)
+	}
+
 	if n.resolvePromptInputs(ctx, nodeResp) {
 		logger.Debug("All required inputs and action are available, returning complete status")
 
 		if ctx.CurrentAction != "" {
-			if nextNode := n.getNextNodeForActionRef(ctx.CurrentAction, logger); nextNode != "" {
+			if nextNode := n.getNextNodeForActionRef(ctx.CurrentAction); nextNode != "" {
 				nodeResp.NextNodeID = nextNode
 			} else {
-				logger.Debug("Invalid action selected", log.String("actionRef", ctx.CurrentAction))
+				logger.Debug(failureReasonInvalidAction, log.String("actionRef", ctx.CurrentAction))
 				nodeResp.Status = common.NodeStatusFailure
-				nodeResp.FailureReason = "Invalid action selected"
+				nodeResp.FailureReason = failureReasonInvalidAction
 				return nodeResp, nil
 			}
 		}
@@ -146,13 +158,113 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 
 	// Include meta in the response if verbose mode is enabled
 	if ctx.Verbose && n.GetMeta() != nil {
-		trimmed := n.trimMetaToRequestedInputs(nodeResp.Inputs, nodeResp.Actions)
+		trimmed := n.trimMetaToRequestedInputs(n.meta, nodeResp.Inputs, nodeResp.Actions)
 		nodeResp.Meta = n.appendSyntheticMetaComponents(trimmed, nodeResp.Inputs)
 	}
 
 	nodeResp.Status = common.NodeStatusIncomplete
 	nodeResp.Type = common.NodeResponseTypeView
 	return nodeResp, nil
+}
+
+// executeLoginOptions handles the LOGIN_OPTIONS variant.
+func (n *promptNode) executeLoginOptions(ctx *NodeContext,
+	nodeResp *common.NodeResponse) (*common.NodeResponse, *serviceerror.ServiceError) {
+	logger := n.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+	authClassToAction := n.authClassToActionMapping()
+
+	if ctx.CurrentAction != "" {
+		if allowedRaw := ctx.RuntimeData[common.RuntimeKeyAllowedLoginOptions]; allowedRaw != "" {
+			if !slices.Contains(strings.Fields(allowedRaw), ctx.CurrentAction) {
+				logger.Debug("Selected action is not in allowed login options",
+					log.String("actionRef", ctx.CurrentAction))
+				nodeResp.Status = common.NodeStatusFailure
+				nodeResp.FailureReason = failureReasonInvalidAction
+				return nodeResp, nil
+			}
+		}
+		if n.resolvePromptInputs(ctx, nodeResp) {
+			return n.finalizeLoginOptionsAction(ctx, nodeResp, authClassToAction)
+		}
+		n.applyMetaForLoginOptions(ctx, nodeResp, nil)
+		nodeResp.Status = common.NodeStatusIncomplete
+		nodeResp.Type = common.NodeResponseTypeView
+		return nodeResp, nil
+	}
+
+	requestedAuthClasses := parseAuthClasses(ctx.RuntimeData[common.RuntimeKeyRequestedAuthClasses])
+	effectivePrompts := n.filterAndOrderPrompts(requestedAuthClasses, authClassToAction)
+	actions := make([]common.Action, 0)
+	for _, p := range effectivePrompts {
+		if p.Action != nil {
+			actions = append(actions, *p.Action)
+		}
+	}
+
+	// Auto-select the sole remaining option so the user skips a single-choice chooser.
+	if len(actions) == 1 && len(requestedAuthClasses) > 0 {
+		ctx.CurrentAction = actions[0].Ref
+		logger.Debug("Auto-selected single login option", log.String("actionRef", ctx.CurrentAction))
+		if n.resolvePromptInputs(ctx, nodeResp) {
+			return n.finalizeLoginOptionsAction(ctx, nodeResp, authClassToAction)
+		}
+		nodeResp.RuntimeData[common.RuntimeKeyAllowedLoginOptions] = ctx.CurrentAction
+		n.applyMetaForLoginOptions(ctx, nodeResp, nil)
+		nodeResp.Status = common.NodeStatusIncomplete
+		nodeResp.Type = common.NodeResponseTypeView
+		return nodeResp, nil
+	}
+
+	nodeResp.Actions = append(nodeResp.Actions, actions...)
+	nodeResp.RuntimeData[common.RuntimeKeyAllowedLoginOptions] = joinActionRefs(effectivePrompts)
+	n.applyMetaForLoginOptions(ctx, nodeResp, effectivePrompts)
+	nodeResp.Status = common.NodeStatusIncomplete
+	nodeResp.Type = common.NodeResponseTypeView
+	return nodeResp, nil
+}
+
+// finalizeLoginOptionsAction completes the chooser once an action is selected and inputs are satisfied.
+func (n *promptNode) finalizeLoginOptionsAction(ctx *NodeContext, nodeResp *common.NodeResponse,
+	authClassToAction map[string]string) (*common.NodeResponse, *serviceerror.ServiceError) {
+	logger := n.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+	nextNode := n.getNextNodeForActionRef(ctx.CurrentAction)
+	if nextNode == "" {
+		logger.Debug(failureReasonInvalidAction, log.String("actionRef", ctx.CurrentAction))
+		nodeResp.Status = common.NodeStatusFailure
+		nodeResp.FailureReason = failureReasonInvalidAction
+		return nodeResp, nil
+	}
+	nodeResp.NextNodeID = nextNode
+	for authClass, ref := range authClassToAction {
+		if ref == ctx.CurrentAction {
+			nodeResp.RuntimeData[common.RuntimeKeySelectedAuthClass] = authClass
+			break
+		}
+	}
+	if actionType := n.getActionTypeForRef(ctx.CurrentAction); actionType != "" {
+		if nodeResp.ForwardedData == nil {
+			nodeResp.ForwardedData = make(map[string]interface{})
+		}
+		nodeResp.ForwardedData[common.ForwardedDataKeyActionType] = actionType
+	}
+	nodeResp.Status = common.NodeStatusComplete
+	nodeResp.Type = ""
+	return nodeResp, nil
+}
+
+// applyMetaForLoginOptions sets verbose-mode meta on the response, reordering it to match
+// effectivePrompts when provided.
+func (n *promptNode) applyMetaForLoginOptions(ctx *NodeContext, nodeResp *common.NodeResponse,
+	effectivePrompts []common.Prompt) {
+	if !ctx.Verbose || n.GetMeta() == nil {
+		return
+	}
+	meta := n.meta
+	if effectivePrompts != nil {
+		meta = n.filteredMeta(effectivePrompts)
+	}
+	trimmed := n.trimMetaToRequestedInputs(meta, nodeResp.Inputs, nodeResp.Actions)
+	nodeResp.Meta = n.appendSyntheticMetaComponents(trimmed, nodeResp.Inputs)
 }
 
 // GetPrompts returns the prompts for the prompt node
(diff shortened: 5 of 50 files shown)


backend/internal/system/config/config_test.go
```diff
@@ -749,3 +749,79 @@ func (suite *ConfigTestSuite) createTempFile(dir, pattern, content string) strin
...
+		},
+		AcrAMR: map[string][]string{
+			"mosip:idp:acr:password": {"Password"},
+			"mosip:idp:acr:empty":    {},
+		},
+	}
+	err := cfg.Validate()
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "mosip:idp:acr:empty")
+	assert.Contains(t, err.Error(), "empty AMR list")
```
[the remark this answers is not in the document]
sacrana0: Isn't assert.Error(t, err) already doing that?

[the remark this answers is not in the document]
sacrana0: @coderabbitai check my reply to your message  `Isn't assert.Error(t, err) already doing that?`

api/application.yaml
```diff
@@ -1148,6 +1148,17 @@ components:
...
+          type: array
+          items:
+            type: string
+          description: |
+            Default ACR (Authentication Context Class Reference) values for the application.
+            Defines the allowed set of ACR values for authorization requests. When acr_values is
+            provided in an authorization request, only values present in this list are accepted;
+            unrecognised values are silently ignored. If filtering removes all requested ACRs,
+            the full default list is used as a fallback.
+          example: ["urn:okta:loa:1fa:pwd", "urn:okta:loa:2fa:any"]
```
ThaminduDilshan: Shall we use different examples? Maybe generic ones

api/application.yaml
```diff
@@ -1221,6 +1232,17 @@ components:
...
```
ThaminduDilshan: Same as previous comment

api/application.yaml
```diff
@@ -1148,6 +1148,17 @@ components:
           example:
             profile: ["name", "family_name", "given_name", "picture"]
             employee: ["emp_id", "department"]
+        defaultAcrValues:
```
ThaminduDilshan: Can you check this gets appeared properly in swagger editor for application create, update and get request responses? Feel like missing some places

backend/internal/system/config/config.go
```diff
@@ -401,6 +401,52 @@ type ConsentConfig struct {
 	MaxRetries int    `yaml:"max_retries" json:"max_retries"` // Max retry attempts for transient errors. Default: 3
 }
 
+// AMRFactor represents a single authentication method reference factor entry.
+type AMRFactor struct {
+	Type  string `yaml:"type" json:"type"`
+	Count int    `yaml:"count,omitempty" json:"count,omitempty"`
```
ThaminduDilshan: What is the usage of this count parameter? Is this for a customization?

tests/integration/oauth/authz/acr_id_token_test.go
```diff
@@ -0,0 +1,510 @@
...
+			"id":        "start",
+			"type":      "START",
+			"onSuccess": "acr_chooser",
+		},
+		{
+			"id":   "acr_chooser",
+			"type": "PROMPT",
+			"properties": map[string]interface{}{
+				"authMethodMapping": map[string]interface{}{
+					"mosip:idp:acr:password":       "pwd_action",
```
ThaminduDilshan: Shall we check all the acr-amr references/ examples added. We should have generic values or thunder specific values rather than adding different vendor related terms

backend/internal/system/config/config.go
sacrana0: This is the structure we have in MOSIP but not used anywhere, maybe kept for future usecases. Will remove it for now.

backend/internal/application/acr_registry.go
```diff
@@ -0,0 +1,30 @@
+/*
```
ThaminduDilshan, line 1: No need to have a util specific file. Shall we remove this file and move this function to the service.go?

backend/internal/application/service_test.go
```diff
@@ -6951,3 +6951,57 @@ func (suite *ServiceTestSuite) TestValidateApplicationForUpdate_ErrorFromValidat
 	assert.NotNil(suite.T(), svcErr)
 	assert.Equal(suite.T(), &ErrorLayoutNotFound, svcErr)
 }
+
+// ---------------------------------------------------------------------------
+// validateDefaultAcrValues
+// ---------------------------------------------------------------------------
+
+func TestValidateDefaultAcrValues_EmptyList(t *testing.T) {
```
ThaminduDilshan: Shall we follow project pattern of using test suites? You can move these methods to ServiceTestSuite

backend/internal/flow/common/constants.go
```diff
@@ -146,6 +146,13 @@ const (
 // DefaultHTTPTimeout defines the default timeout duration for HTTP requests.
 const DefaultHTTPTimeout = 5 * time.Second
 
+const (
```
ThaminduDilshan, line 152: Shall we define a go type for node variant?

backend/internal/flow/common/constants.go
```diff
@@ -158,6 +165,9 @@ const (
 	// NodePropertyOUResolveFrom specifies the strategy for resolving the organization unit.
 	// Supported values: "caller" (use the caller's OU).
 	NodePropertyOUResolveFrom = "resolveFrom"
+	// NodePropertyAuthMethodMapping maps ACR values to action refs on login_options PROMPT nodes.
```
ThaminduDilshan:
```suggestion
	// NodePropertyAuthMethodMapping maps authentication methods to action refs on login_options PROMPT nodes.
```

backend/internal/flow/common/constants.go
```diff
@@ -146,6 +146,13 @@ const (
 // DefaultHTTPTimeout defines the default timeout duration for HTTP requests.
 const DefaultHTTPTimeout = 5 * time.Second
 
+const (
+	// NodeVariantLoginOptions identifies a PROMPT node that presents login method choices to the user.
+	// Each prompt on such a node should carry an ACR tag; the engine filters and reorders the
+	// available actions based on the requested_acr_values in RuntimeData.
+	NodeVariantLoginOptions = "login_options"
```
ThaminduDilshan:
```suggestion
	// NodeVariantLoginOptions identifies a PROMPT node that presents login method choices to the user.
	NodeVariantLoginOptions = "LOGIN_OPTIONS"
```

Let's not mention any acr specific details here keep this as a generic flag.
Check any other such comments as well

backend/internal/flow/common/model.go
```diff
@@ -48,6 +48,7 @@ type Action struct {
 
 // Prompt groups inputs with an action for prompt nodes.
 type Prompt struct {
+	ACR    string  `json:"acr,omitempty"`
```
ThaminduDilshan: Why do we need this? I don't like having spec specific fields at this level

backend/internal/flow/core/prompt_node.go
```diff
@@ -368,11 +414,11 @@ func (n *promptNode) tryAutoSelectSingleAction(ctx *NodeContext) bool {
 	return false
 }
 
-// getAllInputs returns all unique inputs from prompts, deduplicated by Identifier.
-func (n *promptNode) getAllInputs() []common.Input {
+// getAllInputsFrom returns all unique inputs from the given prompts, deduplicated by Identifier.
+func getAllInputsFrom(prompts []common.Prompt) []common.Input {
```
ThaminduDilshan: Why is this change required? 

`getAllInputs()` is a receiver method in prompt node and it already has access to prompts of the current node. No need to pass as a method param. This change only moves `n.prompts` access within method to the outside. I.e. getAllInputsFrom(n.prompts).

Same applies to getAllActions() as well

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -168,6 +168,12 @@ func (a *authAssertExecutor) generateAuthAssertion(ctx *core.NodeContext, logger
 		jwtClaims["authorized_permissions"] = permissions
 	}
 
+	// Include the completed auth class in the JWT so the OAuth2 callback service can
+	// populate the acr claim in the issued ID token.
```
ThaminduDilshan:
```suggestion
	// Include the completed auth class in the JWT
```

backend/internal/flow/common/constants.go
```diff
@@ -195,6 +205,19 @@ const (
...
+	// RuntimeKeySelectedAuthClass holds the ACR value of the authentication method actually used by the user.
+	// Set by the flow engine when a login_options prompt node resolves the chosen action.
+	RuntimeKeySelectedAuthClass = "selected_auth_class"
+	// RuntimeKeyCompletedAuthMethods holds the space-separated list of completed AMR type names.
+	// Accumulated across authentication steps; each authentication executor appends the AMR types it satisfies.
+	RuntimeKeyCompletedAuthMethods = "completed_auth_methods"
+	// ClaimCompletedAuthClass is the JWT claim key used in auth assertion tokens to record
+	// the ACR value that the user satisfied. The OAuth2 callback service reads this claim
+	// to populate the acr claim in the issued ID token.
+	ClaimCompletedAuthClass = "completed_auth_class"
```
ThaminduDilshan: Let's not have this here. This is not a runtime key or flow related constant

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -281,6 +282,17 @@ func (b *graphBuilder) configureNodeInputs(nodeDef *NodeDefinition, node core.No
 	executorNode.SetInputs(inputs)
 }
 
+// configureNodeVariant configures the variant for a prompt node.
+// A node whose Properties contain authMethodMapping is implicitly a login_options variant.
+func (b *graphBuilder) configureNodeVariant(nodeDef *NodeDefinition, node core.NodeInterface) {
+	if _, ok := nodeDef.Properties[common.NodePropertyAuthMethodMapping]; !ok {
```
ThaminduDilshan: Let's not have this validation. Ideally this should go into the flow mgt [validator](https://github.com/asgardeo/thunder/issues/2085) that we're planning to onboard. We can blindly set the variant define in flow definition

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -281,6 +282,17 @@ func (b *graphBuilder) configureNodeInputs(nodeDef *NodeDefinition, node core.No
...
+		return
+	}
+	if promptNode, ok := node.(core.PromptNodeInterface); ok {
+		promptNode.SetVariant(common.NodeVariantLoginOptions)
```
ThaminduDilshan: Rather than setting a hard coded variant like this, let's rely on the variant set in the flow definition

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -320,6 +332,8 @@ func (b *graphBuilder) configureNodePrompts(nodeDef *NodeDefinition, node core.N
 		return nil
 	}
 
+	actionToACR := buildActionToACRMap(nodeDef.Properties)
```
ThaminduDilshan: We shouldn't have these logics. These have to be explicitly define in the flow definition by the flow designer. Designer should set variant and required node properties appropriately when defining the flow definition. Graph builder just blindly set them while building the graph.

Ideally these validations will be moved to the flow validator we're [planning](https://github.com/asgardeo/thunder/issues/2085) to have. And to improve UX, these will be set appropriately by the frontend flow designer

backend/internal/oauth/oauth2/authz/service_test.go
```diff
@@ -1709,3 +1709,206 @@ func determineClaimsForTokens(oidcScopes []string, claimsRequest *oauth2model.Cl
 
 	return accessTokenClaims, idTokenClaims, userInfoClaims
 }
+
+// ---------------------------------------------------------------------------
+// parseACRValues — unit tests
+// ---------------------------------------------------------------------------
+
+func TestParseACRValues_SingleACR(t *testing.T) {
```
ThaminduDilshan: Let's follow test suite pattern

backend/internal/system/config/config.go
ThaminduDilshan: Let's remove it. Doesn't feel valid

backend/internal/system/config/config.go
```diff
@@ -552,6 +598,7 @@ type Config struct {
 	Layout               LayoutConfig           `yaml:"layout" json:"layout"`
 	Email                EmailConfig            `yaml:"email" json:"email"`
 	Consent              ConsentConfig          `yaml:"consent" json:"consent"`
+	ACRAMRMapping        ACRAMRMappingConfig    `yaml:"acr_amr_mapping" json:"acr_amr_mapping"`
```
ThaminduDilshan: Move this inside existing `OAuthConfig` config

backend/internal/system/config/config.go
```diff
@@ -523,6 +523,52 @@ func (c *TrustedIssuerConfig) Validate() error {
...
 }
 
+// AMRFactor represents a single authentication method reference factor entry.
+type AMRFactor struct {
+	Type  string `yaml:"type" json:"type"`
+	Count int    `yaml:"count,omitempty" json:"count,omitempty"`
+}
+
+// ACRAMRMappingConfig holds the ACR-AMR mapping configuration.
+type ACRAMRMappingConfig struct {
```
ThaminduDilshan: Shall we think of a better name. Better to have a general name for the config container rather than specifying to mapping. A field inside this is the mapping.
@thiva-k any suggestions?

backend/internal/system/config/config_test.go
```diff
@@ -993,3 +994,118 @@ func (suite *ConfigTestSuite) createTempFile(dir, pattern, content string) strin
 
 	return tempFile.Name()
 }
+
+// ---------------------------------------------------------------------------
+// ACRAMRMappingConfig.Validate
+// ---------------------------------------------------------------------------
+
+func TestACRAMRMappingValidate_EmptyConfig(t *testing.T) {
```
ThaminduDilshan: Follow test suite pattern

tests/integration/flow/authentication/acr_options_flow_test.go
```diff
@@ -0,0 +1,651 @@
...
+	Nodes: []map[string]interface{}{
+		{
+			"id":        "start",
+			"type":      "START",
+			"onSuccess": "acr_chooser",
+		},
+		// login_options prompt node — the ACR chooser (AC-10)
+		{
+			"id":   "acr_chooser",
+			"type": "PROMPT",
```
ThaminduDilshan: variant should be defined here in the flow definition similar to type and other fields

sacrana0: There is no need of variant field now. Any node with authMethodMapping property in properties will be treated as the login options prompt node.

backend/internal/system/config/config.go
sacrana0: How about ACRConfig or AuthClassConfig?

backend/internal/system/config/config.go
sacrana0: will this acr_values feature only be used by Oauth application. I think we were making it generic to be used by other clients as well? Its already part of flow json which is outside oauth.

ThaminduDilshan: That generalization is only applied to the flow execution layer right? This config is pure oauth config. Even the name suggests the same.
Consider the usage of this config;
1. Application service validate OAuth configs against this (OAuth specific)
2. OAuth layer validates these mappings against the acr values send in authorization request (OAuth specific)

Engine just acts on filtered data in runtime data

tests/integration/flow/authentication/acr_options_flow_test.go
ThaminduDilshan: Refer to https://github.com/asgardeo/thunder/pull/2203#discussion_r3128971808

tests/integration/testutils/test_utils.go
```diff
@@ -118,7 +118,7 @@ func ensureInitialized() {
 		zipFilePattern = os.Getenv("THUNDER_ZIP_PATTERN")
 		dbType = os.Getenv("DB_TYPE")
 		if dbType == "" {
-			dbType = "sqlite"
+			dbType = "postgres"
```
ThaminduDilshan: Do we need these changes?

tests/integration/testutils/test_utils.go
```diff
@@ -815,7 +863,7 @@ func RunSetupScript() error {
 	cmd.Dir = absProductHome // Run from product directory
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
-	cmd.Env = os.Environ()
+	cmd.Env = append(os.Environ(), "WITH_CONSENT=false")
```
ThaminduDilshan: Shall we prevent changing default server setup?. This will affect other development work

tests/integration/testutils/test_utils.go
ThaminduDilshan: Shall we prevent changing default server setup?. This will affect other development work

tests/integration/testutils/test_utils.go
```diff
@@ -790,6 +790,54 @@ func PatchDeploymentConfig(patch map[string]interface{}) error {
...
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, fmt.Errorf("failed to parse deployment.yaml: %w", err)
+	}
+
+	return cfg[key], nil
+}
+
+// GetConfiguredAcrValues reads the running server's deployment.yaml and returns the ACR values
+// defined in the acr_amr_mapping section. Returns an empty slice when no mapping is configured.
+func GetConfiguredAcrValues() ([]string, error) {
```
ThaminduDilshan: I don't think this is needed. Let's configure the integration test server deployment.yaml with the expected values and assume they're present during integration tests

tests/integration/testutils/api_utils.go
```diff
@@ -37,6 +38,21 @@ func GetHTTPClient() *http.Client {
 	return NewHTTPClientWithTokenProvider(GetAccessToken)
 }
 
+// GetNoRedirectHTTPClient returns an HTTP client that does not follow redirects.
+// Use this when testing endpoints that return 302 redirects (e.g. /oauth2/authorize)
+// where the redirect destination is not running (e.g. the frontend gate).
+func GetNoRedirectHTTPClient() *http.Client {
```
ThaminduDilshan, line 42: What's the usage of this method? I don't understand the use case

sacrana0: This is required for integration tests where we are calling /authorize which redirects, so to avoid following the redirect and assert some conditions on it. The same is already created separately in SimulateFederatedOAuthFlow method in the same file.

backend/internal/flow/mgt/graph_builder.go
sacrana0: The flow definition has only one acr related property called authMethodMapping which maps the the amr to the button associated with it.
1. What I am currently doing is, while building the graph I am marking the PromptNode as NodeVariantLoginOptions (there is no variant field in flow json but is there in go prompt node struct) and for each button I am adding the associated acr to it (in flow json each button does not have acr field). So during filter it becomes easy to identify which prompt node to apply filter to and what components to remove.

2. If we want to remove this from graph builder then during filter we will have to read the property again and again to identify which components to remove if there are multiple loginoption prompt nodes.

@ThaminduDilshan 
@coderabbitai what are your thoughts on it?

tests/integration/testutils/test_utils.go
sacrana0: These changes were accidentally merged, removing them.

backend/internal/flow/core/prompt_node.go
sacrana0: removing getAllInputsFrom but cannot do the same for getAllActions as it needs to work on separate set of actions. For LoginOptions node it takes effective actions (filtered based on acrs) and for other nodes it takes all actions on the node.

backend/internal/flow/common/model.go
sacrana0: This is to mark prompts with the acr associated with it. We use it during filtering to identify which prompts to remove.

ThaminduDilshan: No Sachin. Let me explain this with a diagram. In a simple mode, your login flow will be something like this. This is the model that we can support as per the current capabilities of the flow execution engine.

prompt login option filtering

There will be a `PROMPT_NODE` with variant `LOGIN_OPTIONS` in the beginning of the flow definition. Every possible path after this node is a separate branch. So the first two requirements are to hide some paths based on the requested acr values and change their order of appearance (priority).

Next requirement is to ensure that a selected action (or path) is something allowed. I.e. the prompt node in the above example has 4 paths, but requested ACR only contains OTP and PASSWORD. So prompt has to ensure selected action is available in the filtered action list. Once a path is selected, execution cannot switch back to a separate path as you're already in a particular branch. So the same PROMPT_NODE` with variant `LOGIN_OPTIONS` can ensure selected action is a allowed action.

In the simplest form, current `PROMPT_NODE` execution is like this.

```plaintext
//////////////////////////////////////////////////////////////
// Handle a prompt handling a failure
// This is when a prompt node is used as an error page
//////////////////////////////////////////////////////////////
...

//////////////////////////////////////////////////////////////
// Handle a display only prompt
// This is when a prompt node is used just to display some
// information without any user input/ selection
//////////////////////////////////////////////////////////////
...

action ← selectedAction
inputs ← providedInputs

IF action is present THEN
    requiredInputs ← required inputs for action

    IF inputs do NOT contain all requiredInputs THEN
        missingInputs ← requiredInputs - inputs
        PROMPT user for missingInputs
    ELSE
        SET selected action
        COMPLETE execution
    ENDIF
ELSE
    allInputs ← all possible inputs across prompts

    IF inputs contain all required inputs THEN
        inferredAction ← try to auto-select action

        IF inferredAction is found THEN
            SET selected action to inferredAction
            COMPLETE execution
        ENDIF
    ENDIF

    PROMPT user to provide inputs and actions
ENDIF
```

If I understand the requirement correctly, you can simply embed LOGIN_OPTIONS variant handling logic into this as below.

```plaintext
//////////////////////////////////////////////////////////////
// Handle a prompt handling a failure
// This is when a prompt node is used as an error page
//////////////////////////////////////////////////////////////
...


//////////////////////////////////////////////////////////////
// Handle a display only prompt
// This is when a prompt node is used just to display some
// information without any user input/ selection
//////////////////////////////////////////////////////////////
...


//////////////////////////////////////////////////////////////
// Handle login options variant
//////////////////////////////////////////////////////////////

IF variant = LOGIN_OPTIONS THEN

    //////////////////////////////////////////////////////////
    // Subsequent request (user sent inputs and/or action)
    //////////////////////////////////////////////////////////
    IF action is present THEN
        IF runtimeData indicates loginOptionsFiltering THEN
            allowedOptions ← runtimeData.allowedLoginOptions

            IF action is NOT in allowedOptions THEN
                RETURN error ("selected option not allowed")
            ENDIF

            requiredInputs ← required inputs for action

            IF inputs do NOT contain all requiredInputs THEN
                missingInputs ← requiredInputs - inputs
                PROMPT user for missingInputs
            ELSE
                SET selected action
                COMPLETE execution
            ENDIF
        ELSE
            // No filtering → fallback to normal prompt handling
            ...
        ENDIF

    //////////////////////////////////////////////////////////
    // Initial request (no action yet)
    //////////////////////////////////////////////////////////
    ELSE
        IF "requested_auth_classes" exists in runtimeData THEN
            IF "authMethodMapping" exists in runtimeData THEN
                filteredPrompts ← filter prompts using: requested_auth_classes + authMethodMapping

                orderedPrompts ← reorder filteredPrompts based on requested_auth_classes priority

                SET runtimeData.allowedLoginOptions = actions from orderedPrompts

                IF verbose mode THEN
                    meta ← build modified metadata for orderedPrompts
                ENDIF

                PROMPT user with orderedPrompts and meta
            ELSE
                RETURN error (Filtering requested but mapping not available)
            ENDIF
        ELSE
            // No filtering → fallback to normal prompt handling
            ...
        ENDIF

    ENDIF

ENDIF


//////////////////////////////////////////////////////////////
// Existing prompt node execution logic
//////////////////////////////////////////////////////////////
...
```

backend/internal/flow/mgt/graph_builder.go
ThaminduDilshan: To align with the existing flow engine implementation and architecture, I would like to keep engine and graph handling implementation generic. Later when we support other auth methods (ex: SAML), we can use the same  graph and engine implementation and let pre and post layers worry about the spec specific details.
So all configuration details will be a responsibility of the flow designer, not graph builder or engine.

Flow designer will create a PROMPT_NODE with variant LOGIN_OPTIONS. It's his responsibility to set node properties matching to that variant. In future if we need to use this same parameter for some other requirement (Maybe allow filtering login options for native flow execution), this implementation will break as acr is not applicable for app native flows.

Also refer my other comment on the expected approach: https://github.com/asgardeo/thunder/pull/2203#discussion_r3141717662.

@coderabbitai reevaulate

ThaminduDilshan: Maybe it's fine to set variant automatically if the `authMethodMapping` node property is available. But let's not do any processing here

sacrana0: @ThaminduDilshan lets have a variant field in the flow json itself. Will do it the way you have suggested as outlined by coderabbit.

api/application.yaml
```diff
@@ -156,6 +156,7 @@ paths:
                     scopeClaims:
                       profile: ["name", "family_name", "given_name", "picture"]
                       employee: ["emp_id", "department"]
+                    defaultAcrValues: ["urn:thunder:silver", "urn:thunder:gold"]
```
thiva-k: Ideally should we have two fields.. one for defaultAcrValues and one for allowedAcrValues. Seems like currently we are using the defaultAcrValues to filter the values and also as a fallback when the filtered values result is nil. This is wrong behavior IMO.

If the filtered values is nil, we should return an error or do not send acr claims(check the spec for the correct behavior). This filtering should happen based on allowedAcrValues. The defaultAcr values should be used for an application when no acr values are sent in request as a fallback, not for filtering

thiva-k:
Shouldn't we fail the flow, if `claims` parameter has the acr with `essential`: `true` and we cannot satisfy that?

api/application.yaml
sacrana0: As per discussion with @ThaminduDilshan, renaming it to `acrValues`. But using the same for both filtering and as a fallback.

sacrana0:
> Shouldn't we fail the flow, if `claims` parameter has the acr with `essential`: `true` and we cannot satisfy that?

Keeping this out of scope for now. Will create an issue for this.

sacrana0:
@coderabbitai full review

backend/internal/flow/core/prompt_node.go
```diff
@@ -499,3 +565,175 @@ func filterMetaComponents(comps []interface{}, allowedRefs, knownInputActionRefs
...
+		}
+		// Filtered-out ACTION component: drop it by emitting nothing.
+	}
+
+	resultMap := make(map[string]interface{}, len(metaMap))
+	for k, v := range metaMap {
+		resultMap[k] = v
+	}
+	resultMap["components"] = result
+	return resultMap
```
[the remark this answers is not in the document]
sacrana0: For nested meta layouts we had doubts - 
1. what if the actions are in different blocks, how we will reorder the components?
2. Should we atleast remove non-requested acr action components, will skip reordering from nested actions.
Is there are simple solution to this problem @coderabbitai

api/application.yaml
```diff
@@ -1336,6 +1351,17 @@ components:
           example:
             profile: ["name", "family_name", "given_name", "picture"]
             employee: ["emp_id", "department"]
+        acrValues:
```
ThaminduDilshan, line 1376: This is not visible in get application by id response

backend/internal/flow/core/prompt_node.go
```diff
@@ -113,12 +118,44 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
 	}
 
-	if n.resolvePromptInputs(ctx, nodeResp) {
+	// For login_options nodes, compute the filtered+ordered prompt set as a local variable.
+	// This avoids mutating the shared n.prompts field, which would be a race condition when
+	// multiple goroutines execute the same cached graph node concurrently.
+	effectivePrompts := n.prompts
+	var acrToAction map[string]string
+	if n.variant == common.NodeVariantLoginOptions {
+		acrToAction = n.acrToActionMapping()
```
ThaminduDilshan: Can we get rid of acr references from this file? Also feel like login option filtering is too much embedded into the current implementation.

Can't we simplify the approach as proposed in the comment? Did you encounter any issues implementing this?
https://github.com/asgardeo/thunder/pull/2203#discussion_r3141717662

ThaminduDilshan: @coderabbitai can you analyze the prompt_node changes against the suggested approach in https://github.com/asgardeo/thunder/pull/2203#discussion_r3141717662?

sacrana0: @coderabbitai I have updated the prompt_node.go file. Verify if it as per the requirement.

sacrana0: @ThaminduDilshan review this file again, I have made the required changes.

backend/internal/flow/common/constants.go
sacrana0: There is not much benefit creating a type for it currently. There is only one variant currently, maybe considered in future when we have multiple variants.

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 	return nodeResp, nil
 }
 
+// executeLoginOptions handles the LOGIN_OPTIONS variant end-to-end: auth-class-driven filtering
+// and preference ordering on the initial prompt-out leg, allowedLoginOptions enforcement on the
+// subsequent leg, and selectedAuthClass recording on completion. Self-contained from the
+// generic prompt handling so the shared helpers don't need variant-aware parameters.
+func (n *promptNode) executeLoginOptions(ctx *NodeContext, nodeResp *common.NodeResponse,
```
ThaminduDilshan: Let's not pass logger as a method parameter. Instead access it as `n.logger`. Check other places added with your implementation as well.


```go
logger := n.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
```

backend/internal/flow/common/constants.go
ThaminduDilshan: But shall we define a go type to be aligned with the best coding practices?
It would be easier to define it when introducing new such fields, otherwise this will likely to get missed.

backend/internal/flow/core/prompt_node.go
```diff
@@ -706,3 +836,175 @@ func collectMetaComponentMap(comps interface{}, compMap map[string]map[string]in
...
+	for _, p := range prompts {
+		if p.Action != nil && p.Action.Ref != "" {
+			refs = append(refs, p.Action.Ref)
+		}
+	}
+	return strings.Join(refs, " ")
+}
+
+// containsField reports whether target appears as a whitespace-separated field in raw.
+func containsField(raw, target string) bool {
```
ThaminduDilshan: Function implementation can be simplified as below.

```go
return slices.Contains(strings.Fields(target), raw)
```

Can be implemented inline in the caller rather than introducing a separate function

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+	}
+	nodeResp.Status = common.NodeStatusComplete
+	nodeResp.Type = ""
+	return nodeResp, nil
+}
+
+// applyMeta builds the trimmed and synthetic-augmented meta for a verbose response. When
+// effectivePrompts is non-nil the meta is first reordered to match the auth class preference
+// (LOGIN_OPTIONS initial-request leg); otherwise the node's raw meta is used.
+func (n *promptNode) applyMeta(ctx *NodeContext, nodeResp *common.NodeResponse,
```
ThaminduDilshan:
```suggestion
func (n *promptNode) applyMetaForLoginOptions(ctx *NodeContext, nodeResp *common.NodeResponse,
```

backend/internal/flow/core/prompt_node.go
```diff
@@ -706,3 +836,175 @@ func collectMetaComponentMap(comps interface{}, compMap map[string]map[string]in
...
+			result = append(result, p)
+			continue
+		}
+		if _, gated := gatedActions[p.Action.Ref]; !gated {
+			result = append(result, p)
+		}
+	}
+
+	if len(result) == 0 {
+		return n.prompts // graceful fallback for misconfigured flows
```
ThaminduDilshan: This is not just a misconfiguration in flow side right? This can happen even if the runtime requested acrs are invalid.

If that so, shall we update the comments?

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+
+	// When an active filter narrows the chooser to exactly one option, auto-select it so
+	// the user proceeds directly to the credential prompt.
+	if len(actions) == 1 && len(requestedAuthClasses) > 0 {
+		ctx.CurrentAction = actions[0].Ref
+		logger.Debug("Auto-selected single login option", log.String("actionRef", ctx.CurrentAction))
+		if n.resolvePromptInputs(ctx, nodeResp) {
+			return n.finalizeLoginOptionsAction(ctx, nodeResp, authClassToAction, logger)
+		}
+		n.applyMeta(ctx, nodeResp, nil)
```
ThaminduDilshan: Should we set `RuntimeKeyAllowedLoginOptions` for this case as well? Login options are filtered down into a single prompt, but still inputs are not satisfied for the selected prompt

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+		nodeResp.Status = common.NodeStatusIncomplete
+		nodeResp.Type = common.NodeResponseTypeView
+		return nodeResp, nil
+	}
+
+	nodeResp.Actions = append(nodeResp.Actions, actions...)
```
ThaminduDilshan, line 218: This is the case where prompts are filtered out, but need user selection. A prompt can consists of multiple inputs and a action. 
Here we're setting the actions to the node response, but not the inputs. Shouldn't we set inputs too?

backend/internal/application/service.go
```diff
@@ -899,9 +901,36 @@ func validateOAuthParamsForCreateAndUpdate(app *model.ApplicationDTO) (*model.In
...
+	_, ok := mapping.AcrAMR[acr]
+	return ok
+}
+
+// validateAcrValues validates that every ACR value in the provided list is recognized by the global ACR registry.
+func validateAcrValues(acrValues []string) *serviceerror.ServiceError {
+	for _, acr := range acrValues {
+		if !isValidACR(acr) {
+			svcErr := ErrorInvalidAcrValues
+			svcErr.ErrorDescription = core.I18nMessage{
```
ThaminduDilshan: Shall we use `serviceerror.CustomServiceError()` function instead? I'm suspecting this could override the existing error constant

backend/internal/application/service_test.go
```diff
@@ -2647,5 +2647,129 @@ func (suite *ServiceTestSuite) TestValidateApplication_ErrorFromProcessInboundAu
 	assert.Equal(suite.T(), &ErrorInvalidInboundAuthConfig, svcErr)
 }
 
-// TestValidateApplication_ErrorFromValidateAuthFlowID tests error from validateAuthFlowID
-// when an invalid auth flow ID is provided.
+// ---------------------------------------------------------------------------
+// validateAcrValues
+// ---------------------------------------------------------------------------
+
+func TestValidateAcrValues_EmptyList(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/flow/mgt/graph_builder_test.go
```diff
@@ -671,6 +671,108 @@ func (s *GraphBuilderTestSuite) TestBuildGraph_WithMeta() {
...
+	mockGraph.EXPECT().SetStartNode("start").Return(nil)
+
+	graph, err := s.builder.buildGraph(flow)
+
+	s.NotNil(graph)
+	s.Nil(err)
+}
+
+func (s *GraphBuilderTestSuite) TestBuildGraph_AuthMethodMappingWithoutVariant_NoVariantSet() {
+	// Regression: presence of authMethodMapping must NOT auto-mark a node as LOGIN_OPTIONS.
```
ThaminduDilshan: These tests are not needed. Looks like it is added based on the previous review comments.
Shall we remove these unnecessary things?

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 	return nodeResp, nil
 }
 
+// executeLoginOptions handles the LOGIN_OPTIONS variant end-to-end: auth-class-driven filtering
+// and preference ordering on the initial prompt-out leg, allowedLoginOptions enforcement on the
+// subsequent leg, and selectedAuthClass recording on completion. Self-contained from the
+// generic prompt handling so the shared helpers don't need variant-aware parameters.
```
ThaminduDilshan:
```suggestion
// subsequent leg, and selectedAuthClass recording on completion.
```
Shall we remove unnecessary explanations?

backend/internal/flow/core/prompt_node.go
```diff
@@ -155,6 +168,112 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+	effectivePrompts := n.filterAndOrderPrompts(requestedAuthClasses, authClassToAction)
+	actions := make([]common.Action, 0)
+	for _, p := range effectivePrompts {
+		if p.Action != nil {
+			actions = append(actions, *p.Action)
+		}
+	}
+
+	// When an active filter narrows the chooser to exactly one option, auto-select it so
+	// the user proceeds directly to the credential prompt.
```
ThaminduDilshan:
```suggestion
	// the user proceeds directly to the next node.
```

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -284,6 +285,18 @@ func (b *graphBuilder) configureNodeInputs(nodeDef *NodeDefinition, node core.No
 	executorNode.SetInputs(inputs)
 }
 
+// configureNodeVariant configures the variant for a prompt node by reading the explicit
+// Variant field from the node definition. The graph builder is a pure JSON-to-graph translator
+// and does not infer the variant from other properties.
```
ThaminduDilshan:
```suggestion
// Variant field from the node definition.
```
Shall we remove unnecessary comments came from development time?

backend/internal/flow/mgt/model.go
```diff
@@ -143,6 +143,7 @@ type NodeDefinition struct {
 	Layout       *NodeLayout            `json:"layout,omitempty" yaml:"layout,omitempty" jsonschema:"Optional UI layout information for flow composer (position and size on canvas)"`
 	Meta         interface{}            `json:"meta,omitempty" yaml:"meta,omitempty" jsonschema:"Optional metadata. For PROMPT nodes, must include 'components' array for UI rendering. See existing flows for examples."`
 	Prompts      []PromptDefinition     `json:"prompts,omitempty" yaml:"prompts,omitempty" jsonschema:"For PROMPT nodes: defines user inputs and actions. Each prompt has inputs (form fields) and an action (what happens on submit)."`
+	Variant      string                 `json:"variant,omitempty" yaml:"variant,omitempty" jsonschema:"Optional PROMPT node variant. Use 'LOGIN_OPTIONS' to enable ACR-based filtering on this node."`
```
ThaminduDilshan:
```suggestion
	Variant      string                 `json:"variant,omitempty" yaml:"variant,omitempty" jsonschema:"Optional PROMPT node variant. Use 'LOGIN_OPTIONS' to enable login option filtering on this node."`
```

backend/internal/oauth/oauth2/discovery/service.go
```diff
@@ -158,6 +160,16 @@ func (ds *discoveryService) getSupportedSubjectTypes() []string {
 	return constants.GetSupportedSubjectTypes()
 }
 
+func (ds *discoveryService) getSupportedAcrValues() []string {
```
ThaminduDilshan, line 164: Shall we add a method comment?

backend/internal/system/config/config.go
```diff
@@ -576,6 +577,54 @@ func (c *TrustedIssuerConfig) Validate() error {
 	}
 }
 
+// AMRFactor represents a single authentication method reference factor entry.
+type AMRFactor struct {
+	Type string `yaml:"type" json:"type"`
```
ThaminduDilshan: What's the usage of this type? Do we need this?
As far as I checked, this isn't used anywhere except for some unit tests.

If not needed, we can simply use a string map or array for supported amr values within `AuthClassConfig`

tests/integration/flow/authentication/acr_options_flow_test.go
```diff
@@ -0,0 +1,553 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package authentication
+
+// ACR Options Flow Integration Tests
+//
+// Covers acceptance criteria from feature-docs/acr-values.md:
```
ThaminduDilshan: Shall we remove these comments? These are not related to the project implementation right?
If needed, we can generalize and keep them. But `AC-xx` doesn't have any meaning

tests/integration/flow/authentication/acr_options_flow_test.go
```diff
@@ -0,0 +1,553 @@
...
+	Name:     "ACR Options Auth Flow",
+	FlowType: "AUTHENTICATION",
+	Handle:   "auth_flow_acr_options_test",
+	Nodes: []map[string]interface{}{
+		{
+			"id":        "start",
+			"type":      "START",
+			"onSuccess": "acr_chooser",
+		},
+		// login_options prompt node — the ACR chooser (AC-10)
```
ThaminduDilshan:
```suggestion
		// login_options prompt node — the ACR chooser
```

Check other places too

tests/integration/flow/authentication/acr_options_flow_test.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+	}
+	if acrOptionsTestOUID != "" {
+		if err := testutils.DeleteOrganizationUnit(acrOptionsTestOUID); err != nil {
+			ts.T().Logf("failed to delete OU: %v", err)
+		}
+	}
+}
+
+// ------- AC-10: login_options node is valid in a flow graph ----------
```
ThaminduDilshan:
```suggestion
// ------- Test login_options node is valid in a flow graph ----------
```

tests/integration/flow/authentication/acr_options_flow_test.go
```diff
@@ -0,0 +1,553 @@
...
+func (ts *AcrOptionsFlowTestSuite) TestAcrOptions_NodeValidInFlowGraph() {
+	// Initiate flow without any acr_values — engine must present both ACR options.
+	flowStep, err := common.InitiateAuthenticationFlow(acrOptionsTestAppID, false, nil, "")
+	ts.Require().NoError(err, "flow initiation should succeed")
+
+	ts.Require().Equal("INCOMPLETE", flowStep.FlowStatus)
+	ts.Require().Equal("VIEW", flowStep.Type)
+	ts.Require().NotEmpty(flowStep.ExecutionID)
+
+	// Both actions must be present (AC-10).
```
ThaminduDilshan:
```suggestion
	// Both actions must be present
```

Check all places

backend/internal/oauth/oauth2/authz/requestvalidator/acr_test.go
```diff
@@ -0,0 +1,109 @@
...
+package requestvalidator
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseACRValues_SingleACR(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/oauth/oauth2/authz/requestvalidator/acr.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+// ResolveACRValues resolves the effective acr_values for an authorization request.
+//
+// When requestedAcrValues is present and appAcrValues is non-empty, requested ACRs are filtered
+// to only those included in appAcrValues. Requested ACRs not in the app's list are
+// silently ignored. If filtering removes all ACRs (or no ACRs were requested), the full
+// appAcrValues list is returned. The result is a space-separated, order-preserving,
+// deduplicated string suitable for forwarding to the authentication flow engine.
+func ResolveACRValues(requestedAcrValues string, appAcrValues []string) string {
```
thiva-k: Shall we move this to the existing validator.go instead and remove this file?

backend/internal/flow/core/prompt_node.go
sacrana0: LOGIN_OPTIONS node will not have inputs, should we consider that case? @ThaminduDilshan

ThaminduDilshan: Can there be a LOGIN_OPTIONS node that has Basic auth, Google and Github logins shown in the same page? Basic auth prompt will have username and password inputs. Other two will only have actions. No inputs.

So for the first node, expected response is to return username, password inputs and 3 different actions (username/password inputs are only related to basic auth related action). With your change, flow execute API will just request the actions, not the inputs taken in the same page.

If this is not a requirement for LOGIN_OPTIONS, fine to skip for now

sacrana0: This is not our requirement currently. Will skip for now.

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #2217, opened 2026-04-06
Add SMS template support for OTP and flow-based SMS sending

Dilusha-Madushan:
### Purpose

SMS message content was hardcoded in two places — smsExecutor used a static constant and otpService.sendSMSOTP used an inline fmt.Sprintf. This made the message content impossible to customize without a code change and inconsistent with how email already works (full template system).

This PR wires SMS sending into the existing template service, replacing both hardcoded bodies with rendered templates. Three SMS scenarios are introduced — OTP verification, self-registration invite, and user invitation — each backed by a YAML template file following the same declarative resource pattern as email templates.

### Approach

Two rendering sites, different reasons:

otpService.sendSMSOTP renders at the service level because the OTP value is generated and owned by otpService. The executor above it never sees the OTP, so rendering must happen where the value is in scope.
smsExecutor renders at the executor level, matching the emailExecutor pattern — the executor reads the scenario from node properties, builds TemplateData, and calls templateService.Render() before passing the body to the sender.
Template model changes:

Added TemplateTypeSMS constant to model.go — used by the validator to exempt SMS templates from the subject required check (SMS is text-only, no subject concept).
Added three new ScenarioType constants (SMS_OTP, SELF_REGISTRATION_SMS, USER_INVITE_SMS). Separate scenario keys from their email counterparts avoid key collision in the store's lookup map.
Initialization ordering fix:

template.Initialize() was called after notification.Initialize() in servicemanager.go, making it impossible to pass templateService into otpService. No circular dependency exists — this was a pure ordering issue. Fixed by moving template.Initialize() earlier and threading templateService into notification.Initialize() → newOTPService().

160-character warning:

A warning log is emitted (not an error) when a rendered body exceeds 160 characters, since longer messages incur multi-segment SMS billing. The flow continues normally.

### Related Issues
- https://github.com/asgardeo/thunder/issues/1928

## Summary by CodeRabbit

* **New Features**
  * Added SMS template support for OTP verification, user invites, and self-registration.
  * Templates now support SMS type alongside email and can be authored without a subject for SMS.
  * SMS notifications are rendered from templates with context-specific data (OTP, expiry, app name, invite link) instead of fixed default text.

### DIFF
--- a/backend/internal/flow/executor/sms_executor.go
+++ b/backend/internal/flow/executor/sms_executor.go
@@ -29,22 +29,25 @@ import (
 	notifcm "github.com/asgardeo/thunder/internal/notification/common"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/log"
+	"github.com/asgardeo/thunder/internal/system/template"
 )
 
 // phoneNumberRegex matches phone numbers in various formats including optional +, digits, spaces, dashes,
 // dots, and parentheses with a total length of 7 to 20 characters.
 var phoneNumberRegex = regexp.MustCompile(`^\+?[0-9\s\-().]{7,20}$`)
 
-// smsExecutor sends an SMS message using the configured sender from node properties and a default message body.
+// smsExecutor sends an SMS message using the configured sender from node properties and a template-based body.
 type smsExecutor struct {
 	core.ExecutorInterface
-	logger         *log.Logger
-	notifSenderSvc notification.NotificationSenderServiceInterface
+	logger          *log.Logger
+	notifSenderSvc  notification.NotificationSenderServiceInterface
+	templateService template.TemplateServiceInterface
 }
 
 // newSMSExecutor creates a new instance of smsExecutor.
 func newSMSExecutor(flowFactory core.FlowFactoryInterface,
-	notifSenderSvc notification.NotificationSenderServiceInterface) *smsExecutor {
+	notifSenderSvc notification.NotificationSenderServiceInterface,
+	templateService template.TemplateServiceInterface) *smsExecutor {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "SMSExecutor"))
 	base := flowFactory.CreateExecutor(
 		ExecutorNameSMSExecutor,
@@ -58,11 +61,12 @@ func newSMSExecutor(flowFactory core.FlowFactoryInterface,
 		ExecutorInterface: base,
 		logger:            logger,
 		notifSenderSvc:    notifSenderSvc,
+		templateService:   templateService,
 	}
 }
 
 // Execute resolves the recipient from user inputs or runtime data and the sender ID from node properties,
-// then sends the SMS with a default message body.
+// then renders the SMS body from a template and sends it.
 func (e *smsExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse, error) {
 	logger := e.logger.With(log.String(log.LoggerKeyFlowID, ctx.FlowID))
 	logger.Debug("Executing SMS executor")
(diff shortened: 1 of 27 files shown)


backend/internal/flow/executor/sms_executor.go
```diff
@@ -104,16 +108,46 @@ func (e *smsExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse,
...
+	}
+
+	rendered, svcErr := e.templateService.Render(ctx.Context, scenario, templateData)
 	if svcErr != nil {
-		if svcErr.Type == serviceerror.ClientErrorType {
+		return nil, fmt.Errorf("failed to render SMS template: %s", svcErr.Code)
+	}
+
+	if len(rendered.Body) > 160 {
+		logger.Warn("Rendered SMS body exceeds 160 characters; message may be split into multiple segments",
```
ThaminduDilshan: Can we log this from the template service so that we don't have to repeat this check in all the places we use sms templates?

backend/internal/system/template/model.go
```diff
@@ -21,6 +21,11 @@ package template
 // TemplateType represents the type of template (e.g., email).
 type TemplateType string
 
+const (
+	// TemplateTypeSMS represents an SMS template.
+	TemplateTypeSMS TemplateType = "sms"
```
ThaminduDilshan, line 28: Not related to this PR, but don't we have a type for email templates?

backend/internal/system/template/model.go
```diff
@@ -29,12 +34,21 @@ const (
 	ScenarioUserInvite ScenarioType = "USER_INVITE"
 	// ScenarioSelfRegistration represents the self-registration via invite link scenario.
 	ScenarioSelfRegistration ScenarioType = "SELF_REGISTRATION"
+	// ScenarioSMSOTP represents the SMS OTP verification scenario.
+	ScenarioSMSOTP ScenarioType = "SMS_OTP"
```
ThaminduDilshan: Shall we define scenario types delivery method agnostic? I.e. we can simply have `ScenarioOTP`. Anyway since we have a template type, we can distinguish email, sms otp templates

backend/internal/system/template/model.go
Dilusha-Madushan: there's no TemplateTypeEmail constant because it's never needed in code logic. TemplateTypeSMS exists specifically because the code needs to branch on it.

backend/internal/system/template/model.go
Dilusha-Madushan: Addressed

backend/internal/system/template/model.go
Dilusha-Madushan: Added template type email constant

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: Move this into template service

backend/internal/flow/executor/sms_executor.go
```diff
@@ -104,16 +108,46 @@ func (e *smsExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse,
...
+	} else {
+		scenario = template.ScenarioSMSOTP
+	}
+
+	templateData := template.TemplateData{
+		"appName":    ctx.Application.Name,
+		"inviteLink": ctx.RuntimeData[common.RuntimeKeyInviteLink],
+	}
+
+	rendered, svcErr := e.templateService.Render(ctx.Context, scenario, templateData)
```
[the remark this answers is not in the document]
Dilusha-Madushan: Addressed, now not use any default and throw flow error

backend/internal/notification/otp_service.go
```diff
@@ -248,21 +252,26 @@ func (s *otpService) getOTPValidityPeriodInMillis() int64 {
...
-	_client, svcErr := s.clientProvider.GetClient(sender)
+func (s *otpService) sendSMSOTP(ctx context.Context, recipient, otp string,
+	sender common.NotificationSenderDTO, logger *log.Logger) *serviceerror.ServiceError {
+	expiryMinutes := strconv.FormatInt(s.getOTPValidityPeriodInMillis()/60000, 10)
+	templateData := template.TemplateData{"otp": otp, "expiryMinutes": expiryMinutes}
+	rendered, svcErr := s.templateService.Render(ctx, template.ScenarioOTP, template.TemplateTypeSMS, templateData)
 	if svcErr != nil {
-		return svcErr
+		logger.Error("Failed to render SMS OTP template", log.String("error", svcErr.Code))
+		return &ErrorInternalServerError
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai
The current approach is intentional and consistent with the existing patterns in this file. The handler's handleError method does not log errors — it only writes the HTTP response. If the service omits logging here, the failure would be silently dropped with no trace in the logs.

The same pattern is used throughout otp_service.go:

Line 94: logger.Error("Failed to generate OTP", ...) → return nil, &ErrorInternalServerError
Line 119: logger.Error("Failed to create session token", ...) → return nil, &ErrorInternalServerError
Line 277: logger.Error("Failed to send SMS OTP", ...) → return &ErrorInternalServerError
Logging before returning ErrorInternalServerError in the service layer is the established pattern here, not a violation of it. The learned rule referenced may not accurately reflect how this codebase handles 500 paths.

PR #2270, opened 2026-04-09
Secure Thunder with an external identity provider

RushanNanayakkara:
### Purpose
Secure Thunder with an external identity provider. 
Thunder backend will work with a token from a trusted issuer. Thunder console will get the capability to have an external IDP linked to it for authentication. Then the console will use the token from this IDP to access the backend. 

### Related Issue 
- https://github.com/asgardeo/thunder/issues/2257


backend/internal/system/config/config.go
```diff
@@ -406,6 +406,32 @@ type ConsentConfig struct {
...
+//
+// Per RFC 9068 §2.2 and RFC 8707, the Audience field must be set to this server's own
+// identifier (typically its public URL). The frontend must include a matching "resource"
+// parameter in the authorization request so the auth server sets the token's "aud" claim
+// to this server's identifier.
+//
+// RequiredClaims enforces that incoming tokens contain specific claims with expected values.
+// Each entry specifies a claim name and the value it must hold. If any required claim is
+// missing or does not match, the token is rejected.
+type TrustedIssuerConfig struct {
```
rajithacharith, line 453: Can we support an array of trusted issuers? Can you check whether there is an drawback on doing so.

rajithacharith: And do we need enabled attribute. if the configurations are provided, we can say that's the enabled state. WDYT?

backend/internal/system/jose/jwt/service.go
```diff
@@ -348,12 +358,53 @@ func (js *jwtService) VerifyJWTSignatureWithJWKS(jwtToken string, jwksURL string
...
+	// Verify JWT signature
+	if err := js.VerifyJWTSignatureWithPublicKey(jwtToken, pubKey); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// getJWKSKeys returns JWKS keys for the given URL, using a TTL-based cache.
+func (js *jwtService) getJWKSKeys(jwksURL string) ([]map[string]interface{}, *serviceerror.ServiceError) {
```
rajithacharith, line 393: Shall we return an array of defined structs from here 
```
func (js *jwtService) getJWKSKeys(jwksURL string) ([]map[string]JWK{},  // name might differ *serviceerror.ServiceError) {
```

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -59,9 +60,11 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
 		return nil, errInvalidToken
 	}
 
-	// Step 2: Verify JWT signature
+	// Step 2: Verify JWT signature (local key first, then federated JWKS fallback)
 	if err := h.jwtService.VerifyJWTSignature(token); err != nil {
```
rajithacharith: Seems logic inside this is misleading with the current method names. 
Do a jwtSignatureVerification
If fails -> Do a federatedTokenVerification

Can we abstract out the logic into something like this

authenticator will provide the token to jwtService for verification. 
It will do the verification depending on the server configs. 
Authenticator does not need to know about internal details

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -201,7 +201,11 @@ func getValidIssuers(oauthApp *appmodel.OAuthAppConfigProcessedDTO) map[string]b
 	tokenConfig := ResolveTokenConfig(oauthApp, TokenTypeAccess)
 	validIssuers[tokenConfig.Issuer] = true
 
-	// TODO: Add support for external issuers
+	trustedIssuer := config.GetThunderRuntime().Config.TrustedIssuer
```
thiva-k: We may need multiple trusted issuers support for token exchange case
https://github.com/asgardeo/thunder/discussions/2258#discussioncomment-16498995

backend/internal/system/jose/jwt/service.go
```diff
@@ -348,12 +358,53 @@ func (js *jwtService) VerifyJWTSignatureWithJWKS(jwtToken string, jwksURL string
...
+	if cached, ok := js.jwksCache.Load(jwksURL); ok {
+		entry := cached.(*jwksCacheEntry)
+		if time.Now().Before(entry.expiresAt) {
+			return entry.keys, nil
+		}
+	}
+
 	client := httpservice.NewHTTPClientWithTimeout(10 * time.Second)
 	resp, err := client.Get(jwksURL)
```
thiva-k, line 402: Are we doing any validations to enforce this is HTTPS?

backend/internal/system/jose/jwt/service.go
```diff
@@ -55,13 +56,22 @@ type JWTServiceInterface interface {
 	VerifyJWTSignatureWithJWKS(jwtToken string, jwksURL string) *serviceerror.ServiceError
 }
 
+// jwksCacheEntry holds a cached JWKS response with its expiry time.
+type jwksCacheEntry struct {
+	keys      []map[string]interface{}
+	expiresAt time.Time
+}
+
+const jwksCacheTTL = 5 * time.Minute
```
thiva-k: Can't this be a config instead?

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -85,6 +88,53 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
...
 }
 
+// verifyFederatedToken checks if the token is from a trusted external issuer and verifies it via JWKS.
+// Per RFC 9068 §2.2 and RFC 8707, this validates:
+//   - iss: matches the configured trusted issuer
+//   - aud: matches this server's own identifier (the resource server)
+//   - signature: verified via the auth server's JWKS endpoint
+//   - required_claims: each configured claim must match the expected value
+func (h *jwtAuthenticator) verifyFederatedToken(token string) (verified bool) {
+	defer func() {
```
thiva-k: Shouldn't we return/log the appropriate error here instead?

backend/internal/system/config/config.go
RushanNanayakkara: > Can we support an array of trusted issuers? Can you check whether there is an drawback on doing so.

This was discussed and decided that single trusgted issuer is enought for this use case. However we can improve this in the future if needed. 
https://github.com/asgardeo/thunder/discussions/2258#discussioncomment-16488325



> And do we need enabled attribute. if the configurations are provided, we can say that's the enabled state. WDYT?

Having an enable configuration helps with code readabililty and feature disable/enable actioins to be handled. However we can remove it if needed.

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -179,6 +179,17 @@ func (tv *tokenValidator) ValidateSubjectToken(
 		return nil, fmt.Errorf("invalid subject token signature: %w", err)
 	}
 
+	// Validate required claims for trusted issuer tokens.
```
thiva-k: Is this needed for token exchange issuer validation? The protocol specific validations should be sufficient right? IMO token exchange should not require to validate the management api specific validations

backend/internal/oauth/oauth2/tokenservice/utils.go
RushanNanayakkara: Discussed and decided otherwise at https://github.com/asgardeo/thunder/discussions/2258#discussioncomment-16488325. 

Additionally this is not token exchange.

backend/internal/system/jose/jwt/service.go
RushanNanayakkara: Not clear about this comment. Can you please elaborate?

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: error scenario is handled for invlaid tokens above. Will add some logs to identify specific error scenarios.

backend/internal/oauth/oauth2/tokenservice/validator.go
RushanNanayakkara: This feature doesn't use token exchange. 

Claim validation is required to enforce limitations on which users on the auth server could access the resources.

backend/internal/system/jose/jwt/service.go
thiva-k: Shouldn't JWKS URI should be enforced to be HTTPS protocol?

backend/internal/oauth/oauth2/tokenservice/utils.go
thiva-k: This is called only in the token exchange path right?

backend/internal/oauth/oauth2/tokenservice/validator.go
thiva-k: This subject token validation is only token exchange. Is this change needed if token excahnge is  out of scope of the feature?

backend/internal/oauth/oauth2/tokenservice/validator_test.go
```diff
@@ -526,23 +529,206 @@ func (suite *TokenValidatorTestSuite) TestFederationScenario_OnlyThunderIssuerIs
...
+	}
+	token := suite.createTestJWT(claims)
+
+	suite.mockJWTService.On("VerifyJWTSignatureWithJWKS", token, jwksURL).Return(nil)
+
+	result, err := suite.validator.ValidateSubjectToken(token, suite.oauthApp)
+	assert.NoError(suite.T(), err)
+	assert.NotNil(suite.T(), result)
+	suite.mockJWTService.AssertExpectations(suite.T())
 }
```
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

backend/internal/system/config/config.go
```diff
@@ -435,6 +461,7 @@ type Config struct {
 	Layout               LayoutConfig           `yaml:"layout" json:"layout"`
 	Email                EmailConfig            `yaml:"email" json:"email"`
 	Consent              ConsentConfig          `yaml:"consent" json:"consent"`
+	TrustedIssuer        TrustedIssuerConfig    `yaml:"trusted_issuer" json:"trusted_issuer"`
```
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

install/helm/values.yaml
```diff
@@ -326,6 +336,19 @@ configuration:
...
+  # and validates them via its JWKS endpoint. Add the CP origin to cors.allowedOrigins
+  # when using this feature.
+  # trustedIssuer:
+  #   enabled: true
+  #   issuer: "https://cp.example.com"
+  #   jwksUrl: "https://cp.example.com/oauth2/jwks"
+  #   audience: "FEDERATED_CONSOLE"
+  #   requiredClaims:
+  #     - claim: "ouId"
+  #       value: "<tenant-ou-id>"
```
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

backend/internal/system/config/config.go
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

install/helm/values.yaml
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

backend/internal/oauth/oauth2/tokenservice/validator_test.go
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/331a3bcb163cc743082328cefcf12e772023d6b6

backend/internal/oauth/oauth2/tokenservice/validator.go
RushanNanayakkara: +1. Not needed. Removed with https://github.com/asgardeo/thunder/pull/2270/commits/53ccd3044482dfe470cffe8bc2512c97e59e0077

backend/internal/oauth/oauth2/tokenservice/utils.go
RushanNanayakkara: +1. Not needed. Removed with https://github.com/asgardeo/thunder/pull/2270/commits/53ccd3044482dfe470cffe8bc2512c97e59e0077

backend/internal/system/config/config.go
```diff
@@ -406,6 +406,32 @@ type ConsentConfig struct {
...
+// Per RFC 9068 §2.2 and RFC 8707, the Audience field must be set to this server's own
+// identifier (typically its public URL). The frontend must include a matching "resource"
+// parameter in the authorization request so the auth server sets the token's "aud" claim
+// to this server's identifier.
+//
+// RequiredClaims enforces that incoming tokens contain specific claims with expected values.
+// Each entry specifies a claim name and the value it must hold. If any required claim is
+// missing or does not match, the token is rejected.
+type TrustedIssuerConfig struct {
+	Enabled        bool            `yaml:"enabled" json:"enabled"`
```
senthalan: Why we need this enabled config?

If this is not configured, we can fallback to Thunder as the authorization provider. If configured, we can use this

backend/internal/system/config/config.go
senthalan: shall we have this part of the user `server.security.trusted_issuer`

It doesn't make sense to be used as first class configuration

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -59,9 +60,11 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
 		return nil, errInvalidToken
 	}
 
-	// Step 2: Verify JWT signature
+	// Step 2: Verify JWT signature (local key first, then federated JWKS fallback)
```
senthalan: There shouldn't be a fallback here. 

If the server.security.trusted_issuer is not configured, then we should verify using local keys
Else we should try to verify with configured trusted issuer

backend/internal/system/security/jwt_authenticator.go
senthalan: I don't understand why we need to have a `defer func` here and what is the `r` here ?

docs/content/guides/guides/trusted-issuer.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+- You run a multi-tenant deployment with one central Thunder acting as the identity provider and other Thunder instances acting as resource servers.
+- You want tenant instances to delegate login to the central Thunder rather than duplicate user directories.
+- Bearer tokens arriving at the tenant instance are signed by the central Thunder and carry an `iss` claim that identifies it.
+
+If a Thunder instance only serves its own users and issues its own tokens, leave `trusted_issuer.enabled` as `false`.
+
+## Configuration
+
+Add a `trusted_issuer` block to the tenant instance's `deployment.yaml`:
```
senthalan: Shall we add the needed configuration for the console portal as well ?

backend/internal/system/config/config.go
RushanNanayakkara: Added for better readablility and easy disabling of the feature if needed. The same concern was added above as well. Will remove.

docs/content/guides/guides/trusted-issuer.mdx
RushanNanayakkara: `I don't understand why we need to have a defer func here and what is the r here ?`
This is to handle panic scenarios. As discussed offline will remove from here.

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -201,7 +201,6 @@ func getValidIssuers(oauthApp *appmodel.OAuthAppConfigProcessedDTO) map[string]b
 	tokenConfig := ResolveTokenConfig(oauthApp, TokenTypeAccess)
 	validIssuers[tokenConfig.Issuer] = true
 
-	// TODO: Add support for external issuers
```
thiva-k, line 204: Shall we add this TODO back?

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -248,7 +248,6 @@ func (tv *tokenValidator) verifyTokenSignatureByIssuer(
 		return nil
 	}
 
-	// TODO: Implement JWKS-based verification for external federated issuers
```
thiva-k, line 251: Shall we add this TODO back?

RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/oauth/oauth2/tokenservice/utils.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

docs/content/guides/guides/trusted-issuer.mdx
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/system/config/config.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: removed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/system/jose/jwt/service.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/system/config/config.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/5373125a8b927045f9c68005c3f815f9d5135a93

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/d519a4d29233f1a36a81714241c3cc36d9b47830

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: Changed to logic to only use one trusted issuer as discussed offline with https://github.com/asgardeo/thunder/pull/2270/commits/d519a4d29233f1a36a81714241c3cc36d9b47830

backend/internal/system/jose/jwt/service.go
RushanNanayakkara: added with https://github.com/asgardeo/thunder/pull/2270/commits/cf58c55659b58ca03f323c84c3603c7b9d361273

backend/internal/system/security/jwt_authenticator.go
rajithacharith: Search a little bit on this 

### Panic Recovery in Go REST APIs

Use `recover()` at application boundaries (middleware), not inside business logic.

---

#### Middleware Example

```go
func RecoveryMiddleware(next http.Handler) http.Handler {
    return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        defer func() {
            if err := recover(); err != nil {
                log.Println("panic recovered:", err)
                http.Error(w, "Internal Server Error", http.StatusInternalServerError)
            }
        }()
        next.ServeHTTP(w, r)
    })
}
```

#### Usage

```go
mux := http.NewServeMux()
mux.Handle("/api", RecoveryMiddleware(myHandler))
```

---

#### Notes

* `recover()` in `main()` will not catch panics from HTTP handlers
* Do not use `recover()` inside regular functions
* Use it in middleware and in manually spawned goroutines only
* Use `error` for expected failures, not `panic`

---

backend/internal/system/config/config.go
```diff
@@ -406,6 +407,81 @@ type ConsentConfig struct {
...
+type TrustedIssuerConfig struct {
+	Issuer         string          `yaml:"issuer" json:"issuer"`
+	JWKSURL        string          `yaml:"jwks_url" json:"jwks_url"`
+	Audience       string          `yaml:"audience" json:"audience"`
+	JWKSCacheTTL   int             `yaml:"jwks_cache_ttl" json:"jwks_cache_ttl"` // Cache TTL in seconds. Default: 300
+	RequiredClaims []RequiredClaim `yaml:"required_claims" json:"required_claims"`
+}
+
+// DefaultJWKSCacheTTLSeconds is the default JWKS cache TTL when not configured.
+const DefaultJWKSCacheTTLSeconds = 300
```
rajithacharith: default values can be defined using https://github.com/asgardeo/thunder/blob/main/backend/cmd/server/repository/resources/conf/default.json

docs/static/api/next/combined.yaml
rajithacharith, line 1: Shall we remove this file changes. This should come in a separate PR
@thiva-k I think generated file  from docs build was not added

RushanNanayakkara: Removed changes with https://github.com/asgardeo/thunder/pull/2270/commits/18c9a20794cefe2a7d40c836ac8bbfff4e98adfb

backend/internal/system/config/config.go
RushanNanayakkara: Addressed with https://github.com/asgardeo/thunder/pull/2270/commits/18c9a20794cefe2a7d40c836ac8bbfff4e98adfb

rajithacharith (dismissed):

rajithacharith (approved):

PR #2286, opened 2026-04-09
Add localized client metadata support for applications and DCR

### LINKED ISSUE #2106: Localised Client metadata
### Current Limitation

Client name and other human-readable client metadata are only displayed in English. Client metadata should change based on the user-selected `ui_locale`

### Suggested Improvement

Support language-tagged client metadata following the OIDC Dynamic Client Registration 1.0 convention. Refer https://openid.net/specs/openid-connect-registration-1_0.html#LanguagesAndScripts

Proposing the support in both in the application and DCR endpoints. Detailed design is documented here - https://github.com/anushasunkada/thunder/blob/main/feature-docs/localised-client-metadata.md

### Version

_No response_

nandhu-kumar:
fixes #2106

### Purpose

OAuth 2.0 Dynamic Client Registration (RFC 7591) requires clients to register localized metadata
variants using OIDC language-tagged fields (e.g. `"client_name#fr": "Mon Client"`). Previously,
Thunder's DCR endpoint ignored these fields entirely, meaning clients could not register
language-specific display names, logo URIs, terms-of-service, or policy URIs.

This PR adds full support for OIDC language-tagged DCR metadata fields.

### Approach

**Parsing language-tagged fields (RFC 7591 §2.1)**

`DCRRegistrationRequest.UnmarshalJSON` now scans all top-level JSON keys for the `field#lang-tag`
pattern. Valid BCP 47 tags (normalised via `golang.org/x/text/language`) are extracted into four
typed maps: `LocalizedClientName`, `LocalizedLogoURI`, `LocalizedTosURI`, `LocalizedPolicyURI`.
Invalid tags are rejected immediately with a 400 error before any storage occurs.

`DCRRegistrationResponse.MarshalJSON` serializes those maps back as top-level `#`-keyed keys so
clients receive the echoed variants in the 201 response.

**Persisting to i18n store (transaction limitation + compensation)**

The DCR outer transactioner operates on the runtime DB, while both the application store and i18n
store use configDB. Because of this mismatch, they cannot participate in the same transaction.

Localized variants are written to the i18n store *after* the application record commits. On write
failure, a compensation path cleans up any partial i18n rows and deletes the newly-created
application, maintaining consistency without distributed transactions.

Each app's translations are stored under the namespace `app.<id>` with keys `client_name`,
`logo_uri`, `tos_uri`, `policy_uri`.

**i18n integration**

Added `SetTranslationOverridesForNamespace` to perform bulk writes of localized variants during DCR
registration, reducing multiple DB calls into a single operation.

A new `GetTranslationsByNamespace` method on `I18nServiceInterface` fetches all locale rows for a
namespace in a single query, avoiding the N+1 pattern. This is used to support localized metadata
handling within the DCR flow.

**Cleanup on application deletion**

The existing `deleteLocalizedVariants` call in the application service lifecycle ensures i18n rows
are cleaned up when an application is deleted.

### Related Issues
- Fixes #2106

### Related PRs
- N/A

### Checklist
- [x] Tests provided.
    - [x] Unit Tests — `backend/internal/oauth/oauth2/dcr/service_test.go`: covering
      `RegisterClient` with localized variants, i18n write failure compensation, and related scenarios
    - [x] Integration Tests — `tests/integration/oauth/dcr/dcr_test.go`: register with localized
      fields (assert echoed in response), invalid BCP 47 tag rejection (400)

### Security checks
- [x] Confirmed that this PR doesn't commit any secrets.

### DIFF
--- a/backend/cmd/server/servicemanager.go
+++ b/backend/cmd/server/servicemanager.go
@@ -282,7 +282,7 @@ func registerServices(mux *http.ServeMux) jwt.JWTServiceInterface {
 
 	// TODO: Remove entityService dependency after finalizing declarative resource loading pattern
 	applicationService, applicationExporter, err := application.Initialize(
-		mux, mcpServer, entityProvider, entityService, inboundClientService, ouService)
+		mux, mcpServer, entityProvider, entityService, inboundClientService, ouService, i18nService)
 	if err != nil {
 		logger.Fatal("Failed to initialize ApplicationService", log.Error(err))
 	}
@@ -326,7 +326,7 @@ func registerServices(mux *http.ServeMux) jwt.JWTServiceInterface {
 	// Initialize OAuth services.
 	err = oauth.Initialize(mux, applicationService, inboundClientService, authnProvider, jwtService, jweService,
 		flowExecService, observabilitySvc, pkiService, ouService, attributeCacheService, authZService, entityProvider,
-		resourceService)
+		resourceService, i18nService)
 	if err != nil {
 		logger.Fatal("Failed to initialize OAuth services", log.Error(err))
 	}
--- a/backend/internal/application/service.go
+++ b/backend/internal/application/service.go
@@ -36,6 +36,7 @@ import (
 	oupkg "github.com/asgardeo/thunder/internal/ou"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/i18n/core"
+	i18nmgt "github.com/asgardeo/thunder/internal/system/i18n/mgt"
 	"github.com/asgardeo/thunder/internal/system/log"
 	sysutils "github.com/asgardeo/thunder/internal/system/utils"
 )
@@ -62,6 +63,7 @@ type applicationService struct {
 	inboundClientService inboundclient.InboundClientServiceInterface
 	entityProvider       entityprovider.EntityProviderInterface
 	ouService            oupkg.OrganizationUnitServiceInterface
+	i18nService          i18nmgt.I18nServiceInterface
 }
 
 // newApplicationService creates a new instance of ApplicationService.
@@ -69,12 +71,14 @@ func newApplicationService(
 	inboundClient inboundclient.InboundClientServiceInterface,
 	entityProvider entityprovider.EntityProviderInterface,
 	ouService oupkg.OrganizationUnitServiceInterface,
+	i18nService i18nmgt.I18nServiceInterface,
 ) ApplicationServiceInterface {
 	return &applicationService{
 		logger:               log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ApplicationService")),
 		inboundClientService: inboundClient,
 		entityProvider:       entityProvider,
 		ouService:            ouService,
+		i18nService:          i18nService,
 	}
 }
 
@@ -325,7 +329,7 @@ func (as *applicationService) UpdateApplication(ctx context.Context, appID strin
 	if as.inboundClientService.IsDeclarative(ctx, appID) {
 		return nil, &ErrorCannotModifyDeclarativeResource
 	}
-	_, inboundAuthConfig, svcErr := as.validateApplicationForUpdate(ctx, appID, app)
+	existingApp, inboundAuthConfig, svcErr := as.validateApplicationForUpdate(ctx, appID, app)
 
 	if svcErr != nil {
 		return nil, svcErr
@@ -359,6 +363,10 @@ func (as *applicationService) UpdateApplication(ctx context.Context, appID strin
 	}
 
 	if svcErr := as.updateEntityDataForApplicationUpdate(appID, app, inboundAuthConfig); svcErr != nil {
+		return nil, svcErr
+	}
+
+	if svcErr := as.cleanupStaleI18nKeys(ctx, appID, existingApp, app); svcErr != nil {
 		return nil, svcErr
 	}
 
@@ -463,7 +471,7 @@ func (as *applicationService) DeleteApplication(ctx context.Context, appID strin
 		return &serviceerror.InternalServerError
 	}
 
-	return nil
+	return as.deleteLocalizedVariants(ctx, appID)
 }
 
 // isIdentifierTaken checks if an entity with the given identifier already exists.
@@ -1518,4 +1526,62 @@ func (as *applicationService) mapStoreError(err error) *serviceerror.ServiceErro
 	}
 	as.logger.Error("Failed to retrieve application", log.Error(err))
 	return &serviceerror.InternalServerError
+}
+
+// deleteLocalizedVariants removes all i18n translations for an application's fields.
+// All fields are attempted; returns an internal server error if any deletion fails.
+func (as *applicationService) deleteLocalizedVariants(ctx context.Context, appID string) *serviceerror.ServiceError {
+	if as.i18nService == nil {
+		return nil
+	}
+	var hasErr bool
+	for _, field := range []string{"name", "logo_uri", "tos_uri", "policy_uri"} {
+		if svcErr := as.i18nService.DeleteTranslationsByKey(
+			ctx, AppI18nNamespace(), AppI18nKey(appID, field)); svcErr != nil {
+			as.logger.Error("Failed to delete localized variant on app deletion",
+				log.String("appID", appID),
+				log.String("field", field),
+				log.String("namespace", AppI18nNamespace()))
+			hasErr = true
+		}
+	}
+	if hasErr {
+		return &serviceerror.InternalServerError
+	}
+	return nil
+}
+
+// cleanupStaleI18nKeys removes i18n keys for fields that changed from an i18n ref back to plain text.
+// Returns an internal server error if any deletion fails.
+func (as *applicationService) cleanupStaleI18nKeys(
+	ctx context.Context, appID string,
+	existing *model.ApplicationProcessedDTO, updated *model.ApplicationDTO,
+) *serviceerror.ServiceError {
+	if as.i18nService == nil {
+		return nil
+	}
+	type pair struct{ old, updated, field string }
+	fields := []pair{
+		{existing.Name, updated.Name, "name"},
+		{existing.LogoURL, updated.LogoURL, "logo_uri"},
+		{existing.TosURI, updated.TosURI, "tos_uri"},
+		{existing.PolicyURI, updated.PolicyURI, "policy_uri"},
+	}
+	var hasErr bool
+	for _, f := range fields {
+		if isI18nRef(f.old) && !isI18nRef(f.updated) {
+			if svcErr := as.i18nService.DeleteTranslationsByKey(
+				ctx, AppI18nNamespace(), AppI18nKey(appID, f.field)); svcErr != nil {
+				as.logger.Error("Failed to delete stale i18n key",
+					log.String("appID", appID),
+					log.String("field", f.field),
+					log.String("namespace", AppI18nNamespace()))
+				hasErr = true
+			}
+		}
+	}
+	if hasErr {
+		return &serviceerror.InternalServerError
+	}
+	return nil
 }
--- a/backend/internal/application/utils.go
+++ b/backend/internal/application/utils.go
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package application
+
+import "strings"
+
+// AppI18nNamespace returns the i18n namespace for application localized metadata.
+func AppI18nNamespace() string {
+	return "custom"
+}
+
+// AppI18nKey returns the i18n key for an application field.
+func AppI18nKey(appID, field string) string {
+	return "app." + appID + "." + field
+}
+
+// AppI18nRef returns the i18n template reference string for an application field.
+// The returned value is stored as the application's display field so the UI can
+// resolve it to the correct locale at render time.
+func AppI18nRef(appID, field string) string {
+	return "{{t(" + AppI18nNamespace() + ":" + AppI18nKey(appID, field) + ")}}"
+}
+
+// isI18nRef reports whether s is an i18n template reference (e.g. "{{t(custom:app.123.name)}}").
+func isI18nRef(s string) bool {
+	return strings.HasPrefix(s, "{{t(") && strings.HasSuffix(s, ")}}")
+}
--- a/backend/internal/oauth/init.go
+++ b/backend/internal/oauth/init.go
@@ -44,6 +44,7 @@ import (
 	"github.com/asgardeo/thunder/internal/system/crypto/pki"
 	"github.com/asgardeo/thunder/internal/system/database/provider"
 	syshttp "github.com/asgardeo/thunder/internal/system/http"
+	i18nmgt "github.com/asgardeo/thunder/internal/system/i18n/mgt"
 	"github.com/asgardeo/thunder/internal/system/jose/jwe"
 	"github.com/asgardeo/thunder/internal/system/jose/jwt"
 	"github.com/asgardeo/thunder/internal/system/observability"
@@ -65,6 +66,7 @@ func Initialize(
 	authzService authz.AuthorizationServiceInterface,
 	entityProvider entityprovider.EntityProviderInterface,
 	resourceService resource.ResourceServiceInterface,
+	i18nService i18nmgt.I18nServiceInterface,
 ) error {
 	// Fetch runtime transactioner for OAuth services.
 	transactioner, err := provider.GetDBProvider().GetRuntimeDBTransactioner()
@@ -92,6 +94,6 @@ func Initialize(
 			return syshttp.IsSSRFSafeURL(req.URL.String())
 		}),
 		tokenValidator, inboundClient, ouService, attributeCacheSvc, transactioner)
-	dcr.Initialize(mux, applicationService, ouService, transactioner)
+	dcr.Initialize(mux, applicationService, ouService, i18nService, transactioner)
 	return nil
 }
--- a/backend/internal/oauth/oauth2/dcr/error_constants.go
+++ b/backend/internal/oauth/oauth2/dcr/error_constants.go
@@ -19,9 +19,28 @@
 package dcr
 
 import (
+	"strconv"
+
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/i18n/core"
 )
+
+// errInvalidBCP47Tag is returned when a language tag in a DCR request field is not valid BCP 47.
+type errInvalidBCP47Tag struct{ key string }
+
+// Error implements the error interface.
+func (e *errInvalidBCP47Tag) Error() string {
+	return "invalid BCP 47 language tag in field \"" + e.key + "\""
+}
+
+// errTooManyLocalizedVariants is returned when a localizable field exceeds maxLocalizedVariantsPerField.
+type errTooManyLocalizedVariants struct{ field string }
+
+// Error implements the error interface.
+func (e *errTooManyLocalizedVariants) Error() string {
+	return "field \"" + e.field + "\" exceeds the maximum of " +
+		strconv.Itoa(maxLocalizedVariantsPerField) + " localized variants"
+}
 
 // DCR standard service error constants
 var (
--- a/backend/internal/oauth/oauth2/dcr/model.go
+++ b/backend/internal/oauth/oauth2/dcr/model.go
@@ -19,12 +19,17 @@
 package dcr
 
 import (
+	"encoding/json"
+	"strings"
+
 	oauth2const "github.com/asgardeo/thunder/internal/oauth/oauth2/constants"
+	i18nmgt "github.com/asgardeo/thunder/internal/system/i18n/mgt"
 )
 
 // Default values for DCR
 const (
-	ClientSecretExpiresAtNever = 0 // Never expires
+	ClientSecretExpiresAtNever   = 0 // Never expires
+	maxLocalizedVariantsPerField = 20
 )
 
 // DCRRegistrationRequest represents the RFC 7591 Dynamic Client Registration request.
@@ -44,11 +49,78 @@ type DCRRegistrationRequest struct {
 	TosURI                  string                              `json:"tos_uri,omitempty"`
 	PolicyURI               string                              `json:"policy_uri,omitempty"`
 
-	RequirePushedAuthorizationRequests bool `json:"require_pushed_authorization_requests,omitempty"`
+	RequirePushedAuthorizationRequests bool   `json:"require_pushed_authorization_requests,omitempty"`
+	UserInfoSignedResponseAlg          string `json:"userinfo_signed_response_alg,omitempty"`
+	UserInfoEncryptedResponseAlg       string `json:"userinfo_encrypted_response_alg,omitempty"`
+	UserInfoEncryptedResponseEnc       string `json:"userinfo_encrypted_response_enc,omitempty"`
+	// Localized variant maps — populated from #-keyed JSON fields (e.g. "client_name#fr").
+	LocalizedClientName map[string]string `json:"-"`
+	LocalizedLogoURI    map[string]string `json:"-"`
+	LocalizedTosURI     map[string]string `json:"-"`
+	LocalizedPolicyURI  map[string]string `json:"-"`
+}
 
-	UserInfoSignedResponseAlg    string `json:"userinfo_signed_response_alg,omitempty"`
-	UserInfoEncryptedResponseAlg string `json:"userinfo_encrypted_response_alg,omitempty"`
-	UserInfoEncryptedResponseEnc string `json:"userinfo_encrypted_response_enc,omitempty"`
+// UnmarshalJSON decodes DCRRegistrationRequest from JSON, extracting OIDC language-tagged fields
+// (e.g. "client_name#fr") into the localized variant maps.
+func (r *DCRRegistrationRequest) UnmarshalJSON(data []byte) error {
+	type Alias DCRRegistrationRequest
+	if err := json.Unmarshal(data, (*Alias)(r)); err != nil {
+		return err
+	}
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	return parseLocalizedFields(raw, r)
+}
+
+// parseLocalizedFields extracts language-tagged fields (e.g. "client_name#fr") from a raw JSON map
+// and populates the localized variant maps on r.
+func parseLocalizedFields(raw map[string]json.RawMessage, r *DCRRegistrationRequest) error {
+	for key, val := range raw {
+		field, tag, ok := strings.Cut(key, "#")
+		if !ok {
+			continue
+		}
+		canonical, valid := i18nmgt.NormaliseBCP47Tag(tag)
+		if !valid {
+			return &errInvalidBCP47Tag{key: key}
+		}
+		var s string
+		if err := json.Unmarshal(val, &s); err != nil {
+			continue
+		}
+		var target *map[string]string
+		switch field {
+		case "client_name":
+			target = &r.LocalizedClientName
+		case "logo_uri":
+			target = &r.LocalizedLogoURI
+		case "tos_uri":
+			target = &r.LocalizedTosURI
+		case "policy_uri":
+			target = &r.LocalizedPolicyURI
+		}
+		if target == nil {
+			continue
+		}
+		if err := setLocalizedVariant(target, field, canonical, s); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// setLocalizedVariant initializes the map if needed, stores the value, and enforces the variant limit.
+func setLocalizedVariant(m *map[string]string, field, tag, val string) error {
+	if *m == nil {
+		*m = make(map[string]string)
+	}
+	(*m)[tag] = val
+	if len(*m) > maxLocalizedVariantsPerField {
+		return &errTooManyLocalizedVariants{field: field}
+	}
+	return nil
 }
 
 // DCRRegistrationResponse represents the RFC 7591 Dynamic Client Registration response.
@@ -71,11 +143,47 @@ type DCRRegistrationResponse struct {
 	PolicyURI               string                              `json:"policy_uri,omitempty"`
 	AppID                   string                              `json:"app_id,omitempty"`
 
-	RequirePushedAuthorizationRequests bool `json:"require_pushed_authorization_requests,omitempty"`
+	RequirePushedAuthorizationRequests bool   `json:"require_pushed_authorization_requests,omitempty"`
+	UserInfoSignedResponseAlg          string `json:"userinfo_signed_response_alg,omitempty"`
+	UserInfoEncryptedResponseAlg       string `json:"userinfo_encrypted_response_alg,omitempty"`
+	UserInfoEncryptedResponseEnc       string `json:"userinfo_encrypted_response_enc,omitempty"`
+	// Localized variant maps — injected as #-keyed top-level fields during serialization.
+	LocalizedClientName map[string]string `json:"-"`
+	LocalizedLogoURI    map[string]string `json:"-"`
+	LocalizedTosURI     map[string]string `json:"-"`
+	LocalizedPolicyURI  map[string]string `json:"-"`
+}
 
-	UserInfoSignedResponseAlg    string `json:"userinfo_signed_response_alg,omitempty"`
-	UserInfoEncryptedResponseAlg string `json:"userinfo_encrypted_response_alg,omitempty"`
-	UserInfoEncryptedResponseEnc string `json:"userinfo_encrypted_response_enc,omitempty"`
+// MarshalJSON serializes DCRRegistrationResponse to JSON, injecting OIDC language-tagged
+// fields (e.g. "client_name#fr") as top-level keys.
+func (r DCRRegistrationResponse) MarshalJSON() ([]byte, error) {
+	type Alias DCRRegistrationResponse
+	base, err := json.Marshal(Alias(r))
+	if err != nil {
+		return nil, err
+	}
+	var m map[string]interface{}
+	if err := json.Unmarshal(base, &m); err != nil {
+		return nil, err
+	}
+	appendLocalizedFields(m, r)
+	return json.Marshal(m)
+}
+
+// appendLocalizedFields injects localized variant maps from r into m as #-keyed top-level entries.
+func appendLocalizedFields(m map[string]interface{}, r DCRRegistrationResponse) {
+	for tag, val := range r.LocalizedClientName {
+		m["client_name#"+tag] = val
+	}
+	for tag, val := range r.LocalizedLogoURI {
+		m["logo_uri#"+tag] = val
+	}
+	for tag, val := range r.LocalizedTosURI {
+		m["tos_uri#"+tag] = val
+	}
+	for tag, val := range r.LocalizedPolicyURI {
+		m["policy_uri#"+tag] = val
+	}
 }
 
 // DCRErrorResponse represents the RFC 7591 Dynamic Client Registration error response.
--- a/backend/internal/oauth/oauth2/dcr/model_test.go
+++ b/backend/internal/oauth/oauth2/dcr/model_test.go
@@ -1,0 +1,180 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package dcr
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+// DCRModelTestSuite is the test suite for DCR model marshaling/unmarshaling.
+type DCRModelTestSuite struct {
+	suite.Suite
+}
+
+func TestDCRModelTestSuite(t *testing.T) {
+	suite.Run(t, new(DCRModelTestSuite))
+}
+
+func (s *DCRModelTestSuite) TestUnmarshalJSON_BasicFields() {
+	input := `{
+		"client_name": "My App",
+		"redirect_uris": ["https://example.com/cb"],
+		"scope": "openid profile"
+	}`
+	var req DCRRegistrationRequest
+	s.Require().NoError(json.Unmarshal([]byte(input), &req))
+
+	s.Equal("My App", req.ClientName)
+	s.Equal([]string{"https://example.com/cb"}, req.RedirectURIs)
+	s.Equal("openid profile", req.Scope)
+	s.Nil(req.LocalizedClientName)
+}
+
+func (s *DCRModelTestSuite) TestUnmarshalJSON_LocalizedFields() {
+	input := `{
+		"client_name": "My App",
+		"client_name#fr": "Mon Application",
+		"client_name#de": "Meine Anwendung",
+		"logo_uri": "https://example.com/logo.png",
+		"logo_uri#fr": "https://example.com/fr/logo.png",
+		"tos_uri#fr": "https://example.com/fr/tos",
+		"policy_uri#fr": "https://example.com/fr/policy"
+	}`
+	var req DCRRegistrationRequest
+	s.Require().NoError(json.Unmarshal([]byte(input), &req))
+
+	s.Equal("My App", req.ClientName)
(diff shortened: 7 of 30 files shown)


nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

thiva-k:
Hi @nandhu-kumar ,

What is the reason behind including DCR GET in this PR? The i18n resolution in frontend only uses the application id and our i18n implementation via `flow/meta`

nandhu-kumar:
> Hi @nandhu-kumar ,
> 
> What is the reason behind including DCR GET in this PR? The i18n resolution in frontend only uses the application id and our i18n implementation via `flow/meta`

@thiva-k The DCR GET change is not for frontend i18n resolution. /flow/meta already gives the app id/key reference needed by the SDK. DCR GET was included so that localized metadata submitted via DCR is also readable back via DCR, keeping the registration API consistent and preserving the full client representation.

thiva-k:
> @thiva-k The DCR GET change is not for frontend i18n resolution. /flow/meta already gives the app id/key reference needed by the SDK. DCR GET was included so that localized metadata submitted via DCR is also readable back via DCR, keeping the registration API consistent and preserving the full client representation.

What is the requirement to read it via the DCR GET for client metadata if it is already available via i18n apis?

If we are to introduce the DCR GET, better we introduce it in a separate PR adhering to DCRM RFC 7592 spec and other operations too. 
e.g. mandatory dcr registration token issuance and validation, etc.

nandhu-kumar:
> > @thiva-k The DCR GET change is not for frontend i18n resolution. /flow/meta already gives the app id/key reference needed by the SDK. DCR GET was included so that localized metadata submitted via DCR is also readable back via DCR, keeping the registration API consistent and preserving the full client representation.
> 
> What is the requirement to read it via the DCR GET for client metadata if it is already available via i18n apis?
> 
> If we are to introduce the DCR GET, better we introduce it in a separate PR adhering to DCRM RFC 7592 spec and other operations too. e.g. mandatory dcr registration token issuance and validation, etc.

ok @thiva-k I removed the DCR GET changes from this PR and updated it to focus only on localized metadata support

nandhu-kumar:
@coderabbitai review

nandhu-kumar:
@coderabbitai review

backend/internal/system/i18n/mgt/service.go
```diff
@@ -117,7 +123,14 @@ func (s *i18nService) ResolveTranslationsForKey(
...
 	requestedLang := goi18n.Make(language)
 
-	bestTranslation := selectBestTranslation(trans, requestedLang)
+	var bestTranslation Translation
+	if namespace == SystemNamespace {
+		// System namespace: BCP47 best-match fallback so UI strings always render.
+		bestTranslation = selectBestTranslation(trans, requestedLang)
+	} else {
+		// Non-system namespaces (e.g. app.*): exact match only — no server-side fallback.
+		bestTranslation = exactLanguageMatch(trans, language)
```
ThaminduDilshan: @ThumulaPerera this changes the existing translation resolution behaviour. If I understand correctly, for custom namespaces, server will no longer do best matching translations. Will this be a acceptable change?

backend/internal/system/i18n/mgt/service.go
```diff
@@ -241,8 +254,26 @@ func (s *i18nService) ResolveTranslations(
 	}
 
 	result := make(map[string]map[string]string)
-	for _, translations := range allTranslations {
-		translation := selectBestTranslation(translations, requestedLang)
+	for compositeKey, translations := range allTranslations {
```
ThaminduDilshan: Same as previous comment.
cc: @ThumulaPerera

backend/internal/system/i18n/mgt/service.go
```diff
@@ -333,6 +364,68 @@ func (s *i18nService) ClearTranslationOverrides(language string) *serviceerror.I
 	return nil
 }
 
+// DeleteTranslationsByNamespaceAndKey removes all translations for a specific namespace+key pair.
+// Used during application update to fully replace a field's locale variants.
```
ThaminduDilshan:
```suggestion
```
No need to have specific usage comments like this

backend/internal/system/i18n/mgt/service.go
```diff
@@ -333,6 +364,68 @@ func (s *i18nService) ClearTranslationOverrides(language string) *serviceerror.I
...
+	if err := s.store.DeleteTranslationsByNamespaceAndKey(namespace, key); err != nil {
+		s.logger.Error("Failed to delete translations by namespace and key", log.Error(err))
+		return &ErrorInternalServerError
+	}
+	return nil
+}
+
+// DeleteTranslationsByNamespace removes all translations under the given namespace.
+// This is used for lifecycle cleanup (e.g., on application deletion) and intentionally
+// does not check declarative mode since app deletion is guarded at the application layer.
```
ThaminduDilshan:
```suggestion
```

backend/internal/system/i18n/mgt/constants.go
```diff
@@ -37,13 +37,30 @@ var LanguagePreferenceOrder = map[string]int{
...
-var namespaceRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+// Namespaces can contain alphanumeric characters, underscores, hyphens, and dots.
+var namespaceRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
 
 // keyRegex defines the valid format for translation keys.
 // Keys can contain alphanumeric characters, dots, underscores, and hyphens.
 var keyRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
 
+// MaxBCP47TagLength is the maximum allowed length of a BCP 47 language tag (RFC 5646 practical limit).
+const MaxBCP47TagLength = 35
```
ThaminduDilshan: If this doesn't need to be used outside i18n/mgt package, let's make this a private variable

backend/internal/system/i18n/mgt/constants.go
```diff
@@ -37,13 +37,30 @@ var LanguagePreferenceOrder = map[string]int{
...
+
+// NormaliseBCP47Tag parses tag using golang.org/x/text/language and returns its canonical form
```
ThaminduDilshan: Shall we update comment just to explain behaviour/ applicable usage of the method? We don't need to mention external library specific invocations

backend/internal/oauth/oauth2/dcr/model.go
```diff
@@ -43,6 +49,90 @@ type DCRRegistrationRequest struct {
...
+		// Validate the tag eagerly so callers get an error on bad input.
+		canonical, valid := i18nmgt.NormaliseBCP47Tag(tag)
+		if !valid {
+			return &errInvalidBCP47Tag{key: key}
+		}
+		var s string
+		if err := json.Unmarshal(val, &s); err != nil {
+			continue
+		}
+		switch field {
```
ThaminduDilshan, line 94: Shall we extract out localized field resolution to a private helper function to increase readability?
Better to follow same approach in `(r DCRRegistrationResponse) MarshalJSON()` method as well

backend/internal/oauth/oauth2/dcr/model.go
```diff
@@ -43,6 +49,90 @@ type DCRRegistrationRequest struct {
...
+			if len(r.LocalizedPolicyURI) > maxLocalizedVariantsPerField {
+				return &errTooManyLocalizedVariants{field: field}
+			}
+		}
+	}
+	return nil
+}
+
+// errInvalidBCP47Tag is returned when a language tag in a DCR request field is not valid BCP 47.
+type errInvalidBCP47Tag struct{ key string }
```
ThaminduDilshan: Can't we use a go error instead? Also since this is an error, shall we move this to error_constants.go file?
Same with errTooManyLocalizedVariants

backend/internal/oauth/oauth2/dcr/model_test.go
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/oauth/oauth2/dcr/model_test.go
```diff
@@ -0,0 +1,194 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestUnmarshalJSON_BasicFields tests that standard (non-localized) fields are decoded.
+func TestUnmarshalJSON_BasicFields(t *testing.T) {
```
ThaminduDilshan: Shall we follow stretchr/testify test suite pattern aligning with project unit test guidelines?

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -131,6 +140,41 @@ func (ds *dcrService) RegisterClient(ctx context.Context, request *DCRRegistrati
 		return nil, &ErrorServerError
 	}
 
+	// Write localized variants after the runtime-DB transaction commits.
```
ThaminduDilshan: Is this comment correct? I think both applications and i18n translations are stored in config DB. Hence we should be able to perform this within the same transection.

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -258,6 +302,41 @@ func (ds *dcrService) convertApplicationToDCRResponse(appDTO *model.ApplicationD
 	return response, nil
 }
 
+// writeLocalizedVariants persists all localized variants from a DCR request to the i18n table.
+// Called after the application record has been committed to the runtime DB.
+func (ds *dcrService) writeLocalizedVariants(
+	appID string, request *DCRRegistrationRequest) *serviceerror.ServiceError {
+	if ds.i18nService == nil {
+		return nil
```
ThaminduDilshan, line 371: Let's add a debug log here

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -258,6 +302,41 @@ func (ds *dcrService) convertApplicationToDCRResponse(appDTO *model.ApplicationD
...
+		{request.LocalizedLogoURI, "logo_uri", sysutils.IsValidLogoURI},
+		{request.LocalizedTosURI, "tos_uri", sysutils.IsValidURI},
+		{request.LocalizedPolicyURI, "policy_uri", sysutils.IsValidURI},
+	}
+	for _, f := range fields {
+		for tag, val := range f.variants {
+			if f.validateURI != nil && !f.validateURI(val) {
+				return &ErrorInvalidClientMetadata
+			}
+			if _, svcErr := ds.i18nService.SetTranslationOverrideForKey(tag, ns, f.key, val); svcErr != nil {
```
ThaminduDilshan: Isn't it possible to perform this in a single query with a single i18n service invocation? Currently i18n service set method gets executed within two embedded loops making this O(n^2) operation. Even with a transection, this could be expensive?

What change will this require for the i18n service to support it?

@coderabbitai WDYT of this logic?

backend/internal/application/service.go
```diff
@@ -103,10 +106,27 @@ func newApplicationService(
 		layoutMgtService:  layoutMgtService,
 		userSchemaService: userSchemaService,
 		consentService:    consentService,
+		i18nService:       i18nService,
 		transactioner:     transactioner,
 	}
 }
 
+// AppI18nNamespace returns the i18n namespace for an application.
+func AppI18nNamespace(appID string) string {
```
ThaminduDilshan: @ThumulaPerera is this an expected i18n namespace|key pattern?

ThaminduDilshan: What's the usage of this method? If this is not used outside application package, shall we make this a private function and move to the end of this file? If used outside, let's move this to a new utils.go file.

Also shall we rename this to `getI18nNamespace(appID)` or something similar?

backend/internal/application/service.go
```diff
@@ -103,10 +106,27 @@ func newApplicationService(
...
+	return "app." + appID
+}
+
+// deleteLocalizedVariants removes all i18n translations for the application namespace.
+func (as *applicationService) deleteLocalizedVariants(appID string) {
```
ThaminduDilshan: Shall we move this function declaration to somewhere below in the file to increase readability? We usually keep service methods on top and give priority to the public methods

backend/internal/application/service.go
```diff
@@ -115,6 +135,8 @@ func (as *applicationService) deleteEntityCompensation(appID string) {
 }
 
 // CreateApplication creates the application.
+//
+//nolint:gocyclo // orchestration function — branching is inherent to multi-step create flow
```
ThaminduDilshan: Why do we need these comments?

backend/internal/application/service.go
```diff
@@ -646,6 +670,9 @@ func (as *applicationService) DeleteApplication(ctx context.Context, appID strin
 		return &ErrorInternalServerError
 	}
 
+	// Delete all localized variants from the i18n table (AC-19).
+	as.deleteLocalizedVariants(appID)
```
ThaminduDilshan: Shouldn't this be performed within the same transection?
cc: @thiva-k

backend/internal/system/i18n/mgt/store.go
```diff
@@ -36,6 +36,8 @@ type i18nStoreInterface interface {
 	UpsertTranslation(trans Translation) error
 	DeleteTranslationsByLanguage(language string) error
 	DeleteTranslation(language string, key string, namespace string) error
+	DeleteTranslationsByNamespace(namespace string) error
+	DeleteTranslationsByNamespaceAndKey(namespace string, key string) error
```
ThumulaPerera: Shall we rename this as `DeleteTranslationsByKey` to align with `GetTranslationsByKey(key string, namespace string)`

backend/internal/system/i18n/mgt/constants.go
```diff
@@ -37,13 +37,30 @@ var LanguagePreferenceOrder = map[string]int{
 }
 
 // namespaceRegex defines the valid format for namespace strings.
-// Namespaces can contain alphanumeric characters, underscores, and hyphens.
-var namespaceRegex = regexp.MustCompile(`^[a-zA-Z0-9_-]+$`)
+// Namespaces can contain alphanumeric characters, underscores, hyphens, and dots.
+var namespaceRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
```
ThaminduDilshan: @ThumulaPerera noticed that we use hyphens for internal i18n keys. Will this break resolution of sync keys (eg: i18nServiceError)?
Also consider allowing dot too

backend/internal/application/service.go
ThumulaPerera: Current convention is not to have a hierarchy (with dot notation) within a namespace (i.e. namespace should not have dots)
On a separate note, is it necessary to have a namespace per app? Can't we use the `custom` namespace here and keep app as part of the key instead? (But for this, I18n key Deletion logic might need to be improved to support hierarchical deletion)  
@brionmario WDYT?

backend/internal/system/i18n/mgt/service.go
```diff
@@ -45,6 +46,11 @@ type I18nServiceInterface interface {
 	SetTranslationOverrideForKey(language string, namespace string, key string, value string) (
 		*TranslationResponse, *serviceerror.I18nServiceError)
 	ClearTranslationOverrideForKey(language string, namespace string, key string) *serviceerror.I18nServiceError
+	DeleteTranslationsByNamespace(namespace string) *serviceerror.I18nServiceError
+	DeleteTranslationsByNamespaceAndKey(namespace string, key string) *serviceerror.I18nServiceError
+	// GetTranslationsByNamespace returns all raw translations for a namespace as
+	// map[key]map[language]value without locale resolution or best-match logic.
+	GetTranslationsByNamespace(namespace string) (map[string]map[string]string, *serviceerror.I18nServiceError)
```
ThumulaPerera: What is the usage of this method?

backend/internal/system/i18n/mgt/service.go
ThumulaPerera: We should not be doing this as this affects scenarios other than DCR. But even for DCR case, is this needed ? Does the spec restrict falling back to the best available traslation (i.e. if `en-US` is requested and if only `en` is configured, what is the expected behaviour according to the spec ?)

backend/internal/system/i18n/mgt/service.go
ThumulaPerera: please refer my reply in the same

backend/internal/system/i18n/mgt/constants.go
ThumulaPerera: AFAIK, currently we don't use hyphens for namespaces. But I see no harm n allowing hyphens as a word separator.
But I prefer to not have dots because dots are typically used to denote hierarchical ordering. I prefer to keep namespaces flat for simplicity

ThumulaPerera: `Will this break resolution of sync keys` Doesn't look anything will break with this. This is just adding dot as an additional allowed character in namespace name AFAIU

ThumulaPerera:
@nandhu-kumar are we templating the application name so that is translates correctly in the gate app UI? 

For example in the UI when we are displaying the application name (client name), we should display the correct value based on end user's preference correct ? For that to happen, the application name should be set as {{t(`i18nnamespace:i18nkey`)}}

backend/internal/system/i18n/mgt/service.go
nandhu-kumar: Hi @ThumulaPerera , @ThaminduDilshan 

The if namespace == SystemNamespace branch was introduced as part of the DCR localized metadata work to avoid unintended fallback for application-specific data.

In the DCR flow, localized values (e.g., client_name#fr) are stored under app.<appId>. When reading them back (RFC 7592), using selectBestTranslation can return a different language than what was explicitly registered (e.g., request en, only fr exists → returns fr). The intention was to ensure the server returns only the exact language configured and leaves any fallback handling to the SDK/client side.

However, I understand the concern that this changes resolution behavior for all non-system namespaces, not just DCR.

To proceed, I see a few options:

Revert to existing behavior (always selectBestTranslation)
→ Keeps global i18n behavior unchanged (as suggested)
Handle exact-match at the DCR layer
→ Use GetTranslationsByNamespace and pick only the requested language (no fallback), keeping i18n service unchanged
Extend resolver with a flag (e.g., exactMatch)
→ Allow DCR to opt into exact-match without affecting other namespaces

Let me know which approach you prefer — I can proceed accordingly.

Thanks!

backend/cmd/server/servicemanager.go
```diff
@@ -267,7 +267,7 @@ func registerServices(mux *http.ServeMux) jwt.JWTServiceInterface {
 
 	// Initialize OAuth services.
 	err = oauth.Initialize(mux, applicationService, authnProvider, jwtService, flowExecService, observabilitySvc,
-		pkiService, ouService, attributeCacheService, authZService, entityProvider)
+		pkiService, ouService, attributeCacheService, authZService, entityProvider, i18nService)
```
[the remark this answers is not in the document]
nandhu-kumar: The GET /oauth2/dcr/register/{app_id} (RFC 7592 client-read) endpoint is not part of this PR. Only POST /oauth2/dcr/register (RFC 7591 registration) is implemented here. Coderabbit appears to have inferred the full RFC 7592 scope from context, but that endpoint was never added. No documentation is needed for it.

backend/internal/oauth/init.go
```diff
@@ -79,6 +81,6 @@ func Initialize(
 	introspect.Initialize(mux, jwtService, applicationService, authnProvider, discoveryService)
 	userinfo.Initialize(mux, jwtService, tokenValidator, applicationService, ouService, attributeCacheSvc,
 		transactioner)
-	dcr.Initialize(mux, applicationService, ouService, transactioner)
+	dcr.Initialize(mux, applicationService, ouService, i18nService, transactioner)
```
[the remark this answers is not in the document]
nandhu-kumar: The GET /oauth2/dcr/register/{app_id} (RFC 7592 client-read) endpoint is not part of this PR. Only POST /oauth2/dcr/register (RFC 7591 registration) is implemented here. Coderabbit appears to have inferred the full RFC 7592 scope from context, but that endpoint was never added. No documentation is needed for it.

backend/internal/system/i18n/mgt/store.go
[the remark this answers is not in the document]
nandhu-kumar: The GET /oauth2/dcr/register/{app_id} (RFC 7592 client-read) endpoint is not part of this PR — only POST /oauth2/dcr/register (RFC 7591 registration) is implemented here.

This PR does introduce localized metadata fields (client_name#, logo_uri#, etc.), which are user-facing changes. Documentation for these fields, including BCP 47 validation and error behavior, will be added under docs/content/guides/.

backend/internal/oauth/oauth2/dcr/model.go
```diff
@@ -43,6 +47,70 @@ type DCRRegistrationRequest struct {
 	Contacts                []string                            `json:"contacts,omitempty"`
 	TosURI                  string                              `json:"tos_uri,omitempty"`
 	PolicyURI               string                              `json:"policy_uri,omitempty"`
+
+	// Localized variant maps — populated from #-keyed JSON fields (e.g. "client_name#fr").
+	LocalizedClientName map[string]string `json:"-"`
+	LocalizedLogoURI    map[string]string `json:"-"`
+	LocalizedTosURI     map[string]string `json:"-"`
+	LocalizedPolicyURI  map[string]string `json:"-"`
```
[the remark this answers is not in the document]
nandhu-kumar: The GET /oauth2/dcr/register/{app_id} (RFC 7592 client-read) endpoint is not part of this PR — only POST /oauth2/dcr/register (RFC 7591 registration) is implemented here.

This PR introduces localized metadata fields (client_name#, logo_uri#, etc.), which are user-facing changes. Documentation for these fields, including BCP 47 validation and error behavior, will be added under docs/content/guides/.

backend/internal/application/service_test.go
```diff
@@ -7005,3 +7006,64 @@ func (suite *ServiceTestSuite) TestValidateApplicationForUpdate_ErrorFromValidat
 	assert.NotNil(suite.T(), svcErr)
 	assert.Equal(suite.T(), &ErrorLayoutNotFound, svcErr)
 }
+
+// TestAppI18nNamespace verifies the AppI18nNamespace helper returns the correct prefix.
+func TestAppI18nNamespace(t *testing.T) {
+	assert.Equal(t, "app.abc-123", AppI18nNamespace("abc-123"))
+}
```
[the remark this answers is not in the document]
nandhu-kumar: The GET /oauth2/dcr/register/{app_id} (RFC 7592 client-read) endpoint is not part of this PR — only POST /oauth2/dcr/register (RFC 7591 registration) is implemented here.

This PR introduces localized metadata fields (client_name#, logo_uri#, etc.), which are user-facing changes. Documentation for these fields, including BCP 47 validation and error behavior, will be added under docs/content/guides/.

backend/internal/application/service.go
nandhu-kumar: @ThaminduDilshan 

Good point — ideally this should be in the same transaction.

Right now the i18n service doesn’t support transaction context, so it can’t be included in the same transaction. That’s why it’s done outside as a best-effort cleanup.

We’ve added a TODO to move this inside the transaction once i18n supports it.

backend/internal/oauth/oauth2/dcr/service.go
nandhu-kumar: added logger.

backend/internal/system/i18n/mgt/constants.go
nandhu-kumar: Thanks for the clarification — keeping the current regex as it doesn’t introduce any breaking changes.

nandhu-kumar:
> @nandhu-kumar are we templating the application name so that is translates correctly in the gate app UI?
> 
> For example in the UI when we are displaying the application name (client name), we should display the correct value based on end user's preference correct ? For that to happen, the application name should be set as {{t(`i18nnamespace:i18nkey`)}}

@ThumulaPerera  

This PR focuses on the DCR side — accepting and storing localized metadata (e.g., client_name#fr) via the API as per RFC 7591.

Currently, the gate UI doesn’t use the client_name field, and there’s no UI-level i18n resolution happening yet. The t{namespace:key} templating mentioned is more of a UI rendering concern.

When the UI starts displaying the application name (e.g., in consent screens), it should resolve it from the i18n store based on the user’s language. That can be handled as a follow-up, as it’s outside the scope of this PR.

backend/internal/system/i18n/mgt/service.go
ThumulaPerera: `using selectBestTranslation can return a different language than what was explicitly registered (e.g., request en, only fr exists → returns fr)`. Is this going to be a DCR spec violation ? If not, I prefer going ahead in this approach to be consistent. 
If the spec mandates an exact match, we need to check what the expected behavior according to the spec when an unavailable language is requested

ThumulaPerera:
> > @nandhu-kumar are we templating the application name so that is translates correctly in the gate app UI?
> > For example in the UI when we are displaying the application name (client name), we should display the correct value based on end user's preference correct ? For that to happen, the application name should be set as {{t(`i18nnamespace:i18nkey`)}}
> 
> @ThumulaPerera
> 
> This PR focuses on the DCR side — accepting and storing localized metadata (e.g., client_name#fr) via the API as per RFC 7591.
> 
> Currently, the gate UI doesn’t use the client_name field, and there’s no UI-level i18n resolution happening yet. The t{namespace:key} templating mentioned is more of a UI rendering concern.
> 
> When the UI starts displaying the application name (e.g., in consent screens), it should resolve it from the i18n store based on the user’s language. That can be handled as a follow-up, as it’s outside the scope of this PR.

Agreed that resolution logic in the UI is out of scope for this PR.

What I meant was, when persisting the application in the backend, its name should be persisted as a templated string. Otherwise when at some point when this name is displayed in the UI, UI will not have a way to know that this particular field is an internationalised field

backend/internal/application/service.go
ThumulaPerera: @nandhu-kumar we currently have not passed transaction context to i18n service as there was no requirement until now. But as we have a requirement now, shall we update the interface methods to accept the context and use it ? 
cc: @JeethJJ

backend/internal/application/service.go
```diff
@@ -116,6 +120,8 @@ func (as *applicationService) deleteEntityCompensation(appID string) {
 }
 
 // CreateApplication creates the application.
+//
+//nolint:gocyclo
```
ThaminduDilshan: Let's address lint issues and get rid of these cyclo ignore comments

backend/internal/application/service.go
```diff
@@ -647,6 +655,10 @@ func (as *applicationService) DeleteApplication(ctx context.Context, appID strin
 		return &serviceerror.InternalServerError
 	}
 
+	// Delete all localized variants from the i18n table (AC-19).
```
ThaminduDilshan: `(AC-19)` these are acceptance criteria that you have defined for the feature right? But it doesn't make sense to have these committed in the code.
Shall we check all the places and remove them?

backend/internal/application/service.go
```diff
@@ -2587,3 +2599,14 @@ func attributesToPurposeElements(attributes map[string]bool) []consent.PurposeEl
...
 }
+
+// deleteLocalizedVariants removes all i18n translations for the application namespace.
+func (as *applicationService) deleteLocalizedVariants(appID string) {
+	if as.i18nService == nil {
+		return
+	}
+	if svcErr := as.i18nService.DeleteTranslationsByNamespace(AppI18nNamespace(appID)); svcErr != nil {
+		as.logger.Error("Failed to delete localized variants on app deletion",
+			log.String("appID", appID), log.String("error", svcErr.Error.String()))
```
ThaminduDilshan:
```suggestion
			log.String("appID", appID), log.String("error", svcErr.ErrorDescription.String()))
```

backend/internal/application/utils.go
```diff
@@ -0,0 +1,24 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package application
+
+// AppI18nNamespace returns the i18n namespace for an application.
+func AppI18nNamespace(appID string) string {
+	return "app." + appID
```
ThaminduDilshan: @ThumulaPerera hope we're okay with this namespace pattern

backend/internal/application/utils.go
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/oauth/oauth2/dcr/error_constants.go
```diff
@@ -19,10 +19,27 @@
...
+	"strconv"
+
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
 	"github.com/asgardeo/thunder/internal/system/i18n/core"
 )
 
+// errInvalidBCP47Tag is returned when a language tag in a DCR request field is not valid BCP 47.
+type errInvalidBCP47Tag struct{ key string }
+
+func (e *errInvalidBCP47Tag) Error() string {
```
ThaminduDilshan, line 32: Add method comments

backend/internal/system/i18n/mgt/error_constants.go
```diff
@@ -130,3 +130,20 @@ var (
 		},
 	}
 )
+
+// Server errors for i18n operations.
+var (
+	// ErrorInternalServerError is the error returned for unexpected i18n service errors.
+	ErrorInternalServerError = serviceerror.ServiceError{
```
ThaminduDilshan: Let's not have package specific server errors. Instead reuse the system wide internal server error constant defined in system/error package

tests/integration/oauth/dcr/model.go
```diff
@@ -18,8 +18,14 @@
...
 package dcr
 
+import (
+	"encoding/json"
+	"strings"
+)
+
 // DCRRegistrationRequest represents the RFC 7591 Dynamic Client Registration request.
 type DCRRegistrationRequest struct {
+<<<<<<< HEAD
```
ThaminduDilshan: Let's resolve conflicts

backend/internal/application/service.go
```diff
@@ -800,6 +809,11 @@ func (as *applicationService) validateApplicationForUpdate(
 		return nil, nil, existingAppErr
 	}
 
+	// Reject plain-string updates to names that are managed via the i18n system.
+	if isI18nRef(existingApp.Name) && app.Name != existingApp.Name {
```
ThaminduDilshan: Do we need this restriction? I feel like we don't need to restrict updating app display name back to a plain text value. 
If they're updating it to a plain text, we can safely delete that particular i18n key and the associated values right?
@brionmario  @ThumulaPerera WDYT?

backend/internal/application/service.go
```diff
@@ -1519,3 +1533,48 @@ func (as *applicationService) mapStoreError(err error) *serviceerror.ServiceErro
...
+}
+
+// deleteLocalizedVariants removes all i18n translations for the application namespace.
+// Best-effort: errors are logged but not propagated so the delete transaction is not blocked.
+func (as *applicationService) deleteLocalizedVariants(ctx context.Context, appID string) {
+	if as.i18nService == nil {
+		return
+	}
+	if svcErr := as.i18nService.DeleteTranslationsByNamespace(ctx, AppI18nNamespace(appID)); svcErr != nil {
+		as.logger.Error("Failed to delete localized variants on app deletion",
```
ThaminduDilshan: If there's error from i18n service, we don't need to rollback the transection. But IMO we should return an error in the API response. Otherwise client will see a successful operation and will not be aware of the error unless someone (admin/ devops) checks the server logs

backend/internal/application/service.go
```diff
@@ -1519,3 +1533,48 @@ func (as *applicationService) mapStoreError(err error) *serviceerror.ServiceErro
...
+		{"logo_uri", app.LogoURL},
+		{"tos_uri", app.TosURI},
+		{"policy_uri", app.PolicyURI},
+	}
+	entries := make(map[string]map[string]string)
+	for _, f := range fields {
+		if f.val == "" || isI18nRef(f.val) {
+			continue
+		}
+		entries[f.key] = map[string]string{i18nmgt.SystemLanguage: f.val}
```
ThaminduDilshan: If I understand correctly, with this logic we'll be creating i18n key-value pairs for app name, logo_uri, tos_uri and policy_uri for all the apps even if they're not going to use localizations. That will be an additional overhead for some clients right? 

Our expectation is someone will be configuring a i18n key for these fields ONLY if they need localization. If NOT, they can safely use plain text values

backend/internal/application/service.go
```diff
@@ -1519,3 +1533,48 @@ func (as *applicationService) mapStoreError(err error) *serviceerror.ServiceErro
...
+	}
+	if len(entries) == 0 {
+		return
+	}
+	if svcErr := as.i18nService.SetTranslationOverridesForNamespace(
+		ctx, AppI18nNamespace(appID), entries); svcErr != nil {
+		as.logger.Error("Failed to write default translations for app",
```
ThaminduDilshan: We should return an error here too. Otherwise client will not be aware of the error and could lead to an unexpected situation

ThaminduDilshan: IMO rolling back transaction should also be considered here

backend/internal/application/utils.go
```diff
@@ -0,0 +1,38 @@
...
+// AppI18nRef returns the i18n template reference string for an application field.
+// The returned value is stored as the application's display field so the UI can
+// resolve it to the correct locale at render time.
+func AppI18nRef(appID, key string) string {
+	return "{{t(" + AppI18nNamespace(appID) + ":" + key + ")}}"
+}
+
+// isI18nRef reports whether s is an i18n template reference (e.g. "{{t(app.123:name)}}").
+func isI18nRef(s string) bool {
+	return strings.HasPrefix(s, "{{t(") && strings.HasSuffix(s, "}}")
```
ThaminduDilshan:
```suggestion
	return strings.HasPrefix(s, "{{t(") && strings.HasSuffix(s, ")}}")
```

backend/internal/oauth/oauth2/dcr/error_constants.go
```diff
@@ -19,10 +19,28 @@
...
+
+// Error implements the error interface.
+func (e *errInvalidBCP47Tag) Error() string {
+	return "invalid BCP 47 language tag in field \"" + e.key + "\""
+}
+
+// errTooManyLocalizedVariants is returned when a localizable field exceeds maxLocalizedVariantsPerField.
+type errTooManyLocalizedVariants struct{ field string }
+
+func (e *errTooManyLocalizedVariants) Error() string {
```
ThaminduDilshan, line 40: Add method comment. Check any other such places

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -134,6 +143,36 @@ func (ds *dcrService) RegisterClient(ctx context.Context, request *DCRRegistrati
 		return nil, &ErrorServerError
 	}
 
+	// Write localized variants outside the configDB transaction above because the DCR outer
+	// transactioner operates on the runtime DB and cannot wrap the configDB i18n writes.
```
ThaminduDilshan: Is this correct? I don't think runtime DB will be engaged anywhere here

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -303,6 +352,64 @@ func (ds *dcrService) convertApplicationToDCRResponse(appDTO *model.ApplicationD
...
+		}
+	}
+	if len(entries) == 0 {
+		return nil
+	}
+	if svcErr := ds.i18nService.SetTranslationOverridesForNamespace(ctx, ns, entries); svcErr != nil {
+		if svcErr.Type == serviceerror.ClientErrorType {
+			return &ErrorInvalidClientMetadata
+		}
+		return &ErrorServerError
```
ThaminduDilshan, line 425: We should log an error here

backend/internal/system/i18n/mgt/store.go
```diff
@@ -215,6 +237,35 @@ func (s *i18nStore) DeleteTranslationsByLanguage(language string) error {
 	return nil
 }
 
+// DeleteTranslationsByKey deletes all translations for the given namespace and key.
+func (s *i18nStore) DeleteTranslationsByKey(namespace string, key string) error {
```
ThaminduDilshan: Any reason to avoid passing context for this particular method? If there's no special reason, we better add it here as well

backend/internal/application/service.go
ThumulaPerera: +1. We don't need to restrict

backend/internal/system/i18n/mgt/constants.go
ThumulaPerera: Agreed that this won't introduce breaking changes. But any specific reason why we need to allow dots in namespace ?

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -175,6 +220,9 @@ func (ds *dcrService) convertDCRToApplication(request *DCRRegistrationRequest) (
 		}
 		clientID = generatedClientID
 		appName = clientID
+	} else if len(request.LocalizedClientName) > 0 {
```
ThumulaPerera, line 231: DCR spec does not mandate sending a client name without a language tag. Even when `appName == ""`, `len(request.LocalizedClientName) > 0` can be true.
In that case also, we should use the i18n ref as the app name rather than client ID

backend/internal/application/utils.go
ThumulaPerera: We do not intend to have dots (i.e. hierarchy) in namespaces. @nandhu-kumar any reason to have a dot here ? can't we just use `custom` or `app` as the namespace here ? (my preference is towards `custom`)

What have in mind is something like
    namespace: custom
    key: app.[appID].[fieldname]

cc: @brionmario

backend/internal/application/service.go
brionmario: +1 to relax the validation

backend/internal/application/service.go
nandhu-kumar: @ThaminduDilshan Good point. Since application deletion is the primary operation and i18n cleanup is secondary, we kept this as best-effort to avoid blocking deletion.
WDYT?

backend/internal/system/i18n/mgt/constants.go
nandhu-kumar: @ThumulaPerera We’re using the pattern `app.<appID>` for application namespaces, so dot support is required. This doesn’t introduce hierarchy semantics — it’s just a prefix convention for namespacing. Happy to revisit if we standardize namespace patterns differently later.

backend/internal/application/service.go
ThaminduDilshan: Yes I'm okay with not rolling back the transaction as i18n cleanup is secondary. But we should give an indication to the client that something wrong happened right?
Otherwise the admin/ user who deletes the app thinks everything happened successfully. But later on he may find out some stale data is existing in the server. WDYT?

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -134,6 +143,38 @@ func (ds *dcrService) RegisterClient(ctx context.Context, request *DCRRegistrati
...
 	}
 
+	// Write localized variants outside the configDB transaction above because i18n writes
+	// go to the configDB via a separate connection and cannot join the same transaction.
+	// If writing fails, clean up any partial i18n rows and compensate by deleting the created app.
+	// Note: writeLocalizedVariants only returns an error when i18nService is non-nil,
+	// so calling DeleteTranslationsByKey here without a nil guard is safe.
+	// If the compensation DeleteApplication also fails, the app record is left without localized
+	// metadata — an accepted gap that can be cleaned up manually or via a future sweep.
+	if writeErr := ds.writeLocalizedVariants(ctx, createdAppID, request); writeErr != nil {
```
ThaminduDilshan, line 153: I think this approach is followed as context transection propagation is not yet implemented in the i18n service. Hence we have to manually compensate the failures.
@coderabbitai can you create a follow up issue to track this improvement? Mainly we need to add context transection propagation to i18n service and update usages like this.
Validate and add only necessary details. Create the issue content simple, but with the required context

backend/internal/application/service.go
```diff
@@ -1519,3 +1525,53 @@ func (as *applicationService) mapStoreError(err error) *serviceerror.ServiceErro
...
+			as.logger.Error("Failed to delete localized variant on app deletion",
+				log.String("appID", appID),
+				log.String("field", field),
+				log.String("namespace", AppI18nNamespace()))
+			if firstErr == nil {
+				firstErr = svcErr
+			}
+		}
+	}
+	return firstErr
```
ThaminduDilshan: Sorry. Not like this. We should return a server error in this case. You're already logging an error line for each and every key. Lets simply keep a boolean to indicate there's an error and if so, return a internal server error

backend/internal/application/service.go
```diff
@@ -362,6 +366,8 @@ func (as *applicationService) UpdateApplication(ctx context.Context, appID strin
 		return nil, svcErr
 	}
 
+	as.cleanupStaleI18nKeys(ctx, appID, existingApp, app)
```
ThaminduDilshan: This is a update operation. If there's an error during i18n cleanup, we should rollback right?

backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -303,6 +360,73 @@ func (ds *dcrService) convertApplicationToDCRResponse(appDTO *model.ApplicationD
...
+		if _, exists := entries[f.key][i18nmgt.SystemLanguage]; !exists {
+			entries[f.key][i18nmgt.SystemLanguage] = f.defaultVal
+		}
+	}
+	if len(entries) == 0 {
+		return nil
+	}
+	if svcErr := ds.i18nService.SetTranslationOverridesForNamespace(ctx, ns, entries); svcErr != nil {
+		if svcErr.Type == serviceerror.ClientErrorType {
+			logger.Error("Invalid client metadata in localized variants",
```
ThaminduDilshan:
```suggestion
			logger.Debug("Invalid client metadata in localized variants",
```

We shouldn't log client errors. But currently we don't have a clear way to propagate underline service error to the caller (discussed in https://github.com/asgardeo/thunder/discussions/2392).
Let's make this a debug log

ThaminduDilshan (approved):

PR #2305, opened 2026-04-11
Add SMS-based self-invite registration flow

### LINKED ISSUE #1929: Support SMS confirmation during user self-registration
### Current Limitation

Thunder's self-registration flow does not support confirming a user's phone number via SMS before activating their account.

### Suggested Improvement

- Add a new SMS template scenario for confirmation messages.
- Provide a reference/sample registration flow using the new executors:
    - User submits self-registration form with phone number.
    - InviteExecutor in generate mode generates the invite link (see https://github.com/asgardeo/thunder/issues/1926).
    - SMSExecutor in send mode delivers the invite link using the appropriate template.
    - User click on the link to confirm.
    - InviteExecutor in verify mode validates the code/ token.
    - User continues with the remaining registration steps.
    - Creates the account at the end.

Depends on following two tasks:
- https://github.com/asgardeo/thunder/issues/1926
- https://github.com/asgardeo/thunder/issues/1928

### Version

_No response_

Dilusha-Madushan:
### Purpose
Add a new bootstrap registration flow that delivers the self-invite link via SMS instead of email. This enables organizations to onboard users using only a mobile number, without requiring an email address at registration time.

### Approach
A new flow definition registration_flow_basic_invite_sms.json is introduced alongside the existing email-based invite flow. The flow follows the same structural pattern but replaces the email collection and delivery steps with phone-based equivalents:

- **Phone collection** — prompts the user for a mobileNumber (using PHONE_INPUT) instead of an email address
- **Uniqueness validation** — runs AttributeUniquenessValidator against the mobileNumber attribute before proceeding
- **Invite generation** — uses InviteExecutor in generate mode to produce the invite token, unchanged from the email flow
- **SMS delivery** — uses SMSExecutor in send mode with the SELF_REGISTRATION template to dispatch the invite link to the collected mobile number
- **Invite verification** — uses InviteExecutor in verify mode to validate the token when the user follows the link, unchanged from the email flow
- **Provisioning** — collects username, password, given name, and family name, then provisions the user with mobileNumber as the identifying attribute

Two i18n keys (forms.phone.* and forms.sms_sent.*) are added to the signup namespace in en-US.json to support the new prompt and confirmation screens.

No changes are made to existing flows, executors, or backend logic.

### Related Issues
- https://github.com/asgardeo/thunder/issues/1929

### Related PRs
- https://github.com/asgardeo/thunder/pull/2128
- https://github.com/asgardeo/thunder/pull/2217
- https://github.com/asgardeo/thunder/pull/1970

## Summary by CodeRabbit

* **New Features**
  * Multi-step SMS self-invite registration: phone entry, SMS delivery, verification, and account setup.
  * Added phone verification forms and SMS confirmation UI text.

* **Chores**
  * Clarified email registration display names to include an "Email" designation.

### DIFF
--- a/backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite_sms.json
+++ b/backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite_sms.json
@@ -1,6 +1,6 @@
 {
-  "name": "Default Self-Invite Registration Flow",
-  "handle": "default-self-invite-flow",
+  "name": "Default SMS Self-Invite Registration Flow",
+  "handle": "default-self-invite-sms-flow",
   "flowType": "REGISTRATION",
   "nodes": [
     {
@@ -14,7 +14,7 @@
       "executor": {
         "name": "UserTypeResolver"
       },
-      "onSuccess": "prompt_email",
+      "onSuccess": "prompt_phone",
       "onIncomplete": "prompt_usertype"
     },
     {
@@ -81,7 +81,7 @@
       ]
     },
     {
-      "id": "prompt_email",
+      "id": "prompt_phone",
       "type": "PROMPT",
       "meta": {
         "components": [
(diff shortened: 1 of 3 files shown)


backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite_sms.json
```diff
@@ -0,0 +1,412 @@
...
+      "properties": {
+        "senderId": "<your-sender-id>",
+        "smsTemplate": "SELF_REGISTRATION"
+      },
+      "executor": {
+        "name": "SMSExecutor",
+        "mode": "send"
+      },
+      "onSuccess": "registration_sms_sent"
+    },
```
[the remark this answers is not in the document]
Dilusha-Madushan: Server errors in flow execution bypass onFailure — When the SMS executor encounters a server-side failure (e.g. webhook delivery error), it returns a Go error which propagates as an InternalServerError at the task_execution_node level, completely bypassing any configured onFailure route. Additionally, the semantics of a failed registration flow reaching the end node without a successful provisioning need to be clarified — currently the gate has no graceful redirect path for this case, which affects both this flow and registration_flow_basic_invite_email.json. These will be addressed together in a follow-up issue.

backend/cmd/server/bootstrap/flows/registration/registration_flow_basic_invite_sms.json
```diff
@@ -0,0 +1,412 @@
...
+    {
+      "id": "send_registration_sms",
+      "type": "TASK_EXECUTION",
+      "properties": {
+        "senderId": "<your-sender-id>",
+        "smsTemplate": "SELF_REGISTRATION"
+      },
+      "executor": {
+        "name": "SMSExecutor",
+        "mode": "send"
```
thiva-k, line 172: What is the use of modes on SMS executor? Is it used?

Dilusha-Madushan: I added that mode value to be match with email executor. But since this mode value is not using inside the sms executor, we can remove it.

So I have removed that mode value from flow json.

ThaminduDilshan (approved):

PR #2316, opened 2026-04-16
Implement authentication provider manager

### LINKED ISSUE #2045: Implement provider manager layer between executors and authn provider impl

When integrating a new provider architecture, there is no intermediary layer to manage interactions between executors (or services) and the authentication provider implementation, resulting in scattered persistence and management logic.

### Proposed Solution

Implement a provider manager module as an abstraction layer between executors (or corresponding service) and the authentication provider implementation. This manager should encapsulate all related persistence logic for provider state and configuration. For this sub-issue, handle only a single authn provider; no need for multi-provider handling at this stage.

### LINKED ISSUE #2055: Optionally return attributes after authentication through authentication provider

Currently, authentication providers in Thunder do not support returning user attributes as part of the authentication response. This limits the ability of downstream services or clients to obtain relevant user attributes immediately after authentication, leading to extra calls.

### Proposed Solution

- Update the authentication provider interface to optionally allow returning attributes after successful authentication
- Improve provider manager to persist the available attributes if they are returned
- Update provider manager implementation to subsequently respond using these persisted attributes when requested
- Changes should be backward compatible so that providers are not required to return attributes if not needed.

### LINKED ISSUE #2201: Improve passkey authentication to go through authn provider
Currently passkey authentication logic happens in the authn/passkey/service.go. The executors, Atomic APIs call this directly. Hence passkey authentication does not go through authn provider. With the update architecture for user core pluggability, all authentications should go through authn provider. Hence we need to fix this.

### LINKED ISSUE #2202: Improve OTP authentication to go through authn provider
Currently OTP authentication logic happens in the authn/otp/service.go. The executors, Atomic APIs call this directly. Hence passkey authentication does not go through authn provider. With the update architecture for user core pluggability, all authentications should go through authn provider. Hence we need to fix this.

ThumulaPerera:
### Purpose

Introduces an auth provider manager layer that acts as the entry point for all authentication operations. Previously, passkey and OTP authentications bypassed the authn provider and were called directly from executors. This change ensures all authentication — password, passkey, OTP — goes through a unified manager, enabling user core pluggability. 

Note: Federated authentication still bypasses auth provider manager layer (eg: oauth, oidc, google, github). This will be addressed with a separate PR. 

### Approach

  - Added `authnprovider/manager` as an abstraction layer between flow executors and the authn provider implementation. The manager keeps track of authentication history of each authn flow.
  - Updated the authn provider interface to optionally return user attributes after successful authentication. The manager persists these attributes and uses them in subsequent requests. 
  - Routed passkey and OTP authentications through the manager, replacing the previously direct calls from executors and the OAuth module.
  - Removed redundant `credentials` service packages which was simply proxying to authn provider previously.

### Related Issues
  - Fixes https://github.com/asgardeo/thunder/issues/2045
  - Fixes https://github.com/asgardeo/thunder/issues/2055
  - Fixes https://github.com/asgardeo/thunder/issues/2201
  - Fixes https://github.com/asgardeo/thunder/issues/2202

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added a new auth error type "Invalid Request" (AUP-0006).
  * Introduced an authentication provider manager to unify credentials, OTP and passkey flows and preserve per-user auth state.
  * Attribute responses now distinguish included values from names for clearer behavior.

* **Bug Fixes**
  * Improved error mapping and reporting for auth failures, invalid requests, and attribute retrieval.
  * More robust OTP/passkey failure signals and handling.

* **Refactor**
  * Reorganized auth components for cleaner flows and state management.


backend/internal/authnprovider/common/error_constants.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
senthalan, line 2: do we follow the process of updating the year when moving the file

backend/internal/authnprovider/manager/manager.go
```diff
@@ -0,0 +1,126 @@
...
+	return data.attributes, nil
+}
+
+// GetUserAttributes returns attributes for the user, fetching from the provider if not already cached.
+func (m *authnProviderManager) GetUserAttributes(ctx context.Context,
+	requestedAttributes *authnprovidercm.RequestedAttributes,
+	metadata *authnprovidercm.GetAttributesMetadata,
+	authUser AuthUser) (AuthUser, *authnprovidercm.AttributesResponse, *serviceerror.ServiceError) {
+	if !authUser.IsAuthenticated() {
+		m.logger.Error("GetUserAttributes called with unauthenticated authUser")
```
senthalan, line 100: do we need to keep it as error log?

backend/internal/flow/flowexec/engine.go
```diff
@@ -379,6 +380,11 @@ func (fe *flowEngine) updateContextWithNodeResponse(engineCtx *EngineContext, no
 	if len(nodeResp.ForwardedData) > 0 {
 		engineCtx.ForwardedData = nodeResp.ForwardedData
 	}
+
+	// Write back AuthUser from the node response
+	if nodeResp.AuthUser.IsAuthenticated() {
+		engineCtx.AuthUser = nodeResp.AuthUser
```
senthalan, line 386: What happens when we have multiple node in the flow? Last nodes AuthUser will be set to the engine context?

senthalan: And I remember discussing keep the engine decoupled from the authentication flow. Having this attribute might be problem in that case. @ThaminduDilshan FYR

ThumulaPerera: > What happens when we have multiple node in the flow? Last nodes AuthUser will be set to the engine context?

AuthenticateUser method in provider manager takes current AuthUser as an input, augments it with additional data and returns it. So the full history will be preserved.

ThumulaPerera: > And I remember discussing keep the engine decoupled from the authentication flow. Having this attribute might be problem in that case. [@ThaminduDilshan](https://github.com/ThaminduDilshan) FYR

I also prefer this architecturally. Main reason to keep this in flow context is to reduce an additional DB call.

backend/internal/authnprovider/common/error_constants.go
ThumulaPerera: I'm not sure whether we have a convention on this. Created a discussion https://github.com/asgardeo/thunder/discussions/2330.

But our general rule is set year on creation. Do not update year on modification. IMO moving, renaming should be considered modifications. Thinking along that line, I personally think we should keep the year as it is.

backend/internal/authn/service.go
```diff
@@ -209,20 +197,39 @@ func (as *authenticationService) VerifyOTP(ctx context.Context, sessionToken str
...
+		ctx, nil, credentials, nil, nil, authnprovidermgr.AuthUser{})
 	if svcErr != nil {
+		if svcErr.Type == serviceerror.ServerErrorType {
+			return nil, &serviceerror.InternalServerError
+		}
+		if svcErr.Code == authnprovidermgr.ErrorAuthenticationFailed.Code {
+			return nil, &ErrorOTPAuthenticationFailed
+		}
 		return nil, svcErr
 	}
```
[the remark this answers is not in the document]
ThumulaPerera: The mapping is updated

backend/internal/authnprovider/manager/model.go
```diff
@@ -0,0 +1,158 @@
...
+type providerDataJSON struct {
+	Token                     string                              `json:"token"`
+	Attributes                *authnprovidercm.AttributesResponse `json:"attributes,omitempty"`
+	IsAttributeValuesIncluded bool                                `json:"isAttributeValuesIncluded"`
+}
+
+// MarshalJSON implements json.Marshaler. Tokens are encrypted before serialization.
+func (a *AuthUser) MarshalJSON() ([]byte, error) {
+	encryptionService := encrypt.GetEncryptionService()
+
```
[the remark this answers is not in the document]
ThumulaPerera: Very soon (most likely in the same milestone as this PR) we will have an improvement where the entire flow context is encrypted before storing. Then we won't need this specific logic. Hence not addressing this

backend/internal/authnprovider/manager/init.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/asgardeo/thunder/internal/authn/passkey"
+	"github.com/asgardeo/thunder/internal/authnprovider/provider"
+	"github.com/asgardeo/thunder/internal/entity"
+)
+
+// InitializeAuthnProviderManager initializes and returns an AuthnProviderManagerInterface.
+func InitializeAuthnProviderManager(entitySvc entity.EntityServiceInterface,
+	passkeySvc passkey.PasskeyServiceInterface, otpSvc otp.OTPAuthnServiceInterface) AuthnProviderManagerInterface {
+	p := provider.InitializeAuthnProvider(entitySvc, passkeySvc, otpSvc)
+	return newAuthnProviderManager(p)
```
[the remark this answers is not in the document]
ThumulaPerera: This is a partial implementation of a feature. Docs will be added with the complete implementation

backend/internal/authnprovider/manager/interface.go
```diff
@@ -0,0 +1,41 @@
...
+	AuthenticateUser(ctx context.Context, identifiers, credentials map[string]interface{},
+		requestedAttributes *authnprovidercm.RequestedAttributes,
+		metadata *authnprovidercm.AuthnMetadata,
+		authUser AuthUser) (AuthUser, *AuthnBasicResult, *serviceerror.ServiceError)
+	GetUserAvailableAttributes(ctx context.Context,
+		authUser AuthUser) (*authnprovidercm.AttributesResponse, *serviceerror.ServiceError)
+	GetUserAttributes(ctx context.Context,
+		requestedAttributes *authnprovidercm.RequestedAttributes,
+		metadata *authnprovidercm.GetAttributesMetadata,
+		authUser AuthUser) (AuthUser, *authnprovidercm.AttributesResponse, *serviceerror.ServiceError)
```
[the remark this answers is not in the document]
ThumulaPerera: This is a partial implementation of a feature. Docs will be added with the complete implementation

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -0,0 +1,271 @@
...
+		logger:         log.GetLogger().With(log.String(log.LoggerKeyComponentName, "DefaultAuthnProvider")),
+	}
+}
+
+// Authenticate authenticates the user using the internal entity service.
+func (p *defaultAuthnProvider) Authenticate(
+	ctx context.Context,
+	identifiers, credentials map[string]interface{},
+	metadata *authnprovidercm.AuthnMetadata,
+) (*authnprovidercm.AuthnResult, *serviceerror.ServiceError) {
```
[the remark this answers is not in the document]
ThumulaPerera: This is a partial implementation of a feature. Docs will be added with the complete implementation

backend/internal/authnprovider/manager/manager.go
ThumulaPerera: calling getAttributes(...) without authenticated user can happen due to
1. programming error
2. erroneous auth flow definition (eg: having an auth assert executor before engaging any authenticating executors)

In either case, the caller (of `/flow/execute`) cannot correct the error. Hence I treated this case as a server error where we log the error. But I improved this to return a generic internal server error because adding specific error details is pointless in this case.

senthalan (approved):

PR #2322, opened 2026-04-16
Add configurable parentOuId property for OUExecutor

DonOmalVindula:
## Summary

- Add `parentOuId` property to the `OUExecutor` flow node, allowing flow authors to configure where new OUs are created in the hierarchy
- When `parentOuId` is set to a UUID, the OU is created under that specific parent
- When `parentOuId` is set to an empty string, the OU is created at root level (no parent)
- When the property is omitted, existing behavior is preserved (uses `defaultOUID` from RuntimeData)

Resolves #2278

## Test plan

- [x] Unit tests covering all three `parentOuId` scenarios (specific UUID, empty string, omitted)
- [x] Unit tests verifying `parentOuId` overrides `defaultOUID` from RuntimeData
- [x] Manual E2E: verified root-level OU creation with `parentOuId: ""`
- [x] Manual E2E: verified child OU creation with `parentOuId: "<uuid>"`
- [x] Existing tests pass (backward compatible)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced how organization unit parent IDs are determined, now supporting property-based configuration with improved fallback behavior to default values

* **Tests**
  * Added comprehensive test coverage for organization unit creation across multiple parent unit specification scenarios, including custom values, empty values, and default fallback cases

### DIFF
--- a/backend/internal/flow/executor/ou_executor.go
+++ b/backend/internal/flow/executor/ou_executor.go
@@ -20,6 +20,7 @@ package executor
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/asgardeo/thunder/internal/flow/common"
 	"github.com/asgardeo/thunder/internal/flow/core"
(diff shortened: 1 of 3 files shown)


backend/internal/flow/executor/ou_executor.go
```diff
@@ -96,7 +97,11 @@ func (o *ouExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse, e
 	}
 
 	// Create the OU using the OU service.
-	ouRequest := o.getOrganizationUnitRequest(ctx)
+	ouRequest, err := o.getOrganizationUnitRequest(ctx)
+	if err != nil {
+		logger.Error("Invalid node properties", log.String("error", err.Error()))
```
ThumulaPerera: Right now this error makes sense because the only possible error thrown from getOrganizationUnitRequest(...) is related to having invalid node properties. But in future if another error is added to getOrganizationUnitRequest(...), this message becomes invalid. IMO this should be a more generic error. WDYT?

DonOmalVindula: Yes agreed. Will update the error message to be more generic

ThumulaPerera (approved):

Issue #2252, opened 2026-04-08
Cache package performance improvement


rajithacharith:
During the performance tests for Redis cache, we identified that for some cases inmemory cache performance is lower than without even a cache in single replica scenario. 

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-65cfb75f-7fff-b796-079e-a783049a8614"><h4 dir="ltr" style="line-height:1.38;margin-top:14pt;margin-bottom:4pt;"><span style="font-size:12pt;font-family:Arial,sans-serif;color:#666666;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;">OAuth Authorization Code Grant (ac)</span></h4><div dir="ltr" style="margin-left:0pt;" align="left">
Metric | S1 Disabled | S2 In-Memory | S3 Redis @ 10 | S4 Redis @ 25 | S5 Redis @ 50
-- | -- | -- | -- | -- | --
Total Requests | 916 | 431 | 1220 | 596 | 498
Throughput (req/s) | 110.4 | 80.4 | 122.8 | 91.9 | 89.0
Avg Latency (ms) | 37.4 | 35.5 | 40.2 | 85.5 | 152.3
p50 (ms) | 17 | 13 | 20 | 47 | 105
p90 (ms) | 115 | 108 | 118 | 193 | 368
p95 (ms) | 130 | 118 | 132 | 289 | 466
p99 (ms) | 165 | 263 | 152 | 437 | 686
Max (ms) | 200 | 408 | 272 | 720 | 791
Error Rate | 0.00% | 0.00% | 0.00% | 0.00% | 0.00%

</div></b>

Following was identied when analyzed how the locks are used in cache module

The cache package has **4 critical lock-related performance issues** that significantly impact flow execution performance, particularly for in-memory cache operations. These issues cause unnecessary lock contention, longer lock hold times, and redundant locking layers.

---

### Issue 1: Double Locking (Highest Priority - 50% Performance Impact)

**Location**: 
- Wrapper: `backend/internal/system/cache/cache.go` (lines 54-65, 72-80, 88-94, 110-116, 143-148)
- Implementation: `backend/internal/system/cache/inmemorycache.go` (lines 173, 223, 302, 310)

**Problem**: 
The `Cache[T]` wrapper class adds an unnecessary synchronization layer on top of the implementation's own locks:

```go
[...]

PR #2326, opened 2026-04-16
Refactor in-memory cache implementation to improve performance

KaveeshaPiumini:
### Purpose
This pull request refactors the cache implementation to improve concurrency safety and performance. The main changes involve removing unnecessary locking from the generic `Cache` wrapper and making cache statistics in the in-memory cache (`inMemoryCache`) thread-safe by using atomic counters.

**Thread safety and performance improvements:**

* Removed the `sync.RWMutex` lock from the generic `Cache` struct and eliminated all locking/unlocking in its methods, relying on the underlying cache implementation for concurrency control. [[1]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L24) [[2]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L46) [[3]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L60-L62) [[4]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L74-L76) [[5]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L92-L94) [[6]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L111-L113) [[7]](diffhunk://#diff-e58e60779f1960804a306927f19b02c605f886e5cd839fa792e995e5af0e9e18L138-L139)

**Atomic counters for cache statistics:**

* Replaced plain `int64` fields for `hitCount`, `missCount`, and `evictCount` in `inMemoryCache` with `atomic.Int64` to ensure thread-safe updates. All increments and resets now use atomic operations. [[1]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373R26) [[2]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L97-R100) [[3]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L263-R272) [[4]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L280-R281) [[5]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L325-R328) [[6]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L394-R397) [[7]](diffhunk://#diff-bb8949ed5be5609456416529313c6dc071ced240453c6e3147f78120ef6bd373L414-R417)

**Cache statistics calculation:**

* Updated the `GetStats` method in `inMemoryCache` to read statistics using atomic loads, ensuring accurate and safe concurrent access.

These changes ensure that cache operations and statistics are safe in concurrent environments without unnecessary locking overhead.

### Related Issues
- https://github.com/asgardeo/thunder/issues/2252

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * Streamlined cache synchronization and reduced wrapper-level locking around cache operations.
  * Switched cache counters to atomic updates for more accurate, concurrent-safe metrics.
  * Cache statistics are now reported without holding locks for the whole computation, improving responsiveness and accuracy.

### DIFF
--- a/backend/internal/system/cache/cache.go
+++ b/backend/internal/system/cache/cache.go
@@ -21,7 +21,6 @@ package cache
 
 import (
 	"context"
-	"sync"
 
 	"github.com/asgardeo/thunder/internal/system/log"
 )
@@ -43,7 +42,6 @@ type Cache[T any] struct {
 	enabled   bool
 	cacheName string
 	cacheImpl CacheInterface[T]
-	mu        sync.RWMutex
 }
 
 // GetName returns the name of the cache.
@@ -57,9 +55,6 @@ func (c *Cache[T]) Set(ctx context.Context, key CacheKey, value T) error {
 		log.String("cacheName", c.cacheName))
 
 	if c.IsEnabled() && c.cacheImpl.IsEnabled() {
-		c.mu.Lock()
-		defer c.mu.Unlock()
-
 		if err := c.cacheImpl.Set(ctx, key, value); err != nil {
 			logger.Warn("Failed to set value in the cache", log.String("key", key.ToString()), log.Error(err))
 		}
@@ -71,9 +66,6 @@ func (c *Cache[T]) Set(ctx context.Context, key CacheKey, value T) error {
 // Get retrieves a value from the cache.
 func (c *Cache[T]) Get(ctx context.Context, key CacheKey) (T, bool) {
 	if c.IsEnabled() && c.cacheImpl.IsEnabled() {
-		c.mu.RLock()
-		defer c.mu.RUnlock()
-
 		if value, found := c.cacheImpl.Get(ctx, key); found {
 			return value, true
 		}
@@ -89,9 +81,6 @@ func (c *Cache[T]) Delete(ctx context.Context, key CacheKey) error {
 		log.String("cacheName", c.cacheName))
 
 	if c.IsEnabled() && c.cacheImpl.IsEnabled() {
-		c.mu.Lock()
-		defer c.mu.Unlock()
-
 		if err := c.cacheImpl.Delete(ctx, key); err != nil {
 			logger.Warn("Failed to delete value from the cache", log.String("key", key.ToString()), log.Error(err))
 		}
@@ -107,9 +96,6 @@ func (c *Cache[T]) Clear(ctx context.Context) error {
 
 	if c.IsEnabled() && c.cacheImpl.IsEnabled() {
 		logger.Debug("Clearing all entries in the cache")
-
-		c.mu.Lock()
-		defer c.mu.Unlock()
 
 		if err := c.cacheImpl.Clear(ctx); err != nil {
 			logger.Warn("Failed to clear the cache", log.Error(err))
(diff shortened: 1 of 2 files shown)


backend/internal/system/cache/cache.go
```diff
@@ -57,9 +55,6 @@ func (c *Cache[T]) Set(ctx context.Context, key CacheKey, value T) error {
 		log.String("cacheName", c.cacheName))
 
 	if c.IsEnabled() && c.cacheImpl.IsEnabled() {
-		c.mu.Lock()
```
senthalan, line 60: Does removing the complete lock is the option here ?

shouldn't we have the lock in place for the specific key rather than locking completely ?

senthalan: Sorry this is the generic cache function. So we don't need the lock here. But we need key specific cache in the inmemorycache ?

backend/internal/system/cache/inmemorycache.go
```diff
@@ -94,9 +95,9 @@ type inMemoryCache[T any] struct {
 	size           int
 	ttl            time.Duration
 	evictionPolicy evictionPolicy
-	hitCount       int64
-	missCount      int64
-	evictCount     int64
+	hitCount       atomic.Int64
+	missCount      atomic.Int64
+	evictCount     atomic.Int64
```
senthalan, line 100: What is the usecase of these counts?

backend/internal/system/cache/cache.go
KaveeshaPiumini: per-key locking isn't viable for inMemoryCache because every operation touches multiple shared data structures simultaneously — the cache map, the accessOrder linked list, and the lfuHeap. For example, a single Get call; reads from the map, moves an element to the front of the accessOrder list, fixes the position of an item in the lfuHeap. All three must be updated atomically. 

A per-key lock would only guard the value, not the eviction metadata, so two concurrent Get calls on different keys could still race on the shared linked list or heap.

A global mutex is the correct design for this data structure. This PR come from eliminating the redundant second mutex in the wrapper and making the metrics counters lock-free with atomics.

backend/internal/system/cache/inmemorycache.go
KaveeshaPiumini: These counts are used in the GetStats() method. Currently GetStats() is only consumed by unit tests and there's no active production caller.

CC: @ThaminduDilshan

ThaminduDilshan: These were initially added to get some stats on the cache performance. We can have some runtime flag to enable a code block that periodically calls this GetStats() method and print some cache status. This will be helpful to evaluate how much effective a certain cache is during a perf test.
However the code wasn't improved further to support the use case

ThaminduDilshan (approved):

Issue #2347, opened 2026-04-20
Implement new crypto interfaces and migrate both Config and Runtime encrypt/decrypt usages


JeethJJ:
Introduce the new crypto provider interfaces into backend wiring and migrate all encrypt/decrypt usage paths to provider-based calls for both config and runtime flows, without changing existing behaviour. 

Config encrypt migration:
Migrate callers from current encrypt singleton to ConfigCryptoProvider.Encrypt and ConfigCryptoProvider.Decrypt.

Runtime encrypt migration:
Migrate runtime payload encryption/decryption paths to RuntimeCryptoProvider.Encrypt and RuntimeCryptoProvider.Decrypt.

Parent issue : https://github.com/asgardeo/thunder/issues/2047

Issue #2352, opened 2026-04-20
Remove legacy crypto service wiring and dead code after provider migration


JeethJJ:
This issue cleans up the legacy crypto implementation after all runtime consumers have been migrated to the new provider interfaces. It removes old initialisation paths, unused helper packages, and obsolete service wiring so the application has a single crypto integration model going forward.

- Remove old bootstrap wiring for legacy crypto services once all consumers are migrated.
- Delete or isolate obsolete packages/utilities.
- Clean constructor signatures and tests.

Parent issue : https://github.com/asgardeo/thunder/issues/2047

JeethJJ:
Remove EncryptString[1] and DecryptString[2].

[1] https://github.com/asgardeo/thunder/blob/6790e1f115fb0ae946d0de5016fb3bc910639e6e/backend/internal/system/crypto/encrypt/service.go#L228
[2] https://github.com/asgardeo/thunder/blob/6790e1f115fb0ae946d0de5016fb3bc910639e6e/backend/internal/system/crypto/encrypt/service.go#L222

PR #2364, opened 2026-04-20
Introduce ConfigCryptoProvider Interface and adapt it's usage

JeethJJ:
### Summary
This PR introduces config-time crypto abstraction by adding ConfigCryptoProvider and migrating existing symmetric encryption call sites to use the interface instead of calling concrete string-based helpers directly.

### What changed
1. Added config-time crypto contract:
- provider.go
  - Added ConfigCryptoProvider with Encrypt and Decrypt methods.
  - Added RuntimeCryptoProvider placeholder for the next phase.

2. Updated encryption service to implement the provider contract:
- service.go
  - Added provider-compatible Encrypt and Decrypt methods.
  - Kept EncryptString and DecryptString behavior intact through internal helpers.

3. Migrated flow context token crypto to provider calls:
- model.go
  - ToEngineContext now accepts request context and uses provider Decrypt for token restoration.
  - FromEngineContext uses provider Encrypt for token persistence.
  - Engine context propagation remains intact via ToEngineContext(ctx, graph).

4. Migrated auth provider manager token crypto to provider calls:
- model.go
  - MarshalJSON and UnmarshalJSON now use provider Encrypt and Decrypt with context.Background() internally.

5. Migrated common secret property crypto to provider calls:
- property.go
  - GetValue and Encrypt now use provider Decrypt and Encrypt with context.Background().

6. Updated tests for signature and behavior changes:
- service_test.go
- model_test.go
- store_test.go
- model_test.go

### Scope notes
- This PR is config-time crypto only.
- Runtime crypto provider remains a placeholder and is intentionally not implemented in this change.

Issue : https://github.com/asgardeo/thunder/issues/2347

## Summary by CodeRabbit

* **Refactor**
  * Encryption APIs updated to accept context and operate on binary payloads; system components now propagate request context into execution contexts.
  * Secret/property handling switched to the new encryption API and stores ciphertext as binary-backed strings.

* **Tests**
  * Test suites updated to exercise the new context-aware encryption/decryption signatures and serialization round-trips.

### DIFF
--- a/backend/internal/system/crypto/config/service.go
+++ b/backend/internal/system/crypto/config/service.go
@@ -16,10 +16,11 @@
  * under the License.
  */
 
-// Package encrypt provides cryptographic functionality with algorithm agility.
-package encrypt
+// Package config provides cryptographic functionality with algorithm agility.
+package config
 
 import (
+	"context"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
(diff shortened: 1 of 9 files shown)


backend/internal/system/crypto/config/service.go
```diff
@@ -220,13 +238,13 @@ func (cs *EncryptionService) getKeyForDecrypt(kid string) []byte {
 // EncryptString encrypts the given plaintext string and returns a
 // JSON string containing the encrypted data.
 func (cs *EncryptionService) EncryptString(plaintext string) (string, error) {
```
rajithacharith: Do we have any usage of EncryptString and DecryptString methods?

JeethJJ: No, they need to be removed. They will be removed with issue [1]

[1] https://github.com/asgardeo/thunder/issues/2352

rajithacharith (approved):

Issue #2049, opened 2026-03-30
Refactor errors to use I18n errors


rajithacharith:
### Current Limitation

Currently the ServiceError struct is deprecated and the new approach is to use I18n Errors. 

### Suggested Improvement

Replace existing serviceError usage with i18n errors

PR #2371, opened 2026-04-20
Refactor error handling to use I18nServiceError for improved internationalization support

KaveeshaPiumini:
### Purpose
Replaces all usages of the deprecated ServiceError type with I18nServiceError across the Thunder backend service layer. This completes the i18n error handling migration to enable frontend internationalization of error messages.

---
The JSON shape of API error responses has changed across all Thunder backend endpoints.

Before (ServiceError):
```
{
  "code": "IDP-1001",
  "type": "client_error",
  "error": "Identity provider not found",
  "error_description": "The requested identity provider could not be found"
}
```

After (I18nServiceError → I18nErrorResponse):
```
{
  "code": "IDP-1001",
  "message": {
    "key": "error.idpservice.idp_not_found",
    "defaultValue": "Identity provider not found"
  },
  "description": {
    "key": "error.idpservice.idp_not_found_description",
    "defaultValue": "The requested identity provider could not be found"
  }
}
```

Key differences:

- error (string) → message (object with key + defaultValue)
- error_description (string) → description (object with key + defaultValue)
- type field removed from the response body

#### 💥 Impact
Any client (SDK, script, frontend, integration) that reads response.error, response.error_description, or response.type from Thunder API error responses will break.
Integration tests that assert on the old error response shape must be updated (this PR updates all Thunder-owned integration tests accordingly).
Affected modules: Application, Authentication (OAuth, OIDC, GitHub, Google, OTP, Passkey), Authorization, Flow Management, Group, IDP, JWKS, Notification, OAuth DCR, OU, Resource, Role, User, User Schema, Design (Theme, Layout, Resolve).

#### 🔄 Migration Guide
Update any code that reads error fields from Thunder API responses:

- const message = error.error;
- const description = error.error_description;
+ const message = error.message?.defaultValue ?? error.message;
+ const description = error.description?.defaultValue ?? error.description;
If you support i18n, use the key field to look up a translated string and fall back to defaultValue: const message = t(error.message.key, { defaultValue: error.message.defaultValue });

Remove any code that reads error.type — this field is no longer present in the response.

---

### Approach
Replaced ServiceError struct usages with I18nServiceError across all error_constants.go, service.go, handler.go, and related files (357 files, ~18k lines changed).
Updated all handler handleError functions to use WriteI18nErrorResponse instead of WriteErrorResponse.
Regenerated all mocks via make mockery to reflect updated interface signatures.
Updated all integration tests to assert against the new I18nErrorResponse shape (message/description objects).
Removed the deprecated linter suppression in .golangci.yml that was previously silencing ServiceError usage warnings.
### Related Issues
- https://github.com/asgardeo/thunder/issues/2049

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/system/error/serviceerror/error.go
```diff
@@ -43,16 +43,21 @@ type ServiceError struct {
 	ErrorDescription string           `json:"error_description,omitempty"`
 }
 
-// CustomServiceError creates a new service error based on an existing error with custom description.
-func CustomServiceError(svcError ServiceError, errorDesc string) *ServiceError {
-	err := &ServiceError{
+// CustomI18nServiceError creates a new i18n service error based on an existing error with a custom description.
+// The custom description is used as the DefaultValue of the ErrorDescription field,
+// preserving the original translation key.
+func CustomI18nServiceError(svcError I18nServiceError, errorDesc string) *I18nServiceError {
```
ThaminduDilshan: With this, we can get rid of i18n specific implementations like `withI18n` right? We don't need `CustomI18nServiceError`, rather use `CustomServiceError`

ThaminduDilshan: In order to properly support custom service errors, `customServiceError()` function should accept the i18n key as well right? Current implementation uses a single key with multiple values, which will display a different translated value for languages other than english

ThaminduDilshan: Also IMO we can change `I18nServiceError` to `ServiceError` now. We don't need to keep a separate i18n variant anymore

backend/internal/authn/common/error_constants.go
```diff
@@ -21,95 +21,167 @@ package common
...
 
 // APIErrorInvalidRequestFormat is returned when the request body is malformed.
-var APIErrorInvalidRequestFormat = apierror.ErrorResponse{
-	Code:        "AUTHN-1000",
-	Message:     "Invalid request format",
-	Description: "The request body is malformed or contains invalid data",
+var APIErrorInvalidRequestFormat = apierror.I18nErrorResponse{
+	Code: "AUTHN-1000",
+	Message: core.I18nMessage{
+		Key: "error.authncredservice.invalid_request_format", DefaultValue: "Invalid request format",
```
ThaminduDilshan:
```suggestion
		Key: "error.authncredservice.invalid_request_format", 
		DefaultValue: "Invalid request format",
```

backend/internal/authnprovider/provider/error_constants.go
```diff
@@ -0,0 +1,38 @@
...
+
+package provider
+
+import (
+	authnprovidercm "github.com/asgardeo/thunder/internal/authnprovider/common"
+	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+	"github.com/asgardeo/thunder/internal/system/i18n/core"
+)
+
+var errorSystemError = serviceerror.I18nServiceError{
```
ThaminduDilshan: Can't we simply use server wide InternalServerError here too?
@ThumulaPerera Hope we don't need to define specific server error for this case.

backend/internal/authnprovider/provider/rest_authn_provider.go
```diff
@@ -156,11 +153,17 @@ func (p *restAuthnProvider) decodeError(body io.Reader, statusCode int) *service
...
-	return &serviceerror.ServiceError{
-		Type:             errorType,
-		Code:             apiErr.Code,
-		Error:            apiErr.Message,
-		ErrorDescription: apiErr.Description,
+	return &serviceerror.I18nServiceError{
+		Type: errorType,
+		Code: apiErr.Code,
+		Error: core.I18nMessage{
+			Key:          "error.authnproviderservice." + apiErr.Code,
```
ThaminduDilshan, line 160: @ThumulaPerera api error is returned from external authentication provider right? Will this key work in that case? 
I think approach here would be to ask dev to create i18n keys when configuring external auth service

backend/internal/application/handler.go
```diff
@@ -47,12 +47,12 @@ func (ah *applicationHandler) HandleApplicationPostRequest(w http.ResponseWriter
 
 	appRequest, err := sysutils.DecodeJSONBody[model.ApplicationRequest](r)
 	if err != nil {
-		errResp := apierror.ErrorResponse{
+		errResp := apierror.I18nErrorResponse{
```
ThaminduDilshan: Similar to I18nServiceError, we can keep a single `ErrorResponse` here as well right?

backend/internal/application/service.go
```diff
@@ -1151,7 +1151,7 @@ func (as *applicationService) validateThemeID(themeID string) *serviceerror.Serv
 
 	exists, svcErr := as.themeMgtService.IsThemeExist(themeID)
 	if svcErr != nil {
-		return svcErr
+		return &serviceerror.InternalServerError
```
ThaminduDilshan: Is this expected? With this, server will return a internal server error for all type of errors from underline theme mgt or layout service.
If so at least we should log the error here. Otherwise details from underline service will be missed for troubleshooting

backend/internal/authn/handler.go
```diff
@@ -46,7 +46,7 @@ func (ah *authenticationHandler) HandleCredentialsAuthRequest(w http.ResponseWri
 	ctx := r.Context()
 	authRequestPtr, err := sysutils.DecodeJSONBody[AuthenticateWithCredentialsRequestDTO](r)
 	if err != nil {
-		sysutils.WriteErrorResponse(w, http.StatusBadRequest, common.APIErrorInvalidRequestFormat)
+		sysutils.WriteI18nErrorResponse(w, http.StatusBadRequest, common.APIErrorInvalidRequestFormat)
```
ThaminduDilshan: Same here. Should be able to rename `WriteI18nErrorResponse()` and get rid of `WriteErrorResponse`

backend/internal/authz/error_constants.go
```diff
@@ -18,15 +18,24 @@
...
 
 // Server errors for authorization operations.
 var (
 	// ErrorAuthorizationFailed is the error returned when authorization evaluation fails.
 	ErrorAuthorizationFailed = serviceerror.ServiceError{
-		Type:             serviceerror.ServerErrorType,
-		Code:             "AZ-5000",
-		Error:            "Authorization failed",
-		ErrorDescription: "An unexpected error occurred while evaluating authorization",
+		Type: serviceerror.ServerErrorType,
```
ThaminduDilshan, line 30: There are some package where we have defined package specific server error constants instead of using system wide server error constant. We should modify them also to reuse the system wide server error.
@KaveeshaPiumini @rajithacharith shall we track this also with the same effort (with a separate PR)?

ThaminduDilshan: My previous comment on authn provider server error could also be related to this

backend/internal/design/resolve/service.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan: Is this intentional?

backend/internal/design/resolve/service.go
```diff
@@ -95,7 +95,7 @@ func (drs *designResolveService) ResolveDesign(
 		if svcErr.Code == application.ErrorApplicationNotFound.Code {
 			return nil, &common.ErrorApplicationNotFound
 		}
-		return nil, svcErr
+		return nil, &serviceerror.InternalServerError
```
ThaminduDilshan: Let's log error details. Otherwise we'll loose information as we're only returning a generic error.
Better to check other places too

backend/internal/idp/declarative_resource_test.go
```diff
@@ -1,7 +1,6 @@
 /*
  * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
  *
- * WSO2 LLC. licenses this file to you under the Apache License,
```
ThaminduDilshan, line 4: Hope this is not intentional?

backend/internal/idp/service_test.go
```diff
@@ -1,7 +1,6 @@
 /*
  * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
  *
- * WSO2 LLC. licenses this file to you under the Apache License,
```
ThaminduDilshan, line 4: Same here. Can you check whether there are more such places?

backend/internal/notification/mgt_service.go
```diff
@@ -74,7 +74,7 @@ func (s *notificationSenderMgtService) CreateSender(
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "NotificationSenderMgtService"))
 	logger.Debug("Creating notification sender", log.String("name", sender.Name))
 
-	if err := declarativeresource.CheckDeclarativeCreate(); err != nil {
+	if err := declarativeresource.CheckDeclarativeCreateI18n(); err != nil {
```
ThaminduDilshan, line 77: Should be able to rename CheckDeclarativeCreateI18n to CheckDeclarativeCreate and any such similar i18n specific handling.
Okay to split these changes into a second PR

backend/internal/notification/utils.go
```diff
@@ -60,7 +61,10 @@ func validateMessageNotificationSender(sender common.NotificationSenderDTO) *ser
 
 	if err := validateMessageNotificationSenderProperties(sender); err != nil {
 		svcErr := ErrorInvalidRequestFormat
-		svcErr.ErrorDescription = err.Error()
+		svcErr.ErrorDescription = core.I18nMessage{
+			Key:          ErrorInvalidRequestFormat.ErrorDescription.Key,
+			DefaultValue: err.Error(),
```
ThaminduDilshan, line 66: This will also lead to having inconsistent i18n keys right?

backend/internal/system/declarative_resource/errors.go
```diff
@@ -29,13 +29,13 @@ var (
 	ErrorDeclarativeResourceCreateOperation = serviceerror.ServiceError{
 		Type:             serviceerror.ClientErrorType,
 		Code:             "DCR-1001",
-		Error:            "Declarative resource create operation is not allowed",
-		ErrorDescription: "Creating declarative resources is not permitted",
+		Error:            core.I18nMessage{DefaultValue: "Declarative resource create operation is not allowed"},
```
ThaminduDilshan: Here we haven't define i18n keys. Is it intentional?
Check other places too

backend/internal/system/error/serviceerror/error.go
```diff
@@ -33,17 +33,9 @@ const (
...
-// Deprecated: Use I18nServiceError instead for new services or when migrating existing services to i18n.
-type ServiceError struct {
-	Code             string           `json:"code"`
-	Type             ServiceErrorType `json:"type"`
-	Error            string           `json:"error"`
-	ErrorDescription string           `json:"error_description,omitempty"`
-}
-
-// CustomServiceError creates a new service error based on an existing error with custom description.
+// CustomServiceError creates a new i18n service error based on an existing error with a custom description.
```
ThaminduDilshan:
```suggestion
// CustomServiceError creates a new service error based on an existing error with a custom description.
```

backend/internal/system/error/serviceerror/error.go
```diff
@@ -71,28 +66,23 @@ type I18nServiceError struct {
...
 	}
 )
 
 // Server errors
 var (
-	// InternalServerError is the error returned for unexpected server errors.
-	//
-	// Deprecated: Use InternalServerErrorWithI18n instead for new services or when migrating
-	// existing services to i18n.
+	// InternalServerError is the i18n version of InternalServerError.
```
ThaminduDilshan:
```suggestion
	// InternalServerError is the error returned for unexpected server errors.
```

backend/internal/system/error/serviceerror/error.go
ThaminduDilshan, line 96: Probably we need to convert this also into a i18n service error

backend/internal/system/i18n/mgt/declarative_resource.go
```diff
@@ -61,9 +62,12 @@ func (e *translationExporter) GetAllResourceIDs(ctx context.Context) ([]string,
 	languages, err := e.store.GetDistinctLanguages()
 	if err != nil {
 		return nil, &serviceerror.ServiceError{
-			Code:  "I18N_EXPORT_ERROR",
-			Type:  serviceerror.ServerErrorType,
-			Error: "Failed to fetch translation languages for export",
+			Code: "I18N_EXPORT_ERROR",
```
ThaminduDilshan, line 65: @rajithacharith declarative resource error code patterns are different from other packages. Also we haven't defined a error description.

Is this done intentionally? If not shall we track this and get unified with other error codes

backend/internal/user/handler.go
```diff
@@ -310,12 +311,14 @@ func (uh *userHandler) HandleUserPostByPathRequest(w http.ResponseWriter, r *htt
...
 	if err != nil {
-		errResp := apierror.ErrorResponse{
-			Code:        ErrorInvalidRequestFormat.Code,
-			Message:     ErrorInvalidRequestFormat.Error,
-			Description: "Failed to parse request body: " + err.Error(),
+		errResp := apierror.I18nErrorResponse{
+			Code:    ErrorInvalidRequestFormat.Code,
+			Message: ErrorInvalidRequestFormat.Error,
+			Description: core.I18nMessage{
+				DefaultValue: "Failed to parse request body: " + err.Error(),
```
ThaminduDilshan, line 319: We don't use a i18n key here right?

backend/internal/system/error/serviceerror/error.go
KaveeshaPiumini: updated the usages of customServiceErrors to have custom keys where the errors are defined

KaveeshaPiumini: updated

backend/internal/authn/common/error_constants.go
KaveeshaPiumini: updated

backend/internal/application/handler.go
KaveeshaPiumini: updated

backend/internal/authn/handler.go
KaveeshaPiumini: updated

backend/internal/design/resolve/service.go
KaveeshaPiumini: reverted

backend/internal/idp/declarative_resource_test.go
KaveeshaPiumini: reverted

backend/internal/idp/service_test.go
KaveeshaPiumini: reverted

backend/internal/system/error/serviceerror/error.go
KaveeshaPiumini: updated

backend/internal/application/service.go
```diff
@@ -1151,7 +1152,7 @@ func (as *applicationService) validateThemeID(themeID string) *serviceerror.Serv
...
```
rajithacharith: We need to check the service error and translate it to I18nError, without directly using Internal Server Error. 
If the error is a client error, then the behaviour should be pass the client error to the user. This will mask the internal server error

backend/internal/system/error/serviceerror/error.go
KaveeshaPiumini: updated

backend/internal/application/service.go
KaveeshaPiumini: This was unintentional. I refactored the code package by package, and since the theme package was updated later, it ended up changing to match the return type there. I’ve now reverted this change.

backend/internal/application/service.go
KaveeshaPiumini: This was unintentional. I refactored the code package by package, and since the theme package was updated later, it ended up changing to match the return type there. I’ve now reverted this change.

backend/internal/design/resolve/service.go
KaveeshaPiumini: reverted

backend/internal/notification/utils.go
KaveeshaPiumini: Following places were updated.

File | Change
-- | --
backend/internal/notification/utils.go | ErrorInvalidRequestFormat.ErrorDescription.Key → "error.notificationservice.invalid_request_format_description"
backend/internal/idp/utils.go | ErrorInvalidIDPProperty.ErrorDescription.Key → "error.idpservice.invalid_idp_property_description" (×5) and ErrorUnsupportedIDPProperty.ErrorDescription.Key → "error.idpservice.unsupported_idp_property_description"
backend/internal/system/export/service.go | ErrorNoResourcesFound.ErrorDescription.Key → "error.exportservice.no_resources_found_description"
backend/internal/userschema/service.go | err.ErrorDescription.Key → "error.userschemaservice.invalid_user_schema_request_description"

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; color: rgb(204, 204, 204); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(33, 37, 43); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">The <code style="font-family: monospace; color: rgb(209, 154, 102); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">resource/service.go</code> uses of <code style="font-family: monospace; color: rgb(209, 154, 102); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">.ErrorDescription.Key</code> were left untouched — those are correct because their <code style="font-family: monospace; color: rgb(209, 154, 102); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">DefaultValue</code> is <code style="font-family: monospace; color: rgb(209, 154, 102); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">fmt.Sprintf</code> of the same error's format string (<code style="font-family: monospace; color: rgb(209, 154, 102); background-color: rgba(255, 255, 255, 0.1); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">%s</code> placeholder), so key and value are consistent.</p>

backend/internal/system/declarative_resource/errors.go
KaveeshaPiumini: Updated in following cases

File | Changes
-- | --
system/declarative_resource/errors.go | Added Key fields to ErrorDeclarativeResourceCreate/Update/DeleteOperation (keys matching their I18nError counterparts)
oauth/oauth2/dcr/error_constants.go | Added Key fields to all 6 error constants using "error.dcr.<name>" format
group/handler.go | Added Key: "error.groupservice.invalid_request_format_description" to 4 inline Description fields
userschema/handler.go | Added Key: "error.userschemaservice.invalid_request_format_description" to 2 inline Description fields

backend/internal/user/handler.go
KaveeshaPiumini: updated to use i18n key

backend/internal/application/service.go
```diff
@@ -1786,9 +1786,10 @@ func (as *applicationService) deleteApplicationCertificate(
 	if certErr := as.certService.DeleteCertificateByReference(
 		ctx, cert.CertificateReferenceTypeApplication, appID); certErr != nil {
 		if certErr.Type == serviceerror.ClientErrorType {
-			errorDescription := "Failed to delete application certificate: " +
-				certErr.ErrorDescription
-			return serviceerror.CustomServiceError(ErrorCertificateClientError, errorDescription)
+			return serviceerror.CustomServiceError(ErrorCertificateClientError, core.I18nMessage{
+				Key:          "error.applicationservice.delete_certificate_failed_description",
+				DefaultValue: "Failed to delete application certificate: " + certErr.ErrorDescription.DefaultValue,
```
ThaminduDilshan, line 1791: This will not translate correctly right? We'll not be having a fixed value here.
Let's address this separately

ThaminduDilshan (approved):

KaveeshaPiumini:
@coderabbitai, 

create an issue to track the required changes and concerns mentioned in unresolved comments in this PR

backend/internal/authnprovider/provider/error_constants.go
KaveeshaPiumini: I have updated to use ServiceError here.

PR #2377, opened 2026-04-21
Pushed authorization request (PAR) support

### LINKED ISSUE #2363: Pushed Authorization Requests for Thunder (RFC 9126)

Standard OAuth 2.0 authorization requests pass all parameters through the browser as URL query parameters, exposing sensitive data (scopes, redirect URIs, claims) to user-agent logs, referrer headers, and URL length limits. Thunder needs a back-channel mechanism that lets clients submit authorization parameters directly to the server, receiving an opaque `request_uri` reference for the subsequent browser redirect.

Primary:
* OAuth2 / OpenID Connect client application developers
    * Goal: Provide a back channel to send authorize oauth parameters to Authorization server.
    * Pain:  Cannot integrate Thunder with clients or platforms that mandate PAR (e.g., financial, government, or high-security environments).

Secondary:
* Platform / Identity administrators
    * Goal: Enforce more secure communication between relying party and authorization server.
    * Pain: Limited options to align Thunder with modern OAuth2 security best practices and compliance requirements.

Value to Persona:
* Add more security to standards-based oauth/oidc flow.
* Prevent leakage of PII or other regulated data.

Value to Org:
* Increases Thunder’s credibility as a modern, production-ready identity server.
* Reduces integration friction for enterprise and regulated use cases.
* Aligns with OAuth 2.0 Security Best Current Practices and OpenID Connect recommendations.

### Proposed Solution

Add support for PAR providing clients a simple means for a confidential and integrity-protected authorization request.

Refer this discussion -: https://github.com/asgardeo/thunder/discussions/2298

### Alternatives

_No response_

sacrana0:
### Purpose
Support Pushed Authorization Request feature in Thunder.

### Approach
Implementation is based on this [RFC 9126](https://datatracker.ietf.org/doc/html/rfc9126)

### Related Issues
- #2363

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added OAuth2 Pushed Authorization Request (PAR) support (RFC 9126): new PAR endpoint, single-use request_uri, client-binding, expiry, and storage.
  * New per-client flag requirePushedAuthorizationRequests and global PAR config (enable/expiry); PAR advertised in OAuth2 discovery metadata.

* **Tests**
  * End-to-end and unit tests covering PAR flows, validation, storage backends, and interoperability with PKCE and authorization code flow.


backend/internal/oauth/oauth2/par/service.go
```diff
@@ -0,0 +1,174 @@
...
+	// the client that pushed the authorization request.
+	if parRequest.ClientID != clientID {
+		s.logger.Debug("Client ID mismatch for PAR request",
+			log.String("expected", parRequest.ClientID),
+			log.String("actual", clientID))
+		return nil, errClientIDMismatch
+	}
+
+	return &parRequest.OAuthParameters, nil
+}
```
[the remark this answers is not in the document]
sacrana0: @coderabbitai I think its better if we consume the request_uri even if it is used by wrong client. First of all this scenario should not happen, other client should not be able to guess or steal request_uri of another client and it is better to delete that request_uri then keeping it as it is a security concern to keep it.

backend/internal/oauth/oauth2/authz/requestvalidator/validator.go
```diff
@@ -0,0 +1,160 @@
...
+// ValidateAuthorizationRequestParams validates the common authorization request parameters
+// shared by both the standard authorize endpoint and the PAR endpoint.
+//
+// This validates: prompt, grant_type, response_type, PKCE, nonce, and resource.
+// Callers are responsible for validating client_id and redirect_uri before calling this
+// function, since those validations have endpoint-specific error handling semantics
+// (e.g., the authorize endpoint must not redirect errors when the redirect_uri is invalid).
+//
+// Returns (errorCode, errorDescription). Empty errorCode means validation passed.
+func ValidateAuthorizationRequestParams(
```
thiva-k, line 43: Shall we move this function into a file or package(if cyclic dependency) in authz package and remove this? This package is for helper utils

backend/internal/oauth/oauth2/par/service.go
```diff
@@ -0,0 +1,174 @@
...
+		logger: log.GetLogger().With(log.String(log.LoggerKeyComponentName, "PARService")),
+	}
+}
+
+// HandlePushedAuthorizationRequest validates and stores a pushed authorization request.
+// Returns the response on success, or (errorCode, errorDescription) on failure.
+func (s *parService) HandlePushedAuthorizationRequest(
+	ctx context.Context, params map[string]string, oauthApp *appmodel.OAuthAppConfigProcessedDTO,
+) (*parResponse, string, string) {
+	// RFC 9126 §2.1: The request MUST NOT contain a request_uri parameter.
```
thiva-k: Let's not include RFC sections in the comments if not needed. Check other places as well

backend/internal/oauth/oauth2/authz/validator.go
```diff
@@ -49,131 +44,24 @@ func (av *authorizationValidator) validateInitialAuthorizationRequest(msg *OAuth
 	oauthApp *appmodel.OAuthAppConfigProcessedDTO) (bool, string, string) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "AuthorizationValidator"))
 
-	// Extract required parameters.
-	responseType := msg.RequestQueryParams[constants.RequestParamResponseType]
 	clientID := msg.RequestQueryParams[constants.RequestParamClientID]
 	redirectURI := msg.RequestQueryParams[constants.RequestParamRedirectURI]
 
+	// Validate client_id and redirect_uri first. Errors for these must NOT be sent to the
```
thiva-k: Are these comments needed?

backend/internal/oauth/oauth2/par/constants.go
```diff
@@ -0,0 +1,23 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package par provides support for OAuth 2.0 Pushed Authorization Requests (PAR) as defined in RFC 9126.
+package par
+
+// defaultPARExpiresIn is the default PAR request URI lifetime in seconds.
+const defaultPARExpiresIn int64 = 60
```
thiva-k: Do we need this? The default value declared in default.json will be used when reading the config

backend/internal/oauth/oauth2/par/store.go
```diff
@@ -0,0 +1,170 @@
...
+// Returns the request, a boolean indicating if found, and any error.
+func (s *parRequestStore) Consume(
+	ctx context.Context, randomKey string,
+) (pushedAuthorizationRequest, bool, error) {
+	dbClient, err := s.dbProvider.GetRuntimeDBClient()
+	if err != nil {
+		return pushedAuthorizationRequest{}, false, fmt.Errorf("failed to get database client: %w", err)
+	}
+
+	results, err := dbClient.QueryContext(ctx, queryGetPARRequest, randomKey, time.Now(), s.deploymentID)
```
thiva-k: Cant we wrap this get and delete in a transaction to prevent race conditions?

api/application.yaml
```diff
@@ -1139,6 +1139,11 @@ components:
           description: Whether the application is a public client (Mobile apps, SPAs, etc.), which cannot store secrets securely.
           example: false
           default: false
+        requirePushedAuthorizationRequests:
```
thiva-k, line 1146: Shall we check if all places in this file are updated with this new parameter?

sacrana0: Schema is updated but not examples. Api request examples don't need to have all the fields right?

tests/integration/testutils/oauth2_utils.go
```diff
@@ -579,6 +579,166 @@ func refreshAccessToken(clientID, clientSecret, refreshToken string, tokenAuthIn
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed to send authorization request: %w", err)
+	}
+
+	return resp, nil
+}
+
+// GenerateCodeVerifier generates a PKCE code verifier (exported for PAR tests).
+func GenerateCodeVerifier() (string, error) {
```
thiva-k, line 860: Shall we create a `ObtainAccessTokenWithPAR` function in testutils following the existing pattern?

backend/internal/oauth/oauth2/par/constants.go
sacrana0: @thiva-k is default.json the only source of default that is needed? Will there be a scenario where the config is not even in default.json?

thiva-k:
Shall we handle `require_pushed_authorization_requests` in DCR too?

api/application.yaml
thiva-k: Shall we update the examples too to be consistent

backend/internal/oauth/oauth2/par/constants.go
thiva-k: If we are introducing a config, the values in default.json(if we set) will be always used unless it is overridden by the deployment.yaml.

backend/internal/oauth/oauth2/par/store.go
sacrana0: Transaction will not make Get and Delete atomic so it will not prevent race condition. The race condition is already handled in the delete query, we are checking if rowsAffected after Del is not 0. If it is 0 that means another transaction already consumed the request_uri and we return nil.

thiva-k: Shall we then check if we can use `DELETE..RETURNING` here and remove the extra db call ?

sacrana0: Delete..returning is not universally supported, mysql doesn't support it. So its better to not use it.
We could use mutex but that will only make it atomic on one instance of thunder, if multiple instances are running the race condition will be there.

backend/internal/oauth/oauth2/par/handler.go
```diff
@@ -0,0 +1,91 @@
...
+	clientInfo := clientauth.GetOAuthClient(ctx)
+	if clientInfo == nil || clientInfo.OAuthApp == nil {
+		utils.WriteJSONError(w, oauth2const.ErrorInvalidClient, "Client authentication failed",
+			http.StatusUnauthorized, nil)
+		return
+	}
+
+	// Parse form-encoded body.
+	if err := r.ParseForm(); err != nil {
+		utils.WriteJSONError(w, oauth2const.ErrorInvalidRequest, "Failed to parse request body",
```
thiva-k, line 64: This should be a server error if the client is not in context, similar to token endpoint. The client authentication errors are handled in the middleware itself

backend/internal/oauth/oauth2/par/store.go
thiva-k: Let's not use mutex and leave it is as for now

backend/dbscripts/runtimedb/sqlite.sql
```diff
@@ -67,3 +67,14 @@ CREATE TABLE ATTRIBUTE_CACHE (
     EXPIRY_TIME DATETIME NOT NULL,
     CREATED_AT TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
 );
+
+-- Table to store pushed authorization requests (PAR)
+CREATE TABLE PAR_REQUEST (
+    REQUEST_URI VARCHAR(255) PRIMARY KEY,
```
KaveeshaPiumini: Why is the reason to choose `REQUEST_URI` as the primary key here? Do we have other unique identifiers for the `PAR_REQUEST` table?

Also, since it’s `VARCHAR(255)`, it makes the primary key relatively heavy. Do we actually need that length? If this is the only viable option, could you clarify the structure of the `REQUEST_URI` (e.g., does it have any consistent/prefix pattern)?

backend/internal/oauth/oauth2/par/redis_store.go
```diff
@@ -0,0 +1,102 @@
...
+
+// newRedisPARRequestStore creates a new Redis-backed PAR request store.
+func newRedisPARRequestStore(
+	p provider.RedisProviderInterface, validityPeriod time.Duration, deploymentID string,
+) parStoreInterface {
+	return &redisPARRequestStore{
+		client:         p.GetRedisClient(),
+		keyPrefix:      p.GetKeyPrefix(),
+		deploymentID:   deploymentID,
+		validityPeriod: validityPeriod,
```
KaveeshaPiumini: validity period should not be a responsibility of the the store to handle. It should be passed down from the service

backend/internal/oauth/oauth2/par/store.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	// requestURIRandomBytes is the number of random bytes for the request URI (32 bytes = 256 bits).
+	requestURIRandomBytes = 32
+
+	dbColumnRequestURI    = "request_uri"
+	dbColumnRequestParams = "request_params"
+)
+
+var queryInsertPARRequest = dbmodel.DBQuery{
```
KaveeshaPiumini: Better to have the queries in a seperate file

backend/internal/oauth/oauth2/par/store.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to generate request URI: %w", err)
+	}
+
+	data, err := json.Marshal(request)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal PAR request: %w", err)
+	}
+
+	expiryTime := time.Now().Add(s.validityPeriod)
```
KaveeshaPiumini: this should be passed down from the service layer. Store get only the number for exprirySeconds

backend/dbscripts/runtimedb/sqlite.sql
sacrana0: request_uri is a 32 bytes random key identify the reqeust object which is also sent in the response. There is no other unique identifier in the table. 255 characters are not needed 32 bytes will come to 43 chars in base64url, will change the size to 43.

backend/internal/oauth/oauth2/par/handler.go
sacrana0: You meant the clientInfo == nil check? have updated it now

sacrana0:
> Shall we handle `require_pushed_authorization_requests` in DCR too?

Have added it in dcr as well, lets keep parity between application and dcr endpoints.

backend/internal/oauth/oauth2/par/store.go
sacrana0: This is inconsistent in the existing codebase. Some store are following this approach and some are giving the expiry from service layer, refer auth_req_store and auth_code_store.
Will update it to get expiry from service layer.

backend/internal/oauth/oauth2/authz/requestvalidator/validator.go
sacrana0: It cannot be a file as it will create cyclic deps. Adding it as a new package within authz package.

backend/internal/system/config/config.go
```diff
@@ -187,11 +187,18 @@ type DCRConfig struct {
 	Insecure bool `yaml:"insecure" json:"insecure"`
 }
 
+// PARConfig holds the Pushed Authorization Request (RFC 9126) configuration.
+type PARConfig struct {
+	RequirePAR bool  `yaml:"require_par" json:"require_par"`
+	ExpiresIn  int64 `yaml:"expires_in" json:"expires_in"`
```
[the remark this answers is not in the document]
sacrana0: The configuration is not validated in this project. Following the same pattern I am also doing the same.

backend/internal/oauth/oauth2/par/redis_store.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	data, err := json.Marshal(request)
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal PAR request: %w", err)
+	}
+
+	if err := s.client.Set(ctx, s.parKey(randomKey), data, time.Until(expiryTime)).Err(); err != nil {
+		return "", fmt.Errorf("failed to store PAR request in Redis: %w", err)
+	}
```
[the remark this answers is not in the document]
sacrana0: the expiry time is provided from configuration and assumption is configuration are correctly set.

api/application.yaml
sacrana0: updated

backend/internal/oauth/oauth2/par/redis_store.go
```diff
@@ -0,0 +1,100 @@
...
+	"time"
+
+	"github.com/redis/go-redis/v9"
+
+	"github.com/asgardeo/thunder/internal/system/database/provider"
+)
+
+// parRedisClient abstracts the Redis commands used by the PAR store.
+type parRedisClient interface {
+	Set(ctx context.Context, key string, value any, expiration time.Duration) *redis.StatusCmd
```
KaveeshaPiumini, line 35:
```suggestion
	Set(ctx context.Context, key string, value any, expirySeconds int64) *redis.StatusCmd
```

We have been recently refactoring our code base to have the expiry time passed in seconds.(https://github.com/asgardeo/thunder/pull/2120)

sacrana0: Set method signature comes from the redis client go library itself, can't change the type from time.duration to int64.

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -179,6 +165,68 @@ func (as *authorizeService) HandleInitialAuthorizationRequest(ctx context.Contex
...
+				Message: "Failed to process authorization request",
+			}
+		}
+		return nil, &AuthorizationError{
+			Code:    oauth2const.ErrorInvalidRequest,
+			Message: "Invalid, expired, or already used request_uri",
+		}
+	}
+
+	return as.initiateFlowAndStoreRequest(ctx, oauthParams, app)
```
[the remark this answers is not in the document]
sacrana0: Review again with the latest change if this issue is still there. @coderabbitai

backend/internal/oauth/oauth2/par/store.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// requestURIRandomBytes is the number of random bytes for the request URI (32 bytes = 256 bits).
+const requestURIRandomBytes = 32
+
+// parStoreInterface defines the interface for PAR request storage.
+// Implementations operate on opaque random keys; the request_uri URN prefix is
+// added and stripped by the service layer.
+type parStoreInterface interface {
+	Store(ctx context.Context, request pushedAuthorizationRequest, expiryTime time.Time) (string, error)
```
KaveeshaPiumini:
```suggestion
	Store(ctx context.Context, request pushedAuthorizationRequest, expirySeconds int64) (string, error)
```

backend/internal/oauth/oauth2/par/redis_store.go
KaveeshaPiumini: Apologies for the confusion—I commented on the wrong file. Please refer to this instead: [https://github.com/asgardeo/thunder/pull/2377/changes#r3136101685](https://github.com/asgardeo/thunder/pull/2377/changes#r3136101685)

thiva-k (dismissed):

ThaminduDilshan (approved):

Issue #2040, opened 2026-03-30
Dynamic prompt input support


ThaminduDilshan:
## Current limitation
Executors like provisioning and attribute collector cannot return dynamic prompt inputs at runtime (this is applicable only when there's a meta). The current meta structure is fixed, so all defined inputs are always prompted, even if some are already present or not required. This limits flexibility and user experience in flow execution.

## Suggested improvement
Enable executors/ nodes to return dynamic prompt inputs (meta) at runtime. Update the meta structure so that required inputs can be determined and prompted dynamically based on the current context. Executors or some central layer should be able to update meta to avoid prompting already provided or unnecessary inputs, improving flow flexibility and UX.

ThaminduDilshan:
Let's create a design discussion for this

Dilusha-Madushan:
Here is the discussion: https://github.com/asgardeo/thunder/discussions/2321

PR #2388, opened 2026-04-22
Fix prompt node to skip already satisfied inputs in meta

Dilusha-Madushan:
### Purpose
promptNode.appendMissingInputs only checked ctx.UserInputs before marking an input as missing, while the base executor's equivalent method also checks ctx.RuntimeData and ctx.ForwardedData. This caused prompt nodes to re-request inputs (e.g. email set by a prior Google OAuth step) that the executor had already correctly satisfied, breaking flows where upstream steps populate context data.

Additionally, when verbose mode is enabled, the prompt node attached the full n.GetMeta() to the response even when only a subset of inputs were still pending. Clients received a meta describing N fields while inputs listed a smaller subset, causing form rendering mismatches.

### Approach

- Fix 1 — Context-aware missing input detection (appendMissingInputs)
After a ctx.UserInputs miss, the method now checks ctx.RuntimeData (skip if present) and ctx.ForwardedData (skip only if the value is a string — non-string values such as []common.Input stored under ForwardedDataKeyInputs must not be treated as satisfied). This aligns promptNode with the existing executor.appendMissingInputs logic.

- Fix 2 — Verbose meta trimming (trimMetaToRequestedInputs + filterMetaComponents)
A new trimMetaToRequestedInputs method builds two ref sets: allowedRefs (inputs/actions still pending in nodeResp) and knownInputActionRefs (all static inputs/actions defined on the node). It shallow-copies the top-level meta map and replaces "components" with the output of filterMetaComponents, a recursive helper that drops components whose id matches a satisfied input/action, while always preserving structural components (TEXT, BLOCK containers, etc.) and recursing into their children. The display-only node path is unchanged.

### Related Issues
- https://github.com/asgardeo/thunder/issues/2040

## Summary by CodeRabbit

* **Bug Fixes**
  * Better missing-input detection: inputs present in runtime data or forwarded data (when the forwarded value is a string) are treated as satisfied and not re-requested.

* **New Features**
  * Verbose mode now returns streamlined metadata containing only components relevant to the actually requested inputs/actions; non-map metadata is preserved and verbose=false yields no meta.

* **Tests**
  * Added tests covering input availability rules, verbose metadata trimming, and edge cases (non-string forwarded values, non-map meta, verbose off).

### DIFF
--- a/backend/internal/flow/core/prompt_node.go
+++ b/backend/internal/flow/core/prompt_node.go
@@ -146,7 +146,7 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 
 	// Include meta in the response if verbose mode is enabled
 	if ctx.Verbose && n.GetMeta() != nil {
-		nodeResp.Meta = n.GetMeta()
+		nodeResp.Meta = n.trimMetaToRequestedInputs(nodeResp.Inputs, nodeResp.Actions)
 	}
 
 	nodeResp.Status = common.NodeStatusIncomplete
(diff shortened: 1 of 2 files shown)


backend/internal/flow/core/prompt_node.go
```diff
@@ -405,3 +417,85 @@ func (n *promptNode) getActionTypeForRef(actionRef string) string {
...
+// trimMetaToRequestedInputs returns a copy of n.meta with the "components" list trimmed to only
+// include components matching the given inputs and actions (plus structural components like TEXT
+// and BLOCK containers that are not themselves inputs or actions).
+func (n *promptNode) trimMetaToRequestedInputs(inputs []common.Input, actions []common.Action) interface{} {
+	metaMap, ok := n.meta.(map[string]interface{})
+	if !ok {
+		return n.meta
+	}
+
+	allowedRefs := make(map[string]struct{})
```
ThaminduDilshan, line 430: Shall we handle input and action refs in two maps?

ThaminduDilshan: This will complicate the meta filtering logic. Hence fine to keep as it is

ThaminduDilshan (approved):

PR #2394, opened 2026-04-22
add JWE and nested JWT support for UserInfo endpoint

### LINKED ISSUE #2372: Support JWE and nested JWE in userinfo and token endpoint
**Problem Statement**

Thunder supports UserInfo responses in JSON and JWS (signed) formats. However, there is no support for encrypted UserInfo responses (JWE) or nested JWE. Thunder also supports user claims in ID token, this should also support encrypting the signed ID token.

**Goals**
- Reuse the existing application `Certificate` (JWKS/JWKS_URI) under OAuthConfig as the RP's public key source
- Extend `UserInfoConfig` with `signingAlg`, `encryptionAlg` fields
- Extend `IDTokenConfig` with `encryptionAlg` fields
- Auto-derive `ResponseType` from alg config (no need for client to set it explicitly)
- Expose supported algorithm values in the OAuth & OIDC discovery endpoint
- Update DCR to accept `userinfo_signed_response_alg`, `userinfo_encrypted_response_alg`
- On create/update of client(via application & DCR), encryption algorithm can only be selected from the list of supported algorithms.
- Support both JWE and NESTED_JWT as the userinfo response. userinfo endpoint response content-type header should reflect the actual response content type.
- JWE implementation must adhere to https://datatracker.ietf.org/doc/html/rfc7516

**Non-Goals**
- JWKS URI caching for fetched RP keys (deferred; initial implementation fetches on every request)
- `kid` hint selection (initial implementation picks first suitable key from JWKS)

**Tasks**
- [ ] #1219 
- [ ] #1275 

**Acceptance Criteria**

| # | Area | Acceptance Criterion | Pass Condition |
|---|---|---|---|
| AC-1 | Model | `UserInfoConfig` gains `SigningAlg`, `EncryptionAlg`, `EncryptionEnc` fields | Fields present in struct; serialised as `signingAlg`/`encryptionAlg`/`encryptionEnc` in JSON and `signing_alg`/`encryption_alg`/`encryption_enc` in YAML |
| AC-2 | Model | `IDTokenConfig` gains `EncryptionAlg`, `EncryptionEnc` fields | Fields present in struct; serialised as `encryptionAlg`/`encryptionEnc` in JSON and `encryption_alg`/`encryption_enc` in YAML |
| AC-3 | Model | Server-level constant sets define supported JWE key-management algs (`RSA-OAEP`, `RSA-OAEP-256`) and content-encryption algs (`A128CBC-HS256`, `A256GCM`) | Constants exist and are referenced by validation, DCR, discovery, and key-resolution — no duplication |
| AC-4a | Response Type Derivation | `EncryptionAlg` set + `SigningAlg` set → effective type is `NESTED_JWT` | UserInfo service selects `NESTED_JWT` path |
| AC-4b | Response Type Derivation | `EncryptionAlg` set, `SigningAlg` absent → effective type is `JWE` | UserInfo service selects `JWE` path |
| AC-4c | Response Type Derivation | Only `SigningAlg` set → effective type is `JWS` | UserInfo service selects `JWS` path |
| AC-4d | Response Type Derivation | Neither `SigningAlg` nor `EncryptionAlg` set → effective type is `JSON` | UserInfo service selects `JSON` path |
| AC-4e | Response Type Derivation | Caller does not need to set `ResponseType` explicitly; any explicitly set value is ignored | Application create/update does not require `ResponseType` |
| AC-5a | Key Resolution | `Certificate.Type = JWKS` (inline) → server parses the JWKS and selects the first key with `use=enc` or unspecified, matching the configured `kty` | Correct key selected from inline JWKS |
| AC-5b | Key Resolution | `Certificate.Type = JWKS_URI` → server fetches the URI at request time (no caching), parses it, selects key by same rule | Correct key fetched and selected from remote JWKS |
| AC-5c | Key Resolution | No suitable key found in JWKS | HTTP 500, `server_error` |
| AC-5d | Key Resolution | `Certificate` absent but encryption configured | HTTP 500, `server_error` |
| AC-6a | UserInfo — JWE | Builds plain JSON claims, constructs RFC 7516 compact JWE using RP public key | Response body is a valid 5-part compact JWE string |
| AC-6b | UserInfo — JWE | JWE protected header contains `alg`, `enc`, `cty: json` | Header fields present and correct |
| AC-6c | UserInfo — JWE | `Content-Type` response header is `application/jose` | Header set correctly |
| AC-7a | UserInfo — NESTED_JWT | Produces a signed JWS first, then encrypts it as a JWE outer envelope | Response body is a valid compact JWE wrapping a JWS |
| AC-7b | UserInfo — NESTED_JWT | JWE protected header contains `alg`, `enc`, `cty: JWT` | Header fields present and correct |
| AC-7c | UserInfo — NESTED_JWT | `Content-Type` response header is `application/jose` | Header set correctly |
| AC-8a | UserInfo — Content-Type | Effective type `JSON` → `Content-Type: application/json` | Handler sets correct header |
| AC-8b | UserInfo — Content-Type | Effective type `JWS` → `Content-Type: application/jwt` | Handler sets correct header |
| AC-8c | UserInfo — Content-Type | Effective type `JWE` or `NESTED_JWT` → `Content-Type: application/jose` | Handler sets correct header |
| AC-9a | ID Token Encryption | When `IDTokenConfig.EncryptionAlg` is configured, signed ID token is wrapped in a JWE using RP public key | `id_token` field in token response is a valid compact JWE |
| AC-9b | ID Token Encryption | JWE protected header contains `alg`, `enc`, `cty: JWT` | Header fields present and correct |
| AC-10a | App Create/Update Validation | `UserInfoConfig.SigningAlg` provided but not in supported JWS set | HTTP 400, `invalid_configuration` |
| AC-10b | App Create/Update Validation | `UserInfoConfig.EncryptionAlg` provided but not in supported JWE key-management set | HTTP 400, `invalid_configuration` |
| AC-10c | App Create/Update Validation | `UserInfoConfig.EncryptionAlg` set but `EncryptionEnc` absent | HTTP 400, `invalid_configuration` |
| AC-10d | App Create/Update Validation | `IDTokenConfig.EncryptionAlg` set but `EncryptionEnc` absent | HTTP 400, `invalid_configuration` |
| AC-10e | App Create/Update Validation | `EncryptionAlg` set but `Certificate` absent on the application | HTTP 400, `invalid_configuration` |
| AC-11a | DCR | `DCRRegistrationRequest` accepts `userinfo_signed_response_alg`, `userinfo_encrypted_response_alg`, `userinfo_encrypted_response_enc`, `id_token_encrypted_response_alg`, `id_token_encrypted_response_enc` | Fields parsed from request body |
| AC-11b | DCR | `DCRRegistrationResponse` echoes back all provided algorithm fields | Fields present in response |
| AC-11c | DCR | DCR registration with an unsupported algorithm value | HTTP 400, `invalid_client_metadata` (RFC 7591) |
| AC-12 | DCR Mapping | DCR fields mapped to application model: `userinfo_signed_response_alg` → `UserInfoConfig.SigningAlg`, `userinfo_encrypted_response_alg` → `UserInfoConfig.EncryptionAlg`, `userinfo_encrypted_response_enc` → `UserInfoConfig.EncryptionEnc`, `id_token_encrypted_response_alg` → `IDTokenConfig.EncryptionAlg`, `id_token_encrypted_response_enc` → `IDTokenConfig.EncryptionEnc` | Application stored with correct field values after DCR |
| AC-13a | Discovery | `OIDCProviderMetadata` gains `userinfo_signing_alg_values_supported` | Field populated from server constant set and returned in discovery response |
| AC-13b | Discovery | `OIDCProviderMetadata` gains `userinfo_encryption_alg_values_supported` | Field populated and returned |
| AC-13c | Discovery | `OIDCProviderMetadata` gains `userinfo_encryption_enc_values_supported` | Field populated and returned |
| AC-13d | Discovery | `OIDCProviderMetadata` gains `id_token_encryption_alg_values_supported` | Field populated and returned |
| AC-13e | Discovery | `OIDCProviderMetadata` gains `id_token_encryption_enc_values_supported` | Field populated and returned |
| AC-14a | RFC 7516 Compliance | JWE uses compact serialization (5 dot-separated Base64URL parts: header · encrypted-key · IV · ciphertext · tag) | Output parses as valid RFC 7516 compact JWE |
| AC-14b | RFC 7516 Compliance | JWE protected header contains at minimum `alg`, `enc`, `cty` | Header validated on each response |
| AC-14c | RFC 7516 Compliance | CEK and IV are randomly generated per request; IV is never reused | Two JWEs for same payload have different IV and ciphertext |
| AC-15a | Error Handling | Encryption configured, `Certificate` absent at request time | HTTP 500, `server_error` |
| AC-15b | Error Handling | No matching key found in JWKS for configured alg | HTTP 500, `server_error` |
| AC-15c | Error Handling | `JWKS_URI` unreachable at request time | HTTP 500, `server_error` |
| AC-15d | Error Handling | Unsupported algorithm on app create/update | HTTP 400, `invalid_configuration` |
| AC-15e | Error Handling | `EncryptionAlg` set without `EncryptionEnc` on create/update | HTTP 400, `invalid_configuration` |
| AC-15f | Error Handling | DCR request with unsupported algorithm | HTTP 400, `invalid_client_metadata` |
| AC-16a | Tests | Unit tests cover JWE construction for each supported `alg`/`enc` combination | All combinations have passing unit tests |
| AC-16b | Tests | Unit tests cover all four response type auto-derivation cases (AC-4a–d) | Four cases each have a dedicated test |
| AC-16c | Tests | Unit tests cover key selection from inline JWKS and from JWKS_URI | Both key sources tested |
| AC-16d | Tests | Unit tests cover Content-Type header selection for all four response types | All four cases asserted |
| AC-16e | Tests | Integration tests cover end-to-end UserInfo with JWE and NESTED_JWT responses | Full request-to-response flow passes |
| AC-16f | Tests | Integration tests cover ID token encryption via token endpoint | Encrypted `id_token` returned and decodable with RP private key |
| AC-16g | Tests | Integration tests cover discovery metadata fields | All new fields present in `/.well-known/openid-configuration` response |
| AC-16h | Tests | Integration tests cover DCR with encryption alg fields and validation rejection cases | Happy path and rejection cases both covered |

nandhu-kumar:
### Purpose

Adds JWE (JSON Web Encryption) support to the OIDC UserInfo endpoint, enabling clients to request
encrypted and/or signed UserInfo responses as defined in
[OpenID Connect Core §5.3.2](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse)
and [RFC 7518](https://www.rfc-editor.org/rfc/rfc7518).

Previously the UserInfo endpoint only returned a plain JSON payload or a signed JWT (JWS). This PR
adds the remaining two response types so all four modes are supported:

| Mode | Trigger |
|------|---------|
| JSON (plain) | No signing/encryption configured |
| JWS (signed) | `signingAlg` set |
| JWE (encrypted) | `encryptionAlg` + `encryptionEnc` set |
| Nested JWT (sign-then-encrypt) | Both signing and encryption configured |

### Approach

**Application model (`UserInfoConfig`)** — replaced the internal `responseType` enum field with
explicit `signingAlg`, `encryptionAlg`, and `encryptionEnc` fields. The response type is now derived
at runtime from whichever fields are populated, eliminating a redundant field that could go out of
sync. Allowed algorithms are enumerated in
`SupportedUserInfoSigningAlgs` / `SupportedUserInfoEncryptionAlgs` / `SupportedUserInfoEncryptionEncs`.

**Validation (`application/service.go`)** — `validateUserInfoConfig` enforces:
- Only supported algorithm identifiers are accepted.
- `encryptionEnc` is required when `encryptionAlg` is set (RFC 7518 default rule).
- A client JWKS or JWKS URI must be present when encryption is configured (the server needs the
  RP's public key to wrap the CEK).

**UserInfo service** — `deriveResponseType` selects the response mode. The new
`generateJWEUserInfo` and `generateNestedJWTUserInfo` methods resolve the RP's public key from the
application certificate (inline JWKS or remote JWKS URI), then delegate to `JWEService.Encrypt`.
JWKS URI fetching enforces HTTPS, caps the response body at 1 MB, and validates HTTP 200. Key
selection requires `use=enc` (RFC 7517 §4.2).

**DCR (`dcr/service.go`)** — accepts three new registration fields:
`userinfo_signed_response_alg`, `userinfo_encrypted_response_alg`,
`userinfo_encrypted_response_enc`. JWKS marshal failure now returns `server_error` immediately
instead of silently proceeding without a certificate.

**OIDC Discovery (`discovery/service.go`)** — exposes `userinfo_signing_alg_values_supported`,
`userinfo_encryption_alg_values_supported`, and `userinfo_encryption_enc_values_supported` in the
discovery document. The signing list is sourced from `pkiService.GetSupportedSigningAlgorithms()`
(consistent with `id_token_signing_alg_values_supported`).

**JWE utils (`system/jose/jwe/utils.go`)** — `epkToMap` now returns an error for unsupported key
types and curves instead of a silent `nil`, preventing ambiguous nil-map panics in callers.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests (`service_jwe_test.go`, updated `service_test.go`, `utils_test.go`, `dcr/service_test.go`)
    - [x] Integration Tests (`tests/integration/oauth/userinfo/`)

> **Note:** `UserInfoConfig.ResponseType` (serialized as `responseType`) has been removed from the
> model. Applications with this field explicitly stored will silently ignore it after migration — the
> response type is now derived dynamically. No API contract change for consumers of the UserInfo
> endpoint itself.

### Security checks
- [x] Followed secure coding standards in [WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/).
- [x] JWKS URI fetch restricted to HTTPS only; response body capped at 1 MB to prevent SSRF / OOM.
- [x] Encryption key selection restricted to `use=enc` keys per RFC 7517 §4.2.

## Summary by CodeRabbit

* **New Features**
  * Encrypted (JWE) and nested JWT userinfo responses.
  * Per-client configurable userinfo signing and encryption algorithm fields.
  * Discovery endpoint publishes supported userinfo signing/encryption values.
  * Dynamic Client Registration accepts userinfo algorithm parameters.

* **Bug Fixes / Enhancements**
  * Stronger validation of userinfo algorithm combinations and certificate requirements.
  * JWKS fetching hardened with HTTPS/SSRF checks, size limits, safer redirects, and JOSE content-type support.
  * Improved JWT/JWE handling with explicit algorithm selection.

* **Documentation**
  * Updated userinfo algorithm configuration examples.

### DIFF
--- a/api/application.yaml
+++ b/api/application.yaml
@@ -1117,9 +1117,34 @@ components:
           items:
             type: string
           description: |
-            The user attributes to include in the userinfo response. 
+            The user attributes to include in the userinfo response.
             If not specified, falls back to ID token user attributes.
           example: ["email", "name", "given_name", "family_name", "picture"]
+        signingAlg:
+          type: string
+          description: |
+            JWS algorithm used to sign the UserInfo response JWT.
+            When set (without encryptionAlg), the response is a signed JWT (JWS).
+            When set together with encryptionAlg, the response is a Nested JWT (sign-then-encrypt).
+            Omit to return a plain JSON response.
+          enum: ["RS256", "RS512", "PS256", "ES256", "ES384", "ES512", "EdDSA"]
+          example: "RS256"
+        encryptionAlg:
+          type: string
+          description: |
+            JWE key-management algorithm used to encrypt the UserInfo response.
+            Must be set together with encryptionEnc and a client certificate (JWKS).
+            When set without signingAlg, the response is a JWE over the plain JSON payload.
+            When set together with signingAlg, the response is a Nested JWT (sign-then-encrypt).
+          enum: ["RSA-OAEP", "RSA-OAEP-256"]
+          example: "RSA-OAEP-256"
+        encryptionEnc:
+          type: string
+          description: |
+            JWE content-encryption algorithm used to encrypt the UserInfo response payload.
+            Required when encryptionAlg is set.
+          enum: ["A128CBC-HS256", "A256GCM"]
+          example: "A256GCM"
 
     Certificate:
       type: object
--- a/backend/internal/application/model/constants.go
+++ b/backend/internal/application/model/constants.go
@@ -18,7 +18,12 @@
 
 package model
 
-import "errors"
+import (
+	"errors"
+
+	"github.com/asgardeo/thunder/internal/system/jose/jwe"
+	"github.com/asgardeo/thunder/internal/system/jose/jws"
+)
 
 // InboundAuthType represents the type of inbound authentication.
 type InboundAuthType string
@@ -35,9 +40,28 @@ const (
 	// UserInfoResponseTypeJSON represents the JSON userinfo response type.
 	UserInfoResponseTypeJSON UserInfoResponseType = "JSON"
 
-	// UserInfoResponseTypeJWS represents the JWS userinfo response type.
+	// UserInfoResponseTypeJWS represents the signed JWT (JWS) userinfo response type.
 	UserInfoResponseTypeJWS UserInfoResponseType = "JWS"
+
+	// UserInfoResponseTypeJWE represents the encrypted (JWE) userinfo response type.
+	UserInfoResponseTypeJWE UserInfoResponseType = "JWE"
+
+	// UserInfoResponseTypeNESTEDJWT represents the signed-then-encrypted (Nested JWT) userinfo response type.
+	UserInfoResponseTypeNESTEDJWT UserInfoResponseType = "NESTED_JWT"
 )
+
+// SupportedUserInfoSigningAlgs lists JWS algorithms supported for userinfo signing.
+var SupportedUserInfoSigningAlgs = []string{
+	string(jws.RS256), string(jws.RS512), string(jws.PS256),
+	string(jws.ES256), string(jws.ES384), string(jws.ES512),
+	string(jws.EdDSA),
+}
+
+// SupportedUserInfoEncryptionAlgs lists JWE key-management algorithms supported for userinfo encryption.
+var SupportedUserInfoEncryptionAlgs = []string{string(jwe.RSAOAEP), string(jwe.RSAOAEP256)}
+
+// SupportedUserInfoEncryptionEncs lists JWE content-encryption algorithms supported for userinfo encryption.
+var SupportedUserInfoEncryptionEncs = []string{string(jwe.A128CBCHS256), string(jwe.A256GCM)}
 
 // ApplicationNotFoundError is the error returned when an application is not found.
 var ApplicationNotFoundError error = errors.New("application not found")
--- a/backend/internal/application/model/oauth_app.go
+++ b/backend/internal/application/model/oauth_app.go
@@ -47,8 +47,11 @@ type IDTokenConfig struct {
 
 // UserInfoConfig represents the user info endpoint configuration structure.
 type UserInfoConfig struct {
-	ResponseType   UserInfoResponseType `json:"responseType,omitempty" yaml:"response_type,omitempty"`
+	ResponseType   UserInfoResponseType `json:"responseType,omitempty"   yaml:"response_type,omitempty"   jsonschema:"UserInfo response type (JSON, JWS, JWE, NESTED_JWT). Derived from algorithm fields when not explicitly set."`
 	UserAttributes []string             `json:"userAttributes,omitempty" yaml:"user_attributes,omitempty" jsonschema:"User attributes to include in userinfo response."`
+	SigningAlg     string               `json:"signingAlg,omitempty"     yaml:"signing_alg,omitempty"     jsonschema:"JWS algorithm for signed userinfo responses (e.g. RS256)."`
+	EncryptionAlg  string               `json:"encryptionAlg,omitempty"  yaml:"encryption_alg,omitempty"  jsonschema:"JWE key-management algorithm for encrypted userinfo responses (e.g. RSA-OAEP-256)."`
+	EncryptionEnc  string               `json:"encryptionEnc,omitempty"  yaml:"encryption_enc,omitempty"  jsonschema:"JWE content-encryption algorithm (e.g. A256GCM). Required when encryptionAlg is set."`
 }
 
 // OAuthTokenConfig represents the OAuth token configuration structure with access_token and id_token wrappers.
--- a/backend/internal/application/service.go
+++ b/backend/internal/application/service.go
@@ -41,6 +41,7 @@ import (
 	"github.com/asgardeo/thunder/internal/system/config"
 	serverconst "github.com/asgardeo/thunder/internal/system/constants"
 	"github.com/asgardeo/thunder/internal/system/error/serviceerror"
+	syshttp "github.com/asgardeo/thunder/internal/system/http"
 	"github.com/asgardeo/thunder/internal/system/i18n/core"
 	"github.com/asgardeo/thunder/internal/system/log"
 	"github.com/asgardeo/thunder/internal/system/security"
@@ -909,8 +910,11 @@ func toOAuthProcessedDTO(
 
 	if cfg.UserInfo != nil {
 		dto.UserInfo = &model.UserInfoConfig{
-			ResponseType:   cfg.UserInfo.ResponseType,
+			ResponseType:   model.UserInfoResponseType(cfg.UserInfo.ResponseType),
 			UserAttributes: cfg.UserInfo.UserAttributes,
+			SigningAlg:     cfg.UserInfo.SigningAlg,
+			EncryptionAlg:  cfg.UserInfo.EncryptionAlg,
+			EncryptionEnc:  cfg.UserInfo.EncryptionEnc,
 		}
 	}
 
@@ -1304,7 +1308,118 @@ func validateOAuthParamsForCreateAndUpdate(app *model.ApplicationDTO) (*model.In
 		}
 	}
 
+	// Validate userinfo encryption configuration
+	if err := validateUserInfoConfig(oauthAppConfig); err != nil {
+		return nil, err
+	}
+
 	return inboundAuthConfig, nil
+}
+
+// validateUserInfoConfig validates the UserInfo signing and encryption configuration.
+func validateUserInfoConfig(oauthConfig *model.OAuthAppConfigDTO) *serviceerror.ServiceError {
+	if oauthConfig.UserInfo == nil {
+		return nil
+	}
+	cfg := oauthConfig.UserInfo
+
+	if cfg.SigningAlg != "" && !slices.Contains(model.SupportedUserInfoSigningAlgs, cfg.SigningAlg) {
+		return serviceerror.CustomServiceError(
+			ErrorInvalidOAuthConfiguration,
+			core.I18nMessage{DefaultValue: "userinfo signing algorithm '" + cfg.SigningAlg + "' is not supported"},
+		)
+	}
+
+	if cfg.EncryptionEnc != "" && cfg.EncryptionAlg == "" {
+		return serviceerror.CustomServiceError(
+			ErrorInvalidOAuthConfiguration,
+			core.I18nMessage{DefaultValue: "encryptionAlg is required when encryptionEnc is set"},
+		)
+	}
+
+	if cfg.EncryptionAlg != "" {
+		if !slices.Contains(model.SupportedUserInfoEncryptionAlgs, cfg.EncryptionAlg) {
+			return serviceerror.CustomServiceError(
+				ErrorInvalidOAuthConfiguration,
+				core.I18nMessage{
+					DefaultValue: "userinfo encryption algorithm '" + cfg.EncryptionAlg + "' is not supported",
+				},
+			)
+		}
+		if cfg.EncryptionEnc == "" {
+			return serviceerror.CustomServiceError(
+				ErrorInvalidOAuthConfiguration,
+				core.I18nMessage{DefaultValue: "encryptionEnc is required when encryptionAlg is set"},
+			)
+		}
+		if !slices.Contains(model.SupportedUserInfoEncryptionEncs, cfg.EncryptionEnc) {
+			return serviceerror.CustomServiceError(
+				ErrorInvalidOAuthConfiguration,
+				core.I18nMessage{
+					DefaultValue: "userinfo content-encryption algorithm '" + cfg.EncryptionEnc + "' is not supported",
+				},
+			)
+		}
+		hasCert := oauthConfig.Certificate != nil && oauthConfig.Certificate.Type != ""
+		if !hasCert {
+			return serviceerror.CustomServiceError(
+				ErrorInvalidOAuthConfiguration,
+				core.I18nMessage{
+					DefaultValue: "a certificate (JWKS or JWKS_URI) is required when userinfo encryption is configured",
+				},
+			)
+		}
+		if oauthConfig.Certificate.Type == cert.CertificateTypeJWKSURI {
+			if err := syshttp.IsSSRFSafeURL(oauthConfig.Certificate.Value); err != nil {
+				log.GetLogger().Debug("JWKS URI rejected by SSRF policy",
+					log.String("uri", oauthConfig.Certificate.Value), log.Error(err))
+				return serviceerror.CustomServiceError(
+					ErrorInvalidOAuthConfiguration,
+					core.I18nMessage{DefaultValue: "JWKS URI must be a publicly reachable HTTPS URL"},
+				)
+			}
+		}
+	}
+
+	if cfg.ResponseType != "" {
+		switch cfg.ResponseType {
+		case model.UserInfoResponseTypeJWS:
+			if cfg.SigningAlg == "" {
+				return serviceerror.CustomServiceError(
+					ErrorInvalidOAuthConfiguration,
+					core.I18nMessage{DefaultValue: "signingAlg is required when responseType is JWS"},
+				)
+			}
+		case model.UserInfoResponseTypeJWE:
+			if cfg.EncryptionAlg == "" || cfg.EncryptionEnc == "" {
+				return serviceerror.CustomServiceError(
+					ErrorInvalidOAuthConfiguration,
+					core.I18nMessage{
+						DefaultValue: "encryptionAlg and encryptionEnc are required when responseType is JWE",
+					},
+				)
+			}
+		case model.UserInfoResponseTypeNESTEDJWT:
+			if cfg.SigningAlg == "" || cfg.EncryptionAlg == "" || cfg.EncryptionEnc == "" {
+				return serviceerror.CustomServiceError(
+					ErrorInvalidOAuthConfiguration,
+					core.I18nMessage{
+						DefaultValue: "signingAlg, encryptionAlg, and encryptionEnc are required when " +
+							"responseType is NESTED_JWT",
+					},
+				)
+			}
+		case model.UserInfoResponseTypeJSON:
+			// no additional requirements
+		default:
+			return serviceerror.CustomServiceError(
+				ErrorInvalidOAuthConfiguration,
+				core.I18nMessage{DefaultValue: "responseType '" + string(cfg.ResponseType) + "' is not supported"},
+			)
+		}
+	}
+
+	return nil
 }
 
 // validateRedirectURIs validates redirect URIs format and requirements.
@@ -1474,6 +1589,7 @@ func validateTokenEndpointAuthMethod(oauthConfig *model.OAuthAppConfigDTO) *serv
 	}
 
 	hasCert := oauthConfig.Certificate != nil && oauthConfig.Certificate.Type != ""
+	userInfoNeedsCert := oauthConfig.UserInfo != nil && oauthConfig.UserInfo.EncryptionAlg != ""
 
 	switch oauthConfig.TokenEndpointAuthMethod {
 	case oauth2const.TokenEndpointAuthMethodPrivateKeyJWT:
@@ -1490,7 +1606,7 @@ func validateTokenEndpointAuthMethod(oauthConfig *model.OAuthAppConfigDTO) *serv
 			})
 		}
 	case oauth2const.TokenEndpointAuthMethodClientSecretBasic, oauth2const.TokenEndpointAuthMethodClientSecretPost:
-		if hasCert {
+		if hasCert && !userInfoNeedsCert {
 			return serviceerror.CustomServiceError(ErrorInvalidOAuthConfiguration, core.I18nMessage{
 				Key:          "error.applicationservice.client_secret_cannot_have_certificate_description",
 				DefaultValue: "client_secret authentication methods cannot have a certificate",
@@ -1503,7 +1619,7 @@ func validateTokenEndpointAuthMethod(oauthConfig *model.OAuthAppConfigDTO) *serv
 				DefaultValue: "'none' authentication method requires the client to be a public client",
 			})
 		}
-		if hasCert || oauthConfig.ClientSecret != "" {
+		if (hasCert && !userInfoNeedsCert) || oauthConfig.ClientSecret != "" {
 			return serviceerror.CustomServiceError(ErrorInvalidOAuthConfiguration, core.I18nMessage{
 				Key:          "error.applicationservice.none_auth_method_cannot_have_cert_or_secret_description",
 				DefaultValue: "'none' authentication method cannot have a certificate or client secret",
@@ -1712,19 +1828,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
 		oauthInboundAuth.OAuthAppConfig.UserInfo != nil {
 		userInfoConfigInput := oauthInboundAuth.OAuthAppConfig.UserInfo
 		oauthUserInfo.UserAttributes = userInfoConfigInput.UserAttributes
-		responseType := model.UserInfoResponseType(strings.ToUpper(string(userInfoConfigInput.ResponseType)))
-
-		switch responseType {
-		case model.UserInfoResponseTypeJWS:
-			oauthUserInfo.ResponseType = responseType
-		default:
-			oauthUserInfo.ResponseType = model.UserInfoResponseTypeJSON
-		}
+		oauthUserInfo.SigningAlg = userInfoConfigInput.SigningAlg
+		oauthUserInfo.EncryptionAlg = userInfoConfigInput.EncryptionAlg
+		oauthUserInfo.EncryptionEnc = userInfoConfigInput.EncryptionEnc
 	}
 	if oauthUserInfo.UserAttributes == nil {
 		oauthUserInfo.UserAttributes = idTokenConfig.UserAttributes
 	}
-	if oauthUserInfo.ResponseType == "" {
+
+	// Always set ResponseType from algorithm fields so the runtime never has to re-derive it.
+	hasSign := oauthUserInfo.SigningAlg != ""
+	hasEnc := oauthUserInfo.EncryptionAlg != ""
+	switch {
+	case hasSign && hasEnc:
+		oauthUserInfo.ResponseType = model.UserInfoResponseTypeNESTEDJWT
+	case hasEnc:
+		oauthUserInfo.ResponseType = model.UserInfoResponseTypeJWE
+	case hasSign:
+		oauthUserInfo.ResponseType = model.UserInfoResponseTypeJWS
+	default:
 		oauthUserInfo.ResponseType = model.UserInfoResponseTypeJSON
 	}
 
--- a/backend/internal/application/store.go
+++ b/backend/internal/application/store.go
@@ -67,8 +67,11 @@ type idTokenConfig struct {
 
 // userInfoConfig represents user info endpoint configuration for JSON marshaling.
 type userInfoConfig struct {
-	ResponseType   model.UserInfoResponseType `json:"response_type,omitempty"`
-	UserAttributes []string                   `json:"user_attributes,omitempty"`
+	ResponseType   string   `json:"response_type,omitempty"`
+	UserAttributes []string `json:"user_attributes,omitempty"`
+	SigningAlg     string   `json:"signing_alg,omitempty"`
+	EncryptionAlg  string   `json:"encryption_alg,omitempty"`
+	EncryptionEnc  string   `json:"encryption_enc,omitempty"`
 }
 
 // applicationConfigDAO represents the gateway configuration for an application (no identity data).
@@ -549,8 +552,11 @@ func getOAuthConfigJSONBytes(inboundAuth model.InboundAuthConfigProcessedDTO) (j
 	}
 	if oa.UserInfo != nil {
 		cfg.UserInfo = &userInfoConfig{
-			ResponseType:   oa.UserInfo.ResponseType,
+			ResponseType:   string(oa.UserInfo.ResponseType),
 			UserAttributes: oa.UserInfo.UserAttributes,
+			SigningAlg:     oa.UserInfo.SigningAlg,
+			EncryptionAlg:  oa.UserInfo.EncryptionAlg,
+			EncryptionEnc:  oa.UserInfo.EncryptionEnc,
 		}
 	}
 	data, err := json.Marshal(cfg)
--- a/backend/internal/oauth/oauth2/userinfo/error_constants.go
+++ b/backend/internal/oauth/oauth2/userinfo/error_constants.go
@@ -80,4 +80,18 @@ var (
 			DefaultValue: "The 'openid' scope is required for this request",
 		},
 	}
+
+	// errorUserInfoInternalError is returned when an internal error occurs while processing the UserInfo request
+	errorUserInfoInternalError = serviceerror.ServiceError{
+		Type: serviceerror.ServerErrorType,
+		Code: "server_error",
+		Error: core.I18nMessage{
+			Key:          "error.userinfoservice.internal_server_error",
+			DefaultValue: "Internal server error",
+		},
+		ErrorDescription: core.I18nMessage{
+			Key:          "error.userinfoservice.internal_server_error_description",
+			DefaultValue: "An internal server error occurred while processing the UserInfo request",
+		},
+	}
 )
--- a/backend/internal/oauth/oauth2/userinfo/handler.go
+++ b/backend/internal/oauth/oauth2/userinfo/handler.go
@@ -71,11 +71,16 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
 	w.Header().Set(serverconst.CacheControlHeaderName, serverconst.CacheControlNoStore)
 	w.Header().Set(serverconst.PragmaHeaderName, serverconst.PragmaNoCache)
 
-	if result.Type == appmodel.UserInfoResponseTypeJWS {
+	switch result.Type {
+	case appmodel.UserInfoResponseTypeJWS:
 		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJWT)
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte(result.JWTBody))
-	} else {
+	case appmodel.UserInfoResponseTypeJWE, appmodel.UserInfoResponseTypeNESTEDJWT:
+		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJOSE)
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write([]byte(result.JWTBody))
+	default:
 		utils.WriteSuccessResponse(w, http.StatusOK, result.JSONBody)
 	}
 
--- a/backend/internal/oauth/oauth2/userinfo/service.go
+++ b/backend/internal/oauth/oauth2/userinfo/service.go
@@ -21,11 +21,16 @@ package userinfo
 
 import (
 	"context"
+	"crypto"
+	"encoding/json"
+	"io"
+	"net/http"
 	"slices"
 
 	"github.com/asgardeo/thunder/internal/application"
 	appmodel "github.com/asgardeo/thunder/internal/application/model"
 	"github.com/asgardeo/thunder/internal/attributecache"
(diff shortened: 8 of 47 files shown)


backend/internal/application/model/constants.go
```diff
@@ -35,10 +35,25 @@ const (
...
+
+	// UserInfoResponseTypeJWE represents the encrypted (JWE) userinfo response type.
+	UserInfoResponseTypeJWE UserInfoResponseType = "JWE"
+
+	// UserInfoResponseTypeNESTEDJWT represents the signed-then-encrypted (Nested JWT) userinfo response type.
+	UserInfoResponseTypeNESTEDJWT UserInfoResponseType = "NESTED_JWT"
 )
 
+// SupportedUserInfoSigningAlgs lists JWS algorithms supported for userinfo signing.
+var SupportedUserInfoSigningAlgs = []string{"RS256", "RS512", "PS256", "ES256", "ES384", "ES512", "EdDSA"}
```
ThaminduDilshan: For the allowed lists, if possible shall we reuse constants declared in system/jose rather than duplicating the values?

backend/internal/application/model/oauth_app.go
```diff
@@ -47,8 +47,10 @@ type IDTokenConfig struct {
 
 // UserInfoConfig represents the user info endpoint configuration structure.
 type UserInfoConfig struct {
-	ResponseType   UserInfoResponseType `json:"responseType,omitempty" yaml:"response_type,omitempty"`
```
ThaminduDilshan, line 50: I guess you're determining the user info response type based on the configured algorithm parameters. But isn't it better to have a dedicated field for the response type?

Then app management (i.e. app creation and update) layer can validate for the required algorithm fields when appropriate response type is set. Also you're determining response type in runtime in `deriveResponseType()` function based on configured algorithm fields. Having dedicated field can eliminate this as well

backend/internal/application/service.go
```diff
@@ -908,9 +909,15 @@ func toOAuthProcessedDTO(
 	}
 
 	if cfg.UserInfo != nil {
+		signingAlg := cfg.UserInfo.SigningAlg
+		if signingAlg == "" && cfg.UserInfo.ResponseType == "JWS" {
```
ThaminduDilshan: `cfg.UserInfo.ResponseType == "JWS" `

Will this logic work without accepting userinfo.ResponseType in the API payload?

backend/internal/application/store.go
```diff
@@ -67,8 +67,14 @@ type idTokenConfig struct {
...
 // userInfoConfig represents user info endpoint configuration for JSON marshaling.
 type userInfoConfig struct {
-	ResponseType   model.UserInfoResponseType `json:"response_type,omitempty"`
-	UserAttributes []string                   `json:"user_attributes,omitempty"`
+	UserAttributes []string `json:"user_attributes,omitempty"`
+	SigningAlg     string   `json:"signing_alg,omitempty"`
+	EncryptionAlg  string   `json:"encryption_alg,omitempty"`
+	EncryptionEnc  string   `json:"encryption_enc,omitempty"`
+
+	// ResponseType is a legacy read-only field for rows written before the alg-based schema.
```
ThaminduDilshan: We don't need to support backward compatibility at this stage as we're still at pre releases. But refer to https://github.com/asgardeo/thunder/pull/2394/changes#r3147534022

backend/internal/application/service.go
```diff
@@ -1490,7 +1571,7 @@ func validateTokenEndpointAuthMethod(oauthConfig *model.OAuthAppConfigDTO) *serv
 			})
 		}
 	case oauth2const.TokenEndpointAuthMethodClientSecretBasic, oauth2const.TokenEndpointAuthMethodClientSecretPost:
-		if hasCert {
+		if hasCert && !userInfoNeedsCert {
```
ThaminduDilshan, line 1609: Is this returning the correct error when we have user info cert this validation embedded here?
Check handling below as well

backend/internal/application/model/constants.go
```diff
@@ -35,10 +35,25 @@ const (
 	// UserInfoResponseTypeJSON represents the JSON userinfo response type.
 	UserInfoResponseTypeJSON UserInfoResponseType = "JSON"
 
-	// UserInfoResponseTypeJWS represents the JWS userinfo response type.
+	// UserInfoResponseTypeJWS represents the signed JWT (JWS) userinfo response type.
```
ThaminduDilshan, line 43: @thiva-k is this the appropriate place to keep OAuth spec related application config constants?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -103,7 +117,10 @@ func (s *userInfoService) GetUserInfo(
 		return nil, svcErr
 	}
 
-	oauthApp := s.getOAuthApp(ctx, tokenClaims)
+	oauthApp, svcErr := s.getOAuthApp(ctx, tokenClaims)
+	if svcErr != nil {
+		return nil, svcErr
```
ThaminduDilshan, line 122: This results in returning a application error from the user info service right? Instead we should return a userinfo specific error here.
You can handle it inside `getOAuthApp()` function when invoking app service

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,194 @@ func (s *userInfoService) GetUserInfo(
...
+	default:
+		return &UserInfoResponse{Type: appmodel.UserInfoResponseTypeJSON, JSONBody: response}, nil
 	}
+}
 
-	return &UserInfoResponse{
-		Type:     appmodel.UserInfoResponseTypeJSON,
-		JSONBody: response,
-	}, nil
+// deriveResponseType determines the userinfo response type from the alg config (AC-4a–4d).
```
ThaminduDilshan: `(AC-4a–4d).` Is this added intentionally? Didn't get the idea. Is it returning to a spec section?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,194 @@ func (s *userInfoService) GetUserInfo(
...
+		if err := syshttp.IsSSRFSafeURL(certificate.Value); err != nil {
+			s.logger.Error("JWKS URI is not SSRF-safe", log.String("uri", certificate.Value), log.Error(err))
+			return nil, "", &serviceerror.InternalServerError
+		}
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, certificate.Value, nil)
+		if err != nil {
+			s.logger.Error("Failed to build JWKS request", log.Error(err))
+			return nil, "", &serviceerror.InternalServerError
+		}
+		resp, err := s.httpClient.Do(req)
```
ThaminduDilshan: Shall we extract out jwks fetching and response handling to a separate private function to increase readability?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,194 @@ func (s *userInfoService) GetUserInfo(
...
+			return nil, "", &serviceerror.InternalServerError
+		}
+		jwksJSON = string(body)
+	default:
+		s.logger.Error("Unsupported certificate type for userinfo encryption",
+			log.String("type", string(certificate.Type)))
+		return nil, "", &serviceerror.InternalServerError
+	}
+
+	var jwksObj struct {
```
ThaminduDilshan, line 243: Shall we extract out jwks parsing and key finding logic to a separate private function to increase readability?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -163,16 +355,26 @@ func (s *userInfoService) generateJWSUserInfo(
...
 		jwt.TokenTypeJWT,
+		signingAlg,
 	)
 	if err != nil {
-		s.logger.Error("Failed to generate signed UserInfo JWT")
+		if err.Code == jwt.ErrorUnsupportedJWSAlgorithm.Code {
+			s.logger.Error("UserInfo signing algorithm is not supported by the server key",
+				log.String("alg", signingAlg))
+		} else {
+			s.logger.Error("Failed to generate signed UserInfo JWT")
```
ThaminduDilshan: We'll lose error details unless we log the error information. 

Also do we need to handle unsupported jws algorithm error differently? We can log service error default error in a generic way

backend/internal/oauth/oauth2/userinfo/service_jwe_test.go
```diff
@@ -0,0 +1,403 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/oauth/oauth2/userinfo/service_jwe_test.go
```diff
@@ -0,0 +1,403 @@
...
+	"github.com/asgardeo/thunder/internal/system/log"
+	"github.com/asgardeo/thunder/tests/mocks/httpmock"
+	"github.com/asgardeo/thunder/tests/mocks/jose/jwemock"
+	"github.com/asgardeo/thunder/tests/mocks/jose/jwtmock"
+)
+
+const testJWKSURI = "https://rp.example.com/jwks"
+
+// TestDeriveResponseType covers the auto-derivation cases (AC-4a–4d).
+func TestDeriveResponseType(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/system/jose/jwe/model.go
```diff
@@ -36,6 +38,8 @@ const (
 type ContentEncAlgorithm string
 
 const (
+	// A128CBCHS256 represents AES-128-CBC with HMAC-SHA-256 content encryption (RFC 7518 §5.2)
+	A128CBCHS256 ContentEncAlgorithm = "A128CBC-HS256"
```
hwupathum, line 56: Shall we try to support all recommended algorithms from https://datatracker.ietf.org/doc/html/rfc7518#section-5.1?

backend/internal/system/jose/jwt/service.go
```diff
@@ -47,6 +47,11 @@ import (
 type JWTServiceInterface interface {
 	GenerateJWT(sub, iss string, validityPeriod int64, claims map[string]interface{}, typ string) (
 		string, int64, *serviceerror.ServiceError)
+	// GenerateJWTWithAlg is like GenerateJWT but honors an explicit JWS algorithm name (e.g. "RS256").
```
hwupathum: No need to describe the function in the interface since it is mentioned in the implementation anyway

backend/internal/system/http/httpclient.go
```diff
@@ -82,6 +86,112 @@ func NewHTTPClientWithTimeout(timeout time.Duration) HTTPClientInterface {
...
+
+	dialer := &net.Dialer{Timeout: 10 * time.Second}
+	return dialer.DialContext(ctx, network, net.JoinHostPort(safeIP.String(), port))
+}
+
+// privateIPRanges lists CIDR blocks that must not be used as JWKS fetch targets.
+// Covers IPv4/IPv6 loopback, link-local (including cloud metadata services), and
+// RFC1918/unique-local private ranges.
+var privateIPRanges = func() []*net.IPNet {
+	cidrs := []string{
```
ThaminduDilshan, line 151: I don't think it's a good idea to keep them hard coded here. What is the requirement of keeping the private address ranges in the server?
@rajithacharith

backend/internal/system/http/httpclient.go
```diff
@@ -82,6 +86,112 @@ func NewHTTPClientWithTimeout(timeout time.Duration) HTTPClientInterface {
...
+// Use this when redirect behavior must be controlled, e.g. to prevent HTTPS→HTTP downgrades.
+// Requires ThunderRuntime to be initialized before calling (reads TLS config at construction time).
+// ssrfSafeDialContext is wired in to block hostnames that DNS-resolve to private/loopback addresses
+// and to pin the TCP connection to the first validated IP (prevents DNS rebinding).
+func NewHTTPClientWithCheckRedirect(checkRedirect func(*http.Request, []*http.Request) error) HTTPClientInterface {
+	return &HTTPClient{
+		client: &http.Client{
+			Timeout: 30 * time.Second,
+			Transport: &http.Transport{
+				DialContext: ssrfSafeDialContext,
```
ThaminduDilshan, line 99: Is this a requirement from the spec?

backend/internal/system/jose/jwt/service.go
```diff
@@ -144,6 +149,31 @@ func newJWTService(pkiService pki.PKIServiceInterface) (JWTServiceInterface, err
...
 func (js *jwtService) GenerateJWT(
 	sub, iss string, validityPeriod int64, claims map[string]interface{}, typ string,
+) (string, int64, *serviceerror.ServiceError) {
+	return js.generateJWTInternal(sub, iss, validityPeriod, claims, typ, js.signAlg, js.jwsAlg)
+}
+
+// GenerateJWTWithAlg generates a JWT signed with the algorithm identified by alg (a JWS algorithm string
+// such as "RS256"). When alg is empty it behaves identically to GenerateJWT. When alg is set but
+// incompatible with the server's private key, ErrorUnsupportedJWSAlgorithm is returned.
+func (js *jwtService) GenerateJWTWithAlg(
```
hwupathum: Do maintain consistency with JWE service, shall we update GenerateJWT with algorithm instead of creating a new method?

backend/internal/system/jose/jwt/service.go
```diff
@@ -144,6 +149,31 @@ func newJWTService(pkiService pki.PKIServiceInterface) (JWTServiceInterface, err
 // or providing another type is a programmer error and returns InternalServerError.
 func (js *jwtService) GenerateJWT(
 	sub, iss string, validityPeriod int64, claims map[string]interface{}, typ string,
+) (string, int64, *serviceerror.ServiceError) {
+	return js.generateJWTInternal(sub, iss, validityPeriod, claims, typ, js.signAlg, js.jwsAlg)
```
ThaminduDilshan: Instead of having a separate `generateJWTInternal()` function, this can directly call `GenerateJWTWithAlg()` with the expected default params right?

tests/integration/oauth/userinfo/userinfo_test.go
```diff
@@ -1016,7 +1016,7 @@ func (ts *UserInfoTestSuite) TestUserInfo_JWS_Response() {
 			},
 		},
 		"userInfo": map[string]interface{}{
-			"responseType":   "JWS",
```
ThaminduDilshan, line 1019: Let's add some integration tests for the JWE and NESTED_JWT cases as well. I think we already have tests for JWS

backend/internal/application/model/oauth_app.go
```diff
@@ -47,8 +47,10 @@ type IDTokenConfig struct {
...
-	UserAttributes []string             `json:"userAttributes,omitempty" yaml:"user_attributes,omitempty" jsonschema:"User attributes to include in userinfo response."`
+	UserAttributes []string `json:"userAttributes,omitempty" yaml:"user_attributes,omitempty" jsonschema:"User attributes to include in userinfo response."`
```
ThaminduDilshan: Let's update swagger definition in https://github.com/asgardeo/thunder/blob/main/api/application.yaml to reflect these changes and re-generate the API docs in docs/ directory

backend/internal/system/jose/jwt/service.go
ThaminduDilshan: May get outdated with https://github.com/asgardeo/thunder/pull/2394#discussion_r3147829627

backend/internal/system/jose/jwe/model.go
```diff
@@ -22,6 +22,8 @@ package jwe
 type KeyEncAlgorithm string
 
 const (
+	// RSAOAEP uses SHA-1 for OAEP masking (RFC 7518 §4.3). Prefer RSAOAEP256 for new deployments.
+	RSAOAEP KeyEncAlgorithm = "RSA-OAEP"
```
hwupathum, line 26: Shall we try to support all the recommended algoirthms from https://datatracker.ietf.org/doc/html/rfc7518#section-4.1?

backend/internal/application/model/oauth_app.go
nandhu-kumar: @ThaminduDilshan 
We intentionally derive response type from algorithm fields to avoid
inconsistent configurations and align with OIDC metadata behavior.

should i introduce new field for responsetype?

backend/internal/system/http/httpclient.go
nandhu-kumar: These CIDRs are IANA-reserved ranges (RFC 1918 for IPv4 private, RFC 5735 for loopback/link-local, RFC 4291 for IPv6 private). They're protocol constants that don't vary across deployments. Making them configurable would create a security risk: a misconfiguration could open SSRF vectors. Hard-coding security boundaries for fixed standards is intentional.

backend/internal/system/http/httpclient.go
nandhu-kumar: This is a security hardening, not a spec requirement.

Fetching JWKS from attacker-controlled URIs without DNS validation is a well-known SSRF vector. The existing IsSSRFSafeURL check covers IP literals at registration time, but ssrfSafeDialContext adds defense-in-depth at the TCP dial layer to block hostname-based SSRF and DNS rebinding.

backend/internal/system/jose/jwt/service.go
nandhu-kumar: We intentionally introduced GenerateJWTWithAlg() instead of modifying GenerateJWT() to preserve backward compatibility and avoid unintended side effects.

GenerateJWT() continues to use the server’s default signing algorithm (existing behavior unchanged).
GenerateJWTWithAlg() allows explicitly overriding the algorithm when required (e.g., UserInfo JWS based on client configuration).

This keeps a clean separation of concerns:

default path → stable and predictable
override path → explicit and opt-in

Modifying the existing method would couple these behaviors and risk breaking current flows that rely on implicit defaults.

backend/internal/application/service.go
nandhu-kumar: The condition is intentional. A client_secret_* app may now legitimately carry a certificate if it configures userinfo JWE encryption (the cert is the RP's public key, not a client-auth credential). hasCert && !userInfoNeedsCert returns an error only when the cert is present but serves no userinfo purpose — in that case the existing "client_secret authentication methods cannot have a certificate" message is still accurate.

tests/integration/oauth/userinfo/userinfo_test.go
nandhu-kumar: added.

backend/internal/application/model/oauth_app.go
nandhu-kumar: done

backend/internal/oauth/oauth2/userinfo/service.go
nandhu-kumar: @ThaminduDilshan 

Handled it in getOAuthApp() — application errors are now mapped to UserInfo-specific errors instead of being returned directly.

backend/internal/application/model/oauth_app.go
ThaminduDilshan: Let's keep the response type field

thiva-k: Let's use this derivation logic for the DCR-side only and map it to the internal response type

https://github.com/asgardeo/thunder/discussions/2287#discussioncomment-16520903

backend/internal/system/jose/jwe/model.go
ThaminduDilshan: Since this is not related to $subject of this PR, we can extract it out to a followup task.
WDYT @hwupathum ?

backend/internal/system/jose/jwt/service.go
ThaminduDilshan: We can modify the existing `GenerateJWT()` method to accept these parameters and update the usages. There's only few usages right?

backend/internal/system/jose/jwe/model.go
ThaminduDilshan: Since this is not related to $subject of this PR, we can extract it out to a followup task.
WDYT @hwupathum ?

backend/internal/application/model/constants.go
ThaminduDilshan: @thiva-k let's track this with your improvements

backend/internal/system/jose/jwe/model.go
hwupathum: +1

backend/internal/system/jose/jwe/model.go
hwupathum: +1

anushasunkada:
@ThaminduDilshan & @thiva-k 

With this PR merge -> https://github.com/asgardeo/thunder/pull/2455

CryptoProvider now takes keyRefId instead of actual key material. This works fine for the internal keys/internal data encryption, but when we have to use client keys to encrypt data(Eg: userinfo response) we should explicitly pass the public key to the CryptoProvider.

We have 2 options, 
1. On OAuth client registration, we should store encryption key in PKIService and only store the key reference in DB. 
2. We create new encrypt method in CryptoProvider interface to take public key as one of the argument.

I recommend, option-2. 

cc: @nandhu-kumar

api/application.yaml
```diff
@@ -1117,9 +1117,34 @@ components:
           items:
             type: string
           description: |
-            The user attributes to include in the userinfo response. 
+            The user attributes to include in the userinfo response.
             If not specified, falls back to ID token user attributes.
           example: ["email", "name", "given_name", "family_name", "picture"]
+        signingAlg:
```
ThaminduDilshan, line 1123: I think these fields are not reflected in the swagger editor properly. Should have missed to add these new fields to some places

backend/internal/oauth/oauth2/userinfo/error_constants.go
```diff
@@ -80,4 +80,18 @@ var (
 			DefaultValue: "The 'openid' scope is required for this request",
 		},
 	}
+
+	// errorUserInfoInternalError is returned when an internal error occurs while processing the UserInfo request
+	errorUserInfoInternalError = serviceerror.ServiceError{
```
ThaminduDilshan, line 85: We shouldn't have package specific server errors. Let's remove this and re use the existing constants.
@thiva-k this is the existing pattern right? Are we using the same server wide internal server error constant from the system package?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,211 @@ func (s *userInfoService) GetUserInfo(
...
+	default:
+		return appmodel.UserInfoResponseTypeJSON
 	}
+}
 
-	return &UserInfoResponse{
-		Type:     appmodel.UserInfoResponseTypeJSON,
-		JSONBody: response,
-	}, nil
+// resolveRPPublicKey resolves the RP's public key from the application certificate (AC-5a–5d).
```
ThaminduDilshan, line 172: `(AC-5a–5d).` Do we need to keep this part? Does this refer to a spec section?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,211 @@ func (s *userInfoService) GetUserInfo(
...
+		return nil, "", &serviceerror.InternalServerError
+	}
+
+	return s.parseEncryptionKeyFromJWKS(jwksData, encryptionAlg)
+}
+
+// fetchJWKS fetches the JWKS document from the given URI with SSRF protection and a 1 MB size cap.
+func (s *userInfoService) fetchJWKS(ctx context.Context, jwksURI string) ([]byte, *serviceerror.ServiceError) {
+	if err := syshttp.IsSSRFSafeURL(jwksURI); err != nil {
+		s.logger.Error("JWKS URI is not SSRF-safe", log.String("uri", jwksURI), log.Error(err))
```
ThaminduDilshan, line 220: Here we're handling many jwks response issues as server errors. If we're following this approach, IMO we should have validations during app config time to validate these conditions. If validations fails, we shouldn't allow to set the jwks url.
cc: @thiva-k

backend/internal/application/service.go
```diff
@@ -1304,9 +1308,120 @@ func validateOAuthParamsForCreateAndUpdate(app *model.ApplicationDTO) (*model.In
...
+func validateUserInfoConfig(oauthConfig *model.OAuthAppConfigDTO) *serviceerror.ServiceError {
+	if oauthConfig.UserInfo == nil {
+		return nil
+	}
+	cfg := oauthConfig.UserInfo
+
+	if cfg.SigningAlg != "" && !slices.Contains(model.SupportedUserInfoSigningAlgs, cfg.SigningAlg) {
+		return serviceerror.CustomServiceError(
+			ErrorInvalidOAuthConfiguration,
+			core.I18nMessage{DefaultValue: "userinfo signing algorithm '" + cfg.SigningAlg + "' is not supported"},
```
ThaminduDilshan, line 1329: i18n keys are missing for all the custom service error invocations. We need to have dedicated keys for each error.
Check all the places introduced with this PR

backend/internal/application/service.go
```diff
@@ -1712,19 +1828,25 @@ func processUserInfoConfiguration(app *model.ApplicationDTO,
...
+		oauthUserInfo.SigningAlg = userInfoConfigInput.SigningAlg
+		oauthUserInfo.EncryptionAlg = userInfoConfigInput.EncryptionAlg
+		oauthUserInfo.EncryptionEnc = userInfoConfigInput.EncryptionEnc
 	}
 	if oauthUserInfo.UserAttributes == nil {
 		oauthUserInfo.UserAttributes = idTokenConfig.UserAttributes
 	}
-	if oauthUserInfo.ResponseType == "" {
+
+	// Always set ResponseType from algorithm fields so the runtime never has to re-derive it.
```
ThaminduDilshan, line 1839: So with the current approach, we're overriding userinfo response type set in the payload with SigningAlg/ EncryptionAlg and EncryptionEnc fields. I think we better stick to one approach.
We can rely on user to set the expected userinfo response Type and based on that, can validate for the required SigningAlg, EncryptionAlg and EncryptionEnc combinations. If some required fields are missing for the configured response type, we can return a client error.

@thiva-k WDYT is better?

thiva-k: Seems currently only the derived algorithm is set. The response type in request parameter is validated and ignored. Shall we fix @nandhu-kumar ?

tests/integration/oauth/userinfo/userinfo_test.go
```diff
@@ -1016,7 +1020,7 @@ func (ts *UserInfoTestSuite) TestUserInfo_JWS_Response() {
...
```
ThaminduDilshan, line 1019: Let's add this back

ThaminduDilshan (approved):

backend/internal/oauth/oauth2/userinfo/error_constants.go
thiva-k: Yes. Let's use this error @nandhu-kumar https://github.com/nandhu-kumar/thunder/blob/cfb5cf05cf2afae911893f8866cebddc2b485e59/backend/internal/system/error/serviceerror/error.go#L81

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -129,20 +146,196 @@ func (s *userInfoService) GetUserInfo(
...
+	for _, key := range jwksObj.Keys {
+		use, _ := key["use"].(string)
+		if use != "enc" {
+			continue
+		}
+		kty, _ := key["kty"].(string)
+		if kty != "RSA" {
+			continue
+		}
+		if keyAlg, _ := key["alg"].(string); keyAlg != "" && keyAlg != encryptionAlg {
```
thiva-k, line 260: Are we validating if enc is there in registration path, similar to SSRF, etc.?

backend/internal/oauth/oauth2/userinfo/handler.go
```diff
@@ -71,11 +71,16 @@ func (h *userInfoHandler) HandleUserInfo(w http.ResponseWriter, r *http.Request)
...
 
-	if result.Type == appmodel.UserInfoResponseTypeJWS {
+	switch result.Type {
+	case appmodel.UserInfoResponseTypeJWS:
 		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJWT)
 		w.WriteHeader(http.StatusOK)
 		_, _ = w.Write([]byte(result.JWTBody))
-	} else {
+	case appmodel.UserInfoResponseTypeJWE, appmodel.UserInfoResponseTypeNESTEDJWT:
+		w.Header().Set(serverconst.ContentTypeHeaderName, serverconst.ContentTypeJOSE)
```
thiva-k, line 80: Does the spec mandate this must be `jose` ?

backend/internal/oauth/oauth2/userinfo/service.go
thiva-k: Seems some validations are missed. Let's check the JWKS and JWKS_URI validations introduced

backend/internal/oauth/oauth2/userinfo/handler.go
thiva-k: From OIDC spec, userinfo response section:

```
The content-type of the HTTP response MUST be set to application/json if the response body is a text JSON structure; the response body SHOULD be encoded using UTF-8. If the JSON response is signed or encrypted, then the content-type MUST be set to application/jwt.
```

thiva-k:
Merging this to prevent conflicts with https://github.com/asgardeo/thunder/pull/2457. Review comments can be addressed in a follow-up PR

backend/internal/oauth/oauth2/userinfo/error_constants.go
ThaminduDilshan: Looks like this is already addressed

backend/internal/application/service.go
ThaminduDilshan: Already resolved

PR #2488, opened 2026-04-29
Allow boolean schema property to accept string "true"/"false" values

Dilusha-Madushan:
### Purpose
Boolean schema properties only accepted native Go bool values during validation. When a boolean attribute was submitted as a string (e.g. from a JSON payload where the value was serialized as "true" or "false"), validation incorrectly rejected it. This fix allows boolean properties to accept both native booleans and the string representations "true" and "false".

### Approach
Updated validateValue on the boolean property type to use a type switch. Native bool values pass as before; strings are accepted only if they are exactly "true" or "false". All other types (numbers, arbitrary strings, nil) are rejected with a debug log, unchanged from prior behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced boolean schema validation to accept string representations "true" and "false" in addition to native boolean values.

### DIFF
--- a/backend/internal/userschema/model/boolean.go
+++ b/backend/internal/userschema/model/boolean.go
@@ -48,13 +48,17 @@ func (p *boolean) isUnique() bool {
 }
 
 func (p *boolean) validateValue(value interface{}, path string, logger *log.Logger) (bool, error) {
-	_, ok := value.(bool)
-	if !ok {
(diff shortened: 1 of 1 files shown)


Dilusha-Madushan:
Close this PR and handle this issue in separate PR after a discussion.

PR #2494, opened 2026-04-29
Harden CORS origin matching and preflight handling

RushanNanayakkara:
### Purpose

Replace the existing CORS handling with a hardened implementation aligned with the Fetch spec and RFC 6454.

The previous code path in `serverutil.GetAllowedOrigin` used `strings.Contains` for origin matching:

```go
if strings.Contains(redirectURI, allowedOrigin) {
    return allowedOrigin
}
```

A rule of `https://example.com` would match an `Origin: https://example.com.attacker.com` header — i.e., any subdomain or suffix-bearing string an attacker chose. The middleware also did not gate preflight on `Access-Control-Request-Method`, did not emit `Vary: Origin`, accepted the `*` wildcard silently, and did not reject multi-Origin headers.

This PR introduces a new `internal/system/cors` package that performs strict origin parsing, canonical literal matching, and operator-supplied regex matching, and rewrites `internal/system/middleware/cors.go` to enforce preflight gating and `Vary: Origin` correctly on both allow and deny paths.

---
#### ⚠️ Breaking Changes

This PR replaces substring-based origin matching with exact-origin matching, which changes which `allowed_origins` entries are accepted by existing deployments.

### Configuration: `allowed_origins` semantics

Previously the middleware matched the request `Origin` header against configured entries with `strings.Contains` — any entry that appeared as a substring of the request origin was accepted. This PR requires entries to be valid origins (scheme + host + optional port) and matches them by exact equality after canonicalization.

#### 💥 Impact

- **Bare hostnames are rejected at startup.** `example.com` previously matched any `Origin` containing that substring; it is now an invalid literal entry.
- **Port-less and ported entries are distinct.** `https://example.com` previously also matched `https://example.com:8080`; now each port must be listed explicitly.
- **Path-bearing entries are rejected.** `https://example.com/api` is no longer accepted as an entry.
- Out-of-tree code calling `utils.GetAllowedOrigin` will fail to compile.
- Misconfigured entries fail at server start with the offending index and reason, so issues surface before the first cross-origin request.

#### 🔄 Migration Guide

Replace each old `allowed_origins` entry with the exact origin(s) the deployment must accept:

```yaml
# before — relied on substring match
cors:
  allowed_origins:
    - "example.com"

# after — list each scheme + host + port combination explicitly
cors:
  allowed_origins:
    - "https://example.com"
    - "https://example.com:443"
    - "https://app.example.com"
```

For dynamic tenant subdomains, switch to an anchored regex entry:

```yaml
cors:
  allowed_origins:
    - regex: '^https://[a-z0-9-]+\.tenant\.example\.com$'
```

Anchor every regex with `^…$`. The compiler emits a `WARN` log for unanchored patterns at startup; an unanchored pattern accepts attacker-friendly inputs (e.g., `^https://example.com` matches `https://example.com.attacker.com`).

For `*`, decide on the actual list of origins and enumerate them. The boot-time error message points at the offending entry.

---

### Approach

1. **New `internal/system/cors` package.** Origin parsing, canonicalization, rule compilation, and matching are split into single-purpose files:

   - `parser.go` — `ParseOrigin` is the parse gate: rejects empty input, ASCII control bytes (CR, LF, NUL, DEL), unsupported schemes, missing host, userinfo, and any path/query/fragment. Returns `ParseResult{Raw, IsNull, Canonical}` so the matcher hot path never re-parses.
   - `canonical.go` — `Canonicalize` produces the literal-rule key: lowercase scheme + host, strip trailing dot, Punycode IDN labels, bracket IPv6. Port is preserved verbatim — operators decide whether `https://example.com` and `https://example.com:443` are the same origin by listing each variant.
   - `compiler.go` — `CompileAll` turns YAML entries into `[]OriginRule`. Literal entries go through `Canonicalize`; regex entries are compiled with `regexp.Compile` and warned about if missing a leading `^` or trailing `$`. The `*` wildcard is refused with a clear error.
   - `matcher.go` — `Matcher` partitions rules into a canonical-keyed map (O(1) literal lookup) and a regex slice (iterated only on literal miss). The `null` origin is tracked as a separate flag, matched only by literal rules whose value is `null`. The matcher consumes pre-parsed `ParseResult` only; callers must go through `ParseOrigin`.
   - `rule.go`, `errors.go` — `OriginRule` interface, `LiteralRule`/`RegexRule` types, `ErrInvalidOrigin`.

2. **Rewritten `internal/system/middleware/cors.go`.**
   - Preflight gating: `OPTIONS + Access-Control-Request-Method` triggers full preflight headers (`Allow-Methods`, `Allow-Headers`, `Max-Age`); a bare `OPTIONS` falls through to the inner handler.
   - `Vary: Origin` is set on both allow and deny paths to prevent intermediary cache poisoning.
   - Multi-Origin headers (more than one `Origin` value) are rejected per the Fetch spec.
   - `Access-Control-Max-Age` defaults to 600 seconds; `DefaultAllowedHeaders = []string{"Content-Type", "Authorization"}` is exported for reuse across `init.go` call sites.

3. **`internal/system/config/config.go` + `runtimeconfig.go`.**
   - `CORSConfig` carries a cached `compiled *cors.Matcher` populated by `Validate()`. The middleware reads from this cache on every request — no per-request compilation, no per-request canonicalization.
   - `Validate()` emits an `INFO` log at boot summarizing literal/regex counts, refuses `*`, and refuses any compile error.
   - `InitializeThunderRuntime` lazily re-validates so both the production `LoadConfig` path and tests that build a `Config` literal directly get a populated `Matcher`.

4. **Removed `internal/system/utils/serverutil.go` and its test file.** The substring-match implementation has no remaining callers; deleting it closes the only path through which the old behavior could be reintroduced.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
  - 20 curl scenarios covering preflight gating, wildcard rejection, multi-Origin reject, IDN, IPv6, `null` origin
  - Headed Google Chrome 147 verification of six representative scenarios (literal allow/deny, regex allow, preflight + Authorization allow, preflight + custom-header deny, disallowed-origin preflight) against a fresh build of the branch
  - `docs/content/guides/getting-started/configuration.mdx` updated with the new `allowed_origins` format
- [x] Tests provided.
  - [x] Unit Tests
    - `backend/internal/system/cors/parser_test.go`
    - `backend/internal/system/cors/canonical_test.go`
    - `backend/internal/system/cors/compiler_test.go`
    - `backend/internal/system/cors/matcher_test.go`
    - `backend/internal/system/cors/rule_test.go`
    - `backend/internal/system/middleware/cors_test.go` (updated for the new options shape and preflight gating)
  - [ ] Integration Tests — covered at unit level; the cors package functions are pure and the middleware behaviour is exercised by `cors_test.go` against a real `httptest.ResponseRecorder`. No new integration-test suite needed.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * CORS origin configuration now supports regex patterns alongside literal origins for flexible origin matching.

* **Documentation**
  * Updated CORS configuration guide with comprehensive documentation on pattern syntax, matching behavior, and special origin handling.


backend/internal/system/config/config.go
```diff
@@ -279,9 +281,69 @@ type SHA256Config struct {
...
+	if err != nil {
+		return err
+	}
+	c.compiled = cors.NewMatcher(rules)
+
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "CORSConfig"))
+	if len(c.AllowedOrigins) == 0 {
+		logger.Warn("cors.allowed_origins is empty; no cross-origin request will be permitted")
+	} else {
+		logger.Info("cors.allowed_origins compiled",
```
rajithacharith: Shall we change this to debug

rajithacharith: OR no need the log at all

backend/internal/system/config/runtimeconfig.go
```diff
@@ -34,14 +34,24 @@ var (
...
+// middleware path can rely on Matcher() being populated.
 func InitializeThunderRuntime(thunderHome string, config *Config) error {
 	once.Do(func() {
 		runtimeConfig = &ThunderRuntime{
 			ThunderHome: thunderHome,
 			Config:      *config,
 		}
 	})
-
+	if runtimeConfig != nil && runtimeConfig.Config.CORS.Matcher() == nil {
```
rajithacharith: Why do we need this? I think this can be avoided. Configs needs to be validated and taken from the config module

backend/internal/system/middleware/cors.go
```diff
@@ -21,67 +21,142 @@ package middleware
...
-			w.Header().Set("Access-Control-Allow-Headers", opts.AllowedHeaders)
-		}
-		if opts.AllowCredentials {
-			w.Header().Set("Access-Control-Allow-Credentials", "true")
-		}
+	// Vary: Origin must be set on every Origin-bearing response (allow and
+	// deny) so caches key the response by Origin.
+	w.Header().Add("Vary", "Origin")
+
+	matcher := config.GetThunderRuntime().Config.CORS.Matcher()
```
rajithacharith: Shall we get the matcher out from the config. 
Config package -> Validate the deployment.yaml configs 
CORS package -> Have the matcher implementation and uses the config package
Middleware -> uses the matcher from cors package. 

So the core handling logic will be in CORS package

backend/internal/application/init.go
```diff
@@ -75,9 +75,10 @@ func Initialize(
 
 func registerRoutes(mux *http.ServeMux, appHandler *applicationHandler) {
 	opts1 := middleware.CORSOptions{
-		AllowedMethods:   "GET, POST",
-		AllowedHeaders:   "Content-Type, Authorization",
+		AllowedMethods:   []string{"GET", "POST"},
+		AllowedHeaders:   middleware.DefaultAllowedHeaders,
```
ThaminduDilshan, line 79: Why is this required? 
Isn't it good to allow individual handlers to decide which headers they're going to allow? Or can we say server always allows a single set of headers for all endpoints?

RushanNanayakkara: Content-Type and Authorization are the headers always allowed throughout all handlers. This just moves the common configuration into a constant. It is still possible to have custom configurations if needed.

ThaminduDilshan: Existing yes, but might need to allow different methods in future.
Just realized default headers are moved to a constant.
Fine to keep your logic

backend/cmd/server/repository/conf/deployment.yaml
```diff
@@ -51,6 +51,7 @@ jwt:
 cors:
   allowed_origins:
     - "https://localhost:3000"
+    # - regex: '^https://[a-z0-9-]+\.staging\.example\.com(:[0-9]+)?$'
```
ThaminduDilshan, line 54: Using a regex pattern requires explicitly marking it as a regex right?
Is this expected and agreed upon? Trying to understand benefit of this instead of detecting a regex usage at runtime.
Btw I don't have a strong opinion on this.

backend/internal/system/config/config.go
RushanNanayakkara: removed

backend/internal/system/config/runtimeconfig.go
RushanNanayakkara: Done. Removed the nil check and moved the matcher to internal/system/cors. Validation in config.CORSConfig.Validate() is the only path that initializes it.

backend/internal/system/middleware/cors.go
RushanNanayakkara: cors package now owns the matcher (cors/instance.go with InitializeMatcher / GetMatcher). config.Validate() is the only initializer; middleware reads via cors.GetMatcher(). Dependency graph: config → cors → middleware

backend/cmd/server/repository/conf/deployment.yaml
RushanNanayakkara: Differentiating between regex and literals could be ambiguous since both share use of characters like dot and dash, etc. Declaring regex explicitly resolves this issue. 
Also a url detected as a regex at runtime would cause more matches than intended.

backend/internal/system/middleware/cors.go
```diff
@@ -21,67 +21,142 @@ package middleware
...
-// It returns the pattern and wrapped handler that can be registered with http.ServeMux.
+// WithCORS wraps an HTTP handler with CORS handling: origin validation,
+// Vary: Origin emission, and preflight-only headers gated to OPTIONS+ACRM
+// requests. The pattern is returned unchanged so the route registration site
+// keeps the (pattern, handler) shape that http.ServeMux.HandleFunc expects.
 func WithCORS(pattern string, handler http.HandlerFunc, opts CORSOptions) (string, http.HandlerFunc) {
 	return pattern, func(w http.ResponseWriter, r *http.Request) {
 		applyCORSHeaders(w, r, opts)
 		handler(w, r)
 	}
```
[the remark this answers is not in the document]
RushanNanayakkara: The other six are already covered: OPTIONS /applications/ (application/init.go:104) and OPTIONS /groups/ (group/init.go:99) are trailing-slash subtree handlers under Go 1.22+ method-prefixed patterns, so they match /applications/{id} and /groups/{id}. GET /roles/ only dispatches to /roles/{id} and /roles/{id}/assignments, both of which have explicit OPTIONS handlers (role/init.go:153, 162).

backend/internal/system/cors/instance.go
```diff
@@ -0,0 +1,80 @@
...
+// resulting Matcher as the process-wide singleton. Compilation errors are
+// returned with the offending entry's index so a configuration mistake is
+// surfaced at server start rather than on the first cross-origin request; on
+// error the previously installed matcher is left unchanged.
+//
+// InitializeMatcher also emits a boot-time warning for each regex entry whose
+// pattern lacks start/end anchors — a common operator mistake that turns the
+// pattern into a partial-match filter and almost always allows far more
+// origins than intended.
+func InitializeMatcher(entries OriginEntries) error {
```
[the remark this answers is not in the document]
RushanNanayakkara: The CORS section in docs/content/guides/getting-started/configuration.mdx (lines 507-533) already covers the literal-vs-regex entry shapes, the "*" rejection, the 600s Max-Age default, the per-route nature of methods/headers/credentials, regex anchoring guidance, and the null-origin caveat with examples. No migration note since this is pre-1.0. apis.mdx is a stub for the auto-generated API reference, not a place for prose behavior notes. Going to leave the docs as-is.

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+				return fmt.Errorf("cors: allowed_origins[%d]: regex object missing 'regex' field", i)
+			}
+			out = append(out, RegexEntry{Pattern: obj.Regex})
+		default:
+			return fmt.Errorf("cors: allowed_origins[%d]: entry must be a string or { regex: ... } object", i)
+		}
+	}
+	*e = out
+	return nil
+}
```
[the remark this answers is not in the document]
RushanNanayakkara: The CORS section in docs/content/guides/getting-started/configuration.mdx (lines 507-533) already covers the literal-vs-regex entry shapes, the "*" rejection, the 600s Max-Age default, the per-route nature of methods/headers/credentials, regex anchoring guidance, and the null-origin caveat with examples. No migration note since this is pre-1.0. apis.mdx is a stub for the auto-generated API reference, not a place for prose behavior notes. Going to leave the docs as-is.

backend/internal/system/middleware/cors.go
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with latest changes

backend/internal/system/cors/canonical.go
```diff
@@ -0,0 +1,119 @@
...
+func Canonicalize(raw string) (string, error) {
+	if raw == "" {
+		return "", fmt.Errorf("%w: empty origin", ErrInvalidOrigin)
+	}
+	u, err := url.Parse(raw)
+	if err != nil {
+		return "", fmt.Errorf("%w: %w", ErrInvalidOrigin, err)
+	}
+	return canonicalizeFromURL(u)
+}
```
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with latest changes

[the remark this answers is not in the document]
RushanNanayakkara: Canonicalize has exactly one non-test caller: compileLiteral in compiler.go:140-159, which gates the input through ParseOrigin (line 151) before calling Canonicalize (line 154). ParseOrigin already rejects u.User != nil, u.Path != "", u.RawQuery != "", and u.Fragment != "", so values like https://user:pass@example.com/cb?x#y cannot reach Canonicalize in production.

The contract on lines 34-48 documents what canonicalization produces (lowercased scheme/host, IDN Punycoding, IPv6 bracketing, port preservation) and where the function is called from (literal-rule path). It doesn't claim independent strictness against userinfo/path/query/fragment — that's ParseOrigin's job, and compileLiteral invokes it first by design.

docs/content/sdks/overview.mdx
```diff
@@ -11,7 +11,7 @@ import {Search} from '@wso2/oxygen-ui-icons-react';
 import {useState} from 'react';
 import thunderConfig from '@site/docusaurus.thunder.config';
 
-const productName = thunderConfig.project.name;
```
rajithacharith: Do we need this change?

RushanNanayakkara: Yes, without export, the docs build fails with productName is not defined. MDX scopes a bare top-level const inside its implicit content function, so SDKsContent can't see it. Adding export lifts it to module scope and the build goes green. Issue seems to have been introduced with the following PR. 
- https://github.com/asgardeo/thunder/pull/2493

backend/internal/system/cors/error_constants.go
```diff
@@ -0,0 +1,53 @@
+/*
```
ThaminduDilshan, line 1: Shall we rename this file to `error_constants.go` for consistency?

backend/internal/system/cors/errors.go
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package cors provides origin parsing, rule compilation, and matching for
```
ThaminduDilshan: errors.go is not an appropriate place to have the package comment. 
As I understood, this doesn't follow the service pattern that we have in other packages. What's the entry point to the cors package? Is it Matcher?

backend/cmd/server/repository/conf/deployment.yaml
ThaminduDilshan: But just having dot and dash doesn't make it a regex pattern right? It need to have at least one of some other characters (`*`, `[`, `]`, `(`, `)`, etc)

backend/internal/system/cors/errors.go
RushanNanayakkara: This doesn't follow the service patterns since it's a utility package. Entry point is instance go which is called when cors configs are initially read. 
Moved the package comment there.

backend/internal/system/cors/error_constants.go
RushanNanayakkara: Renamed

pavinduLakshan (dismissed):

ThaminduDilshan:
Shall we squash and push?

backend/internal/system/config/config.go
```diff
@@ -279,9 +280,26 @@ type SHA256Config struct {
...
+}
+
+// Validate checks every allowed-origins entry so configuration errors —
+// invalid literals, malformed regexes, the unsupported "*" wildcard — are
+// surfaced at server start rather than on the first cross-origin request,
+// and installs the resulting matcher as the cors package's process-wide
+// singleton. The middleware reads that singleton via cors.GetMatcher; this
+// config layer only owns YAML validation.
+func (c *CORSConfig) Validate() error {
+	return cors.InitializeMatcher(c.AllowedOrigins)
```
ThaminduDilshan, line 302: This method supposed to validate configs. But now while validating it initializes the cors matcher as well. 
Is it expected? 
IMO this initialization is not a good pattern. A caller initializes the service without that's being done in a proper place. We can let main program itself initialize this or since this is a singleton, do a lazy initialization within the GetMatcher method

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+	default:
+		return nil, fmt.Errorf("cors: unknown entry type %T", entry)
+	}
+}
+
+// CompileAll compiles a slice of entries in declaration order. It fails fast
+// on the first invalid entry, reporting the index and underlying cause so the
+// operator can locate the bad entry in deployment.yaml. Empty input yields a
+// nil slice with no error.
+func CompileAll(entries []Entry) ([]OriginRule, error) {
```
ThaminduDilshan, line 124: This is only called within the cors package right? We can make this private and not expose to outside

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+	default:
+		return "unknown"
+	}
+}
+
+// Compile turns one Entry into a compiled OriginRule. Literal entries are
+// gated through ParseOrigin and (for non-null entries) canonicalized; regex
+// entries are compiled via Go's RE2 engine without any additional validation.
+// Operator-supplied regex patterns are taken as-is.
+func Compile(entry Entry) (OriginRule, error) {
```
ThaminduDilshan, line 109: Same as previous comment. We don't need to expose this

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+	"regexp"
+	"strings"
+
+	yaml "gopkg.in/yaml.v3"
+)
+
+// Entry is the discriminated-union type produced by YAML decoding. It carries
+// either a literal allowed-origin string or a regex pattern. Compile turns an
+// Entry into the corresponding compiled OriginRule.
+type Entry interface {
```
ThaminduDilshan, line 32: Can make this private. @RushanNanayakkara can you check all other such fields and methods and only expose what's required for the outside packages?
This way usages will be simple. They can simply import system/cors and list down the available/ intended methods

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+	isOriginEntry()
+}
+
+// LiteralEntry is the bare-string YAML form, e.g. "https://example.com" or
+// the special-case "null".
+type LiteralEntry struct {
+	Value string
+}
+
+func (LiteralEntry) isOriginEntry() {}
```
ThaminduDilshan, line 42: Add method comment

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+
+// RegexEntry is the object YAML form, e.g. { regex: "\\Ahttps://..." }.
+type RegexEntry struct {
+	Pattern string
+}
+
+func (RegexEntry) isOriginEntry() {}
```
ThaminduDilshan, line 49: Add method comment

backend/internal/system/cors/compiler.go
```diff
@@ -0,0 +1,186 @@
...
+
+	yaml "gopkg.in/yaml.v3"
+)
+
+// Entry is the discriminated-union type produced by YAML decoding. It carries
+// either a literal allowed-origin string or a regex pattern. Compile turns an
+// Entry into the corresponding compiled OriginRule.
+type Entry interface {
+	isOriginEntry()
+}
```
ThaminduDilshan, line 34: Also let's move all of these model definitions to a model.go file to be consistent with other packages

rajithacharith (approved):

ThaminduDilshan (approved):
Approving due to urgency. Let's address refactoring comments with a follow up PR

backend/internal/system/cors/matcher.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, re := range m.regexes {
+		if re.MatchString(parsed.Raw) {
+			return true, parsed.Raw
+		}
+	}
+	return false, ""
+}
+
+// Size reports the total number of rules the matcher holds. Intended for
+// boot-time logging and tests.
```
ThaminduDilshan, line 94: Let's not mention about test usages in source code files. True that this is required for test assertion. But still this is a functional method of struct.
Check other places too

backend/internal/system/cors/matcher_test.go
```diff
@@ -0,0 +1,250 @@
...
+func (suite *MatcherTestSuite) TestMatchUsesPreCanonicalizedFastPath() {
+	// When ParseOrigin has populated Canonical, the matcher must not
+	// require the matcher to recompute it. This is the production hot path.
+	m := suite.buildMatcher(LiteralEntry{Value: "https://example.com"})
+	parsed, err := ParseOrigin("HTTPS://Example.COM")
+	suite.Require().NoError(err)
+	allow, echo := m.Match(parsed)
+	assert.True(suite.T(), allow)
+	assert.Equal(suite.T(), "HTTPS://Example.COM", echo)
+}
```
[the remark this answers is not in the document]
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/config/config.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507 

Introduced an initialiser method and validation function no only does validation.

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/matcher.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

backend/internal/system/cors/compiler.go
RushanNanayakkara: Addressed with following PR 
https://github.com/asgardeo/thunder/pull/2507

PR #2498, opened 2026-04-29
fix(userinfo): Align responseType handling and content-type with OIDC spec

nandhu-kumar:
### Purpose
Align UserInfo response handling with OIDC expectations by using explicit `responseType`, correcting response content types, and addressing follow-up review comments from the previous PR.

---

### Approach
- Stopped deriving `responseType` at runtime and now use the explicitly configured value
- Ensured DCR flow remains compatible by deriving `responseType` during DCR processing
- Updated JWE and NESTED_JWT responses to use `application/jwt` as per OIDC spec
- Added `responseType` field to `UserInfoConfig` schema in OpenAPI
- Updated unit tests to include explicit `responseType`
- Removed unused `ContentTypeJOSE` constant
- Cleaned up internal spec references in comments

---

### Related Issues
- N/A

### Related PRs
- N/A

---

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

---

### Security checks
- [x] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets

## Summary by CodeRabbit

* **New Features**
  * Added explicit UserInfo responseType option: JSON, JWS, JWE, or NESTED_JWT; API examples and docs updated.

* **Bug Fixes**
  * Validation now requires an explicit responseType when signing/encryption algorithms are set.
  * UserInfo media type for encrypted and nested JWT responses changed from application/jose to application/jwt.

* **Tests**
  * Integration and unit tests updated to assert explicit responseType and new media type.

### DIFF
--- a/api/application.yaml
+++ b/api/application.yaml
@@ -152,6 +152,8 @@ paths:
                         validityPeriod: 3600
                         userAttributes: ["email", "name", "given_name", "family_name"]
                     userInfo:
+                      responseType: "JWS"
+                      signingAlg: "RS256"
                       userAttributes: ["email", "name", "given_name", "family_name", "picture"]
                     scopeClaims:
                       profile: ["name", "family_name", "given_name", "picture"]
@@ -219,6 +221,8 @@ paths:
                           validityPeriod: 1800
                           userAttributes: ["email", "email_verified", "name", "given_name", "family_name"]
                       userInfo:
+                        responseType: "JWS"
+                        signingAlg: "RS256"
                         userAttributes: ["email", "name", "given_name", "family_name", "picture"]
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
@@ -352,6 +356,8 @@ paths:
                           validityPeriod: 1800
                           userAttributes: ["email", "email_verified", "name", "given_name", "family_name"]
                       userInfo:
+                        responseType: "JWS"
+                        signingAlg: "RS256"
                         userAttributes: ["email", "name", "given_name", "family_name", "picture"]
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
@@ -473,6 +479,8 @@ paths:
                         validityPeriod: 1800
                         userAttributes: ["email", "email_verified", "name", "given_name", "family_name"]
                     userInfo:
+                      responseType: "JWS"
+                      signingAlg: "RS256"
                       userAttributes: ["email", "name", "given_name", "family_name", "picture"]
                     scopeClaims:
                       profile: ["name", "family_name", "given_name", "picture"]
@@ -542,6 +550,8 @@ paths:
                           validityPeriod: 1800
                           userAttributes: ["email", "email_verified", "name", "given_name", "family_name"]
                       userInfo:
+                        responseType: "JWS"
+                        signingAlg: "RS256"
                         userAttributes: ["email", "name", "given_name", "family_name", "picture"]
                       scopeClaims:
                         profile: ["name", "family_name", "given_name", "picture"]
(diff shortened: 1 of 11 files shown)


api/application.yaml
```diff
@@ -1112,6 +1112,16 @@ components:
       type: object
       description: UserInfo endpoint configuration for the OAuth application
       properties:
+        responseType:
```
ThaminduDilshan, line 1125: Still following fields are not visible in the swagger editor.
- responseType
- signingAlg
- encryptionAlg
- encryptionEnc

Missing in;
- Create application payload
- Create application response
- Get application by id response
- Update application payload
- Update application response

I think we have defined examples separately for each of these requests. That's why they're not getting rendered

ThaminduDilshan, line 1125: Shall we define this as a enum in swagger definition?

backend/internal/inboundclient/service.go
```diff
@@ -1025,32 +1029,19 @@ func resolveOAuthTokens(in *inboundmodel.OAuthTokenConfig,
...
 	if in != nil {
 		out.UserAttributes = in.UserAttributes
+		out.ResponseType = in.ResponseType
 		out.SigningAlg = in.SigningAlg
 		out.EncryptionAlg = in.EncryptionAlg
 		out.EncryptionEnc = in.EncryptionEnc
 	}
 	if out.UserAttributes == nil && idToken != nil {
 		out.UserAttributes = idToken.UserAttributes
 	}
```
ThaminduDilshan, line 1049: If the response type is not sent in the request, shall we default it to JSON?

nandhu-kumar: @ThaminduDilshan We intentionally avoid defaulting responseType to JSON to prevent silent fallback behavior.
With the current validation, any configuration that sets algorithm fields requires an explicit responseType, ensuring consistency and avoiding ignored crypto intent.

defaulting silently to JSON would be a regression. Consider this payload:


{
  "userInfo": {
    "signingAlg": "RS256"
  }
}
With a default, this passes validation and the signingAlg is silently ignored — the user asked for signed UserInfo but gets plain JSON. That's a confusing, hard-to-debug failure.

Instead, we rely on validation to enforce correctness: if signingAlg, encryptionAlg, or encryptionEnc are provided without an explicit responseType, the request is rejected with a clear error. This makes the intent unambiguous and prevents silent data loss.

ThaminduDilshan: Agree on that. But that should fail during the OAuth config validation right? Note that this method is called through `applyInboundDefaults()` which is called after invoking the `validateOAuthProfile()`. So we can validate for the above condition inside `validateUserInfoConfig()` right?

My concern is on what happens if someone didn't send userinfo config at all in the payload. It allows creating a oauth app without any userinfo response type right? At runtime we're anyway defaulting to JSON [1]. 
So better to indicate this default in the app payload as well. Otherwise this will allow storing a empty response type even though runtime treats it as a JSON

[1] https://github.com/nandhu-kumar/thunder/blob/4a868370bbbb2158caa3b3bb0d70bd18d707ccd1/backend/internal/oauth/oauth2/userinfo/service.go#L165

api/application.yaml
nandhu-kumar: @ThaminduDilshan is it ok now?

backend/internal/inboundclient/service.go
nandhu-kumar: @ThaminduDilshan 

You're right — when no userInfo config is provided, we were storing an empty responseType even though the runtime treats it as JSON. That inconsistency is misleading.

Fixed in the latest change: resolveUserInfo now defaults ResponseType to JSON when in == nil (i.e., userInfo config is completely absent). The default is scoped to that case only — if the caller sends an explicit config, it's preserved as-is, and validation still catches the algo-without-responseType case before defaults are applied.

backend/internal/inboundclient/service.go
```diff
@@ -1025,32 +1029,21 @@ func resolveOAuthTokens(in *inboundmodel.OAuthTokenConfig,
...
 func resolveUserInfo(in *inboundmodel.UserInfoConfig,
 	idToken *inboundmodel.IDTokenConfig) *inboundmodel.UserInfoConfig {
 	out := &inboundmodel.UserInfoConfig{}
 	if in != nil {
 		out.UserAttributes = in.UserAttributes
+		out.ResponseType = in.ResponseType
 		out.SigningAlg = in.SigningAlg
 		out.EncryptionAlg = in.EncryptionAlg
 		out.EncryptionEnc = in.EncryptionEnc
+	} else {
```
ThaminduDilshan: Shall we split this into a separate if condition? If someone doesn't send response type, but sends user attributes this will fail

ThaminduDilshan (approved):

PR #2507, opened 2026-04-29
Improve hardening cors origin matching and preflight handling

RushanNanayakkara:
### Purpose

Follow-up to #2494 — addresses the unresolved review comments on the merged CORS hardening PR.

Changes are incremental polish on top of the merged work. No behavioral change to the CORS matching algorithm itself; this PR is about API surface, package separation, and a couple of test/doc fixes that came out of review.

### Approach

**Split `CORSConfig.Validate()` into pure validation; install matcher from server bootstrap (#3163140532)**
The config layer now owns _validation only_. A new `cors.Validate(entries)` is the pure-validation entry point used by `(*CORSConfig).Validate`. The runtime matcher is installed from `cmd/server/main.go` via `cors.InitializeMatcher(cfg.CORS.AllowedOrigins)`, immediately after configuration is loaded. Tests that previously relied on `Validate()` to also install the singleton now call `cors.InitializeMatcher` directly (`oauth/oauth2/authz`, `flow/mgt`, `system/export`, `system/middleware`).

**Audit `cors` package exports; unexport everything internal (#3163215656)**
Outside the package, callers only need:

- `OriginEntries` — YAML config slice type (used in `CORSConfig.AllowedOrigins`)
- `Validate`, `InitializeMatcher`, `GetMatcher` — lifecycle entry points
- `*Matcher`, `Match`, `Size`, `LiteralCount`, `RegexCount` — middleware-facing API
- `ParseOrigin`, `ParseResult` — origin parsing used by middleware
- `Err*` sentinels — error matching

Everything else is now unexported: `entry` interface, `literalEntry`/`regexEntry` structs, `originRule`/`literalRule`/`regexRule`, `ruleKind`/`kindLiteral`/`kindRegex`, `compile`, `compileAll`, `compileLiteral`, `compileRegex`, `canonicalize`, `isRegexAnchored`, `newMatcher`. `ResetMatcher` deleted entirely (no callers after external test cleanup). External tests construct `OriginEntries` exclusively through `yaml.Unmarshal` on the existing `UnmarshalYAML` path — same pattern already established by `cors/compiler_test.go`, `flow/mgt/declarative_resource_test.go`, and others.

**Move type definitions to `model.go` (#3163246931)**
Introduced `cors/model.go` holding the YAML-config type defs (`entry`, `literalEntry`, `regexEntry`, `OriginEntries`, and the marker methods), to match the convention used by other packages (`idp/model.go`, `role/model.go`, `entity/model.go`, etc.). Behavior (`UnmarshalYAML`, `Validate`, `compile*`, `isRegexAnchored`, `nodeKindString`) stays in `compiler.go`.

**Lock the pre-canonicalized fast-path test (CodeRabbit)**
`TestMatchUsesPreCanonicalizedFastPath` now constructs `ParseResult` literally so the test would fail if `Match` ever recomputed the canonical form internally. Previously it called `ParseOrigin`, which could mask a regression where `Match` re-canonicalizes on the hot path.

**Doc-comment polish**

- Added doc comments to the two `isOriginEntry()` marker methods explaining the sealed-interface pattern (#3163088005, #3163224530, #3163225503).
- Dropped test-usage mentions from source comments (#3163328747).
- Trimmed `var instance` and (the now-deleted) `ResetMatcher` comments to focus on intent rather than implementation history (#3163203513, #3163174649).

**Test-setup fail-fast fixes**

- `system/export/handler_test.go`: switched the standalone test from `assert.NoError` to `require.NoError` for `cors.InitializeMatcher` so initialization failures are fatal and don't cascade.
- `system/middleware/cors_test.go`: replaced `_ = config.InitializeThunderRuntime(...)` with `suite.Require().NoError(...)` to surface runtime-init failures the same way matcher-init does.

#### Unit tests + lint
- `go test ./...` — **103 packages pass, 0 fail**.
- `make lint_backend` — **0 issues**.

#### Wire-level (curl harness)
A local harness (`target/cors-e2e/run.sh`, not committed) swaps the `cors:` block in `deployment.yaml` across six fixtures, bounces Thunder, and captures every CORS-relevant response header. Every probe matched the unit-test predictions. Spot checks:

| # | Config | Highlights |
|---|---|---|
| 01 | `literal-single` | Allowed origin echoed; uppercase-scheme variant accepted via canonicalization; trailing-slash rejected; preflight allowed and denied paths both correct. |
| 02 | `literal-multi` | All three literals match individually; attacker rejected. |
| 03 | `regex-only` | Anchored `^http://localhost:[0-9]+$` accepts `:9000`/`:1234`/`:8000`; rejects substring/prefix pretenders (`localhost.evil.com`, `evilocalhost:9000`); scheme mismatch (`https://`) denied. |
| 04 | `mixed` | Literal + 2 regex entries all matched; `staging.example.com` (no subdomain) correctly rejected. |
| 05 | `empty` | Every Origin denied; no-Origin still served (no `Vary` because nothing to vary on); preflight denied. |
| 06 | `null-origin` | `Origin: null` echoed; regular literal still works alongside; attacker still rejected. |

#### Browser-level (real Chrome via test harness)
Page origin `http://localhost:9000` (python `http.server` over `harness/index.html`); Thunder on `http://localhost:8090`. Each config swapped the `cors:` block in `deployment.yaml` and bounced Thunder. Each page run executes four cross-origin fetches in real Chrome — matrix below.

| #  | Config           | T1 GET no-creds | T2 GET creds:include[¹] | T3 POST preflight+creds | T4 null-origin iframe |
|----|------------------|:---:|:---:|:---:|:---:|
| 01 | literal-single   | ✓ success    | ✓ blocked | ✓ success | ✓ blocked |
| 02 | literal-multi    | ✓ success    | ✓ blocked | ✓ success | ✓ blocked |
| 03 | regex-only       | ✓ success    | ✓ blocked | ✓ success | ✓ blocked |
| 04 | mixed            | ✓ success    | ✓ blocked | ✓ success | ✓ blocked |
| 05 | empty            | ✓ blocked    | ✓ blocked | ✓ blocked | ✓ blocked |
| 06 | null-origin      | ✓ success    | ✓ blocked | ✓ success | ✓ **success** |

[¹] T2 is by-design blocked across every config: the `/.well-known` endpoint sets `AllowCredentials: false` (public metadata), and Fetch refuses to expose a credentialed-mode response without `Access-Control-Allow-Credentials: true`.

Notable confirmations:
- **Config 03** — page origin matches the regex `^http://localhost:[0-9]+$` end-to-end including the preflight path.
- **Config 05** — every fetch blocked by the browser even though the server returns 2xx (no `Access-Control-Allow-Origin` → browser hides the body). Fail-closed verified.
- **Config 06** — sandboxed iframe produced an opaque `Origin: null`; matcher accepted it because `"null"` was in the allowed list, and the iframe's JS read the response. The single asymmetric cell.

All 24 cells match the predictions. No regressions vs the prior PR #2494 browser run.

### Related Issues
- N/A

### Related PRs
- #2494 (merged — this PR addresses the unresolved review comments)

### Checklist
- [x] Manual test round performed and verified. (See **Testing** above — wire-level curl harness across 6 configs + browser-level Chrome E2E across the same 6 configs, all 24 cells match the unit-test predictions.)
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * CORS matcher installation moved to server startup.
  * Internal CORS processing APIs and types made non-public to tighten internal boundaries.

* **Chores**
  * Test suites updated to use the new CORS initialization flow and removed prior matcher reset calls.

### DIFF
--- a/backend/internal/system/cors/compiler.go
+++ b/backend/internal/system/cors/compiler.go
@@ -26,36 +26,9 @@ import (
 	yaml "gopkg.in/yaml.v3"
 )
 
-// Entry is the discriminated-union type produced by YAML decoding. It carries
-// either a literal allowed-origin string or a regex pattern. Compile turns an
-// Entry into the corresponding compiled OriginRule.
-type Entry interface {
-	isOriginEntry()
-}
-
-// LiteralEntry is the bare-string YAML form, e.g. "https://example.com" or
-// the special-case "null".
-type LiteralEntry struct {
-	Value string
-}
-
-func (LiteralEntry) isOriginEntry() {}
-
-// RegexEntry is the object YAML form, e.g. { regex: "\\Ahttps://..." }.
-type RegexEntry struct {
-	Pattern string
-}
-
-func (RegexEntry) isOriginEntry() {}
-
-// OriginEntries is the slice wrapper that carries the heterogeneous YAML
-// schema for cors.allowed_origins. Custom YAML unmarshaling on this type
-// dispatches between the two entry forms.
-type OriginEntries []Entry
-
 // UnmarshalYAML decodes a YAML sequence whose elements are either scalar
-// strings (LiteralEntry) or mappings of the shape { regex: "..." }
-// (RegexEntry). Anything else is rejected at decode time.
+// strings (literal entries) or mappings of the shape { regex: "..." } (regex
+// entries). Anything else is rejected at decode time.
 func (e *OriginEntries) UnmarshalYAML(node *yaml.Node) error {
 	if node.Kind != yaml.SequenceNode {
 		return fmt.Errorf("cors: allowed_origins must be a list, got %v", nodeKindString(node.Kind))
@@ -64,7 +37,7 @@ func (e *OriginEntries) UnmarshalYAML(node *yaml.Node) error {
 	for i, child := range node.Content {
 		switch child.Kind {
 		case yaml.ScalarNode:
-			out = append(out, LiteralEntry{Value: child.Value})
+			out = append(out, literalEntry{Value: child.Value})
 		case yaml.MappingNode:
 			var obj struct {
 				Regex string `yaml:"regex"`
@@ -75,7 +48,7 @@ func (e *OriginEntries) UnmarshalYAML(node *yaml.Node) error {
 			if obj.Regex == "" {
 				return fmt.Errorf("cors: allowed_origins[%d]: regex object missing 'regex' field", i)
 			}
-			out = append(out, RegexEntry{Pattern: obj.Regex})
+			out = append(out, regexEntry{Pattern: obj.Regex})
 		default:
 			return fmt.Errorf("cors: allowed_origins[%d]: entry must be a string or { regex: ... } object", i)
 		}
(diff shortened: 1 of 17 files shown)


backend/internal/system/cors/compiler.go
ThaminduDilshan: Is this accessed outside cors package? If not can make this too private.
Same with RegexEntry

backend/internal/system/cors/compiler.go
```diff
@@ -39,19 +39,27 @@ type LiteralEntry struct {
...
+// union. The method has no behavior and is never called directly; its sole
+// purpose is to restrict entry implementations to types declared in this
+// package.
 func (RegexEntry) isOriginEntry() {}
 
 // OriginEntries is the slice wrapper that carries the heterogeneous YAML
 // schema for cors.allowed_origins. Custom YAML unmarshaling on this type
 // dispatches between the two entry forms.
-type OriginEntries []Entry
+type OriginEntries []entry
```
ThaminduDilshan: Is this accessed outside cors package?

backend/internal/system/cors/instance.go
```diff
@@ -78,9 +78,8 @@ func GetMatcher() *Matcher {
 	return instance
 }
 
-// ResetMatcher clears the installed singleton. It is intended for tests that
-// exercise multiple configurations within a single process; production code
-// calls InitializeMatcher once at boot and never resets.
+// ResetMatcher clears the installed singleton so a subsequent
+// InitializeMatcher call starts from a clean state.
 func ResetMatcher() {
```
ThaminduDilshan: Is it okay to allow functions outside cors to reset the matcher?
Fine to keep this public if required

backend/internal/system/cors/compiler.go
RushanNanayakkara: This is used outside cors package for unit tests.

backend/internal/system/cors/compiler.go
RushanNanayakkara: This is used in cofig.go as type of allowed origins

ThaminduDilshan (dismissed):

backend/internal/system/cors/compiler.go
RushanNanayakkara: Made private

backend/internal/system/cors/instance.go
RushanNanayakkara: This was used for unit tests only outside the package. Refactored and removed the function.

ThaminduDilshan (approved):

PR #2450, opened 2026-04-26
Drive provisioning inputs from user schema

Dilusha-Madushan:
### Purpose
Registration flows hardcoded user profile attributes (e.g., given_name, family_name, email) directly in flow JSON definitions and in the provisioning executor's input list. This meant every new schema attribute required a flow JSON change, and flows could prompt for fields the user schema didn't even define.

This PR makes provisioning inputs schema-driven: the provisioning executor queries the user schema at runtime to determine which non-credential required attributes are missing, and the prompt node dynamically renders input components for them without any hardcoded field definitions in the flow.

### Approach

**Provisioning executor (provisioning_executor.go):**

Extended HasRequiredInputs to call GetRequiredNonCredentialAttributes on the user schema service after checking node-defined inputs.
Schema-only attributes (not already covered by node inputs) are checked across all context sources (user inputs, runtime data, authenticated user attributes). Missing ones are appended to the executor response and forwarded via ForwardedData to the prompt node.
A single credential attribute (password) is still handled explicitly via the node input definition — schema credential collection is intentionally out of scope (see schema model analysis).

**Prompt node (prompt_node.go):**

Added enrichInputsFromForwardedData: reads schema-derived inputs forwarded by the provisioning executor and appends any that are unsatisfied and not already in the node's prompt definition.
Added appendSyntheticMetaComponents: generates minimal UI component entries (type, label, ref) for schema-derived inputs that have no pre-configured meta component, allowing the frontend to render them without flow JSON changes.

### Related Issues
- https://github.com/asgardeo/thunder/issues/2020

### Related PRs
- https://github.com/asgardeo/thunder/pull/2388

## Summary by CodeRabbit

* **New Features**
  * Added select/dropdown input support, new meta component types (BLOCK/ACTION), synthetic metadata injection for schema-derived inputs, provisioning that validates and surfaces required non-credential user attributes, and internal input display labels hidden from API/JSON. Public user-schema endpoints expose non-credential and required non-credential attributes.

* **Tests**
  * Expanded coverage for input serialization, forwarded-data enrichment, graph serialization, provisioning attribute handling, schema attribute discovery, and added mocks.

PR #2518, opened 2026-04-30
Add dynamic input control properties to ProvisioningExecutor

### LINKED ISSUE #2516: Add controls for optional and batched dynamic inputs in user provisioning node

When flow designers want to collect dynamic inputs during a user provisioning step, they can currently prompt only required inputs, and they cannot control how many dynamic inputs are requested at one time, which causes limited flow flexibility and a poor experience when provisioning requires more structured input collection.

Primary: Flow designers

- Goal: Configure user provisioning steps to collect the right dynamic inputs in a controlled and predictable way.
- Pain: They cannot decide whether provisioning should continue with optional inputs behavior in a configurable way, and they cannot limit the number of dynamic inputs prompted in a single step.

- Value to Persona: Gives flow designers better control over provisioning behavior and makes dynamic input collection more flexible for real-world flows.
- Value to Org: Reduces configuration limitations in provisioning flows, improves product completeness, and lowers the need for workarounds or custom handling in customer setups.

### Proposed Solution

- A property to control whether the node should proceed with optional dynamic inputs behavior.
- A property to control how many dynamic inputs can be requested at one time.

### Alternatives

_No response_

Dilusha-Madushan:
### Purpose
Introduces two optional node properties on the **ProvisioningExecutor** that give flow authors fine-grained control over how schema-driven dynamic inputs are collected during user provisioning. Without these properties the executor behaves exactly as before — only required schema attributes are prompted, all at once.

## Dynamic Node Properties

Two new node properties are read from `NodeProperties` at runtime. Both are fully optional; omitting either one preserves the existing behaviour.

### `dynamicInputsIncludeOptional`

**Type:** `bool`  
**Default:** `false`

Controls whether optional schema attributes are included in the dynamic input prompt alongside required ones.

| Value | Behaviour |
|---|---|
| `false` (default) | Only required schema attributes are fetched and prompted, identical to the previous behaviour |
| `true` | Both required and optional schema attributes are fetched; optional ones are presented to the user but rendered as non-required in the UI, so the user can submit without filling them |

### `maxDynamicInputs`

**Type:** `int`  
**Default:** `0`

Caps how many dynamic inputs are shown in a single prompt iteration, enabling paged collection for schemas with many attributes.

| Value | Behaviour |
|---|---|
| `0` (default) | All missing inputs are prompted at once, identical to the previous behaviour |
| `N > 0` | At most `N` inputs are shown per iteration; the executor re-prompts until all have been presented |

## Key Implementation Decisions

- **Required-before-optional ordering**  
  Within each batch, required missing attributes are always placed before optional ones so the user sees mandatory fields first.

- **Presented-optional tracking**  
  When `dynamicInputsIncludeOptional` is enabled, the identifiers of optional attributes that have been shown to the user are accumulated in a `RuntimeData` key named `provisioningPresentedOptionalAttrs`.

  On each subsequent call to `HasRequiredInputs`, this list is used to skip already-presented optional attributes regardless of whether the user submitted an empty value or omitted the field entirely, preventing infinite re-prompt loops.

- **Backward compatibility**  
  Both properties are absent from all existing flow definitions by default. The executor reads them defensively and falls back to the original code path when they are not present, so no existing flow requires any changes.

## Example Node Configuration

```json
{
  "id": "provisioning",
  "type": "TASK_EXECUTION",
  "properties": {
    "includeOptional": true,
    "maxPerPrompt": 3
  },
  "executor": {
    "name": "ProvisioningExecutor"
  },
  "onSuccess": "auth_assert",
  "onIncomplete": "prompt_schema_attrs"
}
```

### Related Issues
- https://github.com/asgardeo/thunder/issues/2516

### Related PRs
- https://github.com/asgardeo/thunder/pull/2450

## Summary by CodeRabbit

* **New Features**
  * Per-node settings to include optional attributes and to limit how many inputs are requested per iteration; optional prompts are skipped if already presented and required inputs are prioritized.

* **Refactor**
  * SMS OTP runtime state consolidated to use standardized runtime keys for phone attribute and mobile number.

* **Tests**
  * Expanded coverage for dynamic prompting, optional-attribute handling, ordering, limits, and default required-only behavior.

### DIFF
--- a/backend/internal/flow/executor/constants.go
+++ b/backend/internal/flow/executor/constants.go
@@ -77,9 +77,6 @@ const (
 	ouIDKey        = "ouId"
 	defaultOUIDKey = "defaultOUID"
 	userTypeKey    = "userType"
-	// TODO: Revisit when the generic OTP executor is implemented.
-	runtimeKeySMSOTPMobileNumber = "smsOTPMobileNumber"
-	runtimeKeySMSOTPPhoneAttr    = "smsOTPPhoneAttr"
 
 	dataValueTrue  = "true"
 	dataValueFalse = "false"
@@ -87,13 +84,15 @@ const (
 
 // Executor property keys
 const (
-	propertyKeyAssignGroup          = "assignGroup"
-	propertyKeyAssignRole           = "assignRole"
-	propertyKeyRequiredScopes       = "requiredScopes"
-	propertyKeyEmailTemplate        = "emailTemplate"
-	propertyKeySMSTemplate          = "smsTemplate"
-	propertyKeyAllowedUserTypes     = "allowedUserTypes"
-	propertyKeyNotificationSenderID = "senderId"
+	propertyKeyAssignGroup                  = "assignGroup"
+	propertyKeyAssignRole                   = "assignRole"
+	propertyKeyRequiredScopes               = "requiredScopes"
+	propertyKeyEmailTemplate                = "emailTemplate"
+	propertyKeySMSTemplate                  = "smsTemplate"
+	propertyKeyAllowedUserTypes             = "allowedUserTypes"
+	propertyKeyNotificationSenderID         = "senderId"
+	propertyKeyDynamicInputsIncludeOptional = "includeOptional"
+	propertyKeyMaxDynamicInputsPerPrompt    = "maxPerPrompt"
 )
 
 // nonSearchableInputs contains the list of user inputs/ attributes that are non-searchable.
--- a/backend/internal/flow/executor/provisioning_executor.go
+++ b/backend/internal/flow/executor/provisioning_executor.go
@@ -23,6 +23,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"sort"
+	"strings"
 
 	authncm "github.com/asgardeo/thunder/internal/authn/common"
 	"github.com/asgardeo/thunder/internal/entityprovider"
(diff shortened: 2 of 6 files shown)


backend/internal/flow/executor/constants.go
```diff
@@ -77,6 +77,13 @@ const (
 	ouIDKey        = "ouId"
 	defaultOUIDKey = "defaultOUID"
 	userTypeKey    = "userType"
+
+	// runtimeKeyPresentedOptionalAttrs holds a space-separated list of optional schema attribute
+	// identifiers that have already been prompted to the user. ProvisioningExecutor uses this to
+	// skip optional attrs the user has already been shown, even if they left the value empty.
+	// TODO: Revisit optional input tracking — if the flow engine gains a mechanism to detect whether
+	// an optional field was intentionally skipped, remove this key and its associated helper methods.
+	runtimeKeyPresentedOptionalAttrs = "provisioningPresentedOptionalAttrs"
```
ThaminduDilshan: We have a separate location to keep runtime keys. Shall we move all these runtime keys there?
`internal/flow/common/constants.go`

backend/internal/flow/executor/constants.go
```diff
@@ -87,13 +94,15 @@ const (
...
-	propertyKeyAllowedUserTypes     = "allowedUserTypes"
-	propertyKeyNotificationSenderID = "senderId"
+	propertyKeyAssignGroup                  = "assignGroup"
+	propertyKeyAssignRole                   = "assignRole"
+	propertyKeyRequiredScopes               = "requiredScopes"
+	propertyKeyEmailTemplate                = "emailTemplate"
+	propertyKeySMSTemplate                  = "smsTemplate"
+	propertyKeyAllowedUserTypes             = "allowedUserTypes"
+	propertyKeyNotificationSenderID         = "senderId"
+	propertyKeyDynamicInputsIncludeOptional = "dynamicInputsIncludeOptional"
```
ThaminduDilshan:
```suggestion
	propertyKeyDynamicInputsIncludeOptional = "includeOptional"
```

backend/internal/flow/executor/constants.go
```diff
@@ -87,13 +94,15 @@ const (
...
+	propertyKeyMaxDynamicInputs             = "maxDynamicInputs"
```
ThaminduDilshan:
```suggestion
	propertyKeyMaxDynamicInputsPerPrompt             = "maxPerPrompt"
```

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -360,6 +364,70 @@ func (p *provisioningExecutor) fetchSchemaAttributes(
 	return attrs, nil
 }
 
+// getPromptOptionalAttributes reads the dynamicInputsIncludeOptional node property.
+// Returns false when the property is absent, preserving the default behavior of prompting only required attributes.
+func (p *provisioningExecutor) getPromptOptionalAttributes(ctx *core.NodeContext) bool {
```
ThaminduDilshan:
```suggestion
func (p *provisioningExecutor) isPromptOptionalAttributesEnabled(ctx *core.NodeContext) bool {
```

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -285,30 +263,54 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
 		return nodeInputsSatisfied
 	}
 
-	schemaMissingAttrs := make([]common.Input, 0, len(schemaAttrs))
+	// Load the set of optional attrs already prompted in previous iterations so they are not
+	// re-prompted even when the user left them empty.
+	alreadyPromptedOptionals := p.getPresentedOptionalAttrs(ctx)
```
ThaminduDilshan:
```suggestion
	alreadyPromptedOptionalAttrs := p.getPresentedOptionalAttrs(ctx)
```

backend/internal/flow/executor/constants.go
ThaminduDilshan: Or pick something better

backend/internal/flow/executor/constants.go
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/constants.go
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/constants.go
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/provisioning_executor.go

backend/internal/flow/executor/provisioning_executor.go

ThaminduDilshan (approved):

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -360,6 +388,70 @@ func (p *provisioningExecutor) fetchSchemaAttributes(
...
+	}
+	if len(alreadyPresented) == 0 {
+		return
+	}
+	ids := make([]string, 0, len(alreadyPresented))
+	for id := range alreadyPresented {
+		ids = append(ids, id)
+	}
+	sort.Strings(ids)
+	execResp.RuntimeData[common.RuntimeKeyPresentedOptionalAttrs] = strings.Join(ids, " ")
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai

The concern is valid in theory, but doesn't apply to this codebase. Every flow configuration has exactly one provisioning node — the engine resolves executors by node ID and no flow wires up multiple provisioning steps. The flat key is intentional: it tracks optional attribute prompt state across multiple iterations of the same node (e.g., user submits → node runs again → skip already-shown optionals). No cross-node leakage can occur in practice.

PR #2537, opened 2026-05-04
Resolve credential attributes from schema in provisioning executor

### LINKED ISSUE #2492: Use node input identifier for credential key resolution in provisioning executor instead of hardcoded value
### Current Limitation

**_appendNonIdentifyingAttributes_** (lines 415-422) hardcodes the key "**_password_**" to find and store the credential. This is inconsistent with the rest of the executor, which already uses userSchemaService to discover attributes dynamically.

### Suggested Improvement

- Resolve the credential key from node inputs — iterate GetRequiredInputs(ctx), find the first input with Type == InputTypePassword, use its Identifier as the key.

- Fall back to hardcoded — if no PASSWORD_INPUT is declared in the node, default to userAttributePassword ("password").

- Use that key for the lookup — the existing ctx.UserInputs / ctx.RuntimeData lookups stay exactly the same, just using the resolved key instead of the hardcoded constant.

Dilusha-Madushan:
### Purpose
The provisioning executor hardcoded **"password"** as the only credential attribute to collect during user creation. This broke for user type schemas that define multiple credential attributes (e.g. a PIN, an API key, or any field marked credential: true beyond password), and was inconsistent with the rest of the executor which already uses userSchemaService to resolve attributes dynamically.

### Approach
Replaced the hardcoded **appendNonIdentifyingAttributes** with a schema-driven appendCredentialAttributes that calls **userSchemaService.GetCredentialAttributes()** to discover which attributes the user type schema marks as credentials.

The collection logic mirrors the existing non-credential attribute pattern in getAttributesForProvisioning:

- If the node explicitly declares inputs whose identifiers match schema credential attributes, only those are collected — giving the flow author control over which credentials to gather at each step.
- If no such node inputs are declared, all schema credential attributes are collected — ensuring credentials are never silently dropped when a node makes no explicit declaration.
- Values are resolved from UserInputs then RuntimeData, intentionally excluding AuthenticatedUser.Attributes since credentials must be explicitly supplied and never inferred from a prior auth context.

A **fetchCredentialAttributes** helper was added symmetric to the existing fetchAllNonCredentialAttributes. When the schema service is unavailable it returns gracefully with no credentials appended; a service-call failure surfaces as an ExecFailure response rather than a system error, consistent with how other schema fetch failures are handled in this executor.

### Related Issues
- https://github.com/asgardeo/thunder/issues/2492

### DIFF
--- a/backend/internal/flow/executor/provisioning_executor.go
+++ b/backend/internal/flow/executor/provisioning_executor.go
@@ -140,7 +140,9 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
 	}
 
 	// Create the user in the store.
-	p.appendNonIdentifyingAttributes(ctx, &userAttributes)
+	if err := p.appendCredentialAttributes(ctx, &userAttributes); err != nil {
+		return nil, err
+	}
 	createdEntity, err := p.createUserInStore(ctx, userAttributes)
 	if err != nil {
 		logger.Error("Failed to create user in the store", log.Error(err))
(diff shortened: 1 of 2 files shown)


backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -138,7 +138,12 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
 	}
 
 	// Create the user in the store.
-	p.appendNonIdentifyingAttributes(ctx, &userAttributes)
+	if err := p.appendCredentialAttributes(ctx, &userAttributes); err != nil {
+		logger.Error("Failed to append credential attributes", log.Error(err))
+		execResp.Status = common.ExecFailure
+		execResp.FailureReason = "Failed to resolve credential attributes"
```
ThaminduDilshan: Let's handle this as a server error. I.e. let's return a go error without setting failure reason. No need to log at this level

ThaminduDilshan (approved):

PR #2545, opened 2026-05-05
Add state parameter to federated OAuth authorize URL

### LINKED ISSUE #2346: Cannot self signup in gate with social options when there's an input
### Description

Cannot self signup in gate with social options when there's an input. UI enforces entering a value for the inputs.

Image

### Steps to Reproduce

1. Create an application and enable self registration
2. Configure signup flow with basic and google login
3. Try to signup by redirecting to the gate

### Version

v0.33.0

_No response_

### LINKED ISSUE #2368: Social signup doesn't work with Gate/ SDK
### Description

Social signup doesn't seems to be working in gate or SDK. Once click on signup in `gate/signin` page, it goes to a  `gate/signup` page. After clicking on a social login option, browser open a new popup with google login page. Once the social/ google login flow is completed, popup tab opens the `gate/signin` page instead of signup. ence the registration will fail.

https://github.com/user-attachments/assets/dc5cf014-4984-4774-85e7-1e3eca56ecc6

### Steps to Reproduce

1. Create Google IDP and create an application
2. Configure a google login flow with self signup widget and attach that to the app
3. Try to login
4. Click on signup and select Google
5. This opens a new popup. Select the correct google account and continue signup

### Version

v0.34.0

_No response_

DonOmalVindula:
## Summary

- Include a random `state` parameter in the authorization request URL generated by `BuildAuthorizeURL` for federated IDPs (Google, GitHub)
- This is required per OAuth 2.0 spec (RFC 6749 Section 10.12) for CSRF protection
- The Asgardeo SDK's `Callback` component uses the `state` parameter to validate the callback and route OAuth parameters back to the parent window in popup-based social signup flows

## Context

Without the `state` parameter, the SDK's popup-based social signup flow fails because:
1. The `Callback` component throws "Missing OAuth state parameter"
2. The popup URL monitor requires both `code` and `state` to process the callback

The companion SDK fix is in asgardeo/javascript#488.

## Related Issues
- Fixes https://github.com/asgardeo/thunder/issues/2346
- Fixes https://github.com/asgardeo/thunder/issues/2368

## Test plan

- [ ] Test Google social signup flow (popup-based)
- [ ] Test GitHub social signup flow (popup-based)
- [ ] Verify existing OAuth flows (sign-in redirect) still work
- [ ] Run `make test` - unit tests pass

Refs #2368

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced OAuth state parameter generation and validation to strengthen authentication flows.

* **New Features**
  * Improved embedded flow submission handling with refined control-flow logic.

* **Chores**
  * Updated authentication library dependency to version 0.23.4.

### DIFF
--- a/backend/internal/flow/executor/constants.go
+++ b/backend/internal/flow/executor/constants.go
@@ -65,8 +65,10 @@ const (
 	userAttributeGroups   = "groups"
 	userAttributeSub      = "sub"
 
-	userInputCode             = "code"
-	userInputNonce            = "nonce"
+	userInputCode  = "code"
+	userInputNonce = "nonce"
+	userInputState = "state"
+
 	userInputOuName           = "ouName"
 	userInputOuHandle         = "ouHandle"
 	userInputOuDesc           = "ouDescription"
--- a/backend/internal/flow/executor/oauth_executor.go
+++ b/backend/internal/flow/executor/oauth_executor.go
@@ -184,12 +184,20 @@ func (o *oAuthExecutor) BuildAuthorizeFlow(ctx *core.NodeContext, execResp *comm
 		return fmt.Errorf("failed to get idp name: %w", err)
 	}
 
+	// Generate a random state parameter for CSRF protection and append it to the authorize URL.
+	state := systemutils.GenerateUUID()
+	authorizeURL = authorizeURL + "&" + "state=" + state
+
 	// Set the response to redirect the user to the authorization URL.
 	execResp.Status = common.ExecExternalRedirection
 	execResp.RedirectURL = authorizeURL
 	execResp.AdditionalData = map[string]string{
 		common.DataIDPName: idpName,
 	}
+	if execResp.RuntimeData == nil {
+		execResp.RuntimeData = make(map[string]string)
+	}
+	execResp.RuntimeData[common.RuntimeKeyOAuthState] = state
 
 	return nil
 }
@@ -206,6 +214,20 @@ func (o *oAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
 			IsAuthenticated: false,
 		}
 		return nil
+	}
+
+	// Validate the OAuth state parameter to prevent CSRF attacks.
+	// State is validated only when the client sends it back. Clients that handle CSRF
+	// protection client-side (e.g., via sessionStorage) may omit it.
+	if returnedState, ok := ctx.UserInputs[userInputState]; ok && returnedState != "" {
+		expectedState := ctx.RuntimeData[common.RuntimeKeyOAuthState]
+		if returnedState != expectedState {
+			logger.Debug("OAuth state mismatch")
+			execResp.Status = common.ExecFailure
+			execResp.FailureReason = "Invalid OAuth state parameter"
+			return nil
+		}
+		delete(ctx.RuntimeData, common.RuntimeKeyOAuthState)
 	}
 
 	idpID, err := o.GetIdpID(ctx)
--- a/backend/internal/flow/executor/oidc_auth_executor.go
+++ b/backend/internal/flow/executor/oidc_auth_executor.go
@@ -137,6 +137,20 @@ func (o *oidcAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
 		return nil
 	}
 
+	// Validate the OAuth state parameter to prevent CSRF attacks.
+	// State is validated only when the client sends it back. Clients that handle CSRF
+	// protection client-side (e.g., via sessionStorage) may omit it.
+	if returnedState, ok := ctx.UserInputs[userInputState]; ok && returnedState != "" {
+		expectedState := ctx.RuntimeData[common.RuntimeKeyOAuthState]
+		if returnedState != expectedState {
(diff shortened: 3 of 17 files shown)


DonOmalVindula:
Demo of the fixes

## Google Signup
https://github.com/user-attachments/assets/fdacf225-e187-4f31-acd4-7f844351692a

## Github Signup

https://github.com/user-attachments/assets/6cbe79fb-699c-405c-a3a0-e11d9d85b5a9

ThaminduDilshan:
Any specific reason to support state parameter? Even though this PR sets a state parameter in the authz url, IIRC we're not validating it. If we're returning a state, IMO we should validate that during post auth operations.

Previously our thinking was, if a client needs to use state, they can append it to the fed redirect uri before redirecting to the external IDP (Anyway client performs the redirect manually as flow execute returns redirect uri as a JSON body param and client has to parse body snd perform the redirect manually). Once returned with the code, client should validate it against the previously generated state value.
But it would be better to support this in backend. If so we should do the validation as well

brionmario (dismissed):

DonOmalVindula:
> Previously our thinking was, if a client needs to use state, they can append it to the fed redirect uri before redirecting to the external IDP (Anyway client performs the redirect manually as flow execute returns redirect uri as a JSON body param and client has to parse body snd perform the redirect manually). Once returned with the code, client should validate it against the previously generated state value. But it would be better to support this in backend. If so we should do the validation as well

Will add validation logic

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -184,12 +184,20 @@ func (o *oAuthExecutor) BuildAuthorizeFlow(ctx *core.NodeContext, execResp *comm
 		return fmt.Errorf("failed to get idp name: %w", err)
 	}
 
+	// Generate a random state parameter for CSRF protection and append it to the authorize URL.
+	state := systemutils.GenerateUUID()
```
ThaminduDilshan, line 188: I think we should do this in authn/oauth service file. Let me check and revert

backend/internal/flow/executor/constants.go
```diff
@@ -65,8 +65,11 @@ const (
 	userAttributeGroups   = "groups"
 	userAttributeSub      = "sub"
 
-	userInputCode             = "code"
-	userInputNonce            = "nonce"
+	userInputCode  		  	  = "code"
+	userInputNonce 		 	  = "nonce"
+	userInputState 		      = "state"
+
+	runtimeKeyOAuthState      = "oauthState"
```
ThaminduDilshan: Shall we move this to the location where we keep other runtime data keys? Should be in flow/common/constants.go

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -208,6 +216,17 @@ func (o *oAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
 		return nil
 	}
 
+	// Validate the OAuth state parameter to prevent CSRF attacks.
+	if expectedState, ok := ctx.RuntimeData[runtimeKeyOAuthState]; ok && expectedState != "" {
+		returnedState := ctx.UserInputs[userInputState]
+		if returnedState != expectedState {
+			logger.Warn("OAuth state mismatch", log.String("expected", expectedState))
```
ThaminduDilshan:
```suggestion
			logger.Debug("OAuth state mismatch", log.String("expected", expectedState))
```

backend/internal/flow/executor/oidc_auth_executor.go
```diff
@@ -137,6 +137,17 @@ func (o *oidcAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
 		return nil
 	}
 
+	// Validate the OAuth state parameter to prevent CSRF attacks.
+	if expectedState, ok := ctx.RuntimeData[runtimeKeyOAuthState]; ok && expectedState != "" {
+		returnedState := ctx.UserInputs[userInputState]
+		if returnedState != expectedState {
+			logger.Warn("OAuth state mismatch", log.String("expected", expectedState))
```
ThaminduDilshan:
```suggestion
			logger.Debug("OAuth state mismatch", log.String("expected", expectedState))
```

backend/internal/flow/executor/oauth_executor.go
ThaminduDilshan: Currently authn/oauth doesn't share a session between the two interactions. Hence approach would be to;
1. Generate and return a state from `BuildAuthorizeURL()` method in `OAuthAuthnCoreServiceInterface`
2. Store state in the generated session token (in authn/service.go) or in flow context (executor)
3. Accept state as a method param in `ExchangeCodeForToken()` method. Extract and pass the previously stored state value for validation

However since this requires a function signature change, it's fine to do it in executor level for now

DonOmalVindula: Ack.

backend/internal/flow/executor/constants.go
DonOmalVindula: Done

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

Issue #2438, opened 2026-04-24
Support token exchange with external identity provider tokens


sahandilshan:
### What problem are we solving?

When organizations use Thunder alongside other identity providers (Auth0, Keycloak, Entra ID, etc.), they cannot exchange externally-issued tokens for Thunder tokens via the RFC 8693 token exchange grant. Currently, `verifyTokenSignatureByIssuer()` only validates tokens signed by Thunder's own keys and rejects any external issuer. This prevents federated token exchange scenarios where applications need to convert third-party tokens into Thunder tokens.

### Who are we solving this for?

Primary: Platform engineers and developers in multi-IDP environments
- Goal: Exchange tokens from external identity providers (Auth0, Keycloak, Entra ID) for Thunder access tokens
- Pain: Must implement custom token bridging logic outside Thunder since token exchange rejects external tokens

### Why should we solve this now?

- Value to users: Enables seamless integration in multi-IDP environments without custom workarounds
- Value to product: Token exchange with external IDPs is a standard expectation for any OAuth2 server supporting RFC 8693. Thunder already has `TrustedIssuerConfig` infrastructure used by the JWT authenticator — the foundation exists but isn't wired into token exchange

### Proposed Solution

Extend the token exchange flow to support external identity provider tokens by:
1. Reusing the existing `TrustedIssuerConfig` to define trusted external issuers with their JWKS endpoints, expected audiences, and required claims
2. Updating `getValidIssuers()` to include configured external issuers
3. Implementing JWKS-based signature verification in `verifyTokenSignatureByIssuer()` for external issuer tokens
4. Mapping external token claims to Thunder token claims based on configurable rules

PR #2546, opened 2026-05-05
Add external IDP token exchange via JWKS-based verification

sahandilshan:
### Purpose
Enable Thunder's token exchange endpoint to accept JWT tokens from external identity providers. Adds issuer and token_exchange_enabled IDP properties, resolves external issuers at runtime, and verifies signatures against the IDP's JWKS endpoint with audience validation.

### Related Issue(s)
- https://github.com/asgardeo/thunder/issues/2438

## Summary by CodeRabbit

* **New Features**
  * Token exchange support: exchange external IDP JWTs for product-issued tokens (RFC 8693).
  * New IDP config options: issuer and token_exchange_enabled, with conditional required properties when enabled.
  * Context-aware subject-token validation that verifies external issuers and JWKS, and stricter audience checks for auth-assertions.
  * IDP lookups cached for faster retrieval with negative-caching.

* **Documentation**
  * Added guide: Token Exchange with External Identity Providers (workflow, validation, limitations).

### DIFF
--- a/backend/dbscripts/configdb/postgres.sql
+++ b/backend/dbscripts/configdb/postgres.sql
@@ -139,6 +139,9 @@ CREATE TABLE "IDP" (
 -- Composite index for name-based IDP lookups
 CREATE INDEX idx_idp_name_deployment ON "IDP" (DEPLOYMENT_ID, NAME);
 
+-- GIN index for JSONB containment queries on IDP properties (e.g. issuer lookup via @>)
+CREATE INDEX idx_idp_properties ON "IDP" USING GIN ("PROPERTIES" jsonb_path_ops);
+
 -- Table to store notification senders.
 CREATE TABLE "NOTIFICATION_SENDER" (
     DEPLOYMENT_ID VARCHAR(255) NOT NULL,
--- a/backend/internal/idp/constants.go
+++ b/backend/internal/idp/constants.go
@@ -53,6 +53,8 @@ const (
 	PropLogoutEndpoint        = "logout_endpoint"
 	PropJwksEndpoint          = "jwks_endpoint"
 	PropPrompt                = "prompt"
+	PropIssuer                = "issuer"
+	PropTokenExchangeEnabled  = "token_exchange_enabled"
 )
 
 // Known endpoints for Google OAuth2/OIDC.
@@ -111,6 +113,8 @@ var idpPropertyConfigs = map[IDPType]idpPropertyConfig{
 			PropLogoutEndpoint,
 			PropJwksEndpoint,
 			PropPrompt,
+			PropIssuer,
+			PropTokenExchangeEnabled,
 		},
 		Defaults: map[string]string{},
 	},
@@ -128,6 +132,8 @@ var idpPropertyConfigs = map[IDPType]idpPropertyConfig{
 			PropLogoutEndpoint,
 			PropJwksEndpoint,
 			PropPrompt,
+			PropIssuer,
+			PropTokenExchangeEnabled,
 		},
 		Defaults: map[string]string{
 			PropAuthorizationEndpoint: googleAuthorizationEndpoint,
@@ -157,5 +163,17 @@ var idpPropertyConfigs = map[IDPType]idpPropertyConfig{
 			PropUserInfoEndpoint:      gitHubUserInfoEndpoint,
 			PropUserEmailEndpoint:     gitHubUserEmailEndpoint,
 		},
+	},
+}
+
+// tokenExchangeRequiredProps defines the required properties per IDP type when token exchange is enabled.
+var tokenExchangeRequiredProps = map[IDPType][]string{
+	IDPTypeOIDC: {
+		PropIssuer,
+		PropJwksEndpoint,
+	},
+	IDPTypeGoogle: {
+		PropIssuer,
+		PropJwksEndpoint,
 	},
 }
--- a/backend/internal/idp/service.go
+++ b/backend/internal/idp/service.go
@@ -37,6 +37,7 @@ type IDPServiceInterface interface {
 	GetIdentityProviderList(ctx context.Context) ([]BasicIDPDTO, *serviceerror.ServiceError)
 	GetIdentityProvider(ctx context.Context, idpID string) (*IDPDTO, *serviceerror.ServiceError)
 	GetIdentityProviderByName(ctx context.Context, idpName string) (*IDPDTO, *serviceerror.ServiceError)
+	GetIdentityProviderByIssuer(ctx context.Context, issuer string) (*IDPDTO, *serviceerror.ServiceError)
 	UpdateIdentityProvider(ctx context.Context, idpID string, idp *IDPDTO) (*IDPDTO, *serviceerror.ServiceError)
 	DeleteIdentityProvider(ctx context.Context, idpID string) *serviceerror.ServiceError
 }
@@ -155,6 +156,26 @@ func (is *idpService) GetIdentityProviderByName(ctx context.Context,
 			return nil, &ErrorIDPNotFound
 		}
 		logger.Error("Failed to get identity provider by name", log.String("idpName", idpName), log.Error(err))
+		return nil, &serviceerror.InternalServerError
+	}
+
+	return idp, nil
+}
+
+// GetIdentityProviderByIssuer retrieves an identity provider by its issuer property.
+func (is *idpService) GetIdentityProviderByIssuer(ctx context.Context,
+	issuer string) (*IDPDTO, *serviceerror.ServiceError) {
+	logger := is.logger
+	if strings.TrimSpace(issuer) == "" {
+		return nil, &ErrorInvalidIDPID
+	}
+
+	idp, err := is.idpStore.GetIdentityProviderByIssuer(ctx, issuer)
+	if err != nil {
+		if errors.Is(err, ErrIDPNotFound) {
+			return nil, &ErrorIDPNotFound
+		}
+		logger.Error("Failed to get identity provider by issuer", log.String("issuer", issuer), log.Error(err))
 		return nil, &serviceerror.InternalServerError
 	}
 
--- a/backend/internal/idp/store_constants.go
+++ b/backend/internal/idp/store_constants.go
@@ -58,4 +58,15 @@ var (
 		ID:    "IPQ-IDP_MGT-07",
 		Query: `SELECT COUNT(*) AS count FROM "IDP" WHERE DEPLOYMENT_ID = $1`,
 	}
+	// queryGetIdentityProviderByIssuer is the query to get an IDP by its issuer property value.
+	queryGetIdentityProviderByIssuer = model.DBQuery{
+		ID: "IPQ-IDP_MGT-08",
+		PostgresQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
+			"WHERE PROPERTIES @> $1::jsonb AND DEPLOYMENT_ID = $2 LIMIT 1",
+		SQLiteQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
+			"WHERE EXISTS (SELECT 1 FROM json_each(PROPERTIES) " +
+			"WHERE json_extract(value, '$.name') = 'issuer' " +
+			"AND json_extract(value, '$.value') = $1) " +
+			"AND DEPLOYMENT_ID = $2 LIMIT 1",
+	}
 )
--- a/backend/internal/oauth/oauth2/tokenservice/utils.go
+++ b/backend/internal/oauth/oauth2/tokenservice/utils.go
@@ -243,24 +243,9 @@ func ExtractUserAttributes(claims map[string]interface{}) map[string]interface{}
 	return userAttributes
 }
 
-// getValidIssuers collects all valid/trusted issuers for the given OAuth application.
-func getValidIssuers(oauthApp *inboundmodel.OAuthClient) map[string]bool {
-	validIssuers := make(map[string]bool)
-
-	tokenConfig := ResolveTokenConfig(oauthApp, TokenTypeAccess)
-	validIssuers[tokenConfig.Issuer] = true
-
-	// TODO: Add support for external issuers
-	return validIssuers
-}
-
-// validateIssuer validates that a token issuer is trusted by checking against configured issuers.
-func validateIssuer(issuer string, oauthApp *inboundmodel.OAuthClient) error {
-	validIssuers := getValidIssuers(oauthApp)
-	if !validIssuers[issuer] {
-		return fmt.Errorf("token issuer '%s' is not supported", issuer)
-	}
-	return nil
+// isSelfIssuer reports whether the given issuer is the server's own configured issuer.
+func isSelfIssuer(issuer string) bool {
+	return issuer == config.GetServerRuntime().Config.JWT.Issuer
 }
 
 // FetchUserAttributes fetches user attributes and merges default claims and groups into the return map.
--- a/backend/internal/oauth/oauth2/tokenservice/validator.go
+++ b/backend/internal/oauth/oauth2/tokenservice/validator.go
@@ -19,10 +19,12 @@
 package tokenservice
 
 import (
+	"context"
 	"fmt"
 	"slices"
 	"time"
 
+	"github.com/asgardeo/thunder/internal/idp"
 	inboundmodel "github.com/asgardeo/thunder/internal/inboundclient/model"
 	oauth2model "github.com/asgardeo/thunder/internal/oauth/oauth2/model"
 	"github.com/asgardeo/thunder/internal/oauth/oauth2/utils"
@@ -34,18 +36,21 @@ import (
 type TokenValidatorInterface interface {
 	ValidateAccessToken(token string) (*AccessTokenClaims, error)
 	ValidateRefreshToken(token string, clientID string) (*RefreshTokenClaims, error)
-	ValidateSubjectToken(token string, oauthApp *inboundmodel.OAuthClient) (*SubjectTokenClaims, error)
+	ValidateSubjectToken(ctx context.Context, token string, oauthApp *inboundmodel.OAuthClient) (
+		*SubjectTokenClaims, error)
 }
 
 // TokenValidator implements TokenValidatorInterface.
 type tokenValidator struct {
 	jwtService jwt.JWTServiceInterface
+	idpService idp.IDPServiceInterface
 }
 
 // NewTokenValidator creates a new TokenValidator instance.
-func newTokenValidator(jwtService jwt.JWTServiceInterface) TokenValidatorInterface {
+func newTokenValidator(jwtService jwt.JWTServiceInterface, idpService idp.IDPServiceInterface) TokenValidatorInterface {
 	return &tokenValidator{
 		jwtService: jwtService,
+		idpService: idpService,
 	}
 }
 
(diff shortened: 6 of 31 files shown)


backend/internal/idp/service.go
```diff
@@ -226,6 +227,57 @@ func (is *idpService) UpdateIdentityProvider(ctx context.Context, idpID string,
 	return idp, nil
 }
 
+// GetTokenExchangeIssuers returns the list of external IDPs configured for token exchange.
+// An IDP is included when token_exchange_enabled == "true", issuer is non-empty, and
+// jwks_endpoint is non-empty.
+func (is *idpService) GetTokenExchangeIssuers(ctx context.Context) (
```
rajithacharith: Can't we add a store layer function to get this via a single query? Feels like this logic can be optimized.

sahandilshan: This is fixed with 9009938ad7244ec25cd1af075dce7d2e645c0f4b
Add a single query to get values with properties. And once we retrieve, use cache from that point onwards to avoid DB calls again

backend/internal/idp/service.go
```diff
@@ -39,21 +40,26 @@ type IDPServiceInterface interface {
...
+
 // idpService is the default implementation of the IdPServiceInterface.
 type idpService struct {
-	idpStore      idpStoreInterface
-	transactioner transaction.Transactioner
-	logger        *log.Logger
+	idpStore                 idpStoreInterface
+	transactioner            transaction.Transactioner
+	logger                   *log.Logger
+	tokenExchangeIssuerCache cache.CacheInterface[[]TokenExchangeIssuerInfo]
```
senthalan: Why we have this cache in the service layer?

We usually maintain cache in the store level as CacheBackStore

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,68 @@ func (tv *tokenValidator) ValidateSubjectToken(
...
+// resolveExternalIssuer looks up an external IDP whose issuer property matches the given issuer.
+func (tv *tokenValidator) resolveExternalIssuer(ctx context.Context, issuer string) (
+	*idp.TokenExchangeIssuerInfo, error) {
+	if tv.idpService == nil {
+		return nil, fmt.Errorf("no external issuers configured")
 	}
 
-	if err := tv.verifyTokenSignatureByIssuer(token, iss, oauthApp); err != nil {
-		return nil, fmt.Errorf("invalid subject token signature: %w", err)
+	issuers, svcErr := tv.idpService.GetTokenExchangeIssuers(ctx)
```
senthalan: why we need to get all the issuers here? why can't we have 

```
idp, svcErr := tv.idpService.GetIDPByIssuer(ctx, issuer)
issuerInfo := TokenExchangeIssuerInfo{
      .. build issuer from the IDP object
}
return issuerInfo
```

This way the idp package will not have the TokenExchangeIssuerInfo model definition, it will be generic as the IDP and oauth2 package will have the TokenExchangeIssuerInfo

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,68 @@ func (tv *tokenValidator) ValidateSubjectToken(
 		return nil, fmt.Errorf("subject token is missing 'iss' claim: %w", err)
 	}
 
-	if err := validateIssuer(iss, oauthApp); err != nil {
-		return nil, err
+	// Try Thunder's own issuer first.
+	if validateIssuer(iss, oauthApp) == nil {
```
senthalan: Shouldn't this function name indicate that it's checking the Thunder's own issuer

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -190,8 +250,8 @@ func (tv *tokenValidator) ValidateSubjectToken(
 		return nil, err
 	}
 
-	// Determine if this is an auth assertion
-	isAuthAssertion := tv.isAuthAssertion(claims)
+	// Determine if this is an auth assertion (external tokens never have assurance).
+	isAuthAssertion := !isExternal && tv.isAuthAssertion(claims)
```
senthalan: I think we should be able to identity the token as AuthAssertion without isExternal flag is passed

docs/content/guides/guides/trusted-issuer.mdx
```diff
@@ -152,6 +152,120 @@ configuration:
...
+  -H "Content-Type: application/json" \
+  -d '{
+    "name": "Asgardeo",
+    "type": "OIDC",
+    "properties": [
+      {"name": "client_id", "value": "your_client_id_at_asgardeo"},
+      {"name": "client_secret", "value": "your_client_secret"},
+      {"name": "redirect_uri", "value": "https://thunder.example.com/callback"},
+      {"name": "authorization_endpoint", "value": "https://api.asgardeo.io/t/myorg/oauth2/authorize"},
+      {"name": "token_endpoint", "value": "https://api.asgardeo.io/t/myorg/oauth2/token"},
```
senthalan: why does these properties needs to be provided when creating the trusted issuer

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: This is fixed now

backend/internal/idp/service.go
sahandilshan: This is fixed now

docs/content/guides/guides/trusted-issuer.mdx
sahandilshan: This is fixed now. For token exchange IDP, only `issuer`, jwks_endpoint`, client_id and `token_exchange_enabled` configs are required

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: This is fixed now

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: This is fixed now

backend/internal/idp/constants.go
```diff
@@ -74,9 +76,10 @@ const (
 // idpPropertyConfig defines the required and optional properties for an IDP type,
 // along with any default values.
 type idpPropertyConfig struct {
-	Required []string
-	Optional []string
-	Defaults map[string]string
+	Required                         []string
+	Optional                         []string
+	RequiredWhenTokenExchangeEnabled []string
```
senthalan: I hope this is to keep the list of the properties needed when the token exchange is enabled. So shouldn't this be a constants rather than property config

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,79 @@ func (tv *tokenValidator) ValidateSubjectToken(
 		return nil, fmt.Errorf("subject token is missing 'iss' claim: %w", err)
 	}
 
-	if err := validateIssuer(iss, oauthApp); err != nil {
```
rajithacharith, line 175: Following paths have been designed to support external trusted issuers. 
As we are not using it, shall we simplify that as well?

```
// getValidIssuers collects all valid/trusted issuers for the given OAuth application.
func getValidIssuers(oauthApp *inboundmodel.OAuthClient) map[string]bool {
	validIssuers := make(map[string]bool)

	tokenConfig := ResolveTokenConfig(oauthApp, TokenTypeAccess)
	validIssuers[tokenConfig.Issuer] = true

	// TODO: Add support for external issuers
	return validIssuers
}

// isThunderIssuer reports whether the given issuer is one of Thunder's own configured issuers.
func isThunderIssuer(issuer string, oauthApp *inboundmodel.OAuthClient) bool {
	return getValidIssuers(oauthApp)[issuer]
}
```

https://github.com/sahandilshan/thunder/blob/f5c4f297c9d243267683dbb730e135b0ad69088d/backend/internal/oauth/oauth2/tokenservice/utils.go#L247-L248

backend/internal/oauth/oauth2/tokenservice/utils.go
senthalan: Shall we remove this TODO ?

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -254,13 +254,9 @@ func getValidIssuers(oauthApp *inboundmodel.OAuthClient) map[string]bool {
...
 
-// validateIssuer validates that a token issuer is trusted by checking against configured issuers.
-func validateIssuer(issuer string, oauthApp *inboundmodel.OAuthClient) error {
-	validIssuers := getValidIssuers(oauthApp)
-	if !validIssuers[issuer] {
-		return fmt.Errorf("token issuer '%s' is not supported", issuer)
-	}
-	return nil
+// isThunderIssuer reports whether the given issuer is one of Thunder's own configured issuers.
+func isThunderIssuer(issuer string, oauthApp *inboundmodel.OAuthClient) bool {
```
senthalan: Shall we not use the Thunder word here? 
```suggestion
// isSelfIssuer reports whether the given issuer is one of servers own configured issuers.
func isSelfIssuer(issuer string, oauthApp *inboundmodel.OAuthClient) bool {
```

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,79 @@ func (tv *tokenValidator) ValidateSubjectToken(
...
-		return nil, err
+	// Try Thunder's own issuer first.
+	if isThunderIssuer(iss, oauthApp) {
+		if err := tv.verifyTokenSignatureByIssuer(token, iss, oauthApp); err != nil {
+			return nil, fmt.Errorf("invalid subject token signature: %w", err)
+		}
+		return tv.extractSubjectTokenClaims(token, iss, claims, oauthApp)
+	}
+
+	// Not a Thunder-issued token — try external IDP issuers.
```
senthalan: let's remove the thunder word here

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,79 @@ func (tv *tokenValidator) ValidateSubjectToken(
...
+	issuerInfo, resolveErr := tv.resolveExternalIssuer(ctx, iss)
+	if resolveErr != nil {
+		return nil, fmt.Errorf("token issuer '%s' is not supported", iss)
```
rajithacharith: Let's wrap the error and return, 
```
err := validateIssuer(issuer)
	if err != nil {
		// Wrap the original error with more context
		return fmt.Errorf("failed to exchange token for issuer %q: %w", issuer, err)
	}
```

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,79 @@ func (tv *tokenValidator) ValidateSubjectToken(
...
+	if svcErr != nil {
+		return nil, fmt.Errorf("invalid subject token signature: %v", svcErr.Error)
+	}
+
+	// Validate that the token's audience contains the IDP's client_id.
+	if issuerInfo.ClientID != "" {
+		auds, audErr := extractAudiences(claims)
+		if audErr != nil || !slices.Contains(auds, issuerInfo.ClientID) {
+			return nil, fmt.Errorf(
+				"external token audience does not contain expected client_id '%s'", issuerInfo.ClientID)
```
rajithacharith: Shall we handle audErr separately and verification separately?

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -172,14 +178,79 @@ func (tv *tokenValidator) ValidateSubjectToken(
...
+		}
+	}
+
+	return tv.extractSubjectTokenClaims(token, iss, claims, oauthApp)
```
rajithacharith, line 211: Can you check will this work properly with isAuthAssertion path?
```
// For auth assertions, audience is required, must be a single value, and must match
		// config default or requesting client's app_id. Multi-aud auth assertions are rejected
		// as a defense-in-depth measure (auth assertions are a narrow control surface).
		auds, err = extractAudiences(claims)
		if err != nil {
			return nil, fmt.Errorf("auth assertion is missing 'aud' claim: %w", err)
		}
		if len(auds) > 1 {
			return nil, fmt.Errorf("auth assertion must have a single audience")
		}

		defaultAudience := config.GetServerRuntime().Config.JWT.Audience
		clientAppID := oauthApp.AppID

		if !slices.Contains([]string{defaultAudience, clientAppID}, auds[0]) {
			return nil, fmt.Errorf("auth assertion audience mismatch")
		}
```

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -254,13 +254,9 @@ func getValidIssuers(oauthApp *inboundmodel.OAuthClient) map[string]bool {
...
+	return getValidIssuers(oauthApp)[issuer]
```
senthalan: As per @thiva-k we have removed the support to have multiple issuers in Thunder per application. So ideally this should be refactored to check for only one issuer which is configured in the deployment level

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: Fixed

backend/internal/oauth/oauth2/tokenservice/utils.go
sahandilshan: THis is fixed now. Removed the previous dead code as well

backend/internal/oauth/oauth2/tokenservice/utils.go
sahandilshan: This is fixed

backend/internal/idp/constants.go
sahandilshan: This is fixed now

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: This is fixed now

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: This is fixed now

backend/internal/idp/store_constants.go
```diff
@@ -58,4 +58,15 @@ var (
 		ID:    "IPQ-IDP_MGT-07",
 		Query: `SELECT COUNT(*) AS count FROM "IDP" WHERE DEPLOYMENT_ID = $1`,
 	}
+	// queryGetIdentityProviderByIssuer is the query to get an IDP by its issuer property value.
```
KaveeshaPiumini, line 61: Currently, the SQLite query is doing a complex array traversal (json_each) which will result in a full table scan, and the Postgres query relies on the containment operator (@>) which requires either a large GIN index or a full table scan.

We can make these queries significantly faster and unify the Go parameter we pass to them by using targeted expression indexes. By extracting the string directly, we can use standard B-Tree indexes which are much smaller and have less impact on write performance.

Bringing issuer out as a separate, dedicated column is the better approach if possible.

backend/internal/idp/store_constants.go
```diff
@@ -58,4 +58,15 @@ var (
...
+	queryGetIdentityProviderByIssuer = model.DBQuery{
+		ID: "IPQ-IDP_MGT-08",
+		PostgresQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
+			"WHERE PROPERTIES @> $1::jsonb AND DEPLOYMENT_ID = $2 LIMIT 1",
+		SQLiteQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
+			"WHERE EXISTS (SELECT 1 FROM json_each(PROPERTIES) " +
+			"WHERE json_extract(value, '$.name') = 'issuer' " +
+			"AND json_extract(value, '$.value') = $1) " +
+			"AND DEPLOYMENT_ID = $2 LIMIT 1",
+	}
```
KaveeshaPiumini, line 71:
```suggestion
queryGetIdentityProviderByIssuer = model.DBQuery{
    ID: "IPQ-IDP_MGT-08",
    
    // Postgres: Extracts text directly to hit the B-Tree index
    PostgresQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
        "WHERE PROPERTIES->>'issuer' = $1 AND DEPLOYMENT_ID = $2 LIMIT 1",
        
    // SQLite: Extracts text directly to hit the Expression index
    SQLiteQuery: "SELECT ID, NAME, DESCRIPTION, TYPE, PROPERTIES FROM IDP " +
        "WHERE json_extract(PROPERTIES, '$.issuer') = $1 " +
        "AND DEPLOYMENT_ID = $2 LIMIT 1",
}
```

We can pass the exact same simple string (e.g., "https://example.com") as $1 for both databases.

sahandilshan: The suggested queries assume `PROPERTIES` is a flat JSON object (`{"issuer": "..."}`), but it is actually stored as a JSON array of `{name, value}` objects (e.g. `[{"name":"issuer","value":"https://..."}]`) — see `SerializePropertiesToJSONArray` in `cmodels/property.go`. So `PROPERTIES->>'issuer'` and `json_extract(PROPERTIES, '$.issuer')`  would always return NULL and match zero rows.

For the perf concern in the parent comment: adding a GIN index (`jsonb_path_ops`) on the `PROPERTIES` column makes the existing `@>` containment query efficient on Postgres without changing the query or Go code.

backend/internal/idp/store_constants.go
sahandilshan: This is fixed now. Added indexing for postgres DB

backend/dbscripts/configdb/postgres.sql
```diff
@@ -139,6 +139,9 @@ CREATE TABLE "IDP" (
 -- Composite index for name-based IDP lookups
 CREATE INDEX idx_idp_name_deployment ON "IDP" (DEPLOYMENT_ID, NAME);
 
+-- GIN index for JSONB containment queries on IDP properties (e.g. issuer lookup via @>)
+CREATE INDEX idx_idp_properties ON "IDP" USING GIN ("PROPERTIES" jsonb_path_ops);
```
KaveeshaPiumini, line 143: To actually get the performance benefits, we'll need to add the following indexes

-- For PostgreSQL
CREATE INDEX idx_idp_issuer ON IDP ((PROPERTIES->>'issuer'));

-- For SQLite
CREATE INDEX idx_idp_issuer ON IDP (json_extract(PROPERTIES, '$.issuer'));

backend/internal/oauth/oauth2/tokenservice/validator.go
sahandilshan: Hi @rajithacharith,

This is a valid suggestion. If the external token has this assurance claim, it will break with the client id check. We need to fix it. Created a seperate issue to track this https://github.com/asgardeo/thunder/issues/2617

backend/internal/idp/constants.go
```diff
@@ -159,3 +165,15 @@ var idpPropertyConfigs = map[IDPType]idpPropertyConfig{
...
+// tokenExchangeRequiredProps defines the required properties per IDP type when token exchange is enabled.
+var tokenExchangeRequiredProps = map[IDPType][]string{
+	IDPTypeOIDC: {
+		PropIssuer,
+		PropJwksEndpoint,
+	},
+	IDPTypeGoogle: {
+		PropIssuer,
+		PropJwksEndpoint,
+	},
```
rajithacharith, line 178: Is there a specific reason to avoid Github? If not let's get that with a separate PR

rajithacharith (approved):

backend/dbscripts/configdb/postgres.sql
KaveeshaPiumini: @coderabbitai create an issue to track this change

backend/internal/idp/constants.go
sahandilshan: Will add

sahandilshan: @coderabbitai create an issue for this

PR #2549, opened 2026-05-05
Disable auto-inferred registration flow by default

### LINKED ISSUE #2480: Remove auto-inferred registration flow creation
## Summary

When a login (authentication) flow is created, a registration flow is automatically inferred and created as a side effect. This implicit behavior should be removed so that users explicitly create registration flows themselves.

## Current Behavior

- `CreateFlow()` in `backend/internal/flow/mgt/service.go` calls `tryInferRegistrationFlow()` after saving an auth flow.
- The inference logic in `backend/internal/flow/mgt/inference.go` clones the auth flow and transforms it into a registration flow by renaming labels, inserting provisioning/user-type-resolver/phone-input nodes, and stripping auth-specific properties.
- This is gated by the `AutoInferRegistration` config flag but is enabled by default.

### Backend
- Remove the `tryInferRegistrationFlow()` call and method from `backend/internal/flow/mgt/service.go`
- Delete `backend/internal/flow/mgt/inference.go` (inference service and all transformation logic)
- Remove inference-related constants from `backend/internal/flow/mgt/constants.go`
- Remove `AutoInferRegistration` field from `FlowConfig` in `backend/internal/system/config/config.go`
- Remove `auto_infer_registration` from deployment config files
- Remove the inference service interface and regenerate mocks (`make mockery`)
- Remove inference-related test cases from `backend/internal/flow/mgt/service_test.go`
- Review bootstrap flows in `backend/cmd/server/bootstrap/flows/` for any dependencies on inference

### Frontend
- No changes expected. The registration flow section in application settings just lists available flows.

## Notes

- After this change, users will need to explicitly create registration flows through the console.
- Ensure the console UI supports standalone registration flow creation before merging.

DonOmalVindula:
## Summary

- Disable the `auto_infer_registration` config default from `true` to `false` in `default.json`, so registration flows are no longer automatically inferred when creating authentication flows.

## Related Issue

Closes #2480

## Approach

Instead of deleting the auto-inference logic entirely, we disable it by default via the existing `auto_infer_registration` config flag. This preserves the code for cases where users may want to opt-in to auto-inference in the future by setting `flow.auto_infer_registration: true` in their deployment config.

## Test plan

- [x] Verified that creating an authentication flow no longer auto-creates a registration flow with the default config
- [ ] Verify existing registration flows are unaffected

## Summary by CodeRabbit

* **Configuration**
  * Automatic inference of a registration flow is now disabled by default.

* **Database**
  * Registration flow ID for inbound clients is now nullable, allowing empty/unspecified values.

* **Tests**
  * Integration tests updated to reflect disabled auto-inference and now explicitly set or verify registration flow IDs.


brionmario (dismissed):

backend/internal/inboundclient/service.go
```diff
@@ -446,7 +446,7 @@ func (s *inboundClientService) resolveFlowDefaults(ctx context.Context, c *inbou
 		}
 		c.AuthFlowID = flow.ID
 	}
-	if c.RegistrationFlowID == "" && c.AuthFlowID != "" {
+	if c.RegistrationFlowID == "" && c.AuthFlowID != "" && config.GetServerRuntime().Config.Flow.AutoInferRegistration {
```
thiva-k, line 449: Won't this fail if we do not set a registration flow due to NOT NULL on db?

DonOmalVindula: According to [1], this column is nullable (no NOT NULL constraint), so it won't fail

[1] - https://github.com/asgardeo/thunder/blob/c42a984383a721285757b5d21d21d1e1bba4af16/backend/internal/application/init_test.go#L82

thiva-k: Our db scripts have it

https://github.com/asgardeo/thunder/blob/c42a984383a721285757b5d21d21d1e1bba4af16/backend/dbscripts/configdb/postgres.sql#L102

DonOmalVindula: We will discuss this over a discussion - https://github.com/asgardeo/thunder/discussions/2556

However, it won't actually fail because the Go struct's zero value is "" (empty string), not NULL. The store passes `client.RegistrationFlowID` directly as a query param, so the DB receives an empty string which satisfies the `NOT NULL` constraint.

That said, having empty strings in a `NOT NULL` column is not ideal from a data integrity standpoint. That is why I've opened a discussion to decide how we want to handle this properly.

thiva-k (approved):

Issue #2470, opened 2026-04-27
Inconsistent DB table name casing due to SQL keyword quoting


KaveeshaPiumini:
### Problem

Most table names in the DB scripts are written without quotes (e.g. `CREATE TABLE ROLE_PERMISSION`). In PostgreSQL, unquoted identifiers are case-folded to lowercase, so these tables are effectively stored as lowercase (e.g. `role_permission`).

However, a handful of table names clash with SQL reserved keywords — `ROLE` and `GROUP` — and are therefore wrapped in double quotes to avoid parse errors:

| Script | Table | Declaration |
|--------|-------|-------------|
| `configdb/postgres.sql`, `configdb/sqlite.sql` | `ROLE` | `CREATE TABLE "ROLE"` |
| `userdb/postgres.sql`, `userdb/sqlite.sql` | `GROUP` | `CREATE TABLE "GROUP"` |

Because they are quoted, PostgreSQL preserves their case exactly as written (`ROLE`, `GROUP`), while all other tables end up lowercase. This creates an **inconsistent schema** where mixing queries against quoted and unquoted names requires knowing which tables are keywords and which are not.

### Root cause

Quoting is necessary for reserved-word table names and **cannot be removed**. The inconsistency arises because the non-keyword tables are left unquoted.

### Proposed fix

Quote **all** table names across every DB script (`configdb`, `runtimedb`, `userdb`) so that casing is uniformly preserved in uppercase. For example:

```sql
-- Before
CREATE TABLE ROLE_PERMISSION (
    ...
    FOREIGN KEY (ROLE_ID) REFERENCES "ROLE" (ID) ON DELETE CASCADE
);

-- After
CREATE TABLE "ROLE_PERMISSION" (
    ...
    FOREIGN KEY (ROLE_ID) REFERENCES "ROLE" (ID) ON DELETE CASCADE
);
```

This change affects all `CREATE TABLE`, `CREATE INDEX`, `FOREIGN KEY REFERENCES`, and any ad-hoc `INSERT`/`SELECT` statements inside the scripts.

Any Go source files that embed raw SQL table names (e.g. in query strings) will also need to be updated to use the quoted uppercase form.

PR #2550, opened 2026-05-05
Migrate user schema to entity type and introduce `/agent-types` API

thiva-k:
## Summary

Renames the `userschema` package to `entitytype` across the codebase and introduces a parallel `/agent-types` API alongside the renamed `/user-types` endpoints. Both endpoint families share a single `entitytype` handler bound to a `TypeCategory` (`user` / `agent`), with category-scoped validation enforced from the API down to the store query (`ENTITY_TYPES.CATEGORY`). Adds the agent-type API spec to `api/agent.yaml` mirroring the user-type spec.

---
- **Endpoint rename**: `/user-schemas` and `/user-schemas/{id}` → `/user-types` and `/user-types/{id}`.
- **Permission/scope rename**: `system:userschema` → `system:usertype` and `system:userschema:view` → `system:usertype:view`.
- **Helm values rename**: `configuration.consoleClient.scopes` in `install/helm/values.yaml` updated to use the new scope names.

#### 💥 Impact
- API clients calling `/user-schemas*` will receive 404. They must update to `/user-types*`.
- OAuth clients requesting the `system:userschema*` scopes will be denied. Console deployments must update their requested scopes.
- Helm-deployed installations with custom `consoleClient.scopes` referencing `system:userschema:view` need to update the value.

#### 🔄 Migration Guide
1. **API consumers**: replace `/user-schemas` with `/user-types` in all requests. Request/response shapes are unchanged.
2. **OAuth/console scopes**:
   ```diff
   - 'system:userschema:view'
   + 'system:usertype:view'


backend/dbscripts/configdb/sqlite.sql
```diff
@@ -1,19 +1,20 @@
--- Table to store User Schemas
-CREATE TABLE USER_SCHEMAS (
+-- Table to store Entity Schemas (user/agent categories)
+CREATE TABLE ENTITY_TYPES (
```
KaveeshaPiumini:
```suggestion
CREATE TABLE "ENTITY_TYPES" (
```

fixing - https://github.com/asgardeo/thunder/issues/2470

backend/dbscripts/configdb/postgres.sql
```diff
@@ -1,19 +1,20 @@
--- Table to store User Schemas
-CREATE TABLE USER_SCHEMAS (
+-- Table to store Entity Schemas (user/agent categories)
+CREATE TABLE ENTITY_TYPES (
```
KaveeshaPiumini:
```suggestion
CREATE TABLE "ENTITY_TYPES" (
```

fixing - https://github.com/asgardeo/thunder/issues/2470

backend/dbscripts/configdb/sqlite.sql
thiva-k: Updated

backend/dbscripts/configdb/postgres.sql
thiva-k: Updated

backend/internal/system/export/service.go
```diff
@@ -45,7 +45,7 @@ const (
 	resourceTypeApplication        = "application"
 	resourceTypeIdentityProvider   = "identity_provider"
 	resourceTypeNotificationSender = "notification_sender"
-	resourceTypeUserSchema         = "user_schema"
+	resourceTypeEntityType         = "user_type"
```
rajithacharith, line 48: Will there be agent_type in future? If so, Let's keep resourceTypeUserType

rajithacharith (approved):

backend/internal/system/export/service.go
thiva-k: Yes. Will update this in a follow-up PR

PR #2553, opened 2026-05-05
Align credential schema attribute handling in provisioning

### LINKED ISSUE #2633: Persist schema-defined credential attributes during registration provisioning
### Current Limitation

During registration flows, credential attributes can be prompted from the user based on the schema, but provisioning may still miss them when creating the user. This happens because credential persistence is limited to node inputs instead of using the full schema-defined credential attribute set.

### Suggested Improvement

Make credential provisioning schema-backed as well. Use credential attribute metadata from the user schema, with a requiredOnly option similar to non-credential attributes, so registration flows consistently prompt and persist credential fields defined in the schema.

Dilusha-Madushan:
### Purpose
This PR fixes registration provisioning when credential attributes are defined dynamically in the user schema. Previously, credential inputs could be prompted to the user but then be missed during user provisioning because the executor only persisted credential attributes based on node inputs instead of the schema-backed credential set.

This change also makes credential schema handling consistent with non-credential schema handling by exposing credential attribute metadata with a `requiredOnly` option and using that metadata in the provisioning executor when building dynamic prompt inputs.

### Approach
Updated the userschema contract to support `GetCredentialAttributeInfos(..., requiredOnly)` similar to `GetNonCredentialAttributes(..., requiredOnly)`, and propagated that change through the service layer and generated mocks.

In the provisioning executor, credential prompting and credential persistence are now handled as two explicit paths:
- Prompt-time credential resolution uses credential schema metadata, including required/display name information, so dynamic prompt inputs are built correctly.
- Provisioning-time credential persistence fetches schema credential attribute names and persists any available values from the execution context instead of limiting persistence to credential node inputs only.

The executor logic was also clarified by separating prompt-side non-credential and credential schema fetch helpers, and tests were expanded to cover:
- optional credential overridden to required by node input
- capped forwarded dynamic prompt inputs
- credential attribute persistence during provisioning
- required-only vs all-attributes behavior for credential schema metadata in schema and service layers

### Related Issues
- https://github.com/asgardeo/thunder/issues/2633

### Related PRs
- https://github.com/asgardeo/thunder/pull/2537

## Summary by CodeRabbit

* **New Features**
  * Enhanced provisioning prompts: separate required/optional credential and non-credential pools with max-per-prompt batching.
  * Improved password-field synchronization for forwarded inputs.

* **Bug Fixes**
  * Prevent re-prompting of already-present optional attributes.
  * Ensure promotions to password type don't mutate shared meta across runs.
  * Better error handling when credential attribute retrieval fails.

* **Tests**
  * Expanded tests for credential handling, prompting limits, and edge cases.

### DIFF
--- a/backend/internal/entitytype/model/schema.go
+++ b/backend/internal/entitytype/model/schema.go
@@ -113,21 +113,27 @@ func (cs *Schema) ValidateAsDisplayAttribute(name string) DisplayAttributeStatus
 	return DisplayAttributeValid
 }
 
-// AttributeInfo holds an attribute name, its required status, and its human-readable display label.
-// DisplayName may be empty when the schema definition omits the `displayName` field;
+// AttributeInfo holds an attribute name, its required and credential status, and its human-readable
+// display label. DisplayName may be empty when the schema definition omits the `displayName` field;
 // callers should fall back to Attribute when rendering a label.
 type AttributeInfo struct {
 	Attribute   string
 	DisplayName string
 	Required    bool
+	Credential  bool
 }
 
-// GetNonCredentialAttributes returns top-level properties not marked as credentials.
-// When requiredOnly is true, only required properties are included.
-func (cs *Schema) GetNonCredentialAttributes(requiredOnly bool) []AttributeInfo {
+// GetAttributes returns top-level properties filtered by the provided flags.
+// allowCredential includes credential properties; allowNonCredential includes non-credential
+// properties. When requiredOnly is true, only required properties are included.
+func (cs *Schema) GetAttributes(allowCredential, allowNonCredential, requiredOnly bool) []AttributeInfo {
 	result := make([]AttributeInfo, 0, len(cs.properties))
 	for attr, prop := range cs.properties {
-		if prop.isCredential() {
+		isCredential := prop.isCredential()
+		if isCredential && !allowCredential {
+			continue
+		}
+		if !isCredential && !allowNonCredential {
 			continue
 		}
 		if requiredOnly && !prop.isRequired() {
@@ -137,21 +143,10 @@ func (cs *Schema) GetNonCredentialAttributes(requiredOnly bool) []AttributeInfo 
 			Attribute:   attr,
 			DisplayName: prop.getDisplayName(),
 			Required:    prop.isRequired(),
+			Credential:  isCredential,
 		})
 	}
 	return result
-}
-
-// GetCredentialAttributes returns the names of top-level properties marked as credentials.
-func (cs *Schema) GetCredentialAttributes() []string {
-	var fields []string
-	for name, prop := range cs.properties {
-		if prop.isCredential() {
-			fields = append(fields, name)
-		}
-	}
-
-	return fields
 }
 
 // GetUniqueAttributes returns the names of top-level properties marked as unique.
--- a/backend/internal/entitytype/service.go
+++ b/backend/internal/entitytype/service.go
@@ -74,18 +74,16 @@ type EntityTypeServiceInterface interface {
 		attributes json.RawMessage,
 		exists func(map[string]interface{}) (bool, error),
 	) (bool, *serviceerror.ServiceError)
-	GetCredentialAttributes(
+	GetAttributes(
 		ctx context.Context, category TypeCategory, entityType string,
-	) ([]string, *serviceerror.ServiceError)
+		allowCredential, allowNonCredential, requiredOnly bool,
+	) ([]AttributeInfo, *serviceerror.ServiceError)
 	GetUniqueAttributes(
 		ctx context.Context, category TypeCategory, entityType string,
 	) ([]string, *serviceerror.ServiceError)
 	GetDisplayAttributesByNames(
 		ctx context.Context, category TypeCategory, names []string,
 	) (map[string]string, *serviceerror.ServiceError)
-	GetNonCredentialAttributes(
-		ctx context.Context, category TypeCategory, entityType string, requiredOnly bool,
-	) ([]AttributeInfo, *serviceerror.ServiceError)
 }
 
 // entityTypeService is the default implementation of the EntityTypeServiceInterface.
@@ -616,10 +614,13 @@ func (us *entityTypeService) ValidateEntityUniqueness(
 	return true, nil
 }
 
-// GetCredentialAttributes returns the names of schema properties marked as credentials for a given entity type.
-func (us *entityTypeService) GetCredentialAttributes(
+// GetAttributes returns schema properties filtered by the provided flags for the given entity type.
+// allowCredential includes credential properties; allowNonCredential includes non-credential properties.
+// When requiredOnly is true, only required properties are included.
+func (us *entityTypeService) GetAttributes(
 	ctx context.Context, category TypeCategory, entityType string,
-) ([]string, *serviceerror.ServiceError) {
+	allowCredential, allowNonCredential, requiredOnly bool,
+) ([]AttributeInfo, *serviceerror.ServiceError) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, entityTypeLoggerComponentName))
 
 	if svcErr := validateCategory(category); svcErr != nil {
@@ -631,10 +632,10 @@ func (us *entityTypeService) GetCredentialAttributes(
 		if errors.Is(err, ErrEntityTypeNotFound) {
 			return nil, entityTypeNotFoundErr(category)
 		}
-		return nil, logAndReturnServerError(logger, "Failed to load entity type for credential attributes", err)
+		return nil, logAndReturnServerError(logger, "Failed to load entity type for attribute infos", err)
 	}
 
-	return compiledSchema.GetCredentialAttributes(), nil
+	return compiledSchema.GetAttributes(allowCredential, allowNonCredential, requiredOnly), nil
 }
 
 // GetUniqueAttributes returns the names of schema properties marked as unique for a given entity type.
@@ -678,29 +679,6 @@ func (us *entityTypeService) GetDisplayAttributesByNames(
 	}
 
 	return result, nil
-}
-
-// GetNonCredentialAttributes returns non-credential attributes defined in the schema for the given
-// category and entity type. When requiredOnly is true, only required attributes are returned.
-func (us *entityTypeService) GetNonCredentialAttributes(
-	ctx context.Context, category TypeCategory, entityType string, requiredOnly bool,
-) ([]AttributeInfo, *serviceerror.ServiceError) {
-	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, entityTypeLoggerComponentName))
-
-	if svcErr := validateCategory(category); svcErr != nil {
-		return nil, svcErr
-	}
-
-	compiledSchema, err := us.getCompiledSchemaForEntityType(ctx, category, entityType, logger)
-	if err != nil {
-		if errors.Is(err, ErrEntityTypeNotFound) {
-			return nil, entityTypeNotFoundErr(category)
-		}
-		return nil, logAndReturnServerError(logger,
-			"Failed to load entity type for non-credential attributes", err)
-	}
-
-	return compiledSchema.GetNonCredentialAttributes(requiredOnly), nil
 }
 
 func (us *entityTypeService) getCompiledSchemaForEntityType(
--- a/backend/internal/flow/core/prompt_node.go
+++ b/backend/internal/flow/core/prompt_node.go
@@ -448,6 +448,12 @@ func (n *promptNode) enrichInputsFromForwardedData(ctx *NodeContext, nodeResp *c
 				n.logger.Debug("Updated input required flag from ForwardedData",
 					log.String("identifier", fwdInput.Identifier))
 			}
+			if fwdInput.Type == common.InputTypePassword &&
+				nodeResp.Inputs[idx].Type != common.InputTypePassword {
+				nodeResp.Inputs[idx].Type = common.InputTypePassword
+				n.logger.Debug("Updated input type to password from ForwardedData",
+					log.String("identifier", fwdInput.Identifier))
+			}
 			if fwdInput.Type == common.InputTypeSelect &&
 				nodeResp.Inputs[idx].Type == common.InputTypeSelect &&
 				len(fwdInput.Options) > 0 {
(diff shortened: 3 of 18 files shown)


backend/internal/flow/core/prompt_node.go
```diff
@@ -326,6 +326,12 @@ func (n *promptNode) enrichInputsFromForwardedData(ctx *NodeContext, nodeResp *c
 				n.logger.Debug("Updated input required flag from ForwardedData",
 					log.String("identifier", fwdInput.Identifier))
 			}
+			if fwdInput.Type == common.InputTypePassword &&
+				nodeResp.Inputs[idx].Type != common.InputTypePassword {
+				nodeResp.Inputs[idx].Type = common.InputTypePassword
+				n.logger.Debug("Updated input type to password from ForwardedData",
+					log.String("identifier", fwdInput.Identifier))
+			}
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -279,65 +285,121 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
+			continue
+		}
+		nodeInp, inNodeInputs := nodeInputMap[attr.Attribute]
+		effectiveRequired := attr.Required
+		if inNodeInputs {
+			effectiveRequired = attr.Required || nodeInp.Required
+		}
+		if !effectiveRequired && !inNodeInputs {
+			continue
+		}
```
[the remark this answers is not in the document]

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -279,65 +285,121 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
 	}
 
-	if maxInputs := p.getMaxDynamicInputs(ctx); maxInputs > 0 && len(schemaMissingAttrs) > maxInputs {
-		schemaMissingAttrs = schemaMissingAttrs[:maxInputs]
+	// Apply maxPerPrompt cap only to ForwardedData — controls how many attrs are shown per prompt
+	// round. Required fills first so the cap trims from the optional end.
+	toForward := allSchemaMissing
+	if maxInputs := p.getMaxDynamicInputs(ctx); maxInputs > 0 && len(toForward) > maxInputs {
+		toForward = toForward[:maxInputs]
 	}
```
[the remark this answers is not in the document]

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -279,65 +284,123 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
+	// iterations know not to re-prompt them if the user left them empty.
+	p.storePresentedOptionalAttrs(execResp, toForward, alreadyPromptedOptionalAttrs)
 
-	execResp.Inputs = upsertInputs(execResp.Inputs, schemaMissingAttrs)
+	execResp.Inputs = allSchemaMissing
 	if execResp.ForwardedData == nil {
 		execResp.ForwardedData = make(map[string]interface{})
 	}
-	execResp.ForwardedData[common.ForwardedDataKeyInputs] = schemaMissingAttrs
+	execResp.ForwardedData[common.ForwardedDataKeyInputs] = toForward
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai 

The suggestion is not applicable to this implementation.

The provisioning executor is schema-driven — it can only collect and store attributes defined by the entity type. nodeInputMap is not used to introduce independent inputs; it exists solely to apply the optional → required upgrade rule on schema attributes (effectiveRequired = attr.Required || nodeInp.Required) and to gate which optional credential attributes are prompted. Node-defined inputs that have no corresponding entry in schemaAttrs or credAttrs are intentionally ignored because there is no schema field to store their value against.

checkNodeInputs has since been removed for this reason. The execResp.Inputs = allSchemaMissing assignment is correct: allSchemaMissing already reflects any required overrides from node inputs via effectiveRequired, so nothing meaningful is dropped. Merging with upsertInputs would surface schema-absent node inputs in the prompt, which the provisioning executor can never store.

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -276,101 +281,155 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
+		}
+		if !effectiveRequired && alreadyPromptedOptionalAttrs[attr.Attribute] {
 			continue
 		}
 		input := common.Input{
 			Identifier:  attr.Attribute,
 			Type:        common.InputTypeText,
-			Required:    attr.Required,
+			Required:    effectiveRequired,
 			DisplayName: attr.DisplayName,
```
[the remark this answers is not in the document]

backend/internal/entitytype/service.go
```diff
@@ -77,6 +77,9 @@ type EntityTypeServiceInterface interface {
 	GetCredentialAttributes(
 		ctx context.Context, category TypeCategory, entityType string,
 	) ([]string, *serviceerror.ServiceError)
+	GetCredentialAttributeInfos(
```
ThaminduDilshan: Rather than introducing a new `GetCredentialAttributeInfos()` method, shall we improve the existing `GetCredentialAttributes()` to return an array of AttributeInfo?

backend/internal/entitytype/model/schema.go
```diff
@@ -142,6 +142,27 @@ func (cs *Schema) GetNonCredentialAttributes(requiredOnly bool) []AttributeInfo
 	return result
 }
 
+// GetCredentialAttributeInfos returns top-level properties marked as credentials with their
+// required status and display names. When requiredOnly is true, only required properties are
+// included.
+func (cs *Schema) GetCredentialAttributeInfos(requiredOnly bool) []AttributeInfo {
```
ThaminduDilshan: Follow the same approach proposed in previous comment. Let's update existing `GetCredentialAttributes()` method

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -264,9 +264,14 @@ func (p *provisioningExecutor) handleExistingUser(ctx *core.NodeContext, userID
...
-// missing inputs to the executor response. Returns true if all required inputs — both
-// node-defined and schema-derived — are satisfied, otherwise false.
+// HasRequiredInputs checks whether all schema-driven provisioning inputs are satisfied and appends
+// any missing promptable schema attrs to the executor response. Node inputs influence requiredness
+// and prompt metadata for schema attrs, but schema-absent node inputs are ignored.
+//
+// Missing inputs are ordered as: required credentials -> required non-credentials ->
+// optional credentials -> optional non-credentials. maxPerPrompt caps the forwarded
+// prompt batch after this list is built. includeOptional only affects optional
+// non-credential attrs.
```
ThaminduDilshan, line 278: Shouldn't we order credentials to come after non credential attributes?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -276,101 +281,163 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
-	schemaAttrs, err := p.fetchSchemaAttributes(ctx, logger)
+	// Fetch non-credential and credential attributes from the user type.
+	schemaAttrs, err := p.fetchPromptNonCredentialSchemaAttrs(ctx, logger)
 	if err != nil {
 		execResp.Status = common.ExecFailure
 		return false
 	}
-	if len(schemaAttrs) == 0 {
-		return nodeInputsSatisfied
+	credAttrs, err := p.fetchPromptCredentialSchemaAttrs(ctx, logger)
```
ThaminduDilshan: Rather than fetching the schema twice for credential and non-credential attributes, shall we improve this to fetch them at once?

Thinking whether we need to introduce another new method to the schema service.
1. We can fetch the entire schema and process it here rather than introducing a new method.
2. Introduce a new method to the schema service to process and return attributes as `[]AttributeInfo`.
    - Maybe we can introduce a boolean credential field to the `AttributeInfo` struct and take credential attributes filtering as a method parameter
    - GetAttributes(ctx context.Context, category TypeCategory, entityType string, requiredOnly bool, credentialsOnly bool) ([]AttributeInfo, *serviceerror.ServiceError)

In a way it's better to handover this processing responsibility to the schema service. Hence I prefer Option 2

backend/internal/entitytype/service.go
ThaminduDilshan: Since you're anyway going to change the existing method, we can replace the usage of this `GetCredentialAttributes()` method with the new method that we're introducing here. Just have to pass credentials only flag to true

https://github.com/asgardeo/thunder/pull/2553#discussion_r3206050838

backend/internal/flow/executor/provisioning_executor.go
ThaminduDilshan: So logic of `HasRequiredInputs()` method would be;
1. Fetch the entire schema
2. Get already prompted optional attributes from runtime
3. Rather than doing two passes in `buildNonCredentialMissingInputs` and `buildCredentialMissingInputs`, find the ncRequiredMissing, ncOptionalMissing, credRequiredMissing and credOptionalMissing attributes in a single pass. No need to have two such methods
4. Build full schema missing list. Only append ncOptionalMissing and credOptionalMissing if prompt optional enabled
5. Apply max per prompt filtering
6. Record prompted optional attributes

WDYT? This optimizes the current provisioning executor implementation as well

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -276,101 +281,163 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
+		execResp.ForwardedData = make(map[string]interface{})
+	}
+	execResp.ForwardedData[common.ForwardedDataKeyInputs] = toForward
+	logger.Debug("Schema attributes are missing, requesting via prompt",
+		log.Int("missingCount", len(allSchemaMissing)))
+	return false
+}
 
-	// Build required and optional missing lists separately so required ones are always shown first.
+func (p *provisioningExecutor) buildNonCredentialMissingInputs(
```
ThaminduDilshan: Add missing method comments

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -543,24 +631,7 @@ func (p *provisioningExecutor) appendCredentialAttributes(ctx *core.NodeContext,
...
-
-	attrsToPopulate := schemaCredAttrs
-	if len(nodeCredentialInputs) > 0 {
-		attrsToPopulate = nodeCredentialInputs
-	}
-
-	for _, attr := range attrsToPopulate {
 		if value, exists := ctx.UserInputs[attr]; exists {
 			(*attributes)[attr] = value
 		} else if runtimeValue, exists := ctx.RuntimeData[attr]; exists {
```
ThaminduDilshan: Aligning with my previous comment, we can improve the `getAttributesForProvisioning()` method to return two maps; identifying and credential attributes separately. Then the caller can try to identify user with the identifying attributes and append two maps at the end.
This will remove another iteration over the schema attributes

backend/internal/flow/core/prompt_node.go
```diff
@@ -529,44 +535,44 @@ func filterMetaComponents(comps []interface{}, allowedRefs, knownInputActionRefs
...
-
-	// Build a set of refs/ids from the node's own configured prompt inputs —
-	// used to suppress synthesis for node-defined inputs with no meta component.
-	nodeInputRefs := make(map[string]struct{})
-	for _, inp := range n.getAllInputs() {
-		if inp.Ref != "" {
-			nodeInputRefs[inp.Ref] = struct{}{}
-		}
-		nodeInputRefs[inp.Identifier] = struct{}{}
+func cloneBlockWithChildren(compMap map[string]interface{}, children []interface{}) map[string]interface{} {
```
ThaminduDilshan, line 663: Append missing method comments. Check other places too

backend/internal/entitytype/service.go
Dilusha-Madushan: @coderabbitai check

backend/internal/entitytype/model/schema.go
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: @coderabbitai check

backend/internal/flow/executor/provisioning_executor.go

backend/internal/flow/executor/provisioning_executor.go

backend/internal/flow/executor/provisioning_executor.go

backend/internal/flow/core/prompt_node.go
Dilusha-Madushan: @coderabbitai check

backend/internal/entitytype/service.go
Dilusha-Madushan: @coderabbitai I addressed the refactoring, check the current status

backend/internal/entitytype/model/schema.go
Dilusha-Madushan: @coderabbitai check is this still exist or already fixed?

backend/internal/entitytype/service.go
```diff
@@ -74,18 +74,16 @@ type EntityTypeServiceInterface interface {
 		attributes json.RawMessage,
 		exists func(map[string]interface{}) (bool, error),
 	) (bool, *serviceerror.ServiceError)
-	GetCredentialAttributes(
+	GetAttributeInfos(
```
ThaminduDilshan:
```suggestion
	GetAttributes(
```

backend/internal/entitytype/model/schema.go
```diff
@@ -113,21 +113,27 @@ func (cs *Schema) ValidateAsDisplayAttribute(name string) DisplayAttributeStatus
...
+	Credential  bool
 }
 
-// GetNonCredentialAttributes returns top-level properties not marked as credentials.
-// When requiredOnly is true, only required properties are included.
-func (cs *Schema) GetNonCredentialAttributes(requiredOnly bool) []AttributeInfo {
+// GetAttributeInfos returns top-level properties filtered by the provided flags.
+// allowCredential includes credential properties; allowNonCredential includes non-credential
+// properties. When requiredOnly is true, only required properties are included.
+func (cs *Schema) GetAttributeInfos(allowCredential, allowNonCredential, requiredOnly bool) []AttributeInfo {
```
ThaminduDilshan:
```suggestion
func (cs *Schema) GetAttributes(allowCredential, allowNonCredential, requiredOnly bool) []AttributeInfo {
```

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -276,102 +285,142 @@ func (p *provisioningExecutor) HasRequiredInputs(ctx *core.NodeContext,
...
 }
 
-// fetchSchemaAttributes retrieves non-credential attributes from the entity type service.
-// When promptOptionalAttributes is true it fetches all attributes; otherwise only required ones.
-func (p *provisioningExecutor) fetchSchemaAttributes(
-	ctx *core.NodeContext, logger *log.Logger,
+// fetchSchemaAttributeInfos retrieves schema attributes from the entity type service for the
+// current user type. allowCredential and allowNonCredential control which attribute classes
+// are returned.
+func (p *provisioningExecutor) fetchSchemaAttributeInfos(
```
ThaminduDilshan:
```suggestion
func (p *provisioningExecutor) fetchSchemaAttributes(
```

backend/internal/entitytype/service.go
Dilusha-Madushan: @coderabbitai check changes

backend/internal/entitytype/model/schema.go
Dilusha-Madushan: @coderabbitai check changes

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: @coderabbitai check changes

ThaminduDilshan (approved):

Issue #2481, opened 2026-04-28
Expose thunder system resource to support RFC 8707


sahandilshan:
### Current Limitation

Thunder already has the RFC 8707 support. But the thunder system resource server is unable to follow that due to,
- Thunder system resource server identifier is not an URI
- The `system` resource is a hardcoded value and if we change it there will be a breaking change and refactoring that needs to be done.

### Suggested Improvement

Make the thunder resource server identifier a URI value and make the system resource's `handle` param configurable.
Once we do above, we have to identify the places where the `system` scope is being used and do modification to resolve the thunder scope from the runtime

PR #2554, opened 2026-05-05
Treat system resource as a normal resource

sahandilshan:
### Purpose
$subject
Addressed the comments of https://github.com/asgardeo/thunder/pull/2487

### Related Issue(s)
 - https://github.com/asgardeo/thunder/issues/2481

With this PR, we will be removing the existing supported deployement.toml config that was specifically defined to add handle for system resource.

```
resource:
  system_resource_server:
    handle: "mgmt"
    identifier: "https://api.example.com"
```

#### 💥 Impact
If users have used the above config to create the system resource with initial setup script under different handle and identifier, with this change onward those config won't work. They have to manually update the setup script.

#### 🔄 Migration Guide
If the system resource is not change, no need to do anything. 
If the system scope is changed with the above configs, then user's have to add this below config to make sure that thunder do the system resource validation properly
```
server:
    system_permission_prefix: "<handle>"
```

## Summary by CodeRabbit

* **Documentation**
  * Reworded System Resource Server docs and added a setting to prefix system permission strings; added guidance for Console scope updates.

* **Refactor**
  * Unified permission initialization and fallback to consistently use the system root permission and fail fast if uninitialized.

* **Chores**
  * Simplified bootstrap/setup to use a fixed system permission identifier and standardized administrator role assignment.

* **Behavior**
  * Resource server handle is now immutable; attempts to change it return a clear immutable-handle error.

* **Tests**
  * Added tests for root-permission behavior and updated tests to cover handle immutability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/2554?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/cmd/server/bootstrap/01-default-resources.ps1
+++ b/backend/cmd/server/bootstrap/01-default-resources.ps1
@@ -52,16 +52,6 @@ $ErrorActionPreference = 'Stop'
 Log-Info "Creating default $PRODUCT_NAME resources..."
 Write-Host ""
 
-# System resource server configuration from environment variables.
-$SYSTEM_RS_HANDLE = if ($env:SYSTEM_RS_HANDLE) { $env:SYSTEM_RS_HANDLE } else { "" }
-$SYSTEM_RS_IDENTIFIER = if ($env:SYSTEM_RS_IDENTIFIER) { $env:SYSTEM_RS_IDENTIFIER } else { "https://localhost:8090/mcp" }
-
-# Derive the system permission root based on the configured handle.
-if ($SYSTEM_RS_HANDLE) {
-    $SYSTEM_PERMISSION = "${SYSTEM_RS_HANDLE}:system"
-} else {
-    $SYSTEM_PERMISSION = "system"
-}
 
 # ============================================================================
 # Create Default Organization Unit
@@ -350,8 +340,7 @@ if (-not $DEFAULT_OU_ID) {
 $resourceServerData = @{
     name = "System"
     description = "System resource server"
-    handle = $SYSTEM_RS_HANDLE
-    identifier = $SYSTEM_RS_IDENTIFIER
+    identifier = "system"
     ouId = $DEFAULT_OU_ID
 } | ConvertTo-Json -Depth 10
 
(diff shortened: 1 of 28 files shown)


backend/internal/system/security/permissions.go
```diff
@@ -290,6 +290,15 @@ func GetSystemPermissions() *SystemPermissions {
 	return sysPerms
 }
 
+// GetSystemRootPermission returns the root system permission string.
+// Returns UninitializedPermissionSentinel if InitSystemPermissions has not been called.
+func GetSystemRootPermission() string {
+	if sysPerms != nil {
+		return sysPerms.Root
+	}
+	return UninitializedPermissionSentinel
```
rajithacharith: Can't we use nil instead of UninitializedPermissionSentinel?

sahandilshan: We have to return either `UninitializedPermissionSentinel` or the system root permissions, right? What do you mean by return nil?

rajithacharith: Do we need to keep a string called `UninitializedPermissionSentinel`? Are we having any additional benefit of using it?

sahandilshan: We can remove this and return a nil value. Then the go will autmatically handle this with panic error. Having this will just failing this gracefully. That's the only advantage

backend/cmd/server/bootstrap/01-default-resources.ps1
```diff
@@ -350,8 +340,7 @@ if (-not $DEFAULT_OU_ID) {
 $resourceServerData = @{
     name = "System"
     description = "System resource server"
-    handle = $SYSTEM_RS_HANDLE
-    identifier = $SYSTEM_RS_IDENTIFIER
+    identifier = "system"
```
ThaminduDilshan, line 343: Why are we removing the configurable system scope back? Any reason behind this?

(Also better to keep the PR description updated as per the template. It will be helpful to external contributors when checking these changes at a later point)

backend/internal/system/security/permissions_test.go
```diff
@@ -255,6 +255,28 @@ func (s *SecurityContextTestSuite) TestResolveActionPermission_CoversAllMappedAc
 // InitSystemPermissions
 // ---------------------------------------------------------------------------
 
+// ---------------------------------------------------------------------------
+// GetSystemRootPermission
+// ---------------------------------------------------------------------------
+
+func TestGetSystemRootPermission_BeforeInit(t *testing.T) {
```
ThaminduDilshan: Follow test suite pattern

backend/internal/system/security/permissions.go
sahandilshan: @rajithacharith This has been fixed now. Removed the `UninitializedPermissionSentinel` and let this handle with fail closed behaviour

backend/cmd/server/bootstrap/01-default-resources.ps1
sahandilshan: `Why are we removing the configurable system scope back? Any reason behind this?`
As per offline discussion, we shouldn't treat the system resource as a special resource in thunder. So providing a seperate dedicated configurable is not ideal. If anyone want to change it, they have to manually change it from the setup script. That's what we wanted to do here

sahandilshan: > (Also better to keep the PR description updated as per the template. It will be helpful to external contributors when checking these changes at a later point)

Will update the description with better details

backend/internal/system/security/permissions_test.go
sahandilshan: Fixed

rajithacharith (approved):

PR #2687, opened 2026-05-12
Track presented optional inputs across prompt and executor flows

### LINKED ISSUE #2574: Improve optional input handling in the flow execution
### Current Limitation

There are some issues with how we're handling optional inputs in flow execution.

1. Currently for PROMPT nodes, optional inputs are prompted only if at least a single required input is missing from the node input list. `appendMissingInputs()` only returns true if at least one required input is missing [1]. 
    - Prompt node has a logic to auto select the action if there's only one action for the node or inputs are satisfied for a action. If a PROMPT node contains only optional inputs and the action can be auto selected, flow execution will complete without prompting for optional inputs at least once (Also reported in https://github.com/asgardeo/thunder/issues/2194)
    - If all required inputs are satisfied for a PROMPT node with inputs collected over a different node, flow execution will complete without prompting for optional inputs at least once.

    (It appears this is somewhat correctly handled for the task execution nodes/ executors [2].)

2. If a optional input is prompted in a flow and user skips providing a value, if the same input is configured again somewhere later in the flow, we're prompting it again.

[1] https://github.com/asgardeo/thunder/blob/c6feb4a8c7c0dd9ce53938eba86f8fa2c5ed2f1d/backend/internal/flow/core/prompt_node.go#L282
[2] https://github.com/asgardeo/thunder/blob/c6feb4a8c7c0dd9ce53938eba86f8fa2c5ed2f1d/backend/internal/flow/core/executor.go#L214

### Suggested Improvement

Should improve how we're handling optional inputs during the flow execution.

1. Even if all required inputs are satisfied for a node, we should prompt user to enter the optional inputs once
2. If a optional input is prompted once and user skips providing a value, we shouldn't prompt it again later in the flow
    - However if the same input is configured as a required input somewhere later in the flow, we should prompt and treat it as a required input at that time

Dilusha-Madushan:
### Purpose
This PR fixes flow execution behavior for optional inputs that were already shown to the user but left empty.

Previously, the backend only checked whether a value existed. Because of that:
- optional prompt-node inputs could be skipped entirely
- optional inputs that were already shown once could be re-prompted again later
- provisioning had its own executor-local tracking for optional schema inputs instead of using a shared flow-level mechanism

This change introduces shared runtime tracking for optional inputs that were already presented to the user, so optional inputs are prompted once and do not block or reappear again after being skipped.

### Approach
The implementation moves optional-input presentation tracking to the flow engine and makes prompt/executor input checks read from that shared runtime state.

Key changes:
- Added a new generic runtime key, `RuntimeKeyPresentedOptionalInputs`, to store optional input identifiers that were already shown to the user.
- Added shared helper logic in `core/utils.go` to parse, check, and merge the presented-optional-input identifier set.
- Updated the flow engine to record optional inputs into runtime data when a user-visible `INCOMPLETE + VIEW` response is returned.
- Updated `prompt_node` input resolution so optional prompt inputs:
  - block once when first missing
  - stop blocking after they have already been presented
- Updated shared executor input handling with the same presented-optional-input check.
- Updated `ProvisioningExecutor` to stop writing its own provisioning-specific optional-input tracking and instead use the shared runtime-key-based logic only for re-prompt suppression.

This keeps actual input values and prompt-state metadata separate:
- real values still determine whether an attribute is satisfied
- the runtime key only determines whether an optional input has already been shown once

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2574

### Related PRs
- https://github.com/thunder-id/thunderid/pull/2518

## Summary by CodeRabbit

* **New Features**
  * The system records which optional inputs have been presented to the user (including when left empty) to avoid re-prompting.

* **Behavior Changes**
  * Flow engine centrally tracks and updates presented optional inputs; provisioning no longer persists that tracking itself.
  * Missing-input prompting now skips optional inputs already recorded as presented; required inputs remain unaffected.

* **Tests**
  * Added and updated tests covering presented-optional tracking, prompt behavior, merge/update logic, and non-prompt responses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/2687?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/internal/flow/core/executor.go
+++ b/backend/internal/flow/core/executor.go
@@ -186,37 +186,15 @@ func (e *executor) GetExecutionPolicy(mode string) *ExecutionPolicy {
 	return nil
 }
 
-// appendMissingInputs appends the missing required inputs to the executor response.
-// Returns true if any required input is found missing, otherwise false.
+// appendMissingInputs appends the missing executor inputs to the response.
+// Returns true when execution should pause for user input.
 func (e *executor) appendMissingInputs(ctx *NodeContext, execResp *common.ExecutorResponse,
 	requiredInputs []common.Input) bool {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "Executor"),
 		log.String(log.LoggerKeyExecutorName, e.GetName()),
 		log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
 
-	requireData := false
-	for _, input := range requiredInputs {
-		if _, ok := ctx.UserInputs[input.Identifier]; !ok {
-			if _, ok := ctx.RuntimeData[input.Identifier]; ok {
-				logger.Debug("Input available in runtime data, skipping",
-					log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-				continue
-			}
-
-			if value, ok := ctx.ForwardedData[input.Identifier]; ok {
-				if _, isString := value.(string); isString {
-					logger.Debug("Input available in forwarded data, skipping",
-						log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-					continue
-				}
-			}
-
-			requireData = true
-			execResp.Inputs = append(execResp.Inputs, input)
-			logger.Debug("Input not available in the context",
-				log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-		}
-	}
-
-	return requireData
+	missing := collectMissingInputs(ctx, GetPresentedOptionalInputs(ctx.RuntimeData), requiredInputs, logger)
+	execResp.Inputs = append(execResp.Inputs, missing...)
+	return len(missing) > 0
 }
--- a/backend/internal/flow/core/prompt_node.go
+++ b/backend/internal/flow/core/prompt_node.go
@@ -379,37 +379,15 @@ func (n *promptNode) hasRequiredInputs(ctx *NodeContext, nodeResp *common.NodeRe
 	return !n.appendMissingInputs(ctx, nodeResp, n.getAllInputs())
 }
 
-// appendMissingInputs appends the missing required inputs to the node response.
-// Returns true if any required data is found missing, otherwise false.
+// appendMissingInputs appends the missing prompt inputs to the node response.
+// Returns true when the prompt should pause for user input.
 func (n *promptNode) appendMissingInputs(ctx *NodeContext, nodeResp *common.NodeResponse,
 	requiredInputs []common.Input) bool {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
 
-	requireInputs := false
-	for _, input := range requiredInputs {
-		if _, ok := ctx.UserInputs[input.Identifier]; !ok {
-			if _, ok := ctx.RuntimeData[input.Identifier]; ok {
-				logger.Debug("Input available in runtime data, skipping",
-					log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-				continue
-			}
-			if value, ok := ctx.ForwardedData[input.Identifier]; ok {
-				if _, isString := value.(string); isString {
-					logger.Debug("Input available in forwarded data, skipping",
-						log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-					continue
-				}
-			}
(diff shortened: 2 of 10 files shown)


backend/internal/flow/core/executor.go
```diff
@@ -186,8 +186,18 @@ func (e *executor) GetExecutionPolicy(mode string) *ExecutionPolicy {
 	return nil
 }
 
-// appendMissingInputs appends the missing required inputs to the executor response.
-// Returns true if any required input is found missing, otherwise false.
+// IsOptionalInputPrompted returns true when an optional input identifier has already been shown
+// to the user in a prior prompt step.
+func IsOptionalInputPrompted(ctx *NodeContext, identifier string) bool {
```
ThaminduDilshan: Shall we move this method to `flow/core/utils.go`? 
Also I think we can make the `HasPresentedOptionalInput()` function private with this

backend/internal/flow/core/prompt_node.go
```diff
@@ -412,6 +417,10 @@ func (n *promptNode) appendMissingInputs(ctx *NodeContext, nodeResp *common.Node
 	return requireInputs
 }
 
+func (n *promptNode) isOptionalInputPrompted(ctx *NodeContext, identifier string) bool {
```
ThaminduDilshan: This is not required right? We can directly call the `IsOptionalInputPrompted()` in `flow/core/utils.go`

backend/internal/flow/core/utils.go
```diff
@@ -78,3 +82,54 @@ func ResolvePlaceholder(ctx *NodeContext, value string) string {
...
+	if identifier == "" || len(runtimeData) == 0 {
+		return false
+	}
+
+	raw := runtimeData[common.RuntimeKeyPresentedOptionalInputs]
+	if raw == "" {
+		return false
+	}
+
+	_, ok := ParsePresentedOptionalInputIdentifiers(raw)[identifier]
```
ThaminduDilshan: Also rather than parsing presented optional inputs runtime data key for each individual input, shall we parse this once and pass it as a method parameter to the downstream functions?
Otherwise we'll be splitting and parsing the string data array for each input

backend/internal/flow/core/utils.go
```diff
@@ -78,3 +82,54 @@ func ResolvePlaceholder(ctx *NodeContext, value string) string {
...
+	for _, identifier := range identifiers {
+		if identifier != "" {
+			merged[identifier] = struct{}{}
+		}
+	}
+	if len(merged) == 0 {
+		return ""
+	}
+
+	sorted := make([]string, 0, len(merged))
```
ThaminduDilshan: Do we need to sort here? I feel like not needed. 
Is there a advantage we can gain by sorting?

backend/internal/flow/flowexec/engine.go
```diff
@@ -218,6 +219,35 @@ func (fe *flowEngine) Execute(ctx *EngineContext) (FlowStep, *serviceerror.Servi
 	return flowStep, nil
 }
 
+func (fe *flowEngine) trackPresentedOptionalInputs(ctx *EngineContext, nodeResp *common.NodeResponse) {
```
ThaminduDilshan, line 224: Shall we add a method comment?

backend/internal/flow/core/executor.go
Dilusha-Madushan: @coderabbitai check the updated code to confirm is this resolved now?

backend/internal/flow/core/prompt_node.go
Dilusha-Madushan: @coderabbitai check the updated code to confirm is this resolved now?

backend/internal/flow/core/utils.go
Dilusha-Madushan: @coderabbitai check the updated code to confirm is this resolved now?

backend/internal/flow/core/utils.go
Dilusha-Madushan: @coderabbitai check the updated code to confirm is this resolved now?

backend/internal/flow/flowexec/engine.go
Dilusha-Madushan: @coderabbitai check the updated code to confirm is this resolved now?

ThaminduDilshan (approved):

PR #2709, opened 2026-05-13
Add input validation for flow inputs

### LINKED ISSUE #2619: Support Input Validation for Flow Inputs
**Problem Statement**

The flow system currently validates only the presence of required inputs but never checks their content. Invalid data — such as malformed email addresses, passwords below minimum length, or values not matching a required pattern — progresses silently to executors, resulting in cryptic downstream failures or confusing error messages for end users. This problem affects both gate app (built-in UI) customers and API-only customers who build their own interfaces. A declarative, server-enforced validation layer is needed to catch invalid inputs at the prompt node before they reach executor logic.

**Goals**

1. Replace the existing unstructured flat validation properties (`"regex"`, `"maxLength"`) on component objects with a declarative **`validation` array** containing explicit, single-constraint rule objects.
2. Each validation rule must follow a **discriminated `{ "type", "value", "message" }` shape** — exactly one constraint per rule object — replacing the original multi-constraint flat-key approach (where `pattern`, `minLength`, and `maxLength` could all appear together in one rule with a single shared message).
3. Support the following Phase 1 rule types:
   - `"type": "regex"` — value is a regex pattern string
   - `"type": "minLength"` — value is a minimum length number
   - `"type": "maxLength"` — value is a maximum length number
4. Store validation rules in **two locations** within flow definitions:
   - `meta.components[].validation` — for UI metadata consumed by the gate app and SDK
   - `prompts[].inputs[].validation` — for server-side enforcement, also returned to API-only customers in the initial prompt response
5. Execute **server-side validation** in the prompt node, after the required-input presence check passes, and before advancing to the next node.
6. On validation failure, return `flowStatus: INCOMPLETE` with a **`data.fieldErrors` array**, where each entry identifies the failing field and the rule's message. Multiple failing rules on the same field produce separate `fieldErrors` entries.
7. Support both **i18n message keys** (e.g., `{{i18n(validation:email.invalid)}}`) and literal strings as `message` values; the server returns them as-is without resolution.
8. Compile and cache regex patterns at initialization using Go's RE2 engine (via the `regexp` package) to guarantee linear-time matching and prevent ReDoS attacks.
9. Clear invalid user inputs and re-present the prompt when validation fails, reusing the existing `INCOMPLETE` node response path.
10. Apply a **default i18n fallback key** (e.g., `validation.pattern.invalid`) when no `message` is specified in a rule.
11. Migrate existing flow definition JSON files (e.g., `auth_flow_basic.json`) from the current flat-property format to the new `validation` array format.

**Non-goals**

1. User schema attribute constraint inheritance into flow-level validation rules.
2. Attribute uniqueness validation at the flow layer (e.g., checking username uniqueness during registration — handled at the executor/user-type level).
3. Server-side i18n key resolution — messages are returned to the client as-is.

**Acceptance Criteria**

| # | Area | Acceptance Criterion | Pass Condition |
|---|------|----------------------|----------------|
| AC1 | **Rule Schema** | Phase 1 supports three discriminated rule types — `regex` (string value), `minLength` (numeric value), and `maxLength` (numeric value) — each as a single-constraint `{ "type", "value", "message" }` object | Rules of all three types parse and evaluate correctly; inputs are accepted or rejected as expected for each type |
| AC2 | **Dual Data Paths — UI** | Validation rules are returned in `meta.components[].validation` in the prompt response | Gate app and SDK receive the `validation` array under `meta.components` for optional client-side rendering |
| AC3 | **Dual Data Paths — API** | Validation rules are returned in `data.inputs[].validation` in the initial prompt response | An API-only client receives prompt inputs with their `validation` arrays in `data.inputs` upon initiating a flow |
| AC4 | **Flow Definition Migration** | Existing flow JSON files are migrated from flat `"regex"` / `"maxLength"` component properties to the new `validation` array format | `auth_flow_basic.json` and all other flow definitions use `validation` arrays; no bare `"regex"` or `"maxLength"` flat properties remain on component objects |
| AC5 | **Execution Order** | Server-side validation runs in the prompt node only after the required-input presence check passes | Submitting a missing required field returns the existing required-input `INCOMPLETE` response, not a rule validation error |
| AC6 | **Execution Completeness** | All rules across all submitted inputs are evaluated in a single pass | Submitting two invalid fields returns `fieldErrors` for both fields in one response; multiple failing rules on the same field each produce a separate `fieldErrors` entry |
| AC7 | **Error Response** | On validation failure, `flowStatus` is `INCOMPLETE` and `data.fieldErrors` is populated with `{ "identifier", "message" }` entries; `fieldErrors` is placed within `data`, not in a top-level `error` object | Response body is `{ "flowStatus": "INCOMPLETE", "data": { "fieldErrors": [ { "identifier": "...", "message": "..." } ] } }`; no top-level `error` field is present |
| AC8 | **Message Handling** | The rule's `message` is returned as-is (no server-side i18n resolution); if `message` is absent, a default i18n fallback key is used | A rule with `"message": "{{i18n(validation:email.invalid)}}"` returns that exact string; a rule with no `message` returns a predefined default key |
| AC9 | **Input Sanitization** | `SanitizeStringMap()` executes before validation rules are evaluated | XSS payloads are sanitized first; the sanitized value is what gets tested against each rule |
| AC10 | **Reprompt Behavior** | On validation failure, user inputs are cleared and the prompt is re-presented alongside `data.fieldErrors` | Post-failure response includes the original prompt inputs (empty) and `data.fieldErrors`; the user can resubmit corrected values |
| AC11 | **Security — Regex Safety** | Regex patterns are compiled once at flow initialization (cached) using Go's RE2 engine (`regexp` package) | No per-request regex compilation occurs; complex patterns do not cause CPU spikes or timeouts |
| AC12 | **Happy Path** | A submission where all inputs pass validation advances the flow to the next node with no `fieldErrors` | Valid input values cause the flow to transition to the next node; `data.fieldErrors` is absent from the response |

Discussion Link: https://github.com/asgardeo/thunder/discussions/2410

SajidMannikeri17:
### Purpose
 
 The flow system currently validates only the **presence** of required inputs — it never validates their **content**. Invalid data (e.g. a malformed email, a password shorter than the minimum length, a phone number that does not match the required pattern) passes silently through to executors, producing cryptic downstream failures or confusing error messages for end users.
 
 This PR introduces declarative input validation at the prompt-node level:
 
 - Adds a `validation` array on flow input definitions with three Phase 1 rule types: `regex`, `minLength`, `maxLength`.
 - The server enforces these rules after the existing required-input presence check, and returns structured per-field errors (`data.fieldErrors`) on failure.
 - On validation failure the **entire prompt** is re-returned (all inputs + actions, full meta) so clients can re-render the same form the user just submitted with errors attached.
 
 Backward compatible: every new field (`validation`, `fieldErrors`) uses `omitempty`, so flows that don't define validation rules behave exactly as before.
 
 ### Approach
 
 Key design decisions, mirroring the agreed design in [#2410](https://github.com/thunder-id/thunder-id/discussions/2410):
 
 1. **Discriminated `{ type, value, message }` rule shape** — exactly one constraint per rule object. Each failing rule produces its own `fieldErrors` entry. Replaces the original multi-constraint flat-key proposal — unambiguous, extensible to future rule types without struct changes.
 
 2. **Validation runs in the prompt node, after the required-input check.** Missing required fields still take the existing "tell me what's missing" path; validation only fires once all required fields are present, so the two layers don't conflict.
 
 3. **`INCOMPLETE` status pattern is reused.** Validation failures use the existing `INCOMPLETE` + `VIEW` response — `failureReason` stays empty (`fieldErrors` is user feedback, not a system error), `challengeToken` rotates as on any other re-prompt. No new flow status.
 
 4. **`fieldErrors` lives in `data`, not in a top-level `error` object.** It represents user feedback on input, not a server-side failure. Multiple failing rules on the same field produce multiple `fieldErrors` entries, preserving the order declared on the input.
 
 5. **Regex safety.** Patterns are compiled via Go's `regexp` (RE2 engine — linear-time matching, no ReDoS) and cached so repeat requests don't recompile. Invalid patterns are not pre-validated at flow-creation time today; a request hitting an unparseable pattern is treated as a rule failure (defensive — prefers safe-by-default over silent pass).
 
 6. **Dual data paths** for the rule definitions:
    - `meta.components[].validation` — UI/SDK consumption (the Asgardeo JS SDK reads this for optional client-side validation).
    - `prompts[].inputs[].validation` — server-side enforcement and API-only customers.
    The flow author/tooling is responsible for keeping the two in sync; the engine never enriches one from the other.
 
 #### Testing
 
 - **18 unit tests** in `backend/internal/flow/core/input_validation_test.go` covering: each rule type pass/fail, multi-rule per field, default-message fallback, i18n key passthrough, absent input no-op, unknown rule type ignored, regex cache reuse, prompt-node integration on failure, validation-doesn't-run-when-required-missing, and the new full-prompt re-prompt on failure.
 - **6 integration tests** in `tests/integration/flow/authentication/input_validation_test.go` hitting the live `/flow/execute` endpoint and asserting: validation rules present in initial prompt response, fieldErrors populated on invalid submission, multiple rules per field, full prompt re-returned on failure, valid submission advances the flow, and validation runs only after the required-input check.
 - All existing tests across `internal/...` pass.
 
 ### Related Issues
 
 Refs [#2619](https://github.com/asgardeo/thunder/issues/2619).
 
 ### Related PRs
 - N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Per-input validation rules (regex, min/max length) with default fallback messages; field-level errors returned and included in step/view payloads; failing inputs cleared on error
* **Behavior Changes**
  * Validation runs for inputs of the selected action (or all inputs if none); length measured by Unicode code points; unknown rules ignored
* **Tests**
  * Extensive unit and integration tests covering rule handling, error aggregation, prompt re-rendering, and regex caching
* **Documentation**
  * Flow reference and API examples updated to document validation and response schema changes

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/2709)

### DIFF
--- a/.vale/styles/config/vocabularies/vocab/accept.txt
+++ b/.vale/styles/config/vocabularies/vocab/accept.txt
@@ -161,3 +161,4 @@ Twilio
 [Rr]untimedb
 [Uu]serdb
 [Cc]onsentdb
+[Mm]ultibyte
--- a/backend/internal/flow/common/constants.go
+++ b/backend/internal/flow/common/constants.go
@@ -316,3 +316,24 @@ const (
 	// ForwardedDataKeyTemplateData holds template parameters for notification executors
 	ForwardedDataKeyTemplateData = "templateData"
 )
+
+// ValidationType identifies the constraint type of a ValidationRule.
+type ValidationType string
+
+// Validation rule types.
+const (
+	// ValidationTypeRegex matches the submitted value against a regex pattern.
+	ValidationTypeRegex ValidationType = "regex"
+	// ValidationTypeMinLength enforces a minimum string length on the submitted value.
+	ValidationTypeMinLength ValidationType = "minLength"
+	// ValidationTypeMaxLength enforces a maximum string length on the submitted value.
+	ValidationTypeMaxLength ValidationType = "maxLength"
+)
+
+// Default i18n fallback message keys returned in fieldErrors when a validation
+// rule does not specify a message.
+const (
+	DefaultValidationMessageRegex     = "validation.pattern.invalid"
+	DefaultValidationMessageMinLength = "validation.minLength.invalid"
+	DefaultValidationMessageMaxLength = "validation.maxLength.invalid"
+)
--- a/backend/internal/flow/core/input_validation.go
+++ b/backend/internal/flow/core/input_validation.go
@@ -1,0 +1,95 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package core
+
+import (
+	"unicode/utf8"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/utils"
+)
+
+// validateInputValues returns a FieldError per failing rule across the given inputs.
+func validateInputValues(inputs []common.Input, userInputs map[string]string) []common.FieldError {
+	var fieldErrors []common.FieldError
+
+	for _, input := range inputs {
+		if len(input.Validation) == 0 {
+			continue
+		}
+		value, ok := userInputs[input.Identifier]
+		if !ok {
+			continue
+		}
+
+		for _, rule := range input.Validation {
+			if validateInput(rule, value) {
+				continue
+			}
+			fieldErrors = append(fieldErrors, common.FieldError{
+				Identifier: input.Identifier,
+				Message:    resolveRuleMessage(rule),
+			})
+		}
+	}
+
+	return fieldErrors
+}
+
+// validateInput returns false when the value violates the rule. Unknown rule
+// types and regex rules without a CompiledRegex pass through.
+func validateInput(rule common.ValidationRule, value string) bool {
+	switch rule.Type {
+	case common.ValidationTypeRegex:
+		if rule.CompiledRegex == nil {
+			return true
+		}
+		return rule.CompiledRegex.MatchString(value)
+
+	case common.ValidationTypeMinLength:
+		minLen, ok := utils.ToFloat64(rule.Value)
+		if !ok {
+			return true
+		}
+		return utf8.RuneCountInString(value) >= int(minLen)
+
+	case common.ValidationTypeMaxLength:
+		maxLen, ok := utils.ToFloat64(rule.Value)
+		if !ok {
+			return true
+		}
+		return utf8.RuneCountInString(value) <= int(maxLen)
+	}
+	return true
+}
+
+// resolveRuleMessage returns the rule's Message or the default key for its type.
+func resolveRuleMessage(rule common.ValidationRule) string {
+	if rule.Message != "" {
+		return rule.Message
+	}
+	switch rule.Type {
+	case common.ValidationTypeMinLength:
+		return common.DefaultValidationMessageMinLength
+	case common.ValidationTypeMaxLength:
+		return common.DefaultValidationMessageMaxLength
+	default:
+		return common.DefaultValidationMessageRegex
+	}
+}
--- a/backend/internal/flow/core/input_validation_test.go
+++ b/backend/internal/flow/core/input_validation_test.go
@@ -1,0 +1,832 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package core
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+)
+
+type InputValidationTestSuite struct {
+	suite.Suite
+}
+
+func TestInputValidationTestSuite(t *testing.T) {
+	suite.Run(t, new(InputValidationTestSuite))
+}
+
+// preparedRules pre-compiles regex patterns so tests bypassing the graph builder
+// still exercise the regex code path.
+func preparedRules(rules []common.ValidationRule) []common.ValidationRule {
+	if err := common.PrepareValidationRules(rules); err != nil {
+		panic(err)
+	}
+	return rules
+}
+
+func (s *InputValidationTestSuite) TestRegexPass() {
+	inputs := []common.Input{
+		{
+			Identifier: "email",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^[^@]+@[^@]+\\.[^@]+$",
+					Message: "validation.email.invalid"},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"email": "user@example.com"})
+	s.Empty(errs)
+}
+
+func (s *InputValidationTestSuite) TestRegexFail() {
+	inputs := []common.Input{
+		{
+			Identifier: "email",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^[^@]+@[^@]+\\.[^@]+$",
+					Message: "validation.email.invalid"},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"email": "not-an-email"})
+	s.Len(errs, 1)
+	s.Equal("email", errs[0].Identifier)
+	s.Equal("validation.email.invalid", errs[0].Message)
+}
+
+func (s *InputValidationTestSuite) TestMinLengthPass() {
+	inputs := []common.Input{
+		{
+			Identifier: "password",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeMinLength, Value: float64(8),
+					Message: "validation.password.minLength"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"password": "12345678"})
+	s.Empty(errs)
+}
+
+func (s *InputValidationTestSuite) TestMinLengthFail() {
+	inputs := []common.Input{
+		{
+			Identifier: "password",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeMinLength, Value: float64(8),
+					Message: "validation.password.minLength"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"password": "abc"})
+	s.Len(errs, 1)
+	s.Equal("password", errs[0].Identifier)
+	s.Equal("validation.password.minLength", errs[0].Message)
+}
+
+func (s *InputValidationTestSuite) TestMaxLengthPass() {
+	inputs := []common.Input{
+		{
+			Identifier: "username",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeMaxLength, Value: float64(10),
+					Message: "validation.username.maxLength"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"username": "ten_chars_"})
+	s.Empty(errs)
+}
+
+func (s *InputValidationTestSuite) TestMaxLengthFail() {
+	inputs := []common.Input{
+		{
+			Identifier: "username",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeMaxLength, Value: float64(5),
+					Message: "validation.username.maxLength"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"username": "way_too_long"})
+	s.Len(errs, 1)
+	s.Equal("validation.username.maxLength", errs[0].Message)
+}
+
+func (s *InputValidationTestSuite) TestMultipleRulesPerFieldProduceSeparateEntries() {
+	// Perl-style lookahead patterns are not supported by the regex engine; use alternation instead.
+	inputs := []common.Input{
+		{
+			Identifier: "password",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeMinLength, Value: float64(8),
+					Message: "validation.password.minLength"},
+				{Type: common.ValidationTypeRegex,
+					Value:   "^(?:[^A-Z]*[A-Z][^0-9]*[0-9].*|[^0-9]*[0-9][^A-Z]*[A-Z].*)$",
+					Message: "validation.password.complexity"},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"password": "abc"})
+	s.Len(errs, 2)
+	s.Equal("password", errs[0].Identifier)
+	s.Equal("validation.password.minLength", errs[0].Message)
+	s.Equal("password", errs[1].Identifier)
+	s.Equal("validation.password.complexity", errs[1].Message)
+}
+
+func (s *InputValidationTestSuite) TestMultipleFieldsErrorsInSinglePass() {
+	inputs := []common.Input{
+		{
+			Identifier: "username",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^[^@]+@[^@]+$",
+					Message: "validation.email.invalid"},
+			}),
+		},
+		{
+			Identifier: "password",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeMinLength, Value: float64(8),
+					Message: "validation.password.minLength"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{
+		"username": "no-at-sign",
+		"password": "abc",
+	})
+	s.Len(errs, 2)
+}
+
+func (s *InputValidationTestSuite) TestDefaultMessageUsedWhenAbsent() {
+	inputs := []common.Input{
+		{
+			Identifier: "email",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^X+$"},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"email": "abc"})
+	s.Len(errs, 1)
+	s.Equal(common.DefaultValidationMessageRegex, errs[0].Message)
+}
+
+func (s *InputValidationTestSuite) TestMessagePassthroughI18nKey() {
+	i18nKey := "{{i18n(validation:email.invalid)}}"
+	inputs := []common.Input{
+		{
+			Identifier: "email",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^X+$", Message: i18nKey},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"email": "abc"})
+	s.Len(errs, 1)
+	s.Equal(i18nKey, errs[0].Message)
+}
+
+func (s *InputValidationTestSuite) TestAbsentInputSkipsValidation() {
+	inputs := []common.Input{
+		{
+			Identifier: "email",
+			Validation: preparedRules([]common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "^X+$",
+					Message: "validation.email.invalid"},
+			}),
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{})
+	s.Empty(errs)
+}
+
+func (s *InputValidationTestSuite) TestUnknownRuleTypeIgnored() {
+	inputs := []common.Input{
+		{
+			Identifier: "field",
+			Validation: []common.ValidationRule{
+				{Type: "unknownRuleType", Value: "anything", Message: "should.not.appear"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"field": "any value"})
+	s.Empty(errs)
+}
+
+// An empty regex pattern has no effective constraint, so the rule passes any submitted value.
+func (s *InputValidationTestSuite) TestRegexEmptyPatternPasses() {
+	inputs := []common.Input{
+		{
+			Identifier: "field",
+			Validation: []common.ValidationRule{
+				{Type: common.ValidationTypeRegex, Value: "", Message: "should.not.appear"},
+			},
+		},
+	}
+	errs := validateInputValues(inputs, map[string]string{"field": "anything"})
+	s.Empty(errs)
+}
+
+// A non-numeric value (malformed flow definition) is treated as passing rather than failing every submission.
+func (s *InputValidationTestSuite) TestMinLengthNonNumericValuePasses() {
+	inputs := []common.Input{
+		{
+			Identifier: "field",
+			Validation: []common.ValidationRule{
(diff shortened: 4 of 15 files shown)


.vale/styles/config/vocabularies/vocab/accept.txt
```diff
@@ -69,4 +69,6 @@ BCP
 [Aa]llowlist
 Memorystore
 [Uu]pdatedAt
-[Cc]reatedAt
\ No newline at end of file
+[Cc]reatedAt
+[Ll]ookaround[s]?
```
ThaminduDilshan: Lookaround is not a valid word right? Shall we remove this config and fix the usages?

backend/internal/flow/common/constants.go
```diff
@@ -316,3 +316,21 @@ const (
 	// ForwardedDataKeyTemplateData holds template parameters for notification executors
 	ForwardedDataKeyTemplateData = "templateData"
 )
+
+// Validation rule types.
+const (
+	// ValidationTypeRegex matches the submitted value against a regex pattern.
+	ValidationTypeRegex = "regex"
```
ThaminduDilshan: Shall we define a go type for this? 

```go
type ValidationType string
```

backend/internal/flow/core/input_validation.go
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -280,11 +280,29 @@ func (b *graphBuilder) configureNodeInputs(nodeDef *NodeDefinition, node core.No
...
 			Required:   input.Required,
+			Validation: toValidationRules(input.Validation),
 		}
 	}
 	executorNode.SetInputs(inputs)
 }
 
+// toValidationRules converts the mgt-layer rule definitions to runtime ValidationRule values.
+// Returns nil for an empty input so the resulting Input retains a nil Validation slice.
+func toValidationRules(defs []ValidationRuleDefinition) []common.ValidationRule {
```
ThaminduDilshan: Shall we improve graph builder unit tests to cover new validation fields as well?

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -280,11 +280,29 @@ func (b *graphBuilder) configureNodeInputs(nodeDef *NodeDefinition, node core.No
 			Identifier: input.Identifier,
 			Type:       input.Type,
 			Required:   input.Required,
+			Validation: toValidationRules(input.Validation),
```
ThaminduDilshan: Shall we update API swagger definitions to include these changes?

- https://github.com/thunder-id/thunderid/blob/main/api/flow-management.yaml
- https://github.com/thunder-id/thunderid/blob/main/api/flow-execution.yaml

backend/internal/flow/core/input_validation.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+	}
+
+	return fieldErrors
+}
+
+// ruleFails returns true when the given rule's constraint is violated by value.
+// Unknown rule types are treated as a no-op (return false) — they are ignored
+// rather than failing the input, leaving room for forward-compatible additions.
+func ruleFails(rule common.ValidationRule, value string) bool {
```
ThaminduDilshan:
```suggestion
func validateInput(rule common.ValidationRule, value string) bool {
```

Also let's invert the return value. Return false for rule validation failure

backend/internal/flow/core/input_validation.go
```diff
@@ -0,0 +1,150 @@
...
+	case common.ValidationTypeMaxLength:
+		return common.DefaultValidationMessageMaxLength
+	default:
+		return common.DefaultValidationMessageRegex
+	}
+}
+
+// numericRuleValue extracts an int from a JSON-decoded numeric interface{}.
+// JSON numbers decode as float64 by default; ints are also accepted to ease testing.
+func numericRuleValue(v interface{}) (int, bool) {
```
ThaminduDilshan: Instead of this, shall we have a generic util in backend/internal/system/utils/value_util.go similar to `func ToFloat64(v interface{}) (float64, bool)`?
Make sure to add unit tests there

backend/internal/flow/core/input_validation.go
```diff
@@ -0,0 +1,150 @@
...
+import (
+	"regexp"
+	"sync"
+	"unicode/utf8"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+)
+
+var (
+	regexCache   = make(map[string]*regexp.Regexp)
```
ThaminduDilshan: Do we need a separate server wide shared object/ cache to keep these compiled regexes? These validation rules are defined in inputs in a flow definition right? The flow definition itself is compiled to a runtime graph and the graph gets cached in `graphCache`.

Given that, I don't think we need a server-wide regex cache at all. We can just compile the pattern at graph-build time (store it on the rule, or alongside it) and reuse it from ruleFails or the renamed method.

```go
type ValidationRule struct {
	Type    string      `json:"type"`
	Value   interface{} `json:"value"`
    CompiledRegex *regexp.Regexp `json:"-"`
	Message string      `json:"message,omitempty"`
}
```

Do you see any issues with this suggestion? If this doesn't fit, we should at least use the `system/cache` implementation for internal caching

backend/internal/flow/core/input_validation_test.go
```diff
@@ -0,0 +1,670 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/flow/mgt/graph_builder.go
ThaminduDilshan: Here we're applying validation rule configurations to executor backed nodes as well which I believe is correct. ThunderID flow execution engine can support backend only execution (non-verbose) without using any prompt nodes. But we're validating inputs only in the prompt nodes. So this config is not used.

We need to decide whether we're going to allow defining input validations for task execution nodes as well. If so we need to integrate validation logic into task execution nodes/ executors as well.
Or else we need to take a conscious decision that we're not going to allow defining input validations for the task execution nodes (similar to meta. We don't allow defining meta for task execution nodes). If so we should remove this line

backend/internal/flow/core/prompt_node.go
```diff
@@ -126,7 +126,9 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 	}
 
 	if n.resolvePromptInputs(ctx, nodeResp) {
-		logger.Debug("All required inputs and action are available, returning complete status")
```
ThaminduDilshan, line 129: Shall we keep the debug log without the last part?

```go
logger.Debug("All required inputs and action are available")
```

backend/internal/flow/core/prompt_node.go
```diff
@@ -167,6 +169,33 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+// caller should return nodeResp; false when validation passed.
+func (n *promptNode) applyValidationFailureReprompt(ctx *NodeContext, nodeResp *common.NodeResponse) bool {
+	fieldErrors := validateInputValues(n.getInputsForCurrentAction(ctx.CurrentAction), ctx.UserInputs)
+	if len(fieldErrors) == 0 {
+		return false
+	}
+	n.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID)).
+		Debug("Input validation failed", log.Int("errorCount", len(fieldErrors)))
+	for _, fe := range fieldErrors {
+		delete(ctx.UserInputs, fe.Identifier)
```
ThaminduDilshan, line 190: This could be a value satisfied from runtime data as well. Hence just deleting user inputs may not be enough. Engine could still retrieve the same value from runtime data

backend/internal/flow/core/prompt_node.go
```diff
@@ -167,6 +169,33 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
...
+	}
+	ctx.CurrentAction = ""
+	nodeResp.Inputs = n.getAllInputs()
```
ThaminduDilshan: Some inputs defined in a prompt node could be satisfied with previous data collected and hence the initial prompt could be a stripped down set of inputs than what's defined in the flow definition node. With this logic, flow execution will prompt for all inputs upon validation errors despite the initial prompt only prompts for a stripped down set.

IMO this is not correct. We should either;
1. Prompt only those prompted during initial request
or
2. Prompt only those with a validation failure

backend/internal/flow/core/prompt_node.go
```diff
@@ -167,6 +169,33 @@ func (n *promptNode) Execute(ctx *NodeContext) (*common.NodeResponse, *serviceer
 	return nodeResp, nil
 }
 
+// applyValidationFailureReprompt validates the currently-selected action's inputs
+// and, on failure, mutates ctx and nodeResp to re-prompt the whole node with
+// FieldErrors so the client can re-render the same form. Returns true when the
+// caller should return nodeResp; false when validation passed.
+func (n *promptNode) applyValidationFailureReprompt(ctx *NodeContext, nodeResp *common.NodeResponse) bool {
```
ThaminduDilshan:
```suggestion
func (n *promptNode) applyValidationFailureRePrompt(ctx *NodeContext, nodeResp *common.NodeResponse) bool {
```

ThaminduDilshan: Shall we also add unit tests for this?

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
...
+	}
+	if ts.ouID != "" {
+		if err := testutils.DeleteOrganizationUnit(ts.ouID); err != nil {
+			ts.T().Logf("Failed to delete test organization unit during teardown: %v", err)
+		}
+	}
+}
+
+// TestInitialPromptReturnsValidationRules — AC3: validation rules are returned in
+// data.inputs[].validation in the initial prompt response for API-only clients.
```
ThaminduDilshan:
```suggestion
// TestInitialPromptReturnsValidationRules tests whether validation rules are returned in
// data.inputs[].validation in the initial prompt response for API-only clients.
```

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+	passwordInput := findInput(flowStep.Data.Inputs, "password")
+	ts.Require().NotNil(passwordInput, "password input should be present")
+	ts.Require().Len(passwordInput.Validation, 2, "password should have 2 validation rules")
+	ts.Equal("minLength", passwordInput.Validation[0].Type)
+	ts.Equal("regex", passwordInput.Validation[1].Type)
+}
+
+// TestInvalidInputReturnsFieldErrors — AC7: when validation fails the server returns
+// flowStatus=INCOMPLETE and data.fieldErrors with one entry per failed rule.
```
ThaminduDilshan:
```suggestion
// TestInvalidInputReturnsFieldErrors tests when validation fails, the server returns
// flowStatus=INCOMPLETE and data.fieldErrors with one entry per failed rule.
```

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
...
+	ts.NotNil(findInput(resp.Data.Inputs, "username"), "username should be in re-prompted inputs")
+	ts.NotNil(findInput(resp.Data.Inputs, "password"), "password should be in re-prompted inputs")
+	ts.Len(resp.Data.Inputs, 2, "all original inputs should be re-prompted")
+
+	// Action is also re-returned so the submit button can be re-rendered.
+	ts.NotEmpty(resp.Data.Actions, "actions should be re-prompted")
+}
+
+// TestMultipleRulesPerFieldProduceSeparateEntries — AC6: when two rules on the same
+// field both fail, each produces a separate fieldErrors entry.
```
ThaminduDilshan:
```suggestion
// TestMultipleRulesPerFieldProduceSeparateEntries tests when two rules fails on the same
// field, each produces a separate fieldErrors entry.
```

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
...
+
+	ts.Equal("INCOMPLETE", resp.FlowStatus)
+	passwordErrors := filterFieldErrors(resp.Data.FieldErrors, "password")
+	ts.Require().Len(passwordErrors, 2, "two failing rules should produce two entries")
+	ts.Equal(validationPasswordMinMsg, passwordErrors[0].Message)
+	ts.Equal(validationPasswordRegexMsg, passwordErrors[1].Message)
+}
+
+// TestValidSubmissionAdvancesFlow — AC12 happy path: valid inputs pass validation and
+// the flow advances; data.fieldErrors is not present in the response.
```
ThaminduDilshan:
```suggestion
// TestValidSubmissionAdvancesFlow tests valid inputs pass validation and
// the flow advances; data.fieldErrors is not present in the response.
```

tests/integration/flow/authentication/input_validation_test.go
```diff
@@ -0,0 +1,403 @@
...
+	ts.Require().NoError(err, "Failed to submit valid inputs")
+
+	ts.Equal("COMPLETE", resp.FlowStatus, "Expected COMPLETE for valid submission")
+	ts.NotEmpty(resp.Assertion, "JWT assertion should be issued on successful auth")
+	ts.Empty(resp.Data.FieldErrors, "fieldErrors should be absent when all rules pass")
+}
+
+// TestValidationRunsAfterRequiredCheck — AC5: validation runs only after the required-input
+// presence check passes. A submission missing a required field returns the existing
+// missing-input response, not a validation error.
```
ThaminduDilshan:
```suggestion
// TestValidationRunsAfterRequiredCheck tests validation runs only after the required-input
// presence check passes. A submission missing a required field returns the existing
// missing-input response, not a validation error.
```

docs/static/api/next/combined.yaml
```diff
@@ -3988,13 +3988,61 @@ paths:
                           identifier: username
                           type: TEXT_INPUT
                           required: true
+                          validation:
```
ThaminduDilshan: Is this file manually edited? Usual practice is to modify the api swagger definitions in `/api` directory and build the docs. It will auto generate this file

docs/static/api/next/combined.yaml
```diff
@@ -15531,6 +15594,62 @@ components:
...
+          type: string
+          description: Constraint type. Each rule object carries exactly one constraint.
+          enum:
+            - regex
+            - minLength
+            - maxLength
+          example: regex
+        value:
+          description: Constraint parameter. `regex` takes a pattern string compiled with
+            Go's RE2 engine. `minLength` and `maxLength` take a non-negative integer
```
ThaminduDilshan: IMO exposing go lang details is unnecessary here unless it has a specific way to define regex patterns. This is a internal implementation details that a outside user doesn't need to be aware of

docs/content/guides/guides/flows/flow-reference.mdx
```diff
@@ -91,6 +91,67 @@ Components are the individual UI elements inside a View node. Add them to a Blan
 | **Icon** | A small icon element. |
 | **Timer** | A countdown timer, typically used on OTP screens. |
 
+### Input Validation Rules
+
+Input components can declare a `validation` array that the server enforces when the user submits the form. The server evaluates these rules after the required-input presence check and before the View advances to the next node.
+
+The `validation` array attaches to two places in the flow definition:
+
+- `meta.components[].validation` - returned to the Login Gate and the Asgardeo JavaScript SDK so that clients can optionally pre-validate values before submission.
```
ThaminduDilshan:
```suggestion
- `meta.components[].validation` - returned in the verbose flow execution response embedded inside meta so that clients can optionally pre-validate values before submission.
```

docs/static/api/next/combined.yaml
SajidMannikeri17: Removed the Go's RE2 engine mention from the description. Now reads as a generic "regex pattern string" / "regex engine that guarantees linear-time matching."

tests/integration/flow/authentication/input_validation_test.go
SajidMannikeri17: Cleaned up all the instances.

backend/internal/flow/core/prompt_node.go
SajidMannikeri17: Added the direct unit tests for the helper

SajidMannikeri17: Renamed to applyValidationFailureRePrompt.

backend/internal/flow/core/prompt_node.go
SajidMannikeri17: Went with your Option 1. The re-prompt set is now rebuilt as: the action's inputs minus those pre-satisfied via RuntimeData or ForwardedData.the response prompts the exact same subset the user saw initially — never the full node-wide input list.

backend/internal/flow/core/prompt_node.go
SajidMannikeri17: The helper now deletes the failing identifier from all three sources, so the engine cannot re-use a stale value on the next iteration. Added a unit test (TestApplyValidationFailureRePrompt_ClearsRuntimeDataAndForwardedData) that pre-loads stale values in RuntimeData and ForwardedData and asserts both are cleared.

backend/internal/flow/core/prompt_node.go
SajidMannikeri17: Restored with the trimmed wording. The log now reads "All required inputs and actions are available".

backend/internal/flow/mgt/graph_builder.go
SajidMannikeri17: I went with the conscious decision to not allow validation on task execution nodes for now — same treatment as meta. Reasons: the original design ([#2410](https://github.com/thunder-id/thunderid/discussions/2410)) scoped validation to PROMPT nodes; extending to executors needs separate work (wiring validation into each executor, defining the failure response shape for non-verbose flows, tests). Filing as a follow-up if backend-only flow validation becomes a need.

backend/internal/flow/core/input_validation.go
SajidMannikeri17: Implemented the suggestion. Added CompiledRegex *regexp.Regexp (json:"-") to ValidationRule plus a common.PrepareValidationRules helper that compiles in place. graph_builder.toValidationRules calls it during graph build and propagates compile errors, so invalid patterns now fail the build rather than surfacing at request time. The package-level regexCache, its mutex, and getCompiledRegex are gone.

backend/internal/flow/core/input_validation.go
SajidMannikeri17: ToFloat64 already exists in value_util.go (with comprehensive tests in value_util_test.go). Dropped numericRuleValue and switched to utils.ToFloat64; no new util or tests needed.

backend/internal/flow/core/input_validation.go
SajidMannikeri17: Renamed to validateInput as suggested. The return value already returns false on validation failure — function now returns true when the rule passes, false when it fails.

backend/internal/flow/common/constants.go
SajidMannikeri17: Added the type alias and retyped the constants + ValidationRule.Type so the type is actually enforced. mgt.ValidationRuleDefinition.Type stays plain string (JSON deserialization boundary) with a cast at conversion time in toValidationRules.

.vale/styles/config/vocabularies/vocab/accept.txt
SajidMannikeri17: Removed the entry from accept.txt and reworded the docs sentence so the term is no longer used.

docs/content/guides/guides/flows/advanced-configurations.mdx
```diff
@@ -307,6 +307,67 @@ Components are the individual UI elements inside a View node. Add them to a Blan
 | **Timer** | A countdown timer, used on consent screen. |
 | **Dynamic Input Placeholder** | Marks where dynamically forwarded inputs (e.g. from the Provisioning executor's incomplete path) should be rendered inside a View. |
 
+### Input Validation Rules
```
ThaminduDilshan: This doesn't fit under "Building Blocks" section right? Shall we move input validations to a section below "Executor Reference" section in the same level (h2)?
And before "ACR Values in Flows"

SajidMannikeri17: Moved the section.

ThaminduDilshan (approved):

Issue #2750, opened 2026-05-15
Migrate NOTIFICATION_SENDER PROPERTIES from JSON array to JSON object


sahandilshan:
## Summary

Follow-up to #2623. The `NOTIFICATION_SENDER` table uses the same `[{name, value, isSecret}]` JSON array format for its `PROPERTIES` column. It should be migrated to the JSON object format for consistency with the IDP table migration.

## Current format
```json
[
  {"name": "smtp_host", "value": "smtp.gmail.com", "isSecret": false},
  {"name": "smtp_password", "value": "encrypted...", "isSecret": true}
]
```

## New format
```json
{
  "smtp_host": {"value": "smtp.gmail.com", "isSecret": false},
  "smtp_password": {"value": "encrypted...", "isSecret": true}
}
```

## Scope
- Update `NOTIFICATION_SENDER` serialization in `backend/internal/notification/store.go` to use the new JSON object format (shared functions from #2623).
- No index changes needed — there are no property-based queries on this table today.

## Context
- Parent issue: #2623

PR #2752, opened 2026-05-15
Migrate IDP PROPERTIES from JSON array to JSON object format

### LINKED ISSUE #2623: Migrate IDP PROPERTIES from JSON array to JSON object and add expression index
## Summary

The `IDP` table's `PROPERTIES` column currently stores properties as a JSON array of `[{name, value, isSecret}]` objects. This makes targeted indexing difficult — the GIN index added in PR #2546 (`idx_idp_properties`) indexes the entire array (including encrypted secrets), causing unnecessary write amplification and index bloat for what is effectively a single-key lookup.

This issue migrates the `PROPERTIES` column to a JSON object format and adds a targeted expression index for efficient property-based lookups.

## Current format (JSON array)
```json
[
  {"name": "client_id", "value": "abc123", "isSecret": false},
  {"name": "client_secret", "value": "encrypted...", "isSecret": true},
  {"name": "issuer", "value": "https://accounts.google.com", "isSecret": false}
]
```

## New format (JSON object with metadata)
```json
{
  "client_id": {"value": "abc123", "isSecret": false},
  "client_secret": {"value": "encrypted...", "isSecret": true},
  "issuer": {"value": "https://accounts.google.com", "isSecret": false}
}
```

### 1. DB Schema
- Remove the current GIN index: `idx_idp_properties`
- Add a targeted expression index:
  - **PostgreSQL:** `CREATE INDEX idx_idp_issuer ON "IDP" (DEPLOYMENT_ID, (PROPERTIES->'issuer'->>'value'));`
  - **SQLite:** `CREATE INDEX idx_idp_issuer ON IDP (DEPLOYMENT_ID, json_extract(PROPERTIES, '$.issuer.value'));`

### 2. Serialization (`cmodels` or `idp` package)
- Add new serialization/deserialization functions for the JSON object format (keyed by property name, each value is `{value, isSecret}`).
- Keep existing `SerializePropertiesToJSONArray` / `DeserializePropertiesFromJSON` for other consumers (e.g., NOTIFICATION_SENDER — see #2624).

### 3. Store layer
- **Remove** `GetIdentityProviderByIssuer` from the store interface.
- **Add** `GetIdentityProvidersByProperty(ctx, propertyKey string, propertyValue string) ([]IDPDTO, error)` — a generic method that takes a property key, queries using an expression on the JSONB object, and returns all matching IDPs.
- Update the Postgres/SQLite query to use `PROPERTIES->propertyKey->>'value' = $1` (parameterized).

### 4. Cache layer
- Replace the `idpByIssuerCache` (single IDP) with a `idpByPropertyCache` that caches `[]*IDPDTO` keyed by `propertyKey:propertyValue`.
- Update cache invalidation in create/update/delete operations.

### 5. Service layer
- **Remove** `GetIdentityProviderByIssuer` from the service interface.
- **Add** `GetIdentityProvidersByProperty(ctx, propertyKey string, propertyValue string) ([]IDPDTO, *serviceerror.ServiceError)`.

### 6. Callers
- Update `tokenValidator.verifyTokenSignatureByIssuer` in `backend/internal/oauth/oauth2/tokenservice/validator.go` to use the new generic method.

## Why not keep the array format with a targeted index?
The `[{name, value}]` array structure requires iterating array elements to find a specific property. PostgreSQL expression indexes require `IMMUTABLE` functions, but `jsonb_array_elements` is `STABLE` — making a targeted index on the array format would require lying about immutability. The object format allows clean `->>'key'` extraction which is natively immutable and indexable.

## Context
- Related PR: #2546
- Review comment: https://github.com/asgardeo/thunder/pull/2546#discussion_r3207056382
- Raised by @KaveeshaPiumini

sahandilshan:
### Purpose

Migrate the IDP table's `PROPERTIES` column from a JSON array of `[{name, value, isSecret}]` objects to a JSON object keyed by property name with `{value, isSecret}` metadata. This enables clean expression-based indexing and a generic property lookup method.

**Before (JSON array):**
```json
[{"name":"issuer","value":"https://accounts.google.com","isSecret":false}]
```

**After (JSON object):**
```json
{"issuer":{"value":"https://accounts.google.com","isSecret":false}}
```

---
- IDP `PROPERTIES` DB column format changed from JSON array to JSON object. **Existing data requires migration**.

#### 💥 Impact
- Existing IDP rows in the database need a one-time data migration (array → object format).

#### 🔄 Migration Guide
- Run the DB schema migration to replace the GIN index with the expression index.

---

### Approach

1. **New serialization** — Added `SerializePropertiesToJSONObject` / `DeserializePropertiesFromJSONObject` in `cmodels/property.go`. Existing array functions retained for NOTIFICATION_SENDER (separate migration in #2750).
2. **DB index** — Replaced the broad `GIN jsonb_path_ops` index with a targeted expression index on `PROPERTIES->'issuer'->>'value'` (Postgres) / `json_extract(PROPERTIES, '$.issuer.value')` (SQLite).
3. **Generic property lookup** — `GetIdentityProvidersByProperty(key, value)` works across all store layers (DB, file-based, composite, cache-backed). Returns `[]IDPDTO` to support multi-result lookups.
4. **Cache** — Replaced single-IDP `idpByIssuerCache` with list-based `idpByPropertyCache` keyed by `propertyKey:propertyValue`. Lazily populated on reads; invalidated per-property on writes.
5. **Caller update** — Token validator's `resolveExternalIssuer` updated to use the generic method.

### Related Issues
- Fixes #2623

### Related PRs
- #2546 (original token exchange PR that added the GIN index)
- #2750 (follow-up: NOTIFICATION_SENDER migration)

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Query identity providers by arbitrary property key/value pairs; returns multiple matches.

* **Refactor**
  * Replaced issuer-specific single-lookups with property-based multi-lookup across stores and service layers.

* **Performance**
  * Added expression indexes to speed property-based lookups.
  * Cache updated to support property-based queries and cache “not found” results.

* **Tests**
  * Replaced issuer-based tests with property-based tests across service, store, cache, composite, file, and token validation suites.

* **Data**
  * IDP properties now persist as an object-shaped JSON format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/2752?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/internal/idp/cache_backed_store.go
+++ b/backend/internal/idp/cache_backed_store.go
@@ -28,23 +28,30 @@ import (
 
 const cacheBackedIDPStoreLoggerComponentName = "CacheBackedIDPStore"
 
+// cacheablePropertyKeys lists the property keys whose lookup results are cached.
+// Adding a key here means every distinct value for that key gets a cache entry —
+// keep the list small to avoid unbounded cache growth.
+var cacheablePropertyKeys = map[string]bool{
+	PropIssuer: true,
+}
+
 // cacheBackedIDPStore wraps any idpStoreInterface with two in-memory caches.
 type cacheBackedIDPStore struct {
-	idpByIDCache     cache.CacheInterface[*IDPDTO]
-	idpByIssuerCache cache.CacheInterface[*IDPDTO]
-	inner            idpStoreInterface
+	idpByIDCache       cache.CacheInterface[*IDPDTO]
+	idpByPropertyCache cache.CacheInterface[[]IDPDTO]
+	inner              idpStoreInterface
 }
 
 // newCacheBackedIDPStore creates a new cache-backed IDP store wrapping the provided inner store.
 func newCacheBackedIDPStore(
 	idpByIDCache cache.CacheInterface[*IDPDTO],
-	idpByIssuerCache cache.CacheInterface[*IDPDTO],
+	idpByPropertyCache cache.CacheInterface[[]IDPDTO],
 	inner idpStoreInterface,
 ) idpStoreInterface {
 	return &cacheBackedIDPStore{
-		idpByIDCache:     idpByIDCache,
-		idpByIssuerCache: idpByIssuerCache,
-		inner:            inner,
+		idpByIDCache:       idpByIDCache,
+		idpByPropertyCache: idpByPropertyCache,
+		inner:              inner,
 	}
 }
 
@@ -82,7 +89,7 @@ func (s *cacheBackedIDPStore) GetIdentityProvider(ctx context.Context, idpID str
 	return idp, nil
 }
 
-// GetIdentityProviderByName delegates to inner store and populates the ID and issuer caches with the result.
+// GetIdentityProviderByName delegates to inner store and populates the ID cache with the result.
 func (s *cacheBackedIDPStore) GetIdentityProviderByName(ctx context.Context, idpName string) (*IDPDTO, error) {
 	idp, err := s.inner.GetIdentityProviderByName(ctx, idpName)
 	if err != nil || idp == nil {
@@ -92,33 +99,34 @@ func (s *cacheBackedIDPStore) GetIdentityProviderByName(ctx context.Context, idp
 	return idp, nil
 }
 
-// GetIdentityProviderByIssuer retrieves an IDP by its issuer property, using the issuer cache on a hit.
-// A nil cached value means the absence of an IDP for that issuer was previously recorded.
-func (s *cacheBackedIDPStore) GetIdentityProviderByIssuer(ctx context.Context, issuer string) (*IDPDTO, error) {
-	cacheKey := cache.CacheKey{Key: issuer}
-	if cached, ok := s.idpByIssuerCache.Get(ctx, cacheKey); ok {
+// GetIdentityProvidersByProperty retrieves IDPs by property, using the property cache on a hit.
+// Only property keys listed in cacheablePropertyKeys are cached; all others bypass the cache.
+func (s *cacheBackedIDPStore) GetIdentityProvidersByProperty(ctx context.Context,
+	propertyKey, propertyValue string) ([]IDPDTO, error) {
+	if !cacheablePropertyKeys[propertyKey] {
+		return s.inner.GetIdentityProvidersByProperty(ctx, propertyKey, propertyValue)
+	}
+
+	cacheKey := cache.CacheKey{Key: propertyKey + ":" + propertyValue}
+	if cached, ok := s.idpByPropertyCache.Get(ctx, cacheKey); ok {
 		if cached == nil {
 			return nil, ErrIDPNotFound
 		}
 		return cached, nil
 	}
 
-	idp, err := s.inner.GetIdentityProviderByIssuer(ctx, issuer)
+	idps, err := s.inner.GetIdentityProvidersByProperty(ctx, propertyKey, propertyValue)
 	if err != nil {
 		if errors.Is(err, ErrIDPNotFound) {
-			_ = s.idpByIssuerCache.Set(ctx, cacheKey, nil)
+			_ = s.idpByPropertyCache.Set(ctx, cacheKey, nil)
 		}
 		return nil, err
 	}
-	if idp == nil {
-		_ = s.idpByIssuerCache.Set(ctx, cacheKey, nil)
-		return nil, ErrIDPNotFound
-	}
-	s.cacheIDP(ctx, idp)
-	return idp, nil
+	_ = s.idpByPropertyCache.Set(ctx, cacheKey, idps)
+	return idps, nil
 }
 
-// UpdateIdentityProvider fetches the old IDP to capture its issuer, delegates the update, then
+// UpdateIdentityProvider fetches the old IDP to capture its properties, delegates the update, then
 // invalidates old cache entries and caches the new state.
 func (s *cacheBackedIDPStore) UpdateIdentityProvider(ctx context.Context, idp *IDPDTO) error {
 	oldIDP, err := s.inner.GetIdentityProvider(ctx, idp.ID)
@@ -126,21 +134,16 @@ func (s *cacheBackedIDPStore) UpdateIdentityProvider(ctx context.Context, idp *I
 		return err
 	}
 
-	var oldIssuer string
-	if oldIDP != nil {
-		oldIssuer = GetPropertyValue(oldIDP.Properties, PropIssuer)
-	}
-
 	if err := s.inner.UpdateIdentityProvider(ctx, idp); err != nil {
 		return err
 	}
 
-	s.invalidateIDP(ctx, idp.ID, oldIssuer)
+	s.invalidateIDP(ctx, oldIDP)
 	s.cacheIDP(ctx, idp)
 	return nil
 }
 
-// DeleteIdentityProvider fetches the IDP to get its issuer before delegating deletion,
+// DeleteIdentityProvider fetches the IDP to get its properties before delegating deletion,
 // then invalidates the relevant cache entries.
 func (s *cacheBackedIDPStore) DeleteIdentityProvider(ctx context.Context, id string) error {
 	existing, err := s.inner.GetIdentityProvider(ctx, id)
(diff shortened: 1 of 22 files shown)


ThaminduDilshan:
This is a breaking change right? Don't use the skip changelog label.

Also breaking change section in PR description should explain the impact for the product distribution, not the code level changes.

For this PR; DB schema and related data changes and API reauest/ response changes

backend/internal/idp/service.go
```diff
@@ -162,24 +163,27 @@ func (is *idpService) GetIdentityProviderByName(ctx context.Context,
 	return idp, nil
 }
 
-// GetIdentityProviderByIssuer retrieves an identity provider by its issuer property.
-func (is *idpService) GetIdentityProviderByIssuer(ctx context.Context,
-	issuer string) (*IDPDTO, *serviceerror.ServiceError) {
+// GetIdentityProvidersByProperty retrieves identity providers matching a given property key and value.
```
ThaminduDilshan, line 173: Is this only for the service method? Or are we changing the API responses as well?

If we're changing a API response, shouldn't we do same change to all IDP service methods? IMO all endpoints should follow a consistent format.

sahandilshan: We are not changing the API response with this. Just the service layer. API layer will be the same

sahandilshan:
> This is a breaking change right? Don't use the skip changelog label.
> 
> Also breaking change section in PR description should explain the impact for the product distribution, not the code level changes.
> 
> For this PR; DB schema and related data changes and API reauest/ response changes

Yes. I'll add proper description

backend/internal/idp/composite_store_test.go
```diff
@@ -362,3 +362,124 @@ func (s *CompositeIDPStoreTestSuite) TestMergeAndDeduplicateIDPs_PreservesDuplic
...
+	s.fileStore.On("GetIdentityProvidersByProperty", context.Background(), "issuer", "https://example.com").
+		Return([]IDPDTO(nil), errors.New("file read error"))
+
+	result, err := s.compositeStore.GetIdentityProvidersByProperty(
+		context.Background(), "issuer", "https://example.com")
+
+	s.Nil(result)
+	s.Error(err)
+	s.NotErrorIs(err, ErrIDPNotFound)
+}
```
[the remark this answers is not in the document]
sahandilshan: Doc changes are not needed as of now

backend/internal/idp/cache_backed_store.go
```diff
@@ -92,113 +92,95 @@ func (s *cacheBackedIDPStore) GetIdentityProviderByName(ctx context.Context, idp
...
 
-// GetIdentityProviderByIssuer retrieves an IDP by its issuer property, using the issuer cache on a hit.
-// A nil cached value means the absence of an IDP for that issuer was previously recorded.
-func (s *cacheBackedIDPStore) GetIdentityProviderByIssuer(ctx context.Context, issuer string) (*IDPDTO, error) {
-	cacheKey := cache.CacheKey{Key: issuer}
-	if cached, ok := s.idpByIssuerCache.Get(ctx, cacheKey); ok {
+// GetIdentityProvidersByProperty retrieves IDPs by property, using the property cache on a hit.
+func (s *cacheBackedIDPStore) GetIdentityProvidersByProperty(ctx context.Context,
+	propertyKey, propertyValue string) ([]IDPDTO, error) {
+	cacheKey := cache.CacheKey{Key: propertyKey + ":" + propertyValue}
```
ThaminduDilshan, line 110: It's not a good practice to include property values in cache keys right? This can be a sensitive value or credential.

Also won't this key conflict if two IDPs share the same properties?

backend/internal/system/cmodels/property.go
```diff
@@ -154,6 +154,62 @@ func DeserializePropertiesFromJSON(propertiesJSON string) ([]Property, error) {
 	return properties, nil
 }
 
+// SerializePropertiesToJSONObject serializes an array of properties to a JSON object string
+func SerializePropertiesToJSONObject(properties []Property) (string, error) {
```
ThaminduDilshan, line 158: Any reason to introduce duplicate functions? Can't we reuse/ refactor the existing `SerializePropertiesToJSONArray()` and `DeserializePropertiesFromJSON()` functions?

backend/internal/system/cmodels/property_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSerializePropertiesToJSONObject_EmptySlice(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/idp/service.go
```diff
@@ -169,24 +170,27 @@ func (is *idpService) GetIdentityProviderByName(ctx context.Context,
...
-	idp, err := is.idpStore.GetIdentityProviderByIssuer(ctx, issuer)
+	idps, err := is.idpStore.GetIdentityProvidersByProperty(ctx, propertyKey, propertyValue)
 	if err != nil {
 		if errors.Is(err, ErrIDPNotFound) {
 			return nil, &ErrorIDPNotFound
 		}
-		logger.Error("Failed to get identity provider by issuer", log.String("issuer", issuer), log.Error(err))
+		logger.Error("Failed to get identity providers by property",
+			log.String("propertyKey", propertyKey),
+			log.String("propertyValue", propertyValue),
```
ThaminduDilshan: We shouldn't log plain text property values. This maybe specific implementation for issuers, but the function is generic. This can be any IDP property right?

ThaminduDilshan: Check other places as well

backend/internal/idp/cache_backed_store_test.go
```diff
@@ -57,12 +57,39 @@ func (m *mockIDPCache) Delete(ctx context.Context, key cache.CacheKey) error {
 	return args.Error(0)
 }
 
+// mockIDPListCache is a test mock for cache.CacheInterface[[]IDPDTO].
+type mockIDPListCache struct {
```
ThaminduDilshan: Shall we use mockery generated mocks instead of writing custom mocks?

backend/internal/idp/cache_backed_store.go
sahandilshan: > It's not a good practice to include property values in cache keys right? This can be a sensitive value or credential.

Yes, if we use this to retrieve IDPs with secret values, then it's have this issue. But main purpose of this function is to retrieve set of IDPs or a IDP with a given property value. IMO we should mandate that we will not use this function to retrieve IDPs with secret properties

backend/internal/idp/service.go
sahandilshan: Yes. This might print secret values. Will change the log to just to log the key

backend/internal/system/cmodels/property.go
sahandilshan: The old serialize and deserialze resolves two different types json shapes. The old one converts something like this
Array: `[{"name":"k","value":"v","isSecret":false}]`

New function converts something like below,
Object: `{"k":{"value":"v","isSecret":true}}`

once we replaced the `NOTIFICATION_SENDER` table with this new object type, we can remove the old serialize and deserialize methods

backend/internal/system/cmodels/property_test.go
sahandilshan: Fixed

backend/internal/idp/cache_backed_store_test.go
sahandilshan: Fixed

backend/internal/system/cmodels/property.go
sahandilshan: It will be cover with https://github.com/thunder-id/thunderid/issues/2750

backend/internal/idp/cache_backed_store.go
ThaminduDilshan: How are we mandating that? When we have a generic function, someone can use this to retrieve such a property.

ThaminduDilshan: At least shall we define a list of allowed properties that can be retrieved with this method? Then caching will not happen for unintentional key-value pairs. And we should add a comment to the constant explaining the risk of adding new properties to the list

sahandilshan: This is fixed now. Introduced a cachable key set

ThaminduDilshan (approved):

PR #2783, opened 2026-05-18
Migrate frontend from @asgardeo/* to @thunderid/* SDKs

sadilchamishka:
### Purpose
The `@asgardeo/react` and `@asgardeo/react-router` packages have been moved into the monorepo as local workspace packages (`@thunderid/react` and `@thunderid/react-router` under `sdks/react` and `sdks/react-router`). This PR updates the entire frontend to reference the new local packages instead of the old external npm packages.

### Approach
- **`pnpm-workspace.yaml`** — replaced the external pinned catalog entries with `workspace:^` references so all consumers continue using `catalog:` shorthand while resolving to the local packages
- **9 `package.json` files** — renamed `@asgardeo/react`/`@asgardeo/react-router` in `dependencies` and `peerDependencies` across both apps and all shared packages
- **~208 `.ts`/`.tsx` source files** — bulk-renamed all import statements via `find`+`sed`
- **`vite.config.ts`** (console + gate) — updated `server.deps.inline` from `@asgardeo/browser` to `@thunderid/browser`

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
- [ ] Documentation provided.
- [ ] Tests provided.
- [ ] Breaking changes.

### Security checks

### DIFF
--- a/docs/plugins/shims/emotion-css.cjs
+++ b/docs/plugins/shims/emotion-css.cjs
@@ -1,0 +1,17 @@
+// SSR shim for @emotion/css — Emotion creates a DOM cache at module init
+// which fails in Node.js (document is not defined). This no-op shim is used
+// only during Docusaurus server-side rendering; the real package runs in the browser.
+function noop() { return ''; }
+function noopObj() { return {}; }
+
+module.exports = {
+  css: noop,
+  cx: noop,
+  injectGlobal: noop,
+  keyframes: noop,
+  hydrate: noop,
+  flush: noop,
+  merge: noop,
+  getRegisteredStyles: function() { return []; },
+  cache: { key: 'css', registered: {}, inserted: {}, sheet: { tags: [] } },
+};
--- a/docs/plugins/webpackPlugin.ts
+++ b/docs/plugins/webpackPlugin.ts
@@ -18,10 +18,12 @@
 
 /* eslint-disable @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any */
 
+import path from 'path';
+
 export default function () {
   return {
     name: 'product-docs-webpack-plugin',
-    configureWebpack(config: any) {
+    configureWebpack(config: any, isServer: boolean) {
       // url-loader@4.x is incompatible with webpack@5.100+, causing:
       // "TypeError: Cannot read properties of undefined (reading 'date')"
       // Mutate existing rules in-place to replace url-loader with webpack 5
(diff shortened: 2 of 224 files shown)


brionmario (dismissed):

sadilchamishka:
## Why this PR also touches `sdks/react/package.json` and `docs/`

This needs a bit of explanation because the changes look unrelated to a frontend package rename.

### The dependency chain

```
docs → @thunderid/design → @thunderid/react → @emotion/css
```

The docs build (`nx run @thunderid/docs:build`) follows nx's `^build` dependency rule, so it builds `@thunderid/design` first. `@thunderid/design` now imports from `@thunderid/react` (our rename), and its code ends up in Docusaurus's server-side rendering (SSR) bundle.

### Why the docs SSR was crashing

Docusaurus generates HTML by executing a server bundle in Node.js. When that bundle initialises, `@emotion/css` runs `document.createElement` immediately on import (to create its internal CSS stylesheet cache). In Node.js there is no `document`, so it throws `ReferenceError: document is not defined`.

### Why `@emotion/css` was inlined — the root cause

`sdks/react/package.json` had `@emotion/css`, `@floating-ui/react`, and `dompurify` in **`devDependencies`** instead of **`dependencies`**. Rolldown's external function only externalises packages listed in `dependencies` + `peerDependencies`:

```js
const external = (id) =>
  externalPackages.some((dep) => id === dep || id.startsWith(dep + '/'));
// externalPackages = Object.keys(dependencies) + Object.keys(peerDependencies)
```

Because `@emotion/css` was in `devDependencies`, rolldown treated it as internal and **inlined the entire `@emotion/cache` + `@emotion/sheet` source** directly into `sdks/react/dist/index.js`. By the time webpack processed the docs server bundle, there was no `@emotion/css` import to intercept — just raw inlined code calling `document.createElement`.

### Why all three changes are needed together

| File | What it does |
|---|---|
| `sdks/react/package.json` | Moves `@emotion/css`, `@floating-ui/react`, `dompurify` to `dependencies`. Rolldown now keeps them external. The rebuilt dist has `import { css } from "@emotion/css"` — a proper import that webpack can intercept. |
| `docs/plugins/webpackPlugin.ts` | For Docusaurus's server build (`isServer: true`), aliases `@emotion/css` to a no-op shim. Without this, even a properly external `@emotion/css` still crashes SSR because it calls `document.createElement` on import. |
| `docs/plugins/shims/emotion-css.cjs` | The no-op shim — returns empty strings/functions so Emotion's initialisation completes without touching `document`. Only used during SSR; the browser build gets the real package. |

The two fixes must work in tandem: the SDK fix makes the import interceptable; the webpack shim makes the interception safe for SSR.

This was a latent bug in the SDK — before this PR, `@thunderid/design` depended on `@asgardeo/react` (an external npm package that didn't use `@emotion/css`), so the inlining never affected the docs build.

brionmario (approved):

PR #2831, opened 2026-05-19
Trust self-issued tokens on secured APIs when a trusted issuer is configured

### LINKED ISSUE #2830: client_credentials tokens issued by Thunder are rejected by its own secured APIs when a trusted issuer is configured
### Description

When `server.security.trusted_issuer` is configured, Thunder's protected-API security middleware verifies incoming bearer tokens **exclusively** against the external trusted issuer's JWKS and refuses to validate tokens that Thunder itself signed.

As a result, in a federated deployment **any token Thunder issues for itself is rejected by Thunder's own secured APIs**, including:

- access tokens obtained via the `client_credentials` grant (machine-to-machine / service apps), and
- the internal system/admin tokens used for bootstrap and auto-provisioning.

This makes it impossible for an application using `client_credentials` to call the instance's own secured APIs once a trusted issuer is enabled. The token endpoint still issues these tokens and `/oauth2/introspect` still considers them valid, but the API gate rejects them — an internally inconsistent state.

Root cause: `backend/internal/system/security/jwt_authenticator.go` `Authenticate()` branches on `TrustedIssuer.IsConfigured()` with a mutually exclusive `if/else`. When a trusted issuer is configured, only `verifyFederatedToken()` runs; the local-key verification path is unreachable, and a self-issued token is rejected at the `iss != trustedIssuer.Issuer` comparison before signature/audience are ever checked.

The correct model: an API gateway should trust a set of issuers. Thunder must always trust tokens it signed itself; an external trusted issuer should be **additive**, not a replacement.

### Steps to Reproduce

1. Configure a trusted issuer in `deployment.yaml`:
   ```yaml
   server:
     security:
       trusted_issuer:
         issuer: "https://external-idp.example.com"
         jwks_url: "https://external-idp.example.com/oauth2/jwks"
         audience: "FEDERATED_CONSOLE"
   ```
2. Start Thunder and register an application with the `client_credentials` grant, authorized for a `system` permission.
3. Obtain a token: `POST /oauth2/token` with `grant_type=client_credentials`. The token is issued successfully with `iss` = Thunder's own issuer.
4. Call any secured API with that token, e.g. `GET /applications` with `Authorization: Bearer <token>`.
5. **Observed:** `401 Unauthorized` (`AUTH-4010`). The same happens for the admin/system token, so the Console and auto-provisioning are also broken under federation.
6. **Expected:** The token Thunder issued for its own client is accepted (subject only to normal scope/permission authorization); tokens from the external trusted issuer continue to be accepted; unknown issuers are still rejected.

### Version

`main` @ commit `1886efb02` (also present on current `main`).

- OS: macOS (reproducible on Linux; OS-independent — logic defect)
- Database: SQLite (default `deployment.yaml`); not DB-specific
- Go: backend toolchain per `go.mod`
- Affects: any deployment with `server.security.trusted_issuer` set (federated console / cloud).

RushanNanayakkara:
### Purpose

Fixes the limitation where **tokens Thunder issues for itself are rejected by Thunder's own secured APIs when `server.security.trusted_issuer` is configured**. In a federated deployment this broke every `client_credentials` (machine-to-machine) caller, as well as the admin/system tokens used by the Console and by bootstrap/auto-provisioning.

Root cause: `jwtAuthenticator.Authenticate()` branched on `TrustedIssuer.IsConfigured()` with a mutually exclusive `if/else`. With a trusted issuer set, only federated (JWKS) verification ran and the local-key path was unreachable, so a self-issued token failed at the `iss != trustedIssuer.Issuer` check before signature/audience were ever evaluated. The token endpoint still minted these tokens and `/oauth2/introspect` still accepted them — only the API gate rejected them.

### Approach

Verification is now **routed by the token's `iss` claim against an explicit allowlist of accepted issuers**:

- `iss` matches the configured trusted issuer (when set) → verify against its JWKS (existing `verifyFederatedToken`, unchanged).
- `iss` matches this server's own JWT issuer (`jwt.issuer`, which defaults to the server's public URL) → verify with the server's signing key.
- Any other / unreadable `iss` → reject (no verifier invoked).

A configured trusted issuer is **additive, not a replacement**: this server keeps accepting the tokens it issues for itself (e.g. `client_credentials`, system/admin) alongside tokens from the configured trusted issuer. There is **no cross-issuer fallback** — a token routed to one verifier is never retried with the other, so spoofing `iss` gains nothing (the signature must still verify against that issuer's key material). This restores consistency with `/oauth2/introspect` and the token service, which already validate self-issued tokens.

The design follows the OAuth/JWT best practice of issuer allowlisting:
- [RFC 8725 §3.1 (JWT BCP)](https://datatracker.ietf.org/doc/html/rfc8725#section-3.1): *"Always validate the issuer of a JWT and accept it only from trusted issuers."*
- [RFC 9068 §4 (JWT Access Tokens)](https://datatracker.ietf.org/doc/html/rfc9068#section-4): the resource server validates `iss` against a configured allowlist and uses it to select the verifier when multiple authorization servers are accepted.

Security note: this changes a previously intentional, unit-tested "exclusive federation" behavior. That contract (`TestAuthenticate_FederatedTokenFailure`) has been deliberately re-encoded to the new model — a token claiming the trusted issuer that fails JWKS is still rejected with **no fallback to local verification** (asserted explicitly), and a token whose `iss` is outside the allowlist is rejected before any verifier is invoked (`TestAuthenticate_UnknownIssuerUnderFederation`). The change only *adds* acceptance of tokens this server itself signed; no existing configuration changes behavior for external/unknown tokens, and no API or config surface changes.

Verified end-to-end on a real local instance with `trusted_issuer` configured (identical config + seeded data, only the code differing):

| Check (trusted_issuer ON) | Before | After |
| --- | --- | --- |
| self-issued admin token (authorization_code + PKCE, the Console's own flow) → `GET /applications` | 401 | 200 |
| `client_credentials` self-issued token → secured API | 401 | 200 |
| no token / unknown issuer / token spoofing the trusted issuer | 401 | 401 (unchanged) |

Verification was performed by driving Thunder's real OAuth2 endpoints (the same `authorization_code` + PKCE + native authentication flow the Console uses to obtain its token, and the `client_credentials` grant) against a running instance with `trusted_issuer` configured. The Console's user-facing breakage under federation is a direct consequence of that self-issued token being rejected; accepting it (200) is what restores the Console and auto-provisioning.

### Related Issues

- Fixes #2830

### Checklist

    - [x] Unit Tests
    - [ ] Integration Tests

Notes: Documentation updated in `docs/content/guides/guides/trusted-issuer.mdx` (new "Self-Issued Tokens" subsection) and `docs/content/guides/getting-started/configuration.mdx` (operator note on additive behavior); changed regions validated locally with Vale (zero warnings on changed lines). No new integration tests added; the behavior is covered by unit tests in `jwt_authenticator_test.go` (self-issued token accepted under federation, unknown issuer rejected, federated failure with no local fallback) and verified manually end-to-end. Not a breaking change: existing configurations keep working and the change is strictly additive for accepted issuers.

### Security checks

### DIFF
--- a/backend/internal/system/security/jwt_authenticator.go
+++ b/backend/internal/system/security/jwt_authenticator.go
@@ -60,18 +60,11 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
 		return nil, errInvalidToken
 	}
 
-	// Step 2: Verify JWT.
-	// If a trusted issuer is configured, the server delegates token issuance to it
-	// and verifies tokens exclusively against its JWKS. Otherwise, verify with the
-	// server's own signing key.
-	if config.GetServerRuntime().Config.Server.SecurityConfig.TrustedIssuer.IsConfigured() {
-		if !h.verifyFederatedToken(token) {
-			return nil, errInvalidToken
-		}
-	} else {
-		if err := h.jwtService.VerifyJWT(token, "", ""); err != nil {
-			return nil, errInvalidToken
-		}
+	// Step 2: Verify the JWT, routing on its issuer. Tokens this server issued
+	// are verified with its own signing key; Additionally when a trusted issuer is
+	// configured, tokens from that issuer are verified against its JWKS.
+	if err := h.verifyToken(token); err != nil {
+		return nil, err
 	}
 
 	// Step 3: Decode JWT payload to extract attributes
@@ -93,6 +86,29 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
 
 	// Create immutable SecurityContext
 	return newSecurityContext(subject, ouID, token, scopes, attributes), nil
+}
+
+// verifyToken verifies the bearer token by routing on its iss claim against
+// an explicit allowlist of accepted issuers. Tokens from the configured
+// trusted issuer (when set) are verified against its JWKS. Tokens whose iss
+// matches this server's own JWT issuer are verified with the local signing
+// key. Any other iss is rejected. There is no cross-issuer fallback.
+func (h *jwtAuthenticator) verifyToken(token string) error {
+	trustedIssuer := config.GetServerRuntime().Config.Server.SecurityConfig.TrustedIssuer
+	iss := extractIssuer(token)
+	switch {
+	case trustedIssuer.IsConfigured() && iss == trustedIssuer.Issuer:
+		if !h.verifyFederatedToken(token) {
+			return errInvalidToken
+		}
+	case iss == config.GetServerRuntime().Config.JWT.Issuer:
+		if err := h.jwtService.VerifyJWT(token, "", ""); err != nil {
+			return errInvalidToken
+		}
+	default:
+		return errInvalidToken
+	}
+	return nil
 }
 
 // verifyFederatedToken checks if the token is from a trusted external issuer and verifies it via JWKS.
@@ -142,6 +158,17 @@ func extractToken(authHeader string) (string, error) {
 	}
 	token := strings.TrimSpace(utils.TrimPrefixFold(authHeader, constants.AuthSchemeBearer))
 	return token, nil
+}
+
+// extractIssuer returns the iss claim of the token, or an empty string if the
+// token payload cannot be decoded or carries no string iss claim.
+func extractIssuer(token string) string {
+	attributes, err := jwt.DecodeJWTPayload(token)
+	if err != nil {
+		return ""
+	}
+	iss, _ := attributes["iss"].(string)
+	return iss
 }
 
 // extractScopes extracts permissions from JWT claims.
--- a/backend/internal/system/security/jwt_authenticator_test.go
+++ b/backend/internal/system/security/jwt_authenticator_test.go
@@ -459,6 +459,7 @@ const (
 	testFederatedJWKSURL     = "https://external-auth:8090/oauth2/jwks"
 	testFederatedAudience    = "FEDERATED_CONSOLE"
 	testResourceServerAudURL = "https://resource-server:9443"
+	testLocalIssuer          = "https://localhost:8090"
 )
 
 // buildFakeJWT creates a fake JWT string with the given header and payload claims.
@@ -746,9 +747,10 @@ func (suite *JWTAuthenticatorTestSuite) TestVerifyFederatedToken_InvalidPayload(
 }
 
 func (suite *JWTAuthenticatorTestSuite) TestAuthenticate_FederatedTokenFailure() {
-	// When the trusted issuer is configured and federated verification fails
-	// (here, JWKS verification returns an error), Authenticate must reject the
-	// request with errInvalidToken and must NOT fall back to local-key verification.
+	// A token whose issuer is the trusted issuer is routed to federated
+	// verification. When that fails (here, JWKS verification returns an error),
+	// Authenticate must reject the request with errInvalidToken and must NOT
+	// fall back to local-key verification (no cross-issuer fallback).
 	config.ResetServerRuntime()
 	defer config.ResetServerRuntime()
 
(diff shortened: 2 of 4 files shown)


backend/internal/system/security/jwt_authenticator.go
```diff
@@ -60,18 +60,11 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
...
-		}
-	} else {
-		if err := h.jwtService.VerifyJWT(token, "", ""); err != nil {
-			return nil, errInvalidToken
-		}
+	// Step 2: Verify the JWT, routing on its issuer. Tokens this server issued
+	// are verified with its own signing key; when a trusted issuer is
+	// configured, tokens from that issuer are verified against its JWKS.
+	if err := h.verifyToken(token); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
RushanNanayakkara: addressed

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -60,18 +60,11 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
...
-	if config.GetServerRuntime().Config.Server.SecurityConfig.TrustedIssuer.IsConfigured() {
-		if !h.verifyFederatedToken(token) {
-			return nil, errInvalidToken
-		}
-	} else {
-		if err := h.jwtService.VerifyJWT(token, "", ""); err != nil {
-			return nil, errInvalidToken
-		}
+	// Step 2: Verify the JWT, routing on its issuer. Tokens this server issued
+	// are verified with its own signing key; when a trusted issuer is
```
senthalan:
```suggestion
	// are verified with its own signing key; Additionally when a trusted issuer is
```

We should indicate that trusted issuer is additional to the own keys

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -95,6 +88,36 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
 	return newSecurityContext(subject, ouID, token, scopes, attributes), nil
 }
 
+// verifyToken verifies the bearer token, routing on its issuer claim. Tokens
+// issued by this server (iss equals the configured JWT issuer) are verified
+// with the server's own signing key. When a trusted issuer is configured,
+// tokens carrying that issuer are verified against its JWKS. A token whose
+// issuer matches neither is rejected. There is no cross-issuer fallback: a
+// token routed to one verifier is never retried with the other.
```
senthalan: What is this cross-issuer fallback? do we need this comment

backend/internal/system/security/jwt_authenticator.go
```diff
@@ -95,6 +88,36 @@ func (h *jwtAuthenticator) Authenticate(r *http.Request) (*SecurityContext, erro
...
+func (h *jwtAuthenticator) verifyToken(token string) error {
+	trustedIssuer := config.GetServerRuntime().Config.Server.SecurityConfig.TrustedIssuer
+	if !trustedIssuer.IsConfigured() {
+		if err := h.jwtService.VerifyJWT(token, "", ""); err != nil {
+			return errInvalidToken
+		}
+		return nil
+	}
```
senthalan: Do we need this check? verifyFederatedToken already have the `!trustedIssuer.IsConfigured()` this check to return

RushanNanayakkara: The idea was to process the tokens without iss values first so it doesn't fail at the switch. But there's a cleaner implementation for that. Updated the code.

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: It implies that a token is never verify -> fail -> and then verify with other issuer. Not necessary to have here. Removed.

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: Updated

backend/internal/system/security/jwt_authenticator.go
RushanNanayakkara: As per OAuth 2.0 spec the issuer should be mandatory. This was not the behaviour we had. Made the change to align with spec. 
New behaviour : issuer is mandatory. If matched with trusted issuer then validated with JWKS, otherwise validated locally.

senthalan (approved):

PR #2927, opened 2026-05-21
Remove SCIM2/Me request from React SDK sign-in flow

### LINKED ISSUE #2925: SCIM2/Me request fired on every login but result is discarded in React SDK
# SCIM2/Me request fired on login but result is discarded in React SDK

## Description

When a user logs into an application using `@thunderid/react`, a `GET /scim2/Me` request is
made automatically during the sign-in flow. However, the result of this call is never used —
the provider discards it and falls back to extracting user claims directly from the ID token.
This results in a wasted network round-trip on every login.

## Root Cause

The call originates through polymorphic dispatch inside the browser SDK's sign-in path:

1. `ThunderIDProvider` detects the auth code in the URL and calls `client.signIn()`
2. `ThunderIDReactClient.signIn()` delegates to `super.signIn()` → `ThunderIDBrowserClient.signIn()`
3. `ThunderIDBrowserClient._exchangeCodeForTokens()` completes the token exchange and returns `this.getUser()`
4. Because `this` is a `ThunderIDReactClient` instance, the overridden `getUser()` is called, which calls `getScim2Me({ baseUrl })` → **`GET /scim2/Me`**
5. The result is passed to `_onSignInCallback(user)`, which resolves to `updateSession()` in `ThunderIDProvider`
6. `updateSession()` **ignores** the passed user and instead calls `extractUserClaimsFromIdToken()` directly

Relevant code in `sdks/react/src/contexts/ThunderID/ThunderIDProvider.tsx`:

```typescript
// TEMPORARY: SCIM2 and Organizations endpoints are not yet supported.
// Tracker: https://github.com/asgardeo/javascript/issues/212
const claims: Record<string, any> = extractUserClaimsFromIdToken(decodedToken);
setUser(claims);
```

## Expected Behavior

The SCIM2/Me endpoint should either:

- **Be called and its result used** — `updateSession()` should consume the user object
  returned from `getUser()` / the `sign-in` callback argument, so the HTTP call is not wasted.
- **Not be called at all** while the endpoint is not yet fully supported — the TEMPORARY
  workaround in `updateSession()` should be matched by suppressing the `getScim2Me` call
  in `ThunderIDReactClient.getUser()` until SCIM2 support is complete.

## Steps to Reproduce

1. Integrate `@thunderid/react` into an application.
2. Configure `ThunderIDProvider` with a valid `baseUrl` and `clientId`.
3. Sign in through the standard redirect flow.
4. Observe the network tab — a `GET /scim2/Me` request is fired immediately after the
   token exchange, but the user state in the app is populated from ID token claims only.

## Impact

- Unnecessary HTTP request to `/scim2/Me` on every login, adding latency and server load.
- If the SCIM2 endpoint returns richer profile data, that data is silently ignored.
- Confusing discrepancy between what the SDK fetches and what the app actually uses.

rajithacharith:
### Purpose
When a user signs in, the React SDK was firing a `GET /scim2/Me` request automatically via `ThunderIDReactClient.getUser()` (called through polymorphic dispatch inside `ThunderIDBrowserClient._exchangeCodeForTokens`). ThunderID does not support the SCIM2 endpoint yet, so this was an unnecessary network round-trip on every login whose result was silently discarded.

### Approach
- `ThunderIDReactClient.getUser()` now derives user info directly from the ID token via `extractUserClaimsFromIdToken`, matching the existing behaviour in `ThunderIDProvider.updateSession()`.
- `ThunderIDReactClient.getUserProfile()` follows the same approach, removing the `getScim2Me` + `getSchemas` calls.
- Unused imports (`flattenUserSchema`, `generateFlattenedUserProfile`, `generateUserProfile`, `getScim2Me`, `getSchemas`) are removed from `ThunderIDReactClient.ts`.
- `ThunderIDProvider` is unchanged in behaviour — it continues to populate user state from ID token claims.

### Related Issues
- Fixes #2925

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Release Notes

* **Breaking Changes**
  * `getUser()` and `getUserProfile()` methods now have simplified signatures and no longer accept optional parameters
  * Updated user profile data retrieval mechanism

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/2927?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/sdks/react/src/ThunderIDReactClient.ts
+++ b/sdks/react/src/ThunderIDReactClient.ts
@@ -18,12 +18,9 @@
 
 import {
   ThunderIDBrowserClient,
-  flattenUserSchema,
-  generateFlattenedUserProfile,
   UserProfile,
   SignInOptions,
   User,
-  generateUserProfile,
   EmbeddedFlowExecuteResponse,
   SignUpOptions,
   EmbeddedFlowExecuteRequestPayload,
@@ -47,8 +44,6 @@ import {
 } from '@thunderid/browser';
 import getAllOrganizations from './api/getAllOrganizations';
 import getMeOrganizations from './api/getMeOrganizations';
-import getSchemas from './api/getSchemas';
-import getScim2Me from './api/getScim2Me';
 import {ThunderIDReactConfig} from './models/config';
 
 class ThunderIDReactClient<T extends ThunderIDReactConfig = ThunderIDReactConfig> extends ThunderIDBrowserClient<T> {
@@ -117,22 +112,8 @@ class ThunderIDReactClient<T extends ThunderIDReactConfig = ThunderIDReactConfig
     throw new Error('Not implemented');
   }
 
-  override async getUser(options?: any): Promise<User> {
-    try {
-      let baseUrl: string = options?.baseUrl;
-
-      if (!baseUrl) {
-        const configData: any = await (this.getStorageManager() as any).getConfigData();
-        baseUrl = configData?.baseUrl;
-      }
-
-      const profile: User = await getScim2Me({baseUrl});
-      const schemas: any = await getSchemas({baseUrl});
-
-      return generateUserProfile(profile, flattenUserSchema(schemas));
-    } catch (error) {
-      return extractUserClaimsFromIdToken(await this.getDecodedIdToken());
-    }
+  override async getUser(): Promise<User> {
+    return extractUserClaimsFromIdToken(await this.getDecodedIdToken());
   }
 
   override async getDecodedIdToken(sessionId?: string): Promise<IdToken> {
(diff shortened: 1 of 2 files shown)


brionmario (approved):

sdks/react/src/ThunderIDReactClient.ts
```diff
@@ -117,22 +112,8 @@ class ThunderIDReactClient<T extends ThunderIDReactConfig = ThunderIDReactConfig
...
-  override async getUser(options?: any): Promise<User> {
-    try {
-      let baseUrl: string = options?.baseUrl;
-
-      if (!baseUrl) {
-        const configData: any = await (this.getStorageManager() as any).getConfigData();
-        baseUrl = configData?.baseUrl;
-      }
-
-      const profile: User = await getScim2Me({baseUrl});
```
brionmario, line 129: @coderabbitai Create an issue to find and remove all scim2 related logic in the SDKs

PR #3016, opened 2026-05-26
Support single-file declarative resources via -resources startup flag

rajithacharith:
### Purpose

Removes the bash-level YAML document-splitting logic from `start.sh`/`start.ps1` and moves the responsibility into the server binary. Thunder now accepts a `-resources <file>` startup flag and reads all declarative resources from a single multi-document YAML file, eliminating the need for separate per-type directories.

### Approach

**Binary flag (`-resources`)** — `main.go` registers a `-resources` flag, resolves it to an absolute path, and stores it via `config.SetResourcesFile()`. `ServerRuntime` gains a `ResourcesFile` field (added via a separate setter to keep `InitializeServerRuntime`'s signature unchanged).

**Per-type loading** — `ResourceLoader.LoadResources()` checks `ResourcesFile`; when set, it calls the new `GetConfigsFromFile(filePath, resourceType)` instead of `GetConfigs(dir)`. Resource type is derived from the loader's directory name with `strings.TrimSuffix(dir, "s")`, which covers all 14 built-in types.

**Multi-document YAML parsing** — `splitYAMLDocuments` splits on `---` separators. `documentMatchesResourceType` matches a `# resource_type: <type>` comment header (supports `resource_type:`, `resource type:`, and `resourcetype:` prefixes, quoted values). Env-var substitution (`{{.Var}}`) is applied per-document so that non-Go-template expressions in other documents (e.g., `{{ t(signin:title) }}` in flow pages) do not trigger parse errors.

**Non-Go-template expression shielding** — `SubstituteEnvironmentVariables` in `configutil.go` now calls `shieldNonGoTemplateExpressions` before parsing, replacing unknown `{{ ... }}` patterns (UI expressions, bare identifiers like `{{appName}}`) with unique placeholders that are restored after template execution. This fixes the pre-existing `function "t" not defined` error when flow YAML and env-var-using YAML coexisted in the same file.

**Start scripts** — `start.sh` and `start.ps1` are simplified: the old document-splitting logic (~100 lines each) is replaced by a lean env-file loader, and `-resources "$RESOURCES_FILE"` is forwarded to the binary.

**Tests:**
- `manager_file_test.go` — unit tests for `splitYAMLDocuments`, `documentMatchesResourceType`, and `GetConfigsFromFile` (21 cases covering env-var substitution, non-Go-template pass-through, error paths, multi-type files)
- `tests/integration/declarativesinglefile/` — integration test suite that restarts the server with `-resources`, verifies resources are loaded (OU, two IDPs including env-var substitution), checks collection listing, and confirms declarative resources are immutable

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Load declarative resources from a single YAML file via a new -resources startup flag
  * Pass resource file to server startup in both debug and normal modes
  * Load env files into process env (supports JSON-array expansion into indexed vars)

* **Bug Fixes**
  * Prevent accidental interpretation of non-template {{...}} literals during substitution
  * Required env-var placeholders now surface errors if missing

* **Tests**
  * Added unit and integration tests for single-file loading, type filtering, env-var substitution, immutability and startup handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/thunder-id/thunderid/pull/3016?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/internal/system/declarative_resource/loader.go
+++ b/backend/internal/system/declarative_resource/loader.go
@@ -19,7 +19,10 @@
 package declarativeresource
 
 import (
+	"flag"
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/thunder-id/thunderid/internal/system/log"
 )
@@ -48,17 +51,44 @@ func NewResourceLoader(config ResourceConfig, store Storer) *ResourceLoader {
 	}
 }
 
-// LoadResources loads all resources from the configured directory.
-// It reads YAML files, parses them, validates them, and stores them.
+// LoadResources loads all resources using the following precedence:
+//  1. A file supplied via the --resources startup argument.
+//  2. YAML files found directly in repository/resources/ (multi-document format with
+//     "# resource_type: <type>" headers), when present.
+//  3. Individual YAML files inside the repository/resources/<DirectoryName>/ subdirectory.
+//
 // Returns an error if any step fails.
 func (l *ResourceLoader) LoadResources() error {
-	// Read configuration files from the directory
-	configs, err := GetConfigs(l.config.DirectoryName)
+	var configs [][]byte
+	var err error
+
+	resourceType := strings.TrimSuffix(l.config.DirectoryName, "s")
+
+	resourcesFile := ""
+	if f := flag.Lookup("resources"); f != nil {
+		if v := f.Value.String(); v != "" {
+			absPath, absErr := filepath.Abs(v)
+			if absErr == nil {
+				resourcesFile = absPath
+			}
+		}
+	}
+
+	if resourcesFile != "" {
+		configs, err = GetConfigsFromFile(resourcesFile, resourceType)
+	} else {
+		var rootConfigs [][]byte
+		rootConfigs, err = GetConfigsFromRootDir(resourceType)
+		if err == nil && len(rootConfigs) > 0 {
+			configs = rootConfigs
+		} else if err == nil {
+			configs, err = GetConfigs(l.config.DirectoryName)
+		}
+	}
 	if err != nil {
 		return err
 	}
 
-	// Process each configuration file
 	for _, cfg := range configs {
 		if err := l.loadSingleResource(cfg); err != nil {
 			return err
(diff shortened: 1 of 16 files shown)


senthalan (commented):
This works with only one file contains the all the resource definitions and provided as an argument during the server startup. But in the k8s deployment, it's better if we can have a specific directory where we will load all the files during the start up. So the deployments can mount it.

rajithacharith:
> This works with only one file contains the all the resource definitions and provided as an argument during the server startup. But in the k8s deployment, it's better if we can have a specific directory where we will load all the files during the start up. So the deployments can mount it.

Added the support. single files are read and loaded from resources/ directory

backend/internal/system/declarative_resource/loader.go
```diff
@@ -48,17 +50,34 @@ func NewResourceLoader(config ResourceConfig, store Storer) *ResourceLoader {
...
 // Returns an error if any step fails.
 func (l *ResourceLoader) LoadResources() error {
-	// Read configuration files from the directory
-	configs, err := GetConfigs(l.config.DirectoryName)
+	var configs [][]byte
+	var err error
+
+	resourceType := strings.TrimSuffix(l.config.DirectoryName, "s")
+
+	if resourcesFile := config.GetServerRuntime().ResourcesFile; resourcesFile != "" {
```
senthalan: Rather than setting the ResourcesFile location into the  config.GetServerRuntime() object and reading it ,can't we get the value from os flag here?

This way, we don't need to have the SetResourcesFile function. I don't want to have this set func as public

backend/internal/system/declarative_resource/loader.go
```diff
@@ -48,17 +50,34 @@ func NewResourceLoader(config ResourceConfig, store Storer) *ResourceLoader {
 	}
 }
 
-// LoadResources loads all resources from the configured directory.
-// It reads YAML files, parses them, validates them, and stores them.
+// LoadResources loads all resources using the following precedence:
+//  1. A file supplied via the --resources startup argument.
+//  2. YAML files found directly in repository/resources/ (multi-document format with
```
senthalan, line 56: Does our helm support attaching file in this location `repository/resources/` ?

rajithacharith: Added this

senthalan (approved):

Issue #3037, opened 2026-05-28
Create embeddable thunderidengine


anushasunkada:
### Current Limitation

OAuth and orchestration flows are implemented under internal/ and composed only inside the Thunder server binary. External teams that want to embed OAuth and/or flow execution in their own Go process today must fork the repo or duplicate wiring against non-importable internal packages.

### Suggested Improvement

Refer [Discussion](https://github.com/thunder-id/thunderid/discussions/2776)

- [ ] Build a external go application using thunderidengine sdk without any DB dependency. Redis is fine for RuntimeDB.
- [x] Same engine should be used by thunderID application to setup flowmeta, flowexe, oauth.
- [x] ThunderIdengine should only initialize flowmeta, flowexe, oauth packages. so the any external go application is able to work as authorization server with minimal setup.
- [ ] Inject custom executors along with enabling some of the default executors like CredentialsAuthExecutor, AuthorizationExecutor, AuthAssertExecutor and ConsentExecutor
- [x] Way to provide configuration, specifically GateClient configuration, or any configuration used by flow, oauth, executor and its dependent packages for example JWT configurations.
- [ ] Engine SDK should own below services
       - [ ] AttributeCacheServiceInterface
       - [ ] ExecutorsRegistry
       - [ ] InterceptorsRegistry
       - [ ] GraphBuilder
       - [ ] CacheManager
       - [ ] All Factories - Flow
       - [ ] DPoPVerifier
       - [ ] JWTServiceInterface
       - [ ] JWEServiceInterface
       - [ ] AuthAssertGeneratorInterface

PR #3054, opened 2026-05-28
<WIP> Flowexe depends on flowgraph instead of flowmgt

anushasunkada:
### Related Issues
- #3037 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * Restructured flow execution and graph-building responsibilities for clearer separation and maintainability.
  * Unified flow definition types into a shared common model across the system.
* **Bug Fixes / UX**
  * Improved error messaging for flow graph issues with updated localized messages.
* **Tests**
  * Expanded unit and integration test coverage around flow loading, graph construction, and system flow resolution.


senthalan:
The direction is right — `flowmgt` shouldn't own graph compilation. Right now `flowmgt.Initialize`
takes `flowFactory`, `execRegistry`, and `graphCache` purely so it can construct a `GraphBuilder`
internally and expose `GetGraph()` on its interface. That's a mixing of concerns: flow definition
management (CRUD, versioning, handle resolution) and graph compilation are two different things, and
coupling them means every caller of `flowmgt.Initialize` has to pass in graph-building dependencies
they otherwise don't care about. I'd push the decoupling a bit further than what's here.

---

**Rename `flowgraph` → `flowbuilder`**

The package is already doing the right thing — it just needs a name that reflects it.

Give it an explicit `Initialize` that owns `graphCache` internally. Nothing outside needs to know about the cache — it's a compilation detail:

```go
// backend/internal/flow/flowbuilder/builder.go

func Initialize(
    flowFactory      core.FlowFactoryInterface,
    executorRegistry executor.ExecutorRegistryInterface,
    graphCache       core.GraphCacheInterface,
) GraphBuilderInterface
```

With this, `flowcore.Initialize` no longer needs to return `graphCache` as a separate value for callers to thread through. It's consumed right here.

---

**`flowmgt.Initialize` takes the builder instead of the three raw deps**

Since the builder encapsulates everything needed for graph compilation, the flowmgt init signature
simplifies to:

```go
// Before
func Initialize(
    mux             *http.ServeMux,
    mcpServer       *mcp.Server,
    cacheManager    cache.CacheManagerInterface,
    flowFactory     core.FlowFactoryInterface,
    execRegistry    executor.ExecutorRegistryInterface,
    graphCache      core.GraphCacheInterface,
) (FlowMgtServiceInterface, declarativeresource.ResourceExporter, error)

// After
func Initialize(
    mux          *http.ServeMux,
    mcpServer    *mcp.Server,
    cacheManager cache.CacheManagerInterface,
    graphBuilder flowbuilder.GraphBuilderInterface,
) (FlowMgtServiceInterface, declarativeresource.ResourceExporter, error)
```

Internally, `flowmgt` still calls `graphBuilder.InvalidateCache(ctx, flowID)` on create/update/delete — that's fine and correct. But `GetGraph()` should be removed from `FlowMgtServiceInterface` altogether. It doesn't belong on a flow management interface; it's a graph compilation concern that leaked up.

---

**`flowexec` owns the graph building**

Rather than introducing another provider layer, let `flowexec` be responsible for fetching the flow definition and building the graph itself. Define the `FlowProvider` interface locally in `backend/internal/flow/flowexec/provider.go` so `flowexec` owns its own dependency contracts:

```go
// backend/internal/flow/flowexec/provider.go

// FlowProvider provides flow definitions for execution.
type FlowProvider interface {
    GetFlow(ctx context.Context, flowID string) (*common.CompleteFlowDefinition, *serviceerror.ServiceError)
    ResolveFlow(ctx context.Context, handle string, flowType common.FlowType) (string, *serviceerror.ServiceError)
}
```

Then `flowexec.Initialize` takes both the provider and the builder:

```go
// Before
func Initialize(
    mux                  *http.ServeMux,
    flowGraphProvider    flowgraph.FlowGraphProvider,
    inboundClientService inboundclient.InboundClientServiceInterface,
    entityProvider       entityprovider.EntityProviderInterface,
    executorRegistry     executor.ExecutorRegistryInterface,
    observabilitySvc     observability.ObservabilityServiceInterface,
    cryptoSvc            kmprovider.RuntimeCryptoProvider,
) (FlowExecServiceInterface, error)

// After
func Initialize(
    mux                  *http.ServeMux,
    flowProvider         FlowProvider,
    inboundClientService inboundclient.InboundClientServiceInterface,
    entityProvider       entityprovider.EntityProviderInterface,
    executorRegistry     executor.ExecutorRegistryInterface,
    observabilitySvc     observability.ObservabilityServiceInterface,
    cryptoSvc            kmprovider.RuntimeCryptoProvider,
    graphBuilder         flowbuilder.GraphBuilderInterface,
) (FlowExecServiceInterface, error)
```

Internally, when `flowexec` needs to start executing a flow it does:

```go
flowDef, err := s.flowProvider.GetFlow(ctx, flowID)
// ...
graph, err := s.graphBuilder.GetGraph(ctx, toFlowDefinition(flowDef))
```

This removes the need for `FlowGraphProvider` in `flowgraph` entirely, and the `flowGraphProviderAdapter` in `flowmgt` can be deleted.

---

**What `servicemanager.go` looks like after this**

`flowMgtService` already satisfies `FlowProvider` since it has `GetFlow` and `GetFlowByHandle`, so there's no adapter needed — just pass it directly:

```go
flowFactory, graphCache := flowcore.Initialize(cacheManager)

execRegistry := executor.Initialize(flowFactory, ouService, idpService, /* ... */)

flowBuilder := flowbuilder.Initialize(flowFactory, execRegistry, graphCache)

flowMgtService, flowMgtExporter, err := flowmgt.Initialize(mux, mcpServer, cacheManager, flowBuilder)
if err != nil {
    logger.Fatal("Failed to initialize FlowMgtService", log.Error(err))
}

// flowMgtService satisfies flowexec.FlowProvider directly — no adapter needed
flowExecService, err := flowexec.Initialize(
    mux, flowMgtService, inboundClientService, entityProvider,
    execRegistry, observabilitySvc, runtimeCryptoSvc, flowBuilder,
)
```

The result is each component has a clear, single responsibility: 
1. `flowbuilder` compiles definitions into graphs
2. `flowmgt` manages definitions and invalidates the cache on mutations
3. `flowexec` orchestrates fetching + building + executing at runtime.

backend/internal/flow/flowexec/service_test.go
```diff
@@ -60,6 +61,46 @@ func (s *stubTransactioner) Transact(ctx context.Context, txFunc func(context.Co
...
+	return a.svc.GetGraph(ctx, flowID)
+}
+
+func (a flowGraphProviderMockAdapter) ResolveFlow(
+	ctx context.Context, handle string, flowType common.FlowType) (*flowgraph.FlowRef, *serviceerror.ServiceError) {
+	flow, err := a.svc.GetFlowByHandle(ctx, handle, flowType)
+	if err != nil {
+		return nil, err
+	}
+	return &flowgraph.FlowRef{ID: flow.ID}, nil
```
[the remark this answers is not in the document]
anushasunkada: Done

backend/internal/flow/common/model.go
```diff
@@ -16,13 +16,17 @@
  * under the License.
  */
 
+// Package common contains shared flow runtime and definition models.
```
ThaminduDilshan, line 19: Package comment is already added to constants.go file. Shall we keep it in a single file?

backend/internal/flow/common/model.go
```diff
@@ -164,3 +168,170 @@ func getDuration(startTime int64, endTime int64) int64 {
...
+//
+//nolint:lll // Schema tags are intentionally verbose.
+type ConditionDefinition struct {
+	Key    string `json:"key" yaml:"key" jsonschema:"Attribute key to check."`
+	Value  string `json:"value" yaml:"value" jsonschema:"Value to match."`
+	OnSkip string `json:"onSkip" yaml:"onSkip" jsonschema:"Node ID to skip to if condition is not met."`
+}
+
+// nodeDefinitionAlias is used to avoid infinite recursion during marshaling/unmarshaling.
+type nodeDefinitionAlias NodeDefinition
```
ThaminduDilshan, line 286: Do we need this alias? NodeDefinition is also in the same file. Feels redundant

backend/internal/flow/core/init.go
```diff
@@ -21,10 +21,13 @@ package core
...
-// Initialize initializes the core flow package
-func Initialize(cacheManager cache.CacheManagerInterface) (FlowFactoryInterface, GraphCacheInterface) {
-	flowFactory := newFlowFactory()
+// Initialize initializes the core flow package.
+func Initialize() FlowFactoryInterface {
+	return newFlowFactory()
+}
+
+// NewGraphCache creates an in-memory graph cache backed by the cache manager.
+func NewGraphCache(cacheManager cache.CacheManagerInterface) GraphCacheInterface {
```
ThaminduDilshan, line 30: Having this method public violates the dependency injection pattern. Are we able to expose this dependency through the Initialize method?

Alternatively this is the cache for built graphs right? Can't we move this to flowbuilder package and keep it private within that? This is anyway only used by the graph builder right? I think this is a better implementation choice

backend/internal/flow/flowbuilder/error_constants.go
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package flowbuilder provides conversion and graph-building components for flow definitions.
```
ThaminduDilshan, line 19: Shall we move the package comment to the service (builder) or init file?

backend/internal/flow/flowbuilder/builder.go
```diff
@@ -16,7 +16,7 @@
  * under the License.
  */
 
-package flowmgt
```
ThaminduDilshan, line 19: Shall we rename the package folder to `builder`? I'm planning to get rid of "flow" part from `flowexec` package as well.

backend/internal/flow/flowbuilder/error_constants.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	i18ncore "github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+var (
+	// ErrorInvalidFlowData is returned when a flow definition is nil/invalid for graph generation.
+	ErrorInvalidFlowData = serviceerror.ServiceError{
+		Type: serviceerror.ClientErrorType,
+		Code: "FLG-1001",
```
ThaminduDilshan, line 31: Shall we use a flow builder specific error code prefix? Maybe `FLB`

backend/internal/flow/flowbuilder/model.go
```diff
@@ -0,0 +1,24 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package flowbuilder
+
+type segmentBoundary struct {
```
ThaminduDilshan, line 21: Shall we add the struct comment? There's a comment already in the main branch model

backend/internal/flow/mgt/service.go
```diff
@@ -567,6 +563,35 @@ func validateFlowDefinition(flowDef *FlowDefinition) *serviceerror.ServiceError
 	return nil
 }
 
+func completeFlowFromDefinition(flowID string, flowDef *FlowDefinition) *common.CompleteFlowDefinition {
```
ThaminduDilshan, line 566: Shall we add missing method comments to the 3 new methods?

backend/internal/flow/flowexec/provider.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+)
+
+// FlowProvider provides flow definitions for execution.
+type FlowProvider interface {
```
ThaminduDilshan, line 29: Shall we rename this to `FlowProviderInterface` to align with the project convention?

backend/internal/flow/flowexec/provider.go
```diff
@@ -0,0 +1,33 @@
...
+package flowexec
+
+import (
+	"context"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+)
+
+// FlowProvider provides flow definitions for execution.
```
ThaminduDilshan, line 28: This is a consumer interface where the consumer package defines the required methods in their side right? Shall we update the method comment to explain the correct purpose of this interface?

ThaminduDilshan: Also shouldn't the inbound client service also use this method? When creating/ updating inbound clients we need to validate whether a flow exists with the given id. When some thunderid implementer prevents implementing flow mgt service, they will also not able to use inbound client service too.

We have two options;
1. Keep the flow provider in caller package. I.e. flowexec. In this case, we'll need to define another caller interface inside inboundclient package. 
2. Reuse the same flow provider. In this case we should move this interface to a common package such as flow/common.

If our intention is 1), I'm fine with the current implementation. If we're planning to reuse the same provider interface, I'd suggest to move this to the common package

backend/internal/flow/flowexec/service.go
```diff
@@ -489,6 +488,31 @@ func (s *flowExecService) encryptEngineContext(ctx context.Context, engineCtx *E
 	return serialized, nil
 }
 
+func (s *flowExecService) loadGraph(ctx context.Context, flowID string, logger *log.Logger) (
+	core.GraphInterface, *serviceerror.ServiceError) {
+	flowDef, svcErr := s.flowProvider.GetFlow(ctx, flowID)
+	if svcErr != nil {
+		logger.Error("Error retrieving flow definition",
```
ThaminduDilshan, line 495: I think we should handle client errors here. If the flow id is empty or flow is not found, flow mgt service (or flow provider) returns a client type service error. We shouldn't log it and should return a client error from flow exec service.

This seems to be a bug in the existing implementation too. Shall we address it or take this to a followup task?

backend/internal/flow/flowexec/service.go
```diff
@@ -489,6 +488,31 @@ func (s *flowExecService) encryptEngineContext(ctx context.Context, engineCtx *E
 	return serialized, nil
 }
 
+func (s *flowExecService) loadGraph(ctx context.Context, flowID string, logger *log.Logger) (
```
ThaminduDilshan, line 491: Shall we add a method comment?

backend/internal/flow/flowexec/service_graph_test.go
```diff
@@ -0,0 +1,432 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan, line 2: Update license year to 2026

backend/internal/flow/flowexec/service_test.go
```diff
@@ -60,6 +58,93 @@ func (s *stubTransactioner) Transact(ctx context.Context, txFunc func(context.Co
 	txCtx := context.WithValue(ctx, txMarkerKey{}, "tx")
 	return txFunc(txCtx)
 }
+
+type flowProviderMockAdapter struct {
```
ThaminduDilshan, line 62: Shall we use mockery generated mocks rather than writing a custom mock?

Same comment applies to `stubFlowProvider` and `stubGraphBuilder ` as well

backend/internal/flow/mgt/error_constants.go
```diff
@@ -223,22 +223,6 @@ var (
...
-		Error: core.I18nMessage{
-			Key:          "error.flowmgtservice.graph_build_failure",
-			DefaultValue: "Graph build failure",
-		},
-		ErrorDescription: core.I18nMessage{
-			Key:          "error.flowmgtservice.graph_build_failure_description",
-			DefaultValue: "Failed to build executable graph from flow definition",
-		},
-	}
 	// ErrorFlowDeclarativeReadOnly is the error returned when trying to modify a declarative (immutable) flow.
```
ThaminduDilshan, line 226: Shall we update rest of the error codes to return the consecutive IDs? I.e. change `FLM-1017` -> `FLM-1016`, `FLM-1018` -> `FLM-1017`, etc.

We can still do such breaking changes at this level

senthalan:
I don't think moving graph.go and graphCache.go inside flowbuilder would work, because it introduces a cyclic dependency between core and flowbuilder. So I am thinking without creating a flowbuilder package, what if we move the builder logic to core.

The problem with this approach is flowBuilder needs the execRegistry, and the executor package already imports core heavily — ExecutorInterface is used throughout the registry and individual executor files. So if we move the builder into core and have it import executor for ExecutorRegistryInterface, we get a cycle between core and executor.
The fix is to also move ExecutorRegistryInterface into core. It naturally belongs there alongside ExecutorInterface which is already defined in core. The executor package already imports core and can just implement those interfaces — no cycle.
On the flowFactory side — when checking the code, I noticed we can remove the need to pass flowFactory into the executor packages. The current usage is to create executors which don't need the service function. We can have a normal public function to perform this instead.

So the init order would be: create execRegistry first, then flowBuilder (inside core), then flowMgtService, then flowExecService. The flowFactory shouldn't need to be exposed out at all.

ThaminduDilshan:
Closing due to inactivity and the PR being outdated. The same tasks have already been onboarded separately.

PR #3096, opened 2026-06-02
Add declarative resource support for resource server sub resources

### LINKED ISSUE #3094: Resource server export omits sub-resources; file-based store parent filtering broken
## Summary

Two related bugs prevent sub-resources (child resources) from working correctly in the resource server export/declarative-load flow.

## Bug 1: Export skips sub-resources

**File:** `backend/internal/resource/declarative_resource.go` — `GetResourceByID`

`GetResourceList` was called with `parentID = nil`, which only returns root-level resources. Sub-resources were never fetched, so they were absent from the exported YAML. Additionally, the final loop that builds `rs.Resources` filtered to `res.ParentHandle == ""`, again excluding sub-resources.

A secondary issue: `idToHandleMap` (used to resolve parent handles) was built inline in the same pass that consumed it, meaning a child processed before its parent would get an empty `ParentHandle`.

## Bug 2: File-based store ignores `ParentHandle` when filtering by parent

**File:** `backend/internal/resource/file_based_store.go` — `GetResourceListByParent`

Declarative resources loaded from YAML never have `res.Parent` (UUID) set — it is tagged `yaml:"-"`. Only `res.ParentHandle` is populated. The store filtered by `res.Parent == nil` for root-level resources, which matched **all** declarative resources (including sub-resources), making them all appear at root level. The child-resource branch (`res.Parent != nil`) was therefore never entered for declarative resources, so sub-resource queries always returned nothing.

## Fix

**`declarative_resource.go`:**
- Added `fetchAllResources` / `fetchResourcesByParent` helpers that recursively fetch top-level resources and then their children (DFS), ensuring parents appear before children.
- Split into two passes: first build `idToHandleMap` over all resources, then resolve parent handles and collect actions.
- Append all resources (root and sub) to `rs.Resources`.

**`file_based_store.go`:**
- Root-level query: added `&& res.ParentHandle == ""` so sub-resources with a handle-based parent are excluded.
- Child query: when `res.Parent` is nil but `res.ParentHandle` is set, extract the expected parent handle from the synthetic ID (`rsID_handle` format) and match against `res.ParentHandle`.

## Steps to reproduce

1. Create a resource server with a resource that has child resources.
2. Export — child resources are missing from the YAML.
3. Write a YAML with sub-resources (`parent: <handle>`) and load it declaratively — all resources appear at root level via the API.

rajithacharith:
### Purpose

Resource server sub-resources were not included in exports, and the file-based store incorrectly returned all resources as root-level when querying declaratively loaded resource servers.

### Approach

**Export (`declarative_resource.go`):** Added `fetchAllResources`/`fetchResourcesByParent` helpers that recursively fetch resources using DFS (parents before children). Split resource building into two passes — first build the full `idToHandleMap`, then resolve parent handles — to avoid ordering-dependent misses. All resources (root and sub) are now appended to the exported flat list with `parent` set where applicable.

**File-based store (`file_based_store.go`):** Root-level queries now also require `res.ParentHandle == ""` to exclude sub-resources whose `res.Parent` UUID is nil (as it always is for YAML-loaded resources). Child queries now additionally match by extracting the parent handle from the synthetic resource ID (`rsID_handle` format) when `res.Parent` is nil but `res.ParentHandle` is set.

### Related Issues
- Fixes #3094

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * Improved resource retrieval to use a full-list fetch, reliably resolve parent relationships across declarative and DB-backed resources, and stream associated actions incrementally.
* **Tests**
  * Updated and added unit tests to cover full-list retrieval, parent-matching logic, and incremental action-loading behavior.

### DIFF
--- a/backend/internal/resource/declarative_resource.go
+++ b/backend/internal/resource/declarative_resource.go
@@ -113,29 +113,18 @@ func (e *resourceServerExporter) GetResourceByID(ctx context.Context, id string)
 		Resources:   []Resource{},
 	}
 
-	// Get all resources for this server
-	var allResources []Resource
-	resOffset := 0
-	for {
-		resources, err := e.service.GetResourceList(ctx, id, nil, serverconst.MaxPageSize, resOffset)
-		if err != nil {
-			return nil, "", err
-		}
-		if len(resources.Resources) == 0 {
-			break
-		}
-		allResources = append(allResources, resources.Resources...)
-		resOffset += len(resources.Resources)
-		if resOffset >= resources.TotalResults {
-			break
-		}
+	allResources, err := e.service.GetAllResourceList(ctx, id)
+	if err != nil {
+		return nil, "", err
 	}
 
-	// Build map for hierarchical structure keyed by resource ID
-	resourceIDMap := make(map[string]*Resource)
-	// Build separate map for ID to Handle lookups (for parent resolution)
+	// First pass: build ID-to-Handle map so parent handles can be resolved regardless of order
 	idToHandleMap := make(map[string]string)
+	for _, res := range allResources {
+		idToHandleMap[res.ID] = res.Handle
+	}
 
+	// Second pass: build declarative resources with resolved parent handles and actions
 	for _, res := range allResources {
 		resource := Resource{
 			Name:        res.Name,
@@ -144,7 +133,6 @@ func (e *resourceServerExporter) GetResourceByID(ctx context.Context, id string)
 			Actions:     []Action{},
 		}
 
-		// Set parent handle from parent ID using the ID to Handle map
 		if res.Parent != nil && *res.Parent != "" {
 			if parentHandle, ok := idToHandleMap[*res.Parent]; ok {
 				resource.ParentHandle = parentHandle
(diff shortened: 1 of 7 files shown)


backend/internal/resource/declarative_resource.go
```diff
@@ -144,55 +134,77 @@ func (e *resourceServerExporter) GetResourceByID(ctx context.Context, id string)
...
 
-		// Store in map keyed by resource ID (not handle) to avoid duplicates
-		resourceIDMap[res.ID] = &resource
-		// Also store ID to Handle mapping for parent lookup
-		idToHandleMap[res.ID] = res.Handle
+	return rs, server.Name, nil
+}
+
+// fetchAllResources fetches all resources for a resource server, including sub-resources recursively.
+func (e *resourceServerExporter) fetchAllResources(
```
sahandilshan: Shouldn't we add this fetch all resources as a service method in the resource service, not a private function here?

backend/internal/resource/declarative_resource.go
```diff
@@ -144,55 +134,77 @@ func (e *resourceServerExporter) GetResourceByID(ctx context.Context, id string)
...
+	return all, nil
+}
 
-	// Build the hierarchical structure - add root-level resources
-	for _, res := range allResources {
-		if res.ParentHandle == "" {
-			if resource, ok := resourceIDMap[res.ID]; ok {
-				rs.Resources = append(rs.Resources, *resource)
+// fetchResourcesByParent fetches resources for a given parent and recursively fetches their children.
+func (e *resourceServerExporter) fetchResourcesByParent(
```
sahandilshan: Same goes for here, why don't we have fucntions in the resource service to get these. To me this seems like a method that should be available in the resource service

rajithacharith: Fixed

backend/internal/resource/declarative_resource.go
rajithacharith: Fixed

sahandilshan (approved):

PR #3123, opened 2026-06-03
Add pre/post interceptor core logic for flows

### LINKED ISSUE #3122: Pre/ Post Execution Interceptors for Flows

Thunder Flow Engine relies on executors attached to flow nodes, which works well for business logic but creates friction for cross-cutting concerns such as input validation, CAPTCHA verification, rate limiting, and authorization. Today, these concerns either require duplicated nodes across flows or are implemented through hardcoded engine logic and special-purpose executors, resulting in inconsistency and maintenance overhead.

- Flow authors who should not have to add the same validation and protection logic to every flow.
- Platform developers who need a consistent way to implement engine-wide behaviors.
- Security and operations teams that require uniform enforcement of policies across all flows.

- Reduce duplication in flow definitions.
- Keep flow graphs focused on business logic.
- Provide a consistent model for platform-wide concerns.
- Make it easier to add new validations, security controls, and policies without modifying existing flows.

### Proposed Solution

Introduce an interceptor abstraction — lightweight, focused processing units that execute at well-defined lifecycle points during flow execution. Interceptors are distinct from executors in that:

- They are not nodes in the flow graph — they don't occupy graph positions or affect flow topology
- They execute at lifecycle points managed by the engine, not through node traversal
- They can be configured once at the flow level and selectively applied to specific nodes or all nodes

Interceptors engage at specific points in the engine's execution lifecycle:
```
┌──────────────────────────────────────────────────────────────┐
│                     Flow Request Arrives                     │
│                                                              │
│  ┌────────────────────────────────────────────────────────┐  │
│  │         PRE_REQUEST Interceptors                       │  │
│  │  (once per incoming request, before any node runs)     │  │
│  │  Examples: rate limiter, CAPTCHA, input sanitizer,     │  │
│  │           challenge token validation                   │  │
│  └──────────────────────┬─────────────────────────────────┘  │
│                         │                                    │
│  ┌──────────────────────▼─────────────────────────────────┐  │
│  │              Node Execution Loop                       │  │
│  │                                                        │  │
│  │  For each node:                                        │  │
│  │  ┌─────────────────────────────────────────────────┐   │  │
│  │  │      PRE_NODE Interceptors                      │   │  │
│  │  │  (before each node executes)                    │   │  │
│  │  │  Examples: permission validator,                │   │  │
│  │  │    attribute uniqueness validator               │   │  │
│  │  └────────────────────┬────────────────────────────┘   │  │
│  │                       │                                │  │
│  │  ┌────────────────────▼────────────────────────────┐   │  │
│  │  │       Node Execution (current behavior)         │   │  │
│  │  └────────────────────┬────────────────────────────┘   │  │
│  │                       │                                │  │
│  │  ┌────────────────────▼────────────────────────────┐   │  │
│  │  │      POST_NODE Interceptors                     │   │  │
│  │  │  (after each node completes)                    │   │  │
│  │  │  Examples: output sanitizer                     │   │  │
│  │  └─────────────────────────────────────────────────┘   │  │
│  │                                                        │  │
│  └──────────────────────┬─────────────────────────────────┘  │
│                         │                                    │
│  ┌──────────────────────▼─────────────────────────────────┐  │
│  │         POST_REQUEST Interceptors                      │  │
│  │  (once per request, after all node processing)         │  │
│  │  Examples: response enrichment                         │  │
│  └────────────────────────────────────────────────────────┘  │
│                                                              │
└──────────────────────────────────────────────────────────────┘
```

**The detailed solution:** https://github.com/thunder-id/thunderid/discussions/2375

### Alternatives

_No response_

NutharaNR:
### Purpose

The flow engine executes **executors** wired explicitly as nodes in a flow graph. Cross-cutting concerns (challenge-token validation, CAPTCHA, rate limiting, input/output validation, permission and uniqueness checks) don't belong in the per-node path and would require duplicating nodes across every flow. This PR introduces a first-class **interceptor** abstraction: lightweight units that run at well-defined lifecycle points, declared once at the flow level, without polluting the graph.

---
This change modifies the `config_db` schema by adding a new `INTERCEPTORS` column (JSONB on PostgreSQL, TEXT on SQLite) to the `FLOW_VERSION` table to store interceptor definitions. 

And invalid `challengeToken` error is converted to a 200 OK response with `flowStatus:Error` instead of 400 Bad request.
```
{
    "executionId": "019ed469-ab67-735d-bac6-6f4f434161d7",
    "flowStatus": "ERROR",
    "data": {},
    "error": {
        "code": "ICS-1003",
        "message": {
            "key": "error.interceptor.challenge_token_invalid",
            "defaultValue": "Invalid challenge token"
        },
        "description": {
            "key": "error.interceptor.challenge_token_invalid_description",
            "defaultValue": "The challenge token is missing or invalid"
        }
    }
}
```

#### 💥 Impact
Existing deployments running against the previous schema will encounter schema-mismatch errors on flow management and execution operations, since the application now expects a column that is absent from the deployed database. 

In-flight flow execution instances created before the upgrade may also fail, as the persisted engine context will not align with the updated interceptor logic.

#### 🔄 Migration Guide
Before upgrading, apply the schema change to `config_db` so the `FLOW_VERSION` table includes the new `INTERCEPTORS` column. This can be done by re-initializing the database or by running a migration that adds the column to the existing table.

Ongoing flow executions are not guaranteed to survive the upgrade. Where possible, drain or complete active flow instances before migrating.

---

### Approach

Interceptors are **not nodes**. They execute at four engine lifecycle points and fall into two categories: **default** (always enforced, injected by the engine) and **configurable** (declared in the flow definition, scoped to nodes). A registered interceptor is bound to a lifecycle point through an **interceptor unit** (declaration), so one interceptor implementation can serve multiple modes.

#### Lifecycle Points

```
Flow Request
   │
   ▼  PRE_REQUEST        (once per request, before any node)
   │
   ▼  ┌── for each node ──────────────┐
      │  PRE_NODE   (before the node) │
      │  Node.Execute (existing)      │
      │  POST_NODE  (after the node)  │
      └───────────────────────────────┘
   │
   ▼  POST_REQUEST       (once per request, after node processing)
```

#### Interceptor Categories

| Category         | Behaviour                                                                 | Declared in flow? |
| ---------------- | ------------------------------------------------------------------------- | ----------------- |
| **Default**      | Always enforced; injected by the engine; cannot be declared or skipped.   | No                |
| **Configurable** | Declared in the flow definition; scoped to all or selected nodes.         | Yes               |

#### Core Interface

Mode is not part of the interface; it lives on the interceptor unit. The interface is narrow:

```go
type InterceptorInterface interface {
    Execute(ctx *InterceptorContext) (*common.InterceptorResponse, error)
    GetName() string
    IsDefault() bool
    GetPriority() int
}
```

#### Interceptor Unit (binds an interceptor to a lifecycle point)

Analogous to how a task node binds an executor by name. The resolved interceptor instance is attached at init time and not serialized.

```go
type InterceptorUnitInterface interface {
    GetName() string
    GetMode() common.InterceptorMode
    GetScope() common.InterceptorScope
    GetApplyTo() []string
    GetApplyOnStatuses() []common.FlowStatus
    GetProperties() map[string]interface{}
    GetInterceptor() InterceptorInterface
    // + setters
}
```

#### Interceptor Response

```go
type InterceptorResponse struct {
    Status         common.InterceptorStatus
    EngineOutputs  map[string]string
    ChallengeToken string
    FieldErrors    map[string]string
    Error          *serviceerror.ServiceError
}
```

Status values: `COMPLETE`, `INCOMPLETE`, or `FAIL`. Request-level modes (`PRE_REQUEST`/`POST_REQUEST`) reject `INCOMPLETE` as an architectural constraint; node-level modes allow it for mid-flow UI interactions.

#### Package Structure

```
backend/internal/flow/
├── common/
│   ├── constants.go     # InterceptorMode, InterceptorScope, InterceptorStatus, DataKey
│   └── model.go         # InterceptorResponse
├── core/
│   ├── interceptor.go   # InterceptorInterface, InterceptorUnitInterface, base implementations
│   ├── model.go         # InterceptorContext
│   └── factory.go       # CreateInterceptor, CreateInterceptorUnit, CreateDefaultInterceptorUnits
├── interceptor/         # NEW package
│   ├── init.go          # Initialize() — registers interceptors, builds defaults
│   ├── register.go      # Registry, registrars, sanitization & validation
│   ├── defaults.go      # Default interceptor unit declarations
│   ├── constants.go     # Priorities, names, keys
│   ├── error_constants.go
│   └── challenge_token_interceptor.go
├── flowexec/
│   ├── engine.go        # Interceptor invocation hooks woven into the execution loop
│   ├── interceptor_runner.go  # InterceptorRunner — resolve, filter, sort, execute
│   └── model.go         # InterceptorSharedData on EngineContext, serialization
└── mgt/
    ├── model.go         # InterceptorDefinition on flow definition
    └── store.go         # INTERCEPTORS column persistence
```

#### Registration

A **registrar map** is the single source of truth for all known interceptors. Each maps an interceptor name to a dedicated registration function. The configured interceptor list (`FlowConfig.Interceptors`) controls which interceptors to register at startup; when empty, all built-in interceptors are registered.

#### Scoping and Opt-Out

- `scope: ALL` (or unset): runs for every node unless the node opts out via `properties.skipInterceptors`.
- `scope: SELECTED` + `applyTo: [...]`: runs only for listed node IDs.
- Scoping and `skipInterceptors` apply to per-node modes only (`PRE_NODE`/`POST_NODE`). Request-level interceptors are not node-skippable.
- Default interceptors always apply and cannot be scoped or skipped.

#### Cross-Request State

Interceptor state that must survive across requests of a single flow instance is held in `InterceptorSharedData`, serialized into the flow context. The engine's business logic does not read it.

#### Engine Integration

```
Execute(ctx):
    runInterceptors(PRE_REQUEST, ctx, nil)
    for each node:
        runInterceptors(PRE_NODE, ctx, node)
        node.Execute(ctx)
        runInterceptors(POST_NODE, ctx, node)
    flowStep.Status = COMPLETE
    runInterceptors(POST_REQUEST, ctx, nil)   // status-filtered; also runs on early-exit paths
```

The runner builds a self-contained invocation context (clones of `UserInputs` / `ForwardedData` / `AdditionalData`, plus declarations and shared data), so it never touches the live engine context. `RuntimeData` are merged back via `updateContextWithInterceptorResponse()`. Interceptor specific shared data is mutable via interceptor invocation context.

#### Default Interceptor: Challenge Token

Registered once, declared by the engine for two modes:

- **PRE_REQUEST (validate):** compares the incoming token against the stored hash. Skipped on the first request (no stored hash), when the node's `ExecutionPolicy.SkipChallengeValidation` is set, or when segment restart is allowed.
- **POST_REQUEST (rotate):** generates a new secure token, stores its hash in `SharedData`, and returns the token via `EngineOutputs` for the client response. Targets `postRequestStatuses: ["COMPLETE", "INCOMPLETE"]` so token rotation is skipped on error exits.

#### Configuration

```yaml
# Helm values.yaml
configuration:
  flow:
    interceptors: []   # empty = all built-in interceptors
```

#### Persistence

New `INTERCEPTORS` column on `FLOW_VERSION` (JSONB on PostgreSQL, TEXT on SQLite) stores the `InterceptorDefinition` array from the flow definition.

Flow definition example:
```json
{
  "interceptors": [
    {
      "name": "CaptchaInterceptor",
      "mode": "PRE_NODE",
      "scope": "SELECTED",
      "applyTo": ["prompt_credentials"],
      "properties": { "provider": "recaptcha" }
    },
    {
      "name": "AuditLogInterceptor",
      "mode": "POST_REQUEST",
      "applyOnStatuses": ["COMPLETE"],
      "properties": { "logLevel": "info" }
    }
  ],
  "nodes": [...]
}
```
---
### Related Issues
- https://github.com/thunder-id/thunderid/issues/3122

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added interceptor support to flow execution with lifecycle hooks at request and node execution boundaries
  * Implemented built-in challenge token interceptor for automatic token rotation and validation
  * Extended flow definitions to support interceptor configuration and scoping

* **Database**
  * Extended `FLOW_VERSION` table with `INTERCEPTORS` column to persist interceptor configurations

* **Configuration**
  * Added `flow.interceptors` configuration parameter for registering interceptors


NutharaNR:
@coderabbitai review

backend/internal/flow/interceptor/challenge_token_interceptor.go
```diff
@@ -0,0 +1,141 @@
...
+	ctx *core.InterceptorContext) (*core.InterceptorResult, error) {
+	newToken, err := cryptolib.GenerateSecureToken()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate challenge token: %w", err)
+	}
+
+	ctx.SharedData[sharedDataKeyChallengeTokenHash] = cryptolib.HashToken(newToken)
+
+	// Clear the incoming token from shared data after rotation.
+	delete(ctx.SharedData, common.InterceptorDataKeyChallengeToken)
```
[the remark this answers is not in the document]
NutharaNR: caller already guards against nil panic

backend/internal/flow/core/interceptor.go
```diff
@@ -0,0 +1,186 @@
...
+
+// InterceptorResult holds the outcome of an interceptor execution. The service applies
+// EngineOutputs only on PASS.
+type InterceptorResult struct {
+	// Status indicates whether the interceptor passed or failed.
+	Status common.InterceptorStatus
+
+	// FailureError provides the structured error when Status is FAIL.
+	FailureError *serviceerror.ServiceError
+
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/flowexec/engine.go
```diff
@@ -225,6 +242,53 @@ func (fe *flowEngine) Execute(ctx *EngineContext) (FlowStep, *serviceerror.Servi
...
+		UserInputs:          maps.Clone(ctx.UserInputs),
+		ForwardedData:       maps.Clone(ctx.ForwardedData),
+		AdditionalData:      maps.Clone(ctx.AdditionalData),
+		AuthenticatedUser:   ctx.AuthenticatedUser.Clone(),
+		Interceptors:        ctx.InterceptionContext.Interceptors,
+		SharedData:          ctx.InterceptionContext.SharedData,
+	}
+	if execCtx.SharedData == nil {
+		execCtx.SharedData = make(map[string]string)
+	}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/flowexec/service.go
```diff
@@ -606,6 +627,18 @@ func prepareContext(ctx *EngineContext, action string, inputs map[string]string)
...
+// populateInterceptorSharedData copies incoming request data from the engine context into the
+// interceptor shared data so that interceptors can access it during execution.
+func populateInterceptorData(ctx *EngineContext, challengeTokenIn string) {
+	if ctx.InterceptionContext == nil {
+		return
+	}
+
+	if challengeTokenIn != "" {
+		ctx.InterceptionContext.SharedData[common.InterceptorDataKeyChallengeToken] = challengeTokenIn
+	}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/interceptor/defaults.go
```diff
@@ -0,0 +1,37 @@
...
+var defaultInterceptors = []*core.InterceptorDeclaration{
+	{
+		Name: ChallengeTokenInterceptor,
+		Mode: common.InterceptorModePreRequest,
+	},
+	{
+		Name: ChallengeTokenInterceptor,
+		Mode: common.InterceptorModePostRequest,
+	},
+}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/interceptor/service.go
```diff
@@ -0,0 +1,251 @@
...
+	}, nil
+}
+
+// buildDefaultDeclarations creates InterceptorDeclaration entries for every default interceptor
+// defined in defaultInterceptors (defaults.go).
+func (s *interceptorService) buildDefaultDeclarations() []core.InterceptorDeclarationInterface {
+	declarations := make([]core.InterceptorDeclarationInterface, 0, len(defaultInterceptors))
+	for _, b := range defaultInterceptors {
+		declarations = append(declarations, b)
+	}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+
+		// Resolve the interceptor from the registry if not already set on the binding.
+		if b.GetInterceptor() == nil {
+			ic, err := s.registry.GetInterceptor(b.GetName())
+			if err != nil {
+				logger.Error("Interceptor not found in registry",
+					log.String("interceptorName", b.GetName()))
+				continue
+			}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,251 @@
...
+		if result == nil {
+			continue
+		}
+
+		if result.Status == common.InterceptorStatusFail {
+			logger.Debug("Interceptor failed",
+				log.String("interceptorName", name),
+				log.String("errorCode", result.FailureError.Code))
+			return nil, result.FailureError
+		}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/mgt/model.go
```diff
@@ -32,33 +32,36 @@ import (
...
+	ID           string                  `json:"id,omitempty" yaml:"id,omitempty" jsonschema:"Optional explicit ID for the flow. When omitted a UUID is generated."`
+	Handle       string                  `json:"handle" validate:"required" jsonschema:"Unique identifier for the flow (lowercase, alphanumeric with dashes/underscores). Example: 'basic-login', 'invite-registration'"`
+	Name         string                  `json:"name" validate:"required" jsonschema:"Display name for the flow. Example: 'Basic Login Flow', 'Invite Registration'"`
+	FlowType     common.FlowType         `json:"flowType" validate:"required" jsonschema:"Type of flow: 'AUTHENTICATION' for login flows or 'REGISTRATION' for signup flows"`
+	Interceptors []InterceptorDefinition `json:"interceptors,omitempty" yaml:"interceptors,omitempty" jsonschema:"Optional array of interceptor declarations for cross-cutting concerns (e.g., CAPTCHA, rate limiting)."`
+	Nodes        []NodeDefinition        `json:"nodes" validate:"required" jsonschema:"Array of nodes defining the flow steps. Must include START and END nodes. Use get_flow on existing flows to see node structure examples."`
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/flowexec/engine_test.go
```diff
@@ -2271,3 +2087,745 @@ func (s *EngineTestSuite) TestProcessServiceErrorForEventPublish_NilError() {
...
+		Run(func(args mock.Arguments) {
+			execCtx := args.Get(1).(*interceptor.InterceptorInvocationContext)
+			s.NotNil(execCtx.SharedData, "SharedData should be initialized to empty map when nil")
+		}).
+		Return((map[string]string)(nil), nil)
+
+	err := fe.runInterceptors(common.InterceptorModePreNode, ctx, mockNode)
+
+	s.Nil(err)
+}
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/authn/common/model.go
```diff
@@ -40,6 +41,21 @@ type AuthenticatedUser struct {
...
+// cloned so that mutations to the copy do not affect the original.
+func (u AuthenticatedUser) Clone() AuthenticatedUser {
+	clone := u
+	if u.Attributes != nil {
+		clone.Attributes = maps.Clone(u.Attributes)
+	}
+	if u.AvailableAttributes != nil {
+		copied := *u.AvailableAttributes
+		clone.AvailableAttributes = &copied
+	}
```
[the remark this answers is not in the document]
NutharaNR: fixed

docs/content/guides/guides/flows/advanced-configurations.mdx
```diff
@@ -2211,6 +2211,83 @@ When the user selects a gated action, the matched ACR string is included as the
...
+  },
+  ...
+}
+```
+
+### Built-in Interceptors
+
+Some interceptors are always active and cannot be declared in the flow definition. They run at system-level priority before any configurable interceptors.
+
+**Challenge Token** — Active on every flow. On `POST_REQUEST`, generates a per-step challenge token and includes it in the flow step response under the `challengeToken` field. On `PRE_REQUEST`, validates the token submitted with the incoming request to prevent replay and out-of-order submissions. Validation is skipped on the first request of a new flow instance (no prior token issued yet) and when the engine permits a segment restart. The ThunderID JavaScript SDK reads and forwards the challenge token transparently — no client-side handling is required.
```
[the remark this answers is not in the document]
NutharaNR: fixed

backend/internal/flow/interceptor/registrar.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// builtinRegistrars maps each built-in interceptor name to its registration function.
+var builtinRegistrars = map[string]func(InterceptorDependencies, InterceptorRegistryInterface) error{
+	ChallengeTokenInterceptor: registerChallengeTokenInterceptor,
+}
+
+// RegisterInterceptors registers the given interceptors in the registry. If the list is empty,
+// all built-in interceptors are registered.
+func RegisterInterceptors(deps InterceptorDependencies, registry InterceptorRegistryInterface,
```
NutharaNR: Convert to private methods. Check other places as well.

backend/internal/flow/interceptor/register.go
```diff
@@ -0,0 +1,106 @@
...
+	"fmt"
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+// InterceptorDependencies holds the dependencies required for interceptor initialization.
```
NutharaNR, line 110: combine with registry.go

backend/internal/flow/interceptor/service.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+// interceptorService is the default implementation of InterceptorServiceInterface.
+type interceptorService struct {
+	registry InterceptorRegistryInterface
+	logger   *log.Logger
+}
+
+// NewInterceptorService creates a new interceptor service.
+func NewInterceptorService(registry InterceptorRegistryInterface) InterceptorServiceInterface {
```
NutharaNR: convert to private. Move service.go under `flowexec/`

backend/internal/flow/common/constants.go
```diff
@@ -356,3 +359,47 @@ const (
...
+// Interceptor mode constants.
+const (
+	InterceptorModePreRequest  InterceptorMode = "PRE_REQUEST"
+	InterceptorModePreNode     InterceptorMode = "PRE_NODE"
+	InterceptorModePostNode    InterceptorMode = "POST_NODE"
+	InterceptorModePostRequest InterceptorMode = "POST_REQUEST"
+)
+
+// InterceptorCategory distinguishes default (always enforced) from configurable (flow-declared) interceptors.
+type InterceptorCategory string
```
ThaminduDilshan: Is there any use case to have multiple type of categories in future? If not we can just use a boolean in the interceptor struct or constructor method (isDefault or something similar)

backend/internal/flow/common/constants.go
```diff
@@ -356,3 +359,47 @@ const (
...
+	InterceptorStatusFail InterceptorStatus = "FAIL"
+)
+
+// Interceptor shared data keys for incoming request data populated before interceptor execution.
+const (
+	// InterceptorDataKeyChallengeToken is the shared data key for the incoming challenge token.
+	InterceptorDataKeyChallengeToken = "challengeTokenIn"
+)
+
+// InterceptorScope determines which nodes a per-node interceptor applies to.
```
ThaminduDilshan, line 387: Shall we update the method comment here?

backend/internal/flow/flowexec/service.go
```diff
@@ -233,6 +236,24 @@ func (s *flowExecService) initContext(ctx context.Context, appID string, flowTyp
 		return nil, err
 	}
 
+	// Initialize the interception context from the flow definition's interceptors.
+	if s.interceptorService != nil {
+		InterceptorDeclarations, flowErr := s.flowMgtService.GetInterceptorDeclarations(ctx, graphID)
```
ThaminduDilshan: This approach is not going to work with the improvement we're trying to do with https://github.com/thunder-id/thunderid/pull/3054. We should be able to use flow execution without flow mgt dependency

backend/internal/flow/core/interceptor.go
```diff
@@ -0,0 +1,198 @@
...
+	SetApplyTo(applyTo []string)
+	SetProperties(properties map[string]interface{})
+	SetInterceptor(interceptor InterceptorInterface)
+}
+
+// InterceptorDeclaration is the runtime representation of an interceptor declaration used within
+// the interception context. It binds a named interceptor from the registry to its flow-level
+// configuration. The Interceptor field is resolved from the registry at initialization time
+// and is not serialized.
+type InterceptorDeclaration struct {
```
ThaminduDilshan: Shall we make this private?

backend/internal/flow/core/interceptor.go
```diff
@@ -0,0 +1,198 @@
...
+
+import (
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+)
+
+// InterceptorDeclarationInterface defines the contract for an interceptor binding. A binding
+// associates a named interceptor from the registry with its flow-level configuration
+// (mode, scope, properties), similar to how a task execution node binds an executor by name.
+type InterceptorDeclarationInterface interface {
```
ThaminduDilshan: The "Declaration" term is unique to interceptors — the rest of the flow code uses Definition for config (NodeDefinition, ExecutorDefinition, InterceptorDefinition already exists in mgt/model.go) and goes straight from definition → resolved runtime type (taskExecutionNode, ExecutorInterface instance) with no intermediate wrapper. This introduces both a new term and a new layer that the rest of the package doesn't have.

Also, InterceptorDeclarationInterface abstracts nothing — single implementation, plain getters/setters over public fields, no test seam vs. a struct literal. Suggest either:

1. Drop the wrapper entirely: resolve mgt.InterceptorDefinition straight into a configured InterceptorInterface (matches the executor pattern), or
2. If a runtime wrapper is genuinely needed for the positioning metadata, keep it as a plain struct (no interface, no setters) and reuse the existing InterceptorDefinition term to stay consistent with the rest of the package.

I prefer to align this with the existing node/ executor pattern

backend/internal/flow/common/constants.go
NutharaNR: updated.

backend/internal/flow/core/interceptor.go
NutharaNR: Updated to have a holder object --> `InterceptorUnit` (equivalent to taskExecutionNode) and hold `interceptor` (equivalent to `executor`). `InterceptorDefintion` is kept as it is cz it is required in management layer.

backend/cmd/server/servicemanager.go
```diff
@@ -420,23 +415,41 @@ func unregisterServices() {
...
 	// Initialize flow executor registry
-	execRegistry, err := executor.Initialize(deps, config.GetServerRuntime().Config.Flow)
+	execRegistry, err := executor.Initialize(execDeps, config.GetServerRuntime().Config.Flow)
 	if err != nil {
 		logger.Fatal(ctx, "Failed to register flow executors", log.Error(err))
 	}
-	return flowFactory, graphCache, execRegistry
+	interceptorRegistry, err := interceptor.Initialize(interceptorDeps, config.GetServerRuntime().Config.Flow)
+	if err != nil {
+		logger.Fatal(ctx, "Failed to initialize InterceptorRegistry", log.Error(err))
```
```suggestion
		logger.Fatal(ctx, "Failed to initialize interceptor registry", log.Error(err))
```

backend/internal/flow/common/constants.go
```diff
@@ -358,3 +358,39 @@ const (
...
+)
+
+// InterceptorStatus represents the outcome of an interceptor execution.
+type InterceptorStatus string
+
+// Interceptor status constants.
+const (
+	InterceptorStatusComplete   InterceptorStatus = "COMPLETE"
+	InterceptorStatusIncomplete InterceptorStatus = "INCOMPLETE"
+	InterceptorStatusFail       InterceptorStatus = "FAIL"
```
ThaminduDilshan: Shall we rename this to Failure to align with existing node/ executor statuses?

backend/internal/flow/core/interceptor.go
```diff
@@ -0,0 +1,194 @@
...
+	SetInterceptor(interceptor InterceptorInterface)
+	Clone() InterceptorUnitInterface
+}
+
+// interceptorUnit is the runtime representation of an interceptor declaration used within
+// the interception context. It binds a named interceptor from the registry to its flow-level
+// configuration. The Interceptor field is resolved from the registry at initialization time
+// and is not serialized.
+type interceptorUnit struct {
+	Name        string                  `json:"name"`
```
ThaminduDilshan: Shall we keep these fields private since we already have getters and setters?

backend/internal/flow/core/interceptor.go
```diff
@@ -0,0 +1,194 @@
...
+package core
+
+import (
+	"github.com/thunder-id/thunderid/internal/flow/common"
+)
+
+// InterceptorUnitInterface defines the contract for an interceptor binding. A binding
+// associates a named interceptor from the registry with its flow-level configuration
+// (mode, scope, properties), similar to how a task execution node binds an executor by name.
+type InterceptorUnitInterface interface {
```
ThaminduDilshan: Shall we split two interfaces and implementations into two files for clarity?
Maybe `interceptor_unit.go` and `interceptor.go`

backend/internal/flow/mgt/service.go
```diff
@@ -564,6 +574,82 @@ func validateFlowDefinition(flowDef *FlowDefinition) *serviceerror.ServiceError
...
+	if err := validateInterceptors(flowDef.Interceptors); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// validateInterceptors validates the interceptor definitions declared in a flow.
+func validateInterceptors(interceptors []InterceptorDefinition) *serviceerror.ServiceError {
+	validModes := map[common.InterceptorMode]bool{
```
ThaminduDilshan: Shall we move these two maps to a constants file?

backend/internal/system/config/config.go
```diff
@@ -316,11 +316,12 @@ func (c *OTPConfig) Validate() error {
...
-	DefaultAuthFlowHandle    string `yaml:"default_auth_flow_handle" json:"default_auth_flow_handle"`
-	UserOnboardingFlowHandle string `yaml:"user_onboarding_flow_handle" json:"user_onboarding_flow_handle"`
-	MaxVersionHistory        int    `yaml:"max_version_history" json:"max_version_history"`
-	AutoInferRegistration    bool   `yaml:"auto_infer_registration" json:"auto_infer_registration"`
-	Store                    string `yaml:"store" json:"store"`
+	DefaultAuthFlowHandle    string   `yaml:"default_auth_flow_handle" json:"default_auth_flow_handle"`
+	UserOnboardingFlowHandle string   `yaml:"user_onboarding_flow_handle" json:"user_onboarding_flow_handle"`
+	MaxVersionHistory        int      `yaml:"max_version_history" json:"max_version_history"`
+	AutoInferRegistration    bool     `yaml:"auto_infer_registration" json:"auto_infer_registration"`
+	Interceptors             []string `yaml:"interceptors" json:"interceptors"`
```
ThaminduDilshan: Shall we reorder to have this just below executors?

backend/internal/flow/flowexec/model.go
```diff
@@ -62,8 +62,35 @@ type EngineContext struct {
 	Assertion         string
 	ExecutionHistory  map[string]*common.NodeExecutionRecord
 
-	ChallengeTokenIn   string
-	ChallengeTokenHash string
+	InterceptorSharedData map[string]string
+}
+
+// MergeRuntimeData merges the given data into RuntimeData.
+func (ec *EngineContext) MergeRuntimeData(data map[string]string) {
```
ThaminduDilshan: If possible, let's make this a private method in the flow exec package

backend/internal/flow/flowexec/service.go
```diff
@@ -218,7 +217,6 @@ func (s *flowExecService) initContext(ctx context.Context, appID string, flowTyp
 	if err := s.setApplicationToContext(&engineCtx, logger); err != nil {
 		return nil, err
 	}
-
 	return &engineCtx, nil
```
ThaminduDilshan, line 255: Shall we keep the empty line?

backend/internal/flow/flowexec/service.go
```diff
@@ -520,13 +518,27 @@ func prepareContext(ctx *EngineContext, action string, inputs map[string]string)
...
 
 	// Set the action if provided
 	if action != "" {
 		ctx.CurrentAction = action
 	}
 }
 
+// setChallengeToken copies incoming request data from the engine context into the
+// interceptor shared data so that interceptors can access it during execution.
+func setChallengeToken(ctx *EngineContext, challengeTokenIn string) {
```
ThaminduDilshan: Shall we rename method name to something like `setChallengeTokenInCtx` or similar?

backend/internal/flow/flowexec/interceptor_runner_test.go
```diff
@@ -0,0 +1,550 @@
...
+
+func (r *testInterceptorRegistry) IsRegistered(name string) bool {
+	_, ok := r.interceptors[name]
+	return ok
+}
+
+// --- Stubs ---
+
+// stubInterceptor implements core.InterceptorInterface for testing.
+type stubInterceptor struct {
```
ThaminduDilshan: Can't we use mockery to generate mocks rather than writing custom mocks?

backend/internal/flow/interceptor/constants.go
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package interceptor provides the interceptor abstraction for cross-cutting flow concerns.
```
ThaminduDilshan: Shall we keep package comment only in init.go file?

backend/internal/flow/interceptor/constants.go
```diff
@@ -0,0 +1,40 @@
...
+package interceptor
+
+const (
+	// PriorityDefault is the priority for default (always-enforced) interceptors.
+	PriorityDefault = 100
+
+	// PriorityConfigurable is the base priority for configurable (flow-declared) interceptors.
+	PriorityConfigurable = 200
```
ThaminduDilshan: Shall we rename this constant to contain base part too?

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/flow/interceptor"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	i18ncore "github.com/thunder-id/thunderid/internal/system/i18n/core"
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+const (
+	propertyKeySkipInterceptors = "skipInterceptors"
```
ThaminduDilshan: There should be a place where we list down all the property keys. Shall we move this also to that place?

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+// InterceptorRunnerInterface owns interceptor orchestration for flow instances.
+type InterceptorRunnerInterface interface {
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+	return mode == common.InterceptorModePreNode || mode == common.InterceptorModePostNode
+}
+
+// validateConfiguredInterceptors checks that none of the configured interceptors are default ones.
+func (s *interceptorRunner) validateConfiguredInterceptors(
+	execCtx *InterceptorRunnerContext, logger *log.Logger,
+) *serviceerror.ServiceError {
+	for _, b := range execCtx.Interceptors {
+		if isDefaultInterceptorUnit(b.GetName()) {
+			logger.Error(execCtx.Ctx, "Default interceptor cannot be configured",
```
ThaminduDilshan: We can just print a warn log and ignore it right? No need to fail the execution. Since it's anyway getting appended via the default interceptor merge, it will run with default configurations

backend/internal/flow/interceptor/defaults.go
```diff
@@ -0,0 +1,59 @@
...
+
+// DefaultInterceptors lists all default (always-enforced) interceptors.
+// These declarations are automatically injected into every flow instance.
+var DefaultInterceptors []core.InterceptorUnitInterface
+
+// DefaultInterceptorNames holds the unique names of all default interceptors for quick lookup.
+var DefaultInterceptorNames map[string]struct{}
+
+// initDefaultInterceptorUnits builds the default interceptor execution units using the flow factory.
+func initDefaultInterceptorUnits(factory core.FlowFactoryInterface) {
```
ThaminduDilshan: What if we move these two methods to the registry? We can remove this file completely

backend/internal/flow/interceptor/defaults.go
```diff
@@ -0,0 +1,59 @@
...
+package interceptor
+
+import (
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/flow/core"
+)
+
+// DefaultInterceptors lists all default (always-enforced) interceptors.
+// These declarations are automatically injected into every flow instance.
+var DefaultInterceptors []core.InterceptorUnitInterface
```
ThaminduDilshan: These variables can also be held inside the registry right?

backend/internal/flow/interceptor/register_test.go
```diff
@@ -0,0 +1,311 @@
...
+	err := registerChallengeTokenInterceptor(deps, registry)
+
+	assert.NoError(s.T(), err)
+	assert.True(s.T(), registry.IsRegistered(ChallengeTokenInterceptor))
+}
+
+// --- Test stubs ---
+
+// stubInterceptor is a minimal test implementation of InterceptorInterface.
+type stubInterceptor struct {
```
ThaminduDilshan: Can't we use mockery to generate mocks rather than writing custom implementations?

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+
+		// Resolve the interceptor from the registry if not already set on the unit.
+		if b.GetInterceptor() == nil {
+			ic, err := s.registry.GetInterceptor(b.GetName())
+			if err != nil {
+				logger.Error(execCtx.Ctx, "Interceptor not found in registry",
+					log.String("interceptorName", b.GetName()), log.Error(err))
+				return nil, serviceerror.CustomServiceError(interceptor.ErrorInterceptorExecution, i18ncore.I18nMessage{
+					Key:          "error.interceptor.execution_error_description",
```
ThaminduDilshan: This exceeds 120 line length right? Not sure why the linter passed

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+	logger := s.logger.With(log.String(log.LoggerKeyExecutionID, execCtx.ExecutionID))
+
+	node := execCtx.CurrentNode
+	var currentNodeID string
+	if node != nil {
+		currentNodeID = node.GetID()
+	}
+
+	// Validate that configured interceptors are not default ones.
+	if err := s.validateConfiguredInterceptors(execCtx, logger); err != nil {
```
ThaminduDilshan: One disadvantage with the current logic is that it validates and build interceptor list on each request or node execution right? It may also run beyond that when called at PRE and POST hooks too.
If we can improve the logic to only perform some part of this once for the graph, it should improve the performance. Similar logic should have already been followed with executors.

Registry can keep registered executors separately; default and rest (configurable) (Loop in `GetDefaultInterceptorUnits()` method should also get improved with this).
At graph construction, we can get the applicable interceptor list for the graph with the interceptor list in the flow definition. At this level we just need to pass the applicable criteria and get the filtered list.

WDYT?

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -0,0 +1,340 @@
...
+		case common.InterceptorStatusIncomplete:
+			s.mergeInterceptorResponse(resp, result)
+			resp.Status = common.InterceptorStatusIncomplete
+			return resp, nil
+		case common.InterceptorStatusFail:
+			s.handleInterceptorError(resp, result)
+			resp.Status = common.InterceptorStatusFail
+			return resp, nil
+		default:
+			logger.Error(execCtx.Ctx, "Interceptor returned invalid status",
```
ThaminduDilshan, line 154: This is due to a implementation issue right? Shall we handle this as a server error?

backend/internal/flow/core/model.go
```diff
@@ -69,3 +69,34 @@ type ExecutionPolicy struct {
...
+	AppID       string
+	FlowType    common.FlowType
+
+	// Mode is the lifecycle point at which this interceptor is executing.
+	Mode common.InterceptorMode
+
+	// Engine state
+	FlowStatus          common.FlowStatus
+	UserInputs          map[string]string
+	CurrentNode         NodeInterface
```
ThaminduDilshan: Shall we set the node's execution policy or inner fields similar to `AllowSegmentRestart` field? Passing node is unnecessary to the interceptors and could allow access to unwanted node data. So better to remove node if possible

backend/internal/flow/flowexec/engine.go
```diff
@@ -82,11 +84,29 @@ func (fe *flowEngine) Execute(ctx *EngineContext) (FlowStep, *serviceerror.Servi
 		return flowStep, err
 	}
 
-	skipChallengeValidation := fe.validateSegmentResumePolicy(ctx, logger)
 	currentNode := ctx.CurrentNode
 
+	// Run PRE_REQUEST interceptors once before any node executes.
+	interceptorExecSuccess, svcErr := fe.runInterceptors(
```
ThaminduDilshan:
```suggestion
	itSuccess, svcErr := fe.runInterceptors(
```

backend/internal/flow/interceptor/challenge_token_interceptor_test.go
```diff
@@ -0,0 +1,376 @@
...
+	assert.NotEqual(s.T(), "old-hash", ctx.SharedData[sharedDataKeyChallengeTokenHash])
+
+	newToken := result.ChallengeToken
+	assert.True(s.T(), cryptolib.ValidateTokenHash(newToken, ctx.SharedData[sharedDataKeyChallengeTokenHash]))
+}
+
+// --- Stubs ---
+
+// stubFlowFactory implements core.FlowFactoryInterface for testing.
+type stubFlowFactory struct{}
```
ThaminduDilshan: Shall we avoid writing custom stubs and try to use mockery as much as possible?

backend/internal/flow/flowexec/interceptor_runner.go
NutharaNR: Yes. It is already handled as a server error with;
```
var ErrorInterceptorExecution = serviceerror.ServiceError{
	Code: "ICS-1001",
	Type: serviceerror.ServerErrorType,
	Error: core.I18nMessage{
		Key:          "error.interceptor.execution_error",
		DefaultValue: "Interceptor execution error",
	},
	ErrorDescription: core.I18nMessage{
		Key:          "error.interceptor.execution_error_description",
		DefaultValue: "An error occurred while executing an interceptor",
	},
}
```

Converted this to generic InternalServer Error at engine.

backend/internal/flow/interceptor/error_constants.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	"github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+// ErrorInterceptorExecution defines the error for interceptor execution failures.
+var ErrorInterceptorExecution = serviceerror.ServiceError{
+	Code: "ICS-1001",
+	Type: serviceerror.ServerErrorType,
```
ThaminduDilshan: We shouldn't use package specific server errors. Let's use generic InternalServerError struct from system package instead

backend/internal/flow/flowexec/interceptor_runner.go
NutharaNR: The `executor` is bound to a `task_execution_node` only at execution time (lazy executor binding). To keep `interceptor` resolution consistent with this pattern, interceptor handling is split into two phases:

- Graph build time: Merge defaults interceptor units into the graph's interceptor list, validate names, group by mode, sort by priority.
- Runner (runtime): Filter by node scoping, then resolve the `interceptor` from the registry (lazy, like executors).

backend/internal/flow/core/interceptor.go
NutharaNR: converted to private

backend/internal/flow/core/interceptor.go
NutharaNR: separated into two files

backend/internal/flow/mgt/service.go
NutharaNR: moved to constants file

backend/internal/system/config/config.go
NutharaNR: updated

backend/internal/flow/flowexec/model.go
NutharaNR: converted to private

backend/internal/flow/flowexec/interceptor_runner_test.go
NutharaNR: updated to use mockery

backend/internal/flow/flowexec/service.go
NutharaNR: updated

backend/internal/flow/interceptor/constants.go
NutharaNR: updated

backend/internal/flow/flowexec/interceptor_runner.go
NutharaNR: moved to node property keys under common/constants.go

backend/internal/flow/flowexec/interceptor_runner.go
NutharaNR: this is moved to graph builder https://github.com/thunder-id/thunderid/pull/3123

backend/internal/flow/interceptor/defaults.go
NutharaNR: moved to register.go

backend/internal/flow/interceptor/defaults.go
NutharaNR: moved to register.go

backend/internal/flow/interceptor/register_test.go
NutharaNR: updated

backend/internal/flow/flowexec/interceptor_runner.go
NutharaNR: updated.

backend/internal/flow/core/model.go
NutharaNR: The interceptorRunnerContext modified to follows to include required node info instead of passing the whole node:
```
type InterceptorRunnerContext struct {
	Ctx                  context.Context
	ExecutionID          string
	AppID                string
	FlowType             common.FlowType
	FlowStatus           common.FlowStatus
	CurrentNodeID        string
	NodeType             common.NodeType
	SkipInterceptors     []string
	ExecutionPolicy      *core.ExecutionPolicy
	AllowSegmentRestart  bool
	UserInputs           map[string]string
	ForwardedData        map[string]interface{}
	AdditionalData       map[string]string
	ResolvedInterceptors []core.InterceptorUnitInterface
	SharedData           map[string]string
}
```

backend/internal/flow/interceptor/constants.go
NutharaNR: remained to `BasePriorityConfigurable`

backend/internal/flow/core/graph.go
```diff
@@ -46,16 +46,19 @@ type GraphInterface interface {
 	SetSegments(segments []Segment)
 	GetSegmentByID(segmentID string) *Segment
 	GetSegmentByStartNode(nodeID string) *Segment
+	GetResolvedInterceptors(mode common.InterceptorMode) []InterceptorUnitInterface
```
ThaminduDilshan: Shall we keep terms simple? No need to mention resolved part in method names and struct variables too right?
This is just returning the list of interceptors applicable or configured for the graph

backend/internal/flow/flowexec/engine.go
```diff
@@ -416,6 +563,59 @@ func (fe *flowEngine) updateContextWithNodeResponse(engineCtx *EngineContext, no
...
+	case common.InterceptorStatusFailure:
+		flowStep.Status = common.FlowStatusError
+		flowStep.Error = resp.Error
+		return false
+	default:
+		return true
+	}
+}
+
+func (fe *flowEngine) updateFlowStepWithInterceptorResponse(flowStep *FlowStep, resp *common.InterceptorResponse) {
```
ThaminduDilshan, line 606: Shall we add a method comment?

backend/internal/flow/flowexec/interceptor_runner_test.go
```diff
@@ -0,0 +1,439 @@
...
+		if args.Get(0) != nil {
+			ic = args.Get(0).(core.InterceptorInterface)
+		}
+	}).Return().Maybe()
+	return m
+}
+
+// testInterceptorUnit implements core.InterceptorUnitInterface for testing. Retained because
+// engine_test.go in the same package depends on it.
+type testInterceptorUnit struct {
```
ThaminduDilshan: We can replace this also with mockery generated mock right? 
Disadvantage of having custom written mocks is that they need to be updated each time when a interface is modified. By using mockery, we can avoid that as mocks will be auto generated and only mocks the required method behaviour

backend/internal/flow/mgt/constants.go
```diff
@@ -48,6 +50,20 @@ const (
 	defaultNodeYPos = 0
 )
 
+// validInterceptorModes contains the set of valid interceptor modes for validation.
+var validInterceptorModes = map[common.InterceptorMode]bool{
```
ThaminduDilshan: Shall we reuse the data structures in flow common rather than duplicating the same lists here?

backend/internal/flow/interceptor/error_constants.go
```diff
@@ -0,0 +1,52 @@
...
+package interceptor
+
+import (
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	"github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+// ErrorInterceptorFailed defines the error for interceptor validation failures.
+var ErrorInterceptorFailed = serviceerror.ServiceError{
+	Code: "ICS-1002",
```
ThaminduDilshan: Shall we check error codes are with correct identifiers? `ICS-1001` seems to be missing

NutharaNR: updated

backend/internal/flow/mgt/constants.go
NutharaNR: updated

backend/internal/flow/flowexec/interceptor_runner_test.go
NutharaNR: updated to use mockery generated mocks

backend/internal/flow/core/graph.go
NutharaNR: renamed the method to `Get/SetInterceptors` and graph field to `interceptors`

backend/internal/flow/flowexec/engine.go
NutharaNR: added method comment

ThaminduDilshan (approved):

PR #3127, opened 2026-06-03
Add IDP attribute configuration for token exchange and federated login

### LINKED ISSUE #3126: Support attribute configuration on external IDPs (token exchange + federated login)
### Problem

External IDPs emit claims under their own names. These names often differ from what ThunderID expects locally — URI-style claims (`http://schemas.xmlsoap.org/.../emailaddress`), vendor-specific names, or nested structures. Two places are affected:

- **Token exchange (RFC 8693)** — external subject-token claims pass through verbatim, filtered only by `userAttributes`. If the claim names don't match the local names, those claims are either lost or misinterpreted.
- **Federated login (OIDC/OAuth/Google/GitHub)** — the authenticated user's attributes and the identifier used for consistency checks are taken directly from the external claims. Mismatched names mean incorrect or missing attributes on the local user, and just-in-time (JIT) provisioning has no signal for which user type to target.

### Requirements

- External-to-local attribute mapping must be configurable **per external IDP** and applied consistently to **both token exchange and federated login**.
- The configuration must be **common to all IDP types**, not a type-specific property.
- It must **resolve the local user type** for an incoming identity — at minimum a fixed default, with room to add claim-driven resolution later without a breaking change.
- It must **map external attribute names to local user attributes** for the resolved user type:
  - Source names may be URI-style or nested (addressable via a path).
  - Targets must be valid attributes of the resolved user type's schema.
  - A single external attribute may map to multiple local attributes; mapping two external attributes to the same local attribute must be rejected as a conflict.
- **Unmapped claims must pass through unchanged** (backward compatible).
- In federated login, the **resolved user type must drive JIT provisioning** of new users.
- **Self-issued tokens must be unaffected.**

sadilchamishka:
## Purpose

External IDP claims were passed through verbatim, so issued tokens (token exchange) and federated-login user attributes only carried the right claims if the external names matched what ThunderID expects.

This PR adds a structured **attribute configuration** on the identity provider that resolves the local user type and maps external attributes to local attributes, used by both token exchange and federated login.

## Model

`attributeConfiguration` is a **dedicated `ATTRIBUTE_CONFIGURATION` JSONB column** on the IDP — common to every IDP type, separate from the type-specific `properties`. It has two parts: how the incoming identity resolves to a local **user type**, and per-user-type **attribute mappings**.

```json
"attributeConfiguration": {
  "userTypeResolution": {
    "default": "Person"
  },
  "userTypeAttributeMappings": [
    {
      "userType": "Person",
      "attributes": [
        { "externalAttribute": "given_name",    "localAttribute": "firstName" },
        { "externalAttribute": "address.email", "localAttribute": "email" }
      ]
    }
  ]
}
```

- **`userTypeResolution` (required)** — resolves which local user type an incoming identity maps to. This iteration supports only **`default`** — a single fixed user type. The object shape reserves room for claim-driven resolution later, when an IDP can signal the user type from a token claim.
- **`userTypeAttributeMappings`** — an array of per-user-type entries. Each entry:
  - **`userType` (required)** — the local user type whose schema defines the valid target attributes. The resolved user type also selects which entry's mappings apply.
  - **`attributes`** — array of mapping objects, each with:
    - **`externalAttribute` (required)** — source attribute name. May be a dot-notation path into a nested claim; a name containing dots is matched literally first (URI-style claims).
    - **`localAttribute` (required)** — target attribute. Must be a non-credential attribute in the user type's schema (validated at write time).

**Validation:** a single `externalAttribute` may map to multiple local attributes; mapping two external attributes to the same `localAttribute` is a conflict and is rejected (after trimming); at most one entry per user type; unmapped claims pass through unchanged.

### Why this shape

`attributeConfiguration` is named for what it describes — how an external identity becomes (and later stays) a local identity — rather than just "mapping", because the envelope is built to grow. Each addition below is a new optional key on an existing object/array, so the roadmap is **purely additive — no breaking change**:

| Future capability | Where it slots in |
|---|---|
| Claim-driven user-type resolution | `userTypeResolution.externalAttribute` + `valueMapping` |
| Value transforms / value mapping | `transform`, `valueMapping` keys on each attribute object |
| Sync & account-linking policy | `sync`, `useForAccountLinking` keys on each attribute object |
| Multiple user types per IDP | additional entries in `userTypeAttributeMappings` |

See the design discussion (#3157) for the full roadmap and how it maps to reviewer feedback.

## Where it applies

Mapping is applied at the **authn service layer**, not the executor — so the transformation is in one place for atomic auth APIs and the flow executor stays lightweight (per review feedback).

- **Token exchange (RFC 8693)** — external subject-token claims mapped in the token-exchange validator before becoming token attributes.
- **Federated login (OIDC/OAuth/Google/GitHub)** — external claims mapped inside the per-IDP-type authn service's `Authenticate()` (OAuth/OIDC; GitHub and Google inherit it by wrapping those services) before the claims propagate up to the flow. The executors are unchanged from `main`.

`userTypeResolution.default` selects which mapping entry applies and validates targets against that user type's schema at write time; JIT provisioning resolves the user type through the existing flow logic (app-allowed + self-registration), not from IDP claims.

## Also retained

- Property values are no longer HTML-escaped on write (JSON/URL values round-trip correctly).
- Internal properties (`issuer`, `token_exchange_enabled`) are excluded from the external authorize request.

## Testing

DB store round-trip; `validateAttributeConfiguration` (valid / unknown user type / target-not-in-schema / duplicate target / duplicate-target whitespace variant / duplicate user type / one-source-to-multiple-targets / empty); `ApplyAttributeMappings` (rename, nested source, literal dotted keys, collision precedence, passthrough); the OAuth/OIDC authn-service `resolveAttributeMappings` paths (mapping applied, client/server error wrapping, nil IDP); token-exchange consumer tests. Docs updated (OIDC/OAuth/GitHub provider guides + trusted-issuer).

Fixes #3126


backend/internal/idp/utils.go
```diff
@@ -45,6 +46,30 @@ func GetPropertyValue(properties []cmodels.Property, name string) string {
...
+		if strings.TrimSpace(externalClaim) == "" || strings.TrimSpace(localClaim) == "" {
+			return serviceerror.CustomServiceError(ErrorInvalidIDPProperty, core.I18nMessage{
+				Key: "error.idpservice.claim_mappings_empty_entry_description",
+				DefaultValue: fmt.Sprintf(
+					"property '%s' must not contain empty claim names", PropClaimMappings),
+			})
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — validateClaimMappings now rejects JSON null (nil map after unmarshal) and also rejects duplicate mapping targets. Tests added for both cases.

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -252,9 +254,17 @@ func (tv *tokenValidator) resolveExternalIssuer(ctx context.Context, issuer stri
 		return nil, fmt.Errorf("no JWKS endpoint configured for issuer '%s'", issuer)
 	}
 
+	var claimMappings map[string]string
+	if mappingsValue := idp.GetPropertyValue(idpDTO.Properties, idp.PropClaimMappings); mappingsValue != "" {
+		if err := json.Unmarshal([]byte(mappingsValue), &claimMappings); err != nil {
+			return nil, fmt.Errorf("invalid claim mappings configured for issuer '%s': %w", issuer, err)
+		}
+	}
+
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — added a reservedClaimNames blocklist in validateClaimMappings that rejects mapping targets matching standard JWT/OAuth2 claim names (sub, iss, aud, exp, nbf, iat, jti, scope, client_id, act, grant_type). Enforced at IDP write time so invalid mappings are caught at creation/update. Test for mapping to a reserved name (sub) included.

backend/internal/idp/utils.go
```diff
@@ -45,6 +46,30 @@ func GetPropertyValue(properties []cmodels.Property, name string) string {
...
+		})
+	}
+	for externalClaim, localClaim := range mappings {
+		if strings.TrimSpace(externalClaim) == "" || strings.TrimSpace(localClaim) == "" {
+			return serviceerror.CustomServiceError(ErrorInvalidIDPProperty, core.I18nMessage{
+				Key: "error.idpservice.claim_mappings_empty_entry_description",
+				DefaultValue: fmt.Sprintf(
+					"property '%s' must not contain empty claim names", PropClaimMappings),
+			})
+		}
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — validateClaimMappings now trims leading/trailing whitespace from all keys and values before validation and returns a normalized JSON string. The call site in validateIDPProperties replaces the property with this normalized value, so the stored mapping always has clean keys that match at runtime. A test covering whitespace normalization is included.

backend/internal/idp/utils.go
```diff
@@ -45,6 +46,107 @@ func GetPropertyValue(properties []cmodels.Property, name string) string {
...
+			return "", serviceerror.CustomServiceError(ErrorInvalidIDPProperty, core.I18nMessage{
+				Key: "error.idpservice.claim_mappings_duplicate_target_description",
+				DefaultValue: fmt.Sprintf(
+					"property '%s': local claim name '%s' appears as a target more than once",
+					PropClaimMappings, localClaim),
+			})
+		}
+		seenTargets[localClaim] = true
+		normalized[externalClaim] = localClaim
+	}
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — validateClaimMappings now tracks seenSources and rejects two external claim names that collapse to the same value after trimming. Added TestValidateIDPProperties_ClaimMappings_DuplicateSourceAfterTrim covering the whitespace-collision case.

docs/content/guides/guides/trusted-issuer.mdx
```diff
@@ -259,6 +260,44 @@ Response:
...
+```
+
+Mapping behavior:
+
+- A claim whose name appears in `claim_mappings` is renamed to its local name; the original name is dropped.
+- Claims without a mapping entry pass through unchanged.
+- When a mapped (local) name collides with an existing claim of the same name, the mapped value takes precedence.
+- The mapped claims are still subject to the application's `userAttributes` configuration, which determines which attributes appear on the issued token.
+
+Claim mapping applies only to claims carried in the external subject token. It does not perform a local user lookup. To register an IDP with claim mappings:
```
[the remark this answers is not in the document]
sadilchamishka: Done — added 'Mapping is applied only to top-level claims; nested claim objects are not traversed.' to both the trusted-issuer guide and the OIDC provider guide claim-mapping sections.

backend/internal/flow/executor/utils.go
```diff
@@ -28,9 +29,38 @@ import (
...
+
+	idpDTO, svcErr := idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		logger.Error("Failed to load identity provider for claim mapping",
+			log.String("idpID", idpID), log.String("errorCode", svcErr.Code))
+		return claims
+	}
+
+	mappings, err := idp.ParseClaimMappings(idpDTO.Properties)
+	if err != nil {
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — added a nil guard on idpDTO after GetIdentityProvider returns. If the service returns (nil, nil), the function now returns claims unchanged instead of panicking. Added TestApplyIDPClaimMappings_NilIDPDTO_Passthrough to cover this case.

backend/internal/authn/oauth/utils.go
```diff
@@ -71,6 +71,9 @@ func parseIDPConfig(idp *idpPkg.IDPDTO) (*OAuthClientConfig, error) {
 			oAuthClientConfig.OAuthEndpoints.LogoutEndpoint = value
 		case idpPkg.PropJwksEndpoint:
 			oAuthClientConfig.OAuthEndpoints.JwksEndpoint = value
+		case idpPkg.PropIssuer, idpPkg.PropTokenExchangeEnabled:
```
sadilchamishka, line 75: @sahandilshan Is it okay to consider this change.
The authorization request has appended the issuer and token-exchange-enabled as query parameters.

backend/internal/idp/attribute_configuration_test.go
```diff
@@ -0,0 +1,166 @@
...
+		UserTypeAttributeMappings: []UserTypeAttributeMapping{{
+			Attributes: []AttributeMapping{{ExternalAttribute: "given_name", LocalAttribute: "firstName"}},
+		}},
+	}}
+	svcErr := svc.validateAttributeConfiguration(context.Background(), idp)
+	assert.NotNil(t, svcErr)
+	assert.Equal(t, ErrorInvalidAttributeConfiguration.Code, svcErr.Code)
+}
+
+func TestValidateAttributeConfiguration_EmptyMappings(t *testing.T) {
```
sahandilshan: Check if these tests are following the test suite pattern

backend/internal/flow/executor/utils.go
```diff
@@ -28,9 +29,34 @@ import (
...
+// the IDP is created or updated).
+func applyIDPAttributeMappings(ctx context.Context, idpService idp.IDPServiceInterface, idpID string,
+	claims map[string]interface{}, logger *log.Logger) map[string]interface{} {
+	if idpService == nil || idpID == "" || len(claims) == 0 {
+		return claims
+	}
+
+	idpDTO, svcErr := idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		logger.Error("Failed to load identity provider for attribute mapping",
```
sahandilshan: We need to use the logger with context. Check the other places as well

backend/internal/idp/attribute_configuration_test.go
sadilchamishka: Good catch — converted to the testify `suite.Suite` pattern to match the rest of the package (`service_test.go`, `utils_test.go`, etc.). Added `AttributeConfigurationTestSuite` with a `SetupTest` that builds the service with a fresh entity-type mock per test, switched all assertions to the `s.*` style, and run it via `suite.Run`. Done in de37a1791.

backend/internal/flow/executor/utils.go
sadilchamishka: Fixed — `applyIDPAttributeMappings` now logs via `logger.ErrorWithContext(ctx, ...)`. I also swept the other Go files changed in this PR: the only remaining non-context call is in `servicemanager.go` (startup/init path, no request context available), so that one is left as-is. Done in de37a1791.

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -504,6 +509,28 @@ func (o *oAuthExecutor) resolveUserTypeForAutoProvisioning(ctx *core.NodeContext
...
+	if mappedUserType != "" {
+		entityType, svcErr := o.entityTypeService.GetEntityTypeByName(ctx.Context,
+			entitytype.TypeCategoryUser, mappedUserType)
+		if svcErr != nil {
+			if svcErr.Type == serviceerror.ClientErrorType {
+				execResp.Status = common.ExecFailure
+				execResp.Error = svcErr
+				return nil
+			}
+			logger.ErrorWithContext(ctx.Context, "Error while retrieving IDP-configured user type",
```
ThaminduDilshan: We don't need to log at executor level. Return the detailed go error. Task execution node will log it and return a generic internal server error

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -504,6 +509,28 @@ func (o *oAuthExecutor) resolveUserTypeForAutoProvisioning(ctx *core.NodeContext
...
+				log.String("errorCode", svcErr.Code),
```
ThaminduDilshan: Btw `WithContext` log methods are removed now

backend/internal/idp/error_constants.go
```diff
@@ -124,6 +124,20 @@ var (
 			DefaultValue: "One or more identity provider properties are not supported",
 		},
 	}
+	// ErrorInvalidAttributeConfiguration is the error returned when the identity provider's identity
+	// profile configuration is invalid.
+	ErrorInvalidAttributeConfiguration = serviceerror.ServiceError{
```
ThaminduDilshan: Shall we add error codes in order? Move this to the last in the list. Otherwise will be hard to find the next error code when doing a code change and could possibly duplicate the same code

backend/internal/idp/handler.go
```diff
@@ -55,46 +54,47 @@ func (ih *idpHandler) HandleIDPPostRequest(w http.ResponseWriter, r *http.Reques
...
 	idpDTO := &IDPDTO{
-		Name:        sysutils.SanitizeString(createRequest.Name),
-		Description: sysutils.SanitizeString(createRequest.Description),
-		Type:        IDPType(sysutils.SanitizeString(createRequest.Type)),
-		Properties:  properties,
+		Name:                   sysutils.SanitizeString(createRequest.Name),
+		Description:            sysutils.SanitizeString(createRequest.Description),
+		Type:                   IDPType(sysutils.SanitizeString(createRequest.Type)),
+		Properties:             properties,
+		AttributeConfiguration: sanitizeAttributeConfiguration(createRequest.AttributeConfiguration),
```
ThaminduDilshan, line 74: Shall we update the `api/idp.yaml` with the new changes and regenerate the api docs?

ThumulaPerera:
Haven't gone through detail yet. But at glance it looks like the attribute mapping happens at the executor level. Shouldn't that happen at the corresponding authn service level ?

backend/internal/idp/attribute_configuration_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/thunder-id/thunderid/internal/entitytype"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	"github.com/thunder-id/thunderid/internal/system/i18n/core"
+	"github.com/thunder-id/thunderid/tests/mocks/entitytypemock"
+)
+
+type AttributeConfigurationTestSuite struct {
```
ThaminduDilshan: Rather than introducing a new test suite and file, can't we incorporate this into the existing service test suite?

backend/internal/idp/model.go
```diff
@@ -20,13 +20,42 @@ package idp
...
+// UserTypeAttributeMapping holds the external-to-local attribute mappings for a single local user type.
+type UserTypeAttributeMapping struct {
+	UserType   string             `json:"userType,omitempty" yaml:"user_type,omitempty"`
+	Attributes []AttributeMapping `json:"attributes,omitempty" yaml:"attributes,omitempty"`
+}
+
+// AttributeConfiguration holds the user-type resolution and per-user-type attribute mappings for an
+// identity provider.
+type AttributeConfiguration struct {
+	UserTypeResolution        *UserTypeResolution        `json:"userTypeResolution,omitempty" yaml:"user_type_resolution,omitempty"`                //nolint:lll
```
ThaminduDilshan, line 47: @thiva-k can a agent login through IDPs? This will be unlikely for social connectors, but could be there for enterprise login cases. Do we need to consider renaming these to generic terms as user type won't align with agents?
Open for discussion

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -504,6 +509,28 @@ func (o *oAuthExecutor) resolveUserTypeForAutoProvisioning(ctx *core.NodeContext
...
+				log.String("description", svcErr.ErrorDescription.DefaultValue))
+			return errors.New("error while retrieving user type")
+		}
+		execResp.RuntimeData[userTypeKey] = entityType.Name
```
ThaminduDilshan: We can't rely on idp claims to resolve the user type. This could be self-registration disabled or even non allowed user type for the app. 
We should first find the user type and default OU based on the existing logic (app allowed and self-reg enabled). Then this can be added as a additional validation to check whether attribute mappings are configured for the resolved user type

backend/internal/authn/oauth/utils.go
```diff
@@ -72,6 +72,9 @@ func parseIDPConfig(idp *idpPkg.IDPDTO) (*OAuthClientConfig, error) {
...
```
ThaminduDilshan, line 75: Is it intentional to ignore these two configs?

backend/internal/flow/executor/utils.go
```diff
@@ -28,9 +29,34 @@ import (
...
+// the IDP is created or updated).
+func applyIDPAttributeMappings(ctx context.Context, idpService idp.IDPServiceInterface, idpID string,
+	claims map[string]interface{}, logger *log.Logger) map[string]interface{} {
+	if idpService == nil || idpID == "" || len(claims) == 0 {
+		return claims
+	}
+
+	idpDTO, svcErr := idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		logger.ErrorWithContext(ctx, "Failed to load identity provider for attribute mapping",
```
ThaminduDilshan: Is it intentional to log and return the existing claim set here? Should we treat this as a error?

(If treating it as a error, let's modify this util to return a go error and return it directly from executor without logging)

backend/internal/authn/oauth/utils.go
sadilchamishka: Yes, If not those properties will be added in the authorization request we send to the federated IDP, which is unnecessary. This is not related to the PR, but addressed as I noticed that concern.

docs/content/guides/guides/identity-providers/add-oidc-provider.mdx
```diff
@@ -86,6 +86,52 @@ curl -kL -X POST https://localhost:8090/identity-providers \
 | `prompt` | The `prompt` parameter to send in authorization requests (for example, `login`, `consent`). |
 | `token_exchange_enabled` | Set to `true` to enable token exchange with this provider. Requires `issuer` and `jwks_endpoint`. |
 
+## Attribute Configuration
```
ThaminduDilshan, line 89: I prefer to take this out from oidc provider doc as attribute mappings are applicable for all the IDP types. But since we only support oauth and oidc, fine to keep it here for now

ThaminduDilshan:
We might need to add the breaking-change label and update PR description since we're changing the DB schema

backend/internal/idp/attribute_configuration_test.go
sadilchamishka: Merged into IDPServiceTestSuite in service_test.go — the attribute configuration tests now use the shared suite with mockET injected.

backend/internal/flow/executor/utils.go
sadilchamishka: Changed to `logger.ErrorWithContext(ctx, ...)` — `ctx context.Context` is available in the function signature.

backend/internal/flow/executor/oauth_executor.go
sadilchamishka: Removed the executor-level log; the server-error path now returns `fmt.Errorf("error while retrieving user type %s: %s", ...)` so the task-execution node logs it.

backend/internal/flow/executor/oauth_executor.go
sadilchamishka: The executor-level log is removed entirely (see #3401018603). The `WithContext` call in utils.go is for the best-effort IDP-load failure path, which keeps the log package's recommended pattern.

backend/internal/idp/error_constants.go
sadilchamishka: Moved `ErrorInvalidAttributeConfiguration` (IDP-1012) to the end of the list, after IDP-1011.

backend/internal/idp/handler.go
sadilchamishka: Added `attributeConfiguration` to `IDPRequest` and `IDPResponse` in api/idp.yaml, along with new `AttributeConfiguration`, `UserTypeResolution`, `UserTypeAttributeMapping`, and `AttributeMapping` schema definitions.

backend/internal/idp/attribute_configuration_test.go
sadilchamishka: Done — removed `AttributeConfigurationTestSuite` and merged all tests into `IDPServiceTestSuite` (added `mockET` field and inject it in `SetupTest`).

backend/internal/idp/model.go
sadilchamishka: Good point — leaving this open for discussion. The `userType` field reflects the current local-user provisioning model; if agent login through IDPs is needed in future, the field name (or the whole shape) can be revisited then.

backend/internal/flow/executor/oauth_executor.go
sadilchamishka: Fixed — the IDP-configured user type no longer bypasses app logic. It now acts as a filter: `resolveUserTypeForAutoProvisioning` iterates `allowedUserTypes`, skipping entries that don't match the IDP-configured type, then applies the existing self-registration check. Self-registration disabled or non-allowed types are correctly rejected.

backend/internal/authn/oauth/utils.go
sadilchamishka: Yes, intentional — `PropIssuer` and `PropTokenExchangeEnabled` are server-side config consumed by the token-exchange path; they must not be forwarded as query params on the external authorize request. The comment on line 76 explains this.

backend/internal/flow/executor/utils.go
sadilchamishka: Intentional best-effort behaviour — if the IDP fails to load, the original claims are returned unchanged so the authentication flow isn't blocked by a mapping misconfiguration. The comment above the function documents this. Changed `logger.Error` to `logger.ErrorWithContext` to include the correlation ID.

docs/content/guides/guides/identity-providers/add-oidc-provider.mdx
sadilchamishka: Noted — once we have a generic IDP docs page, the attribute mapping section can move there. Keeping it under the OIDC guide for now as agreed.

api/idp.yaml
```diff
@@ -521,6 +525,61 @@ components:
...
+
+    EntityTypeResolution:
+      type: object
+      description: Resolves which local entity type an incoming federated identity maps to.
+      properties:
+        default:
+          type: string
+          description: Default local entity type for federated just-in-time provisioning.
+          example: "person"
+
```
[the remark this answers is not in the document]
sadilchamishka: Done — marked \`entityTypeResolution\` required on \`AttributeConfiguration\` and \`default\` required on \`EntityTypeResolution\` so the contract matches the service-side validation. (0e6e26ac1)

backend/internal/authn/oidc/service.go
```diff
@@ -244,9 +247,14 @@ func (s *oidcAuthnService) Authenticate(ctx context.Context, idpID, code string)
...
 
+	idpDTO, svcErr := s.idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		return nil, svcErr
+	}
+
 	result := &authncm.FederatedAuthResult{
 		Sub:    sub,
-		Claims: claims,
+		Claims: idp.ApplyAttributeMappings(claims, idp.GetAttributeMappings(idpDTO)),
```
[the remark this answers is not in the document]
sadilchamishka: Not adding this guard — the path is already nil-safe and cannot panic: `GetAttributeMappings(nil)` returns nil and `ApplyAttributeMappings(claims, nil)` returns the claims unchanged, and `GetIdentityProvider` never returns `(nil, nil)` (it returns `ErrorIDPNotFound` otherwise). A nil-DTO guard here would be error handling for a case that cannot occur.

api/idp.yaml
```diff
@@ -521,6 +525,65 @@ components:
       items:
         $ref: '#/components/schemas/BasicIDPResponse'
 
+    AttributeConfiguration:
```
ThaminduDilshan, line 578: These new fields are not visible in the swagger editor request, response payloads. Probably we have defined sepsrate examples

sadilchamishka: Good catch — the request/response payloads use inline `example:` blocks that overrode the schema, so the new field was hidden. Added `attributeConfiguration` to all five IDP request/response examples (create/get/update), so it now shows in the Swagger editor. (ca1af9725)

api/idp.yaml
```diff
@@ -521,6 +525,65 @@ components:
...
+      type: object
+      description: External-to-local attribute mapping configuration for the identity provider.
+      required:
+        - entityTypeResolution
+      properties:
+        entityTypeResolution:
```
ThaminduDilshan: We should verify whether having `entity` term is okay with external facing layers. Previous expectation was to consider this a internal implementation detail.
Better to validate with @darshanasbg

backend/internal/authn/oauth/service.go
```diff
@@ -328,9 +328,14 @@ func (s *oAuthAuthnService) Authenticate(ctx context.Context, idpID, code string
 		return nil, &common.ErrorSubClaimNotFound
 	}
 
+	idpDTO, svcErr := s.idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		return nil, svcErr
```
ThaminduDilshan, line 333: We should return a wrapped service error here. Returning idp error code from authn service is wrong IMO.
Also that is the current pattern followed in other places

backend/internal/authn/oidc/service.go
```diff
@@ -244,9 +247,14 @@ func (s *oidcAuthnService) Authenticate(ctx context.Context, idpID, code string)
 		}
 	}
 
+	idpDTO, svcErr := s.idpService.GetIdentityProvider(ctx, idpID)
+	if svcErr != nil {
+		return nil, svcErr
```
ThaminduDilshan, line 252: Same as previous comment

backend/internal/idp/attribute_configuration_test.go
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package idp
```
ThaminduDilshan: Shall we move this to the existing service_test.go file?

backend/internal/idp/service.go
```diff
@@ -45,22 +48,91 @@ type IDPServiceInterface interface {
...
+		transactioner:     transactioner,
+		logger:            log.GetLogger().With(log.String(log.LoggerKeyComponentName, "IdPService")),
+		uuidGenerator:     utils.GenerateUUIDv7,
 	}
 }
 
+// validateAttributeConfiguration validates the IDP's attribute configuration: a required default entity type, and
+// for each entity type's attributes a valid claim-mapping shape with every local (target) claim a
+// non-credential attribute defined in that entity type's schema. No-op when no profile is configured.
+func (is *idpService) validateAttributeConfiguration(ctx context.Context, idp *IDPDTO) *serviceerror.ServiceError {
```
ThaminduDilshan: Shall we have this some where below public struct methods? It's easy to read when there're public struct methods on top

backend/internal/authn/oauth/service.go
sadilchamishka: Done — now wraps the IDP error in the authn domain (client → `ErrorClientErrorWhileRetrievingIDP` via `CustomServiceError`, server → `InternalServerError`), matching the existing pattern in `GetOAuthClientConfig`; no longer leaks the IDP error code. (a281e33ef)

backend/internal/authn/oidc/service.go
sadilchamishka: Done — same wrapping applied here (`authncm.ErrorClientErrorWhileRetrievingIDP` for client errors, `InternalServerError` for server errors). (a281e33ef)

backend/internal/idp/attribute_configuration_test.go
sadilchamishka: Done — merged into `service_test.go` (same `IDPServiceTestSuite`) and removed the separate file. (a281e33ef)

backend/internal/idp/service.go
sadilchamishka: Done — moved `validateAttributeConfiguration` below the public `idpService` methods. (a281e33ef)

backend/internal/authn/oauth/service.go
```diff
@@ -328,9 +328,22 @@ func (s *oAuthAuthnService) Authenticate(ctx context.Context, idpID, code string
...
+		logger.Error(ctx, "Error while retrieving identity provider", log.String("errorCode", svcErr.Code),
+			log.String("description", svcErr.ErrorDescription.DefaultValue))
+		return nil, &serviceerror.InternalServerError
+	}
+
 	result := &common.FederatedAuthResult{
 		Sub:    sub,
-		Claims: userInfo,
+		Claims: idp.ApplyAttributeMappings(userInfo, idp.GetAttributeMappings(idpDTO)),
 	}
```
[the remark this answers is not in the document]
sadilchamishka: Added — `if idpDTO == nil { return nil, &ErrorInvalidIDP }` after the error check, mirroring the `GetOAuthClientConfig` guard in this file (and an equivalent nil guard returning `InternalServerError` in the OIDC service). (f483d9381)

backend/internal/idp/utils.go
```diff
@@ -46,6 +46,111 @@ func GetPropertyValue(properties []cmodels.Property, name string) string {
...
+		}
+		if seenTargets[m.LocalAttribute] {
+			return serviceerror.CustomServiceError(ErrorInvalidAttributeConfiguration, core.I18nMessage{
+				Key: "error.idpservice.attribute_configuration_duplicate_target_description",
+				DefaultValue: fmt.Sprintf(
+					"local attribute name '%s' appears as a mapping target more than once", m.LocalAttribute),
+			})
+		}
+		seenTargets[m.LocalAttribute] = true
+	}
```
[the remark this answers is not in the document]
sadilchamishka: Fixed — `validateAttributeMappingShape` now trims `LocalAttribute` before the duplicate-target check (and uses the trimmed value in the message), consistent with the emptiness check, so whitespace variants no longer bypass dedup. Added a `DuplicateTargetWhitespaceVariant` test. (7eaee6156)

api/idp.yaml
```diff
@@ -521,6 +575,65 @@ components:
...
+          $ref: '#/components/schemas/EntityTypeResolution'
+        entityTypeAttributeMappings:
+          type: array
+          description: Per-entity-type attribute mapping profiles.
+          items:
+            $ref: '#/components/schemas/EntityTypeAttributeMapping'
```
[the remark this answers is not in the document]
sadilchamishka: Not adding `maxItems` for now — the server-side validation enforces no array size cap, so a spec-only limit would create the same spec/server drift flagged earlier on this PR, and adding a server-enforced cap without a product-defined limit would be speculative. Happy to add it on both sides once a concrete bound is defined.

api/idp.yaml
sadilchamishka: Fallback to the previous model.
```
"attributeConfiguration": {
  "userTypeResolution": {
    "default": "Person"
  },
  "userTypeAttributeMappings": [
    {
      "userType": "Person",
      "attributes": [
        { "externalAttribute": "given_name",    "localAttribute": "firstName" },
        { "externalAttribute": "address.email", "localAttribute": "email" }
      ]
    }
  ]
}
```

We can bring agentType related things if there are use cases for AI agents has use cases.
Then we don't have the exposing of entityType concern.

ThaminduDilshan (approved):

PR #3141, opened 2026-06-04
Add correlation IDs to the logs with context-aware logging methods

UdeshAthukorala:
### Purpose
Resolves the inability to trace all log entries belonging to a single request. The correlation ID infrastructure already existed (`CorrelationIDMiddleware` extracts/generates a trace ID per request and stores it in the request context), but application log statements never consumed it — only 2 call sites in the backend attached the trace ID, so logs could not be filtered per request when investigating errors.

This PR introduces context-aware logging so that every log entry made with a request context is automatically stamped with `trace_id=<correlation-id>`, and migrates the `idp` package as the first consumer.

### Approach

**1. Context-aware logging in `internal/system/log` (the foundation):**

  - A `contextHandler` (`slog.Handler` decorator) wraps the existing `TextHandler`. Its `Handle(ctx, record)` reads the trace ID from the context and appends it to every record — one central choke point, so no call site can "forget" the trace ID once it passes a context.
  - New `InfoContext` / `DebugContext` / `WarnContext` / `ErrorContext`methods on the `Logger` wrapper, mirroring the stdlib `slog` API, deliver the request context to the handler at log time.
  - The handler reads `ctx.Value(TraceIDKey)` directly instead of`sysContext.GetTraceID`, because the latter generates a new random UUID when the context has none — which would stamp unrelated records (startup, lifecycle) with distinct, misleading trace IDs.

**2. Deprecation of the non-context methods (migration driver):**

  - `Info`/`Debug`/`Warn`/`Error` are marked `// Deprecated:` pointing to their replacements. End state (follow-up PRs): once all ~1,861 call sites are migrated, the deprecated methods will be deleted so an uncorrelated log statement becomes a compile error.
  - ⚠️  **Maintainer attention — lint config change:** static check's SA1019 flags every call to a deprecated method as a lint error (~1,861 hits repo-wide). A temporary `exclude-rules` entry in `backend/.golangci.yml` suppresses SA1019 **only for these four methods** so CI stays green during the incremental migration, while IDEs still surface strikethrough deprecation warnings. The exclusion will be removed right before deleting the methods, so any stragglers fail loudly.

 **3. First package migration — `internal/idp`:**

  - All handler/service/store/cache log calls switched to the `*Context` variants using the `ctx` already in scope.
  - `ctx` threaded through the validation helper chain (`validateIDP` → `validateIDPProperties` → `ensureOpenIDScope` → `createAndAppendProperty`).
  - Declarative-resource validation passes `context.Background()` with a justification comment (runs at server startup — no request exists).
  - One intentional leftover: `ValidateResource` in `declarative_resource.go` logs via the deprecated method, as its signature is owned by the `declarativeresource` package interface — it will migrate with that package.

Remaining packages will be migrated in follow-up PRs (one package/group per PR), after which the deprecated methods get deleted.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2038

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Improved context-aware logging to include request trace IDs for better observability.
  * Responses now mask secret property values when returning created identity-provider data.

* **Tests**
  * Added tests for context-aware logging behavior.
  * Added tests for secret-property handling in IDP requests.
  * Added tests for cache error-handling to ensure resilient behavior.

* **Chores**
  * Temporary lint suppression to ease logging migration.

### DIFF
--- a/backend/internal/system/log/log.go
+++ b/backend/internal/system/log/log.go
@@ -41,6 +41,37 @@ type Logger struct {
 	internal *slog.Logger
 }
 
+// contextHandler decorates a slog.Handler to add the trace ID (correlation ID)
+// from the context to every log record, when present. The trace ID is set in
+// the request context by the CorrelationIDMiddleware.
+type contextHandler struct {
+	slog.Handler
+}
+
+// Handle adds the trace ID from the context to the record before delegating
+// to the wrapped handler. The trace ID is only added when it is actually
+// present in the context; sysContext.GetTraceID is not used here as it
+// generates a new ID when absent, which would stamp unrelated log records
+// with distinct, misleading trace IDs.
+func (h *contextHandler) Handle(ctx context.Context, record slog.Record) error {
+	if ctx != nil {
+		if traceID, ok := ctx.Value(sysContext.TraceIDKey).(string); ok && traceID != "" {
+			record.AddAttrs(slog.String(LoggerKeyTraceID, traceID))
+		}
+	}
+	return h.Handler.Handle(ctx, record)
+}
+
+// WithAttrs preserves the context decoration on loggers derived via With.
+func (h *contextHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	return &contextHandler{Handler: h.Handler.WithAttrs(attrs)}
+}
+
+// WithGroup preserves the context decoration on loggers derived via WithGroup.
+func (h *contextHandler) WithGroup(name string) slog.Handler {
+	return &contextHandler{Handler: h.Handler.WithGroup(name)}
+}
+
 // GetLogger creates and returns a singleton instance of the logger.
 func GetLogger() *Logger {
 	once.Do(func() {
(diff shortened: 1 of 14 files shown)


backend/internal/system/log/log.go
```diff
@@ -109,25 +140,61 @@ func (l *Logger) IsDebugEnabled() bool {
...
 
 // Error logs an error message with custom fields.
+//
+// Deprecated: Error does not propagate the request context, so log entries
+// miss the trace ID (correlation ID). Use ErrorContext instead. Refs #2038.
 func (l *Logger) Error(msg string, fields ...Field) {
 	l.internal.Error(msg, convertFields(fields)...)
 }
 
+// InfoContext logs an informational message with custom fields, automatically
```
ThaminduDilshan:
```suggestion
// InfoWithContext logs an informational message with custom fields, automatically
```

UdeshAthukorala: Addressed with: https://github.com/thunder-id/thunderid/pull/3141/commits/3425b9793246c12d99c3826c379a92b3dc85ce43

rajithacharith (approved):

Issue #3131, opened 2026-06-04
M2M token `sub` claim should be the resource entity ID, not the OAuth client ID


sahandilshan:
## Description

Currently, when a token is issued via the `client_credentials` grant, the `sub` claim in the M2M token is set to the **OAuth Client ID**. This applies to all M2M token consumers — agents, applications, etc.

The `sub` claim should instead be the **resource entity ID** — i.e., the identity of the resource that the OAuth client represents:

- If the OAuth client belongs to an **agent**, `sub` should be the agent entity ID
- If the OAuth client belongs to an **application**, `sub` should be the application ID

The current behavior makes it impossible for downstream resource servers to identify which resource entity a token belongs to without additional lookups.

## Related

- Discussion #3128

PR #3142, opened 2026-06-04
Add implicit on-behalf-of act claim to authorization code user tokens

### LINKED ISSUE #3133: Support implicit OBO `act` claim in user tokens issued through an application
## Description

When a user logs in to an application via the authorization code flow, the issued token should optionally include the `act` claim identifying the application (or agent) that mediated the login. This represents an implicit on-behalf-of (OBO) relationship — the user authenticated **through** the application.

### Expected behavior

The token issued to the user should look like:

```json
{
  "sub": "user-123",
  "act": {
    "sub": "<application-or-agent-entity-id>"
  }
}
```

### Default behavior

- **Agents:** Enabled by default — when a user logs in through an agent's OAuth client, the `act` claim is automatically included
- **Applications:** Disabled by default — applications can opt-in to append the `act` claim in user tokens

## Related

- Discussion #3128
- #3131

sahandilshan:
### Purpose

Adds an implicit on-behalf-of (OBO) `act` claim to user tokens issued through the OAuth2 authorization code flow.

When a user logs in through an OAuth client that belongs to an agent or application, the issued **access token** now optionally carries an `act` claim identifying the mediating entity, turning a normal login into an implicit OBO token without a separate token exchange step:

```json
{ "sub": "<user-id>", "act": { "sub": "<agent-or-application-entity-id>" } }
```

- **Agents** include the claim **by default**.
- **Applications** opt in via a new `include_act_claim` flag on the OAuth profile (default `false`).
- `act.sub` is the resource **entity ID** (consistent with #3131), not the OAuth client ID.
- The claim is added to **access tokens only** — never the ID token.
- `client_credentials` and `token_exchange` token shapes are unaffected.

Part of the agent identity story discussed in #3128.

### Approach

- **Entity resolution:** the resolved `OAuthClient` now carries the entity category (typed `entity.EntityCategory`) and the `IncludeActClaim` flag. A helper `ShouldAppendActorClaim()` returns true for agent clients, or application clients that opted in.
- **authorization_code grant:** when `ShouldAppendActorClaim()` is true, the access token build context is given `ActorClaims{Sub: oauthApp.ID}`. The existing claim builder emits the RFC 8693 §4.1 nested `act: { sub }` shape. The ID token path is untouched.
- **Frozen at issuance + refresh:** the decision is captured at issuance as an internal `act_sub` marker embedded in the refresh token. The `refresh_token` grant replays the `act` claim from that stored marker rather than re-evaluating the live flag — so toggling `include_act_claim` never changes tokens already issued (verified both directions).
- **Config surface:** `include_act_claim` is exposed only through the management API for applications (no Console UI, no DCR self-registration in this PR). Documented in `api/application.yaml`.

### Related Issues
- Fixes #3133
- Related: #3131, #3128

## Summary by CodeRabbit

* **New Features**
  * Added includeActClaim option for OAuth applications (default: false) to control whether an implicit actor (act) claim is added to issued access tokens.
  * Agent clients always receive actor claims; application clients receive them only when includeActClaim is enabled.
  * Actor subject is preserved and replayed across authorization-code and refresh-token flows so access tokens reflect stored actor info.
  * ID tokens remain unchanged and never include the act claim.


api/application.yaml
```diff
@@ -1384,6 +1389,14 @@ components:
...
           example: false
           default: false
+        includeActClaim:
+          type: boolean
+          description: >-
+            When true, access tokens issued through this client's authorization_code flow include
+            an implicit on-behalf-of `act` claim identifying the application entity. Agent clients
+            always include it regardless of this setting. Default false.
+          example: false
+          default: false
```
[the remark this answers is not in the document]
sahandilshan: We should not document this as of now. This is mostly document when the UI comes for this

api/application.yaml
```diff
@@ -1384,6 +1389,14 @@ components:
           description: Whether DPoP-bound access tokens (RFC 9449) are required for this application.
           example: false
           default: false
+        includeActClaim:
+          type: boolean
+          description: >-
+            When true, access tokens issued through this client's authorization_code flow include
+            an implicit on-behalf-of `act` claim identifying the application entity. Agent clients
```
ThaminduDilshan: Do we need to mention about agent clients at this level? This is the application payload right? Agent's will share the same inbound auth configs, but it's a internal detail

api/application.yaml
```diff
@@ -1483,6 +1496,14 @@ components:
           description: Whether DPoP-bound access tokens (RFC 9449) are required for this application.
           example: false
           default: false
+        includeActClaim:
+          type: boolean
+          description: >-
+            When true, access tokens issued through this client's authorization_code flow include
```
ThaminduDilshan, line 1501: Same as previous comment

backend/internal/application/service.go
```diff
@@ -743,10 +744,12 @@ func toProcessedDTO(
 		}
 
 		var ouID string
+		var entityCategory entity.EntityCategory
 		if e != nil {
 			ouID = e.OUID
+			entityCategory = entity.EntityCategory(e.Category)
```
ThaminduDilshan: Isn't this a fixed category always for the apps? Do we need to read it from the entity?
Not a hard suggestion. Fine to read it from entity as well since anyway we've hardcoded entity type during app creation

backend/internal/inboundclient/service.go
```diff
@@ -325,17 +326,17 @@ func (s *inboundClientService) resolveClientID(entityID string) string {
 	if s.entityProvider == nil {
 		return ""
 	}
-	entity, epErr := s.entityProvider.GetEntity(entityID)
```
ThaminduDilshan, line 328: Any particular reason to rename this? 

Agree that common go pattern is to have shorter vaiables/ terms, etc. But I prefer to keep `entity` over `e`. Sometimes it is harder to interpret the meaning with just `e`. 
@coderabbitai WDYT? Provide short answer

backend/internal/application/service.go
sahandilshan: Valid suggestion and fixed with d90d2f3438bc1f3ed673eb7b575a772f3e5dd3a7

api/application.yaml
sahandilshan: Fixed with d90d2f3438bc1f3ed673eb7b575a772f3e5dd3a7

api/application.yaml
sahandilshan: Fixed with d90d2f3438bc1f3ed673eb7b575a772f3e5dd3a7

ThaminduDilshan (approved):

PR #3175, opened 2026-06-05
Update README.md parsing logic to extract introduction before Architecture section

KaveeshaPiumini:
### Purpose
This pull request updates the logic in the release workflow for extracting the introduction section from the `README.md` file. Instead of extracting everything before the first `---` divider, it now extracts everything before the `## Architecture` section. This makes the extraction more robust and less dependent on markdown divider formatting.

- Workflow improvement:
  * In `.github/workflows/release-builder.yml`, changed the method for extracting the introduction from `README.md` to use the `## Architecture` heading as the delimiter instead of the first `---` divider.

## Summary by CodeRabbit

* **Chores**
  * Adjusted release notes extraction to stop at the "## Architecture" heading.
  * Bumped release metadata across the project (charts, sample package.json files, and version.txt) from 0.42.0 → 0.43.0.


darshanasbg (dismissed):

pavinduLakshan (commented):
Wouldn't the ideal fix be to add the dividers back? If it brings too many visual clutter, we could consider a static html comment as a divider so that the release note automation can use that to extract the relevant section/s from the readme for the release note.

ThaminduDilshan (dismissed):

KaveeshaPiumini:
> Wouldn't the ideal fix be to add the dividers back? If it brings too many visual clutter, we could consider a static html comment as a divider so that the release note automation can use that to extract the relevant section/s from the readme for the release note.

Adding dividers to the README.md makes it not appealing. Hence, let's proceed with this approach as of now.

PR #3194, opened 2026-06-08
Make redirect_uri at token endpoint conditional per RFC 6749 §4.1.3

### LINKED ISSUE #3198: Token endpoint rejects requests without redirect_uri even when the authorize request omitted it (RFC 6749   §4.1.3)
### Description
  `backend/internal/oauth/oauth2/granthandlers/authorization_code.go:92` unconditionally rejects token requests
  that omit `redirect_uri`. Per [RFC 6749 §4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3),
  `redirect_uri` at the token endpoint is REQUIRED only when it was included in the authorization request, and
  their values MUST be identical.

  Clients that legitimately omit `redirect_uri` at the authorize endpoint (allowed by [RFC 6749 
  §3.1.2.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.1.2.3) when exactly one redirect URI is
  registered) cannot exchange their authorization code, because the token endpoint always demands
  `redirect_uri`.

  ### Expected behavior
  - `redirect_uri` provided at authorize → token endpoint MUST send an identical value (unchanged).
  - `redirect_uri` omitted at authorize → token endpoint is not required to send it.

  ### Current behavior
  The token endpoint always requires `redirect_uri`, regardless of what the authorize request contained.

  ### Reproduction
  1. Register an OAuth client with exactly one `redirect_uri`.
  2. Send `GET /oauth2/authorize?client_id=...&response_type=code&scope=...` without a `redirect_uri` parameter.
  3. Authorization completes, code is issued (ThunderID falls back to the registered URI per §3.1.2.3 —
  correct).
  4. `POST /oauth2/token` with `grant_type=authorization_code&code=...&client_id=...` (also without
  `redirect_uri`).
  5. Token endpoint returns `400 invalid_request` with `"Redirect URI is required"`. Per RFC, this should
  succeed.

  ### Existing TODO in the codebase
  The file flags this at the call site:

  // TODO: Redirect uri is not mandatory when excluded in the authorize request and is valid scenario.
  //  This should be removed when supporting other means of authorization.

  ### Related

  A fix is proposed in #3194

HesandaLiyanage:
RFC 6749 §4.1.3 requires redirect_uri at the token endpoint only when the authorization request included it. The authorization_code grant handler previously rejected any token request that omitted redirect_uri, even for clients that did not send one at the authorize endpoint (a valid case when exactly one redirect URI is registered).

Track whether redirect_uri was provided at the authorize endpoint on OAuthParameters and AuthorizationCode, persist it through the request and code stores, and gate the token-endpoint match check on it. The resolved URI continues to drive the actual client redirect.

### Purpose
  The authorization_code grant handler currently rejects any token request that
  omits `redirect_uri`, even when the original authorization request didn't
  include one — a valid scenario per [RFC 6749 
  §4.1.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)
  when the client has exactly one registered redirect URI. This fixes that
  behaviour and resolves the `// TODO: Redirect uri is not mandatory when
  excluded in the authorize request` comment in
  `backend/internal/oauth/oauth2/granthandlers/authorization_code.go`.

  RFC 6749 §4.1.3:
  > `redirect_uri`: REQUIRED, if the "redirect_uri" parameter was included in
  > the authorization request as described in Section 4.1.1, and their values
  > MUST be identical.

  ### Approach
  - Add `RedirectURIProvided bool` to `OAuthParameters` and `AuthorizationCode`
    to track whether the authorization request actually included `redirect_uri`.
  - Set it in the standard authorize path (`authz/service.go`) and the PAR path
    (`par/service.go`) *before* the registered-URI fallback runs, so the
    effective `RedirectURI` continues to drive the real client redirect while
    the original "was it provided?" signal is preserved.
  - Persist the flag through both the auth request store and the auth code
    store (JSON serialisation in `auth_req_store.go` and `auth_code_store.go`,
    with new keys `redirect_uri_provided` in both `store_constants.go` and
    `auth_code_store.go`).
  - Drop the unconditional pre-fetch rejection in `ValidateGrant`.
  - Gate the post-fetch match check on `code.RedirectURIProvided` in
    `validateAuthorizationCode` — when `true`, identical values are still
    required; when `false`, the comparison is skipped entirely.

  Behaviour matrix:

  | `redirect_uri` at authorize | `redirect_uri` at token | Result |
  |---|---|---|
  | present | matches | success (unchanged) |
  | present | mismatch | `invalid_grant` (unchanged) |
  | present | omitted | `invalid_grant` (new — was previously rejected at `ValidateGrant`) |
  | omitted | omitted | success (new — was previously rejected) |
  | omitted | any value | success (skipped per spec) |

  ### Why this validation lives in the grant handler

  Client authentication (including the `WWW-Authenticate` challenge for failed
  `invalid_client` cases) is handled in `clientauth/middleware.go`, upstream of
  the grant handler. This change does not touch that — it adds no new logic to
  client auth or to the token endpoint's HTTP layer.

  The `redirect_uri` check in this PR is a different concern: it validates the
  token request payload against the persisted authorization code. That
  validation cannot run earlier in the pipeline because the authorization code
  is loaded inside the grant handler (`retrieveAndValidateAuthCode` →
  `validateAuthorizationCode`). The check is added alongside the existing
  cross-field validations (client ID, resource, expiry) in
  `validateAuthorizationCode`, which is the established place for code-bound
  request validation.

  ### Related Issues
  - Fixes #3198 

  ### Related PRs
  - N/A

  ### Checklist
      - [x] Unit Tests
      - [ ] Integration Tests

  ### Security checks
  - [x] Followed secure coding standards in [WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/la
  test/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)

## Summary by CodeRabbit

* **New Features**
  * Persistently track whether a redirect URI was supplied in the original authorization request.

* **Bug Fixes**
  * Token exchange enforces redirect-URI validation only when a redirect URI was originally provided (matches RFC 6749 §4.1.3).

* **Tests**
  * Updated and added tests to validate redirect-URI semantics across request storage, authorization codes, and grant handling.

* **Documentation**
  * Clarified docs to explain when redirect-URI is required and how matching is enforced.


ThaminduDilshan:
Is there a github issue for this? General practice is to have a github issue created, get it assigned to you and only start the implementation after that's being assigned. Maintainers will review the issue and assign it after confirming this is a bug or needed improvement. 

Since the PR is already created, shall we create an issue and get it assigned to you?

HesandaLiyanage:
@ThaminduDilshan  Thanks for the heads-up. No, there's no issue. I came across the  
// TODO: Redirect uri is not mandatory when excluded in the authorize request 
comment at
  backend/internal/oauth/oauth2/granthandlers/authorization_code.go:92 
while reading through the OAuth code path to learn the codebase, and went straight to a PR because the TODO was self-contained and explicit about the missing RFC 6749 §4.1.3 behavior.

Totally fine to do this properly. Happy to file an issue retroactively describing the bug and link it back to this PR. Want me to go ahead and create it? I'll keep the issue-first flow in mind for future contributions.

HesandaLiyanage:
@ThaminduDilshan Filed as #3198 . Updated the PR description with the link.

backend/internal/oauth/oauth2/authz/model.go
```diff
@@ -35,10 +35,16 @@ type OAuthMessage struct {
...
-	CodeID              string
-	Code                string
-	ClientID            string
-	RedirectURI         string
+	CodeID   string
+	Code     string
+	ClientID string
+	// RedirectURI is the effective redirect URI bound to this authorization code.
+	RedirectURI string
+	// RedirectURIProvided records whether redirect_uri was present in the original
```
thiva-k: Let's remove these verbose comments

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -270,10 +270,14 @@ func (as *authorizeService) handleStandardAuthorizationRequest(
 	}
 
 	// Construct authorization request context.
+	// RedirectURIProvided records whether the original request carried redirect_uri,
```
thiva-k: Let's remove these comments

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -22,9 +22,16 @@ import "github.com/thunder-id/thunderid/internal/system/utils"
...
-	State               string
-	ClientID            string
-	RedirectURI         string
+	State    string
+	ClientID string
+	// RedirectURI is the effective redirect URI used to send the user back to the
+	// client. It may have been resolved from the client's registered URIs when the
+	// authorization request omitted the redirect_uri parameter.
+	RedirectURI string
+	// RedirectURIProvided records whether the redirect_uri parameter was present in
```
thiva-k: Let's remove this too

HesandaLiyanage:
@thiva-k  Done , comments removed.

HesandaLiyanage:
@thiva-k CodeRabbit flagged that this change makes the "always validated with exact
  matching" wording in docs/content/guides/guides/applications/application-settings.mdx:42
  slightly imprecise. That line is now only true when the authorize request included redirect_uri.

  Happy to update that sentence and add a short note in
  docs/content/guides/guides/protocols/oauth-oidc/authorization-code.mdx
  about the conditional behavior. Should that land in this PR, or as a follow-up so this stays scoped to the code change?

backend/internal/oauth/oauth2/granthandlers/authorization_code.go
```diff
@@ -89,14 +89,10 @@ func (h *authorizationCodeGrantHandler) ValidateGrant(ctx context.Context, token
...
 
-	// TODO: Redirect uri is not mandatory when excluded in the authorize request and is valid scenario.
-	//  This should be removed when supporting other means of authorization.
-	if tokenRequest.RedirectURI == "" {
-		return &model.ErrorResponse{
-			Error:            constants.ErrorInvalidRequest,
-			ErrorDescription: "Redirect URI is required",
-		}
-	}
+	// Per RFC 6749 §4.1.3, redirect_uri at the token endpoint is REQUIRED only if
```
ThaminduDilshan: Shall we remove this comment? Same comment has been repeated again in lines 300-303

ThaminduDilshan:
> @thiva-k CodeRabbit flagged that this change makes the "always validated with exact matching" wording in docs/content/guides/guides/applications/application-settings.mdx:42 slightly imprecise. That line is now only true when the authorize request included redirect_uri.
> 
> Happy to update that sentence and add a short note in docs/content/guides/guides/protocols/oauth-oidc/authorization-code.mdx about the conditional behavior. Should that land in this PR, or as a follow-up so this stays scoped to the code change?

Let's update the documentation in this PR

HesandaLiyanage:
@ThaminduDilshan Done. Removed the duplicate RFC comment and trimmed a few related ones in the tests for consistency. Also updated application-settings.mdx and authorization-code.mdx to reflect the conditional behavior.

ThaminduDilshan (approved):
Thanks for your contribution

thiva-k (approved):

PR #3206, opened 2026-06-08
Update usages of Authenticated User

ThumulaPerera:
### Purpose

Decouple the flow engine from the internal representation of an authenticated user. Previously, authentication results carried explicit identity fields (userID, userType, ouID, token) that leaked provider-specific knowledge into every layer of the system. This commit makes authentication results opaque so that the flow engine doesn't need to know how a user is represented — only that one exists. This is a prerequisite for supporting pluggable/external authentication providers.

### Approach

1. **Replaced the `AuthnResult` model with a token-based design** — Instead of flat identity fields, the result now carries opaque `EntityReferenceToken` / `AttributeToken` values alongside optional resolved `EntityReference` / `Attributes`. This lets the provider decide what form the result takes.

2. **Simplified `AuthUser` to match** — Removed the internal `providerKey` map and identity fields. `AuthUser` now just holds the four token/resolved pairs. `IsAuthenticated()` checks that both entity reference and attribute data exist in some form (token or resolved).

3. **Added lazy entity reference resolution** — New `GetEntityReference` method on the manager. Instead of every executor extracting user identity directly, they call through the manager which resolves the token on demand via the provider.

4. **Removed `AuthenticatedUser` from the flow layer** — Deleted the field from `NodeContext`, `NodeResponse`, and `ExecutorResponse`. All executors now work exclusively through `AuthUser` and the manager interface.

5. **Moved runtime data to a dedicated map** — Authentication side-effects (like federated auth claims) are returned as `RuntimeAttributes` rather than being baked into the identity model.

6. **Updated all executors, tests, and mocks** to use the new signatures and stop accessing identity fields directly.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2361

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Refactor**
  * Authentication now returns a unified token+claims result and a single AuthUser model; executors and flows consistently use the authn-provider for user resolution, attributes, and placeholder/runtime enrichment.
* **Bug Fixes**
  * Added handling and localized messages for ambiguous-user cases to surface clear client errors and descriptions.


backend/internal/flow/common/model.go
```diff
@@ -94,35 +93,33 @@ type Prompt struct {
...
+	Status         ExecutorStatus             `json:"status"`
+	Inputs         []Input                    `json:"inputs,omitempty"`
+	AdditionalData map[string]string          `json:"additionalData,omitempty"`
+	RedirectURL    string                     `json:"redirectUrl,omitempty"`
+	RuntimeData    map[string]string          `json:"runtimeData,omitempty"`
+	ForwardedData  map[string]interface{}     `json:"forwardedData,omitempty"`
+	Assertion      string                     `json:"assertion,omitempty"`
+	Error          *serviceerror.ServiceError `json:"error,omitempty"`
+	AuthUser       authnprovidermgr.AuthUser  `json:"-"`
 }
```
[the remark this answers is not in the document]
ThumulaPerera: This is not a user facing change

backend/internal/authnprovider/common/model.go
```diff
@@ -24,49 +24,18 @@ type AuthnMetadata struct {
...
-	UserType string `json:"userType"`
-
-	AttributesResponse *AttributesResponse `json:"attributeResponse,omitempty"`
+	AuthenticatedClaims AuthenticatedClaims `json:"authenticatedClaims,omitempty"`
+	// EntityReferenceToken can be nil, iff entity reference is included
+	EntityReferenceToken any              `json:"entityReferenceToken"`
+	EntityReference      *EntityReference `json:"entityReference,omitempty"`
+	// AttributeToken can be nil, iff attribute values are included
+	AttributeToken any                 `json:"attributeToken"`
+	Attributes     *AttributesResponse `json:"attributes,omitempty"`
```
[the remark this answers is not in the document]
ThumulaPerera: Docs will be added with completed feature

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -418,64 +446,31 @@ func (a *authAssertExecutor) appendComputedAttributes(
...
-			if attrResp != nil {
-				attrs[attrName] = attrResp.Value
-			}
-		}
-	}
-	return attrs, nil
-}
-
 // getUserAttributesFromUserProvider retrieves user attributes from the user provider.
 func (a *authAssertExecutor) getUserAttributesFromUserProvider(ctx context.Context, userID string) (
```
sadilchamishka: `getUserAttributesFromUserProvider` looks dead now — after this PR removes the `getUserAttributesFromAuthnProvider` path, the only remaining callers are in `auth_assert_executor_test.go`. Production attribute fetching now goes exclusively through `authnProvider.GetUserAttributes`. Per the project rule on deleting dead code cleanly, this method (and its test cases) should be removed unless it's intended for an upcoming change. Note `entityProvider` itself is still used by `fetchAllUserGroups`, so only this method is dead, not the field.

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -71,61 +83,184 @@ func (p *defaultAuthnProvider) Authenticate(
...
+
+	return attributesResponse, nil
 }
 
 func (p *defaultAuthnProvider) resolveCredentials(
 	ctx context.Context,
 	identifiers, credentials map[string]interface{},
-) (*credentialOutcome, *serviceerror.ServiceError) {
+) (*authnResult, *serviceerror.ServiceError) {
+	if provisioned, ok := credentials["provisionedEntityID"]; ok {
```
sadilchamishka, line 253: `"provisionedEntityID"` is a magic string that forms a cross-package contract: the provisioning executor writes this exact key (`provisioning_executor.go:220`) and this provider reads it. A typo on either side would silently fall through to the other credential branches and break provisioning with no compile-time error. Worth promoting to a single shared exported constant referenced from both sites.

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -71,61 +83,184 @@ func (p *defaultAuthnProvider) Authenticate(
...
+	parsedToken, ok := attributeToken.(map[string]interface{})
+	if !ok {
+		return nil, p.logAndReturnServerError(ctx, "Invalid token format")
+	}
+	entityID := ""
+	if idVal, ok := parsedToken[UserAttributeUserID]; ok {
+		entityID, _ = idVal.(string)
+	}
+	if entityID == "" {
+		identifiedEntityID, identifyErr := p.entitySvc.IdentifyEntity(ctx, parsedToken)
```
sadilchamishka: The `token[userID]` → `IdentifyEntity` → `GetEntity` resolution block is now duplicated almost verbatim three times: here in `GetAttributes`, in `GetEntityReference` (~line 149), and in `Authenticate` (~line 100). Extracting a private helper (e.g. `resolveEntityID(ctx, token)` returning the entity or a mapped client error) would remove the triplication and keep the error mapping consistent across all three entry points.

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -71,61 +83,184 @@ func (p *defaultAuthnProvider) Authenticate(
...
+		if identifyErr != nil {
+			if errors.Is(identifyErr, entity.ErrEntityNotFound) {
+				return nil, newClientError(authnprovidercm.ErrorCodeUserNotFound,
+					"User not found", "No user found matching the provided entity reference token")
```
sadilchamishka: Error description says "entity reference token" but this is `GetAttributes`, which operates on the attribute token. Minor, but the message will be misleading in logs/responses for an attribute fetch. Suggest "...matching the provided attribute token".

backend/internal/authnprovider/manager/manager.go
```diff
@@ -47,108 +51,196 @@ func newAuthnProviderManager(p provider.AuthnProviderInterface) AuthnProviderMan
...
+			m.logger.Debug(ctx, "disambiguation requested but current user is not authenticated")
+			return authUser, nil, &ErrorAuthenticationFailed
+		}
+
+		if authUser.entityReferenceToken == nil {
+			m.logger.Debug(ctx, "disambiguation requested but current user's entity reference token is missing")
+			return authUser, nil, &ErrorAuthenticationFailed
+		}
+
+		if entityRefToken, ok := authUser.entityReferenceToken.(map[string]interface{}); !ok || entityRefToken == nil {
```
sadilchamishka: This disambiguation block re-asserts `authUser.entityReferenceToken.(map[string]interface{})` four separate times (lines 73, 79, 86). Assert once into a local `entityRefToken` and reuse it — clearer and avoids the repeated type assertions. Also note this whole branch keys on `credentials["sub"]` and short-circuits before ever calling the provider; the comment flags it as temporary, but it's worth a follow-up issue since it bakes provider-specific (`sub`) knowledge back into the manager that the rest of this PR is trying to remove.

backend/internal/authnprovider/manager/manager.go
```diff
@@ -47,108 +51,196 @@ func newAuthnProviderManager(p provider.AuthnProviderInterface) AuthnProviderMan
...
+		}
+		valStr, ok := val.(string)
+		if !ok || valStr == "" {
+			m.logger.Debug(ctx, "disambiguation requested but userID is missing or invalid in identifiers")
+			return authUser, nil, &ErrorAuthenticationFailed
+		}
+		userID := valStr
+		authUser = AuthUser{
+			entityReferenceToken: map[string]interface{}{
+				"userID": userID,
```
sadilchamishka: Uses the literal `"userID"` here while `userAttributeUserID` is defined at the top of this file (line 32) and used a few lines above at line 93. Use the constant for both keys to avoid drift.

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: This issue is there for all credential types. That will be addressed separately in a PR to update the interface to support explicitly providing the credential type rather than through a key in credentials map.

backend/internal/authnprovider/manager/manager.go
ThumulaPerera: Fixed

backend/internal/authnprovider/manager/manager.go
ThumulaPerera: Fixed re-assertion.

Regarding the comment on, the whole branch keys on credentials["sub"] and bypasses the provider, that is intensional and temporary until there is a better approach

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: Fixed

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: Done

backend/internal/flow/executor/auth_assert_executor.go
ThumulaPerera: Fixed

sadilchamishka (dismissed):

backend/internal/flow/executor/ou_executor.go
```diff
@@ -82,15 +90,39 @@ func (o *ouExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse, e
...
+			if svcErr.Code != authnprovidermgr.ErrorUserNotFound.Code &&
+				svcErr.Code != authnprovidermgr.ErrorAmbiguousUser.Code {
+				logger.Error("Failed to get entity reference", log.String("errorCode", svcErr.Code),
+					log.String("errorDescription", svcErr.ErrorDescription.DefaultValue))
+				execResp.Status = common.ExecFailure
+				execResp.Error = &ErrFailedToIdentifyUser
+				return execResp, nil
+			}
+			logger.Debug("User not found or ambiguous, proceeding with OU creation")
+		}
```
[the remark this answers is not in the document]
ThumulaPerera: provisioning executor allows ambiguous user to be provisioned when cross OU provisioning is enabled. in such scenarios, OU executor should also create the target OU. Hence this behaviour is correct

backend/internal/flow/core/node.go
```diff
@@ -73,7 +73,7 @@ func (n *node) ShouldExecute(ctx *NodeContext) bool {
 		return true
 	}
 
-	resolvedKey := ResolvePlaceholder(ctx, n.condition.Key)
+	resolvedKey := ResolvePlaceholder(ctx, n.condition.Key, nil, nil, nil)
 	return resolvedKey == n.condition.Value
```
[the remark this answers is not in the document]
ThumulaPerera: The comment says "pass the actual authn provider and executor response (the same objects used elsewhere)." Those objects don't exist at this call site. 

In practice we have not come across a usecase for conditionally executing a node based on userID or ouID. This will be treated as a known limitation

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -796,3 +841,46 @@ func (p *provisioningExecutor) assignToRole(
...
+	// Filter allowed user types to only those with self-registration enabled
+	selfRegEnabledSchemas := make([]entitytype.EntityType, 0)
+	for _, userType := range ctx.Application.AllowedUserTypes {
+		entityType, svcErr := p.entityTypeService.GetEntityTypeByName(ctx.Context,
+			entitytype.TypeCategoryUser, userType)
+		if svcErr != nil {
+			// TODO: throw an error here
+			logger.Warn("Failed to retrieve entity type for allowed user type", log.String("userType", userType))
+			return nil
+		}
```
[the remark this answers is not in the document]
ThumulaPerera: Fixed

backend/internal/flow/executor/sms_auth_executor.go
```diff
@@ -313,24 +327,46 @@ func (s *smsOTPAuthExecutor) resolvePhoneInput(ctx *core.NodeContext, fallback c
...
+				if strVal, isString := attr.Value.(string); isString && strVal != "" {
+					mobileNumber = strVal
+				}
+			}
+		}
 	}
+
+	// Store the resolved mobile number in runtime data for downstream use.
+	ctx.RuntimeData[mobileAttrName] = mobileNumber
+
```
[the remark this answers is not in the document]
ThumulaPerera: Fixed

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -181,7 +181,49 @@ func (a *authAssertExecutor) generateAuthAssertion(ctx *core.NodeContext, logger
...
+		for attrName, attrResp := range attrResp.Attributes {
+			if attrResp != nil {
+				fetchedAttributes[attrName] = attrResp.Value
+			}
+		}
+	}
+
+	resolvedAttributes, attrErr := a.resolveUserAttributes(ctx, requiredAttributes, fetchedAttributes,
+		entityRef.EntityID, entityRef.EntityType, entityRef.OUID)
 	if attrErr != nil {
```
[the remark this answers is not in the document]
ThumulaPerera: GetEntityReference ensures entityRef is non-nil if svcErr is nil.

backend/internal/authn/oidc/service_test.go
```diff
@@ -462,3 +448,198 @@ func (suite *OIDCAuthnServiceTestSuite) TestValidateIDTokenWithoutJWKSEndpoint()
...
+			"sub":   "different_user",
+			"email": "other@example.com",
+		}, nil)
+
+	result, svcErr := suite.service.Authenticate(context.Background(), testOIDCIDPID, "auth_code")
+	suite.Nil(svcErr)
+	suite.NotNil(result)
+	suite.Equal("1234567890", result.Token["sub"])
+	suite.Nil(result.AuthenticatedClaims["email"])
+}
```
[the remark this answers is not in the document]
ThumulaPerera: This is an existing test

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -186,13 +198,40 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
...
+	execResp.AuthUser = authUser
+	if err != nil {
+		logger.ErrorWithContext(ctx.Context, "Failed to authenticate user for provisioning",
+			log.String("error", err.ErrorDescription.DefaultValue))
+		execResp.Status = common.ExecFailure
+		execResp.Error = &ErrUserAuthFailed
+	}
+	for key, value := range authenticatedClaims {
+		execResp.RuntimeData[key] = systemutils.ConvertInterfaceValueToString(value)
+	}
```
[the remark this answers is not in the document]
ThumulaPerera: Fixed

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -141,6 +151,8 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
 	if execResp.Status == common.ExecFailure && (execResp.Error == nil || execResp.Error.Code != ErrUserNotFound.Code) {
 		return execResp, nil
 	}
+	// clear execResp set by IdentifyUser
+	execResp.Status = ""
 	if userID != nil && *userID != "" {
```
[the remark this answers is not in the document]
ThumulaPerera: Fixed

backend/internal/flow/executor/oidc_auth_executor.go
```diff
@@ -158,15 +154,29 @@ func (o *oidcAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
...
+		authUser, attributes, err := o.authnProvider.GetUserAttributes(ctx.Context, nil, nil, execResp.AuthUser)
+		execResp.AuthUser = authUser
+		if err != nil {
+			logger.Warn("Failed to fetch user attributes for authenticated user, proceeding without attributes")
+		} else {
+			for key, value := range attributes.Attributes {
+				existingCtxUserAttributes[key] = value
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ThumulaPerera: Fixed

backend/internal/authn/common/model.go
```diff
@@ -84,6 +84,11 @@ type FederatedAuthCredential struct {
 	Code    string
 }
 
+// OpenID4VPCredential carries the presentation session state to the authn provider.
+type OpenID4VPCredential struct {
```
ThaminduDilshan, line 88: Unrelated question to this PR, but shouldn't the openid4vp have a authn service?

backend/internal/authn/magiclink/service.go
```diff
@@ -33,13 +33,8 @@ import (
...
 
-// MagicLinkAuthnResult holds the result of a magic link authentication attempt.
-// When InternalEntity is nil the token was valid but no local user exists yet (registration flow);
-// VerifiedIdentifiers then carries the proven channel identifiers (e.g. email).
-type MagicLinkAuthnResult struct {
-	InternalEntity      *entityprovider.Entity
-	VerifiedIdentifiers map[string]interface{}
-}
+// userAttributeUserID is the user attribute key for user ID in authentication results.
+const userAttributeUserID = "userID"
```
ThaminduDilshan: Why do we need to hardcode user attribute here?

sadilchamishka (dismissed):

backend/internal/authn/common/model.go
ThumulaPerera: Ideally it should. I had a chat with @thiva-k on this as well. I think we might be able to move the package itself inside authn package. Not planning to get that into this PR

backend/internal/authn/magiclink/service.go
ThumulaPerera: It is used as the fallback subject attribute

backend/internal/authnprovider/manager/error_constants.go
```diff
@@ -54,6 +54,21 @@ var (
 		},
 	}
 
+	// ErrorGetEntityReferenceClientError is returned when the underlying provider rejects the
+	// entity reference fetch due to a client-side reason.
+	ErrorGetEntityReferenceClientError = serviceerror.ServiceError{
```
ThaminduDilshan: Shall we keep error constants in ascending order? Having them randomly could lead to having duplicate error codes in future

backend/internal/authnprovider/manager/manager.go
```diff
@@ -26,8 +26,12 @@ import (
 	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
 	"github.com/thunder-id/thunderid/internal/system/i18n/core"
 	"github.com/thunder-id/thunderid/internal/system/log"
+	systemutils "github.com/thunder-id/thunderid/internal/system/utils"
 )
 
+const userAttributeUserID = "userID"
```
ThaminduDilshan: These constants are defined in multiple packages. Is it expected?

backend/internal/authnprovider/manager/interface.go
```diff
@@ -31,7 +31,9 @@ type AuthnProviderManagerInterface interface {
 	AuthenticateUser(ctx context.Context, identifiers, credentials map[string]interface{},
```
ThaminduDilshan, line 31: This should be coming from existing implementation. But what is the expectation of passing a `AuthUser` and getting back a `AuthUser` in response?

backend/internal/authnprovider/manager/manager.go
```diff
@@ -47,108 +51,197 @@ func newAuthnProviderManager(p provider.AuthnProviderInterface) AuthnProviderMan
 func (m *authnProviderManager) AuthenticateUser(ctx context.Context, identifiers, credentials map[string]interface{},
 	requestedAttributes *authnprovidercm.RequestedAttributes,
 	metadata *authnprovidercm.AuthnMetadata,
-	authUser AuthUser) (AuthUser, *AuthnBasicResult, *serviceerror.ServiceError) {
-	result, svcErr := m.provider.Authenticate(ctx, identifiers, credentials, metadata)
+	authUser AuthUser) (AuthUser, authnprovidercm.AuthenticatedClaims, *serviceerror.ServiceError) {
+	if sub, ok := credentials["sub"]; ok {
+		// Temporary handling of disambiguation after a federated authentication step.
```
ThaminduDilshan, line 53: Didn't get the disambiguation handling inside AuthenticateUser method when the sub is present in credentials. What's the thinking behind this?

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -71,61 +83,179 @@ func (p *defaultAuthnProvider) Authenticate(
...
+				return nil, newClientError(authnprovidercm.ErrorCodeAmbiguousUser,
+					"Ambiguous user", "Multiple users found matching the provided "+tokenLabel)
+			}
+			return nil, p.logAndReturnServerError(ctx, "Failed to identify entity after authentication",
+				log.String("error", identifyErr.Error()))
+		}
+		entityID = *identifiedEntityID
+	}
+
+	entityResult, getErr := p.entitySvc.GetEntity(ctx, entityID)
```
ThaminduDilshan, line 241: Is it expected to perform this via two entity svc invocations? First to get entity id from token and then to get attributes from the entity id

backend/internal/authn/service.go
```diff
@@ -138,15 +138,15 @@ func (as *authenticationService) AuthenticateWithCredentials(ctx context.Context
...
+	newAuthUser, _, svcErr := as.authnProvider.AuthenticateUser(ctx, identifiers, credentials, nil, nil,
 		authnprovidermgr.AuthUser{})
 	if svcErr != nil {
 		return nil, as.mapCredentialsAuthnError(ctx, svcErr, logger)
 	}
 
-	if basicResult == nil {
-		logger.Error(ctx, "Credentials authenticate response is nil")
-		return nil, &serviceerror.InternalServerError
+	newAuthUser, entityRef, svcErr := as.authnProvider.GetEntityReference(ctx, newAuthUser)
```
ThaminduDilshan, line 147: Here also I'm not clear on the requirement to do this with two authn provider invocations

backend/internal/authnprovider/common/model.go
```diff
@@ -133,3 +102,17 @@ type AttributeResponse struct {
...
+// EntityReference contains the reference to an entity.
+type EntityReference struct {
+	EntityID       string `json:"entityId"`
+	EntityCategory string `json:"entityCategory"`
+	EntityType     string `json:"entityType"`
+	OUID           string `json:"ouId"`
+}
+
+// GetAttributesMetadata holds metadata used when retrieving entity attributes.
+type GetAttributesMetadata struct {
```
ThaminduDilshan, line 115: GetAttributesMetadata is a verb right? It doesn't sound right for a struct/ data structure

backend/internal/flow/core/executor.go
```diff
@@ -124,61 +129,94 @@ func (e *executor) ValidatePrerequisites(ctx *NodeContext, execResp *common.Exec
...
+		log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+
+	if val, ok := ctx.RuntimeData[userAttributeUserID]; ok && val != "" {
+		return val
+	}
+
+	if authnProvider != nil && ctx.AuthUser.IsAuthenticated() {
+		authUser, entityRef, err := authnProvider.GetEntityReference(ctx.Context, ctx.AuthUser)
+		if err != nil {
+			logger.Warn(ctx.Context,
```
ThaminduDilshan: Is it intentional to have a warn log here? Similar logs are treated as debug few code blocks above

backend/internal/flow/core/utils.go
```diff
@@ -83,6 +95,23 @@ func ResolvePlaceholder(ctx *NodeContext, value string) string {
 	})
 }
 
+func fetchContextUserRef(
```
ThaminduDilshan, line 99: Shall we add a method comment?

backend/internal/flow/core/executor.go
```diff
@@ -124,61 +129,94 @@ func (e *executor) ValidatePrerequisites(ctx *NodeContext, execResp *common.Exec
 		return true
 	}
 
+	authenticatedUserAttributes := make(map[string]string)
+	if authnProvider != nil && ctx.AuthUser.IsAuthenticated() {
+		authUser := ctx.AuthUser
+		providerAuthUser, entityRef, err := authnProvider.GetEntityReference(ctx.Context, ctx.AuthUser)
```
ThaminduDilshan: Can use the authUser variable

backend/internal/flow/core/executor.go
```diff
@@ -124,61 +129,94 @@ func (e *executor) ValidatePrerequisites(ctx *NodeContext, execResp *common.Exec
...
+		if err != nil {
+			logger.Debug(ctx.Context,
+				"Failed to get entity reference for authenticated user, proceeding without user id")
+		} else {
+			authUser = providerAuthUser
+			if entityRef.EntityID != "" {
+				authenticatedUserAttributes[userAttributeUserID] = entityRef.EntityID
+			}
+		}
+		providerAuthUser, authAttributes, err := authnProvider.GetUserAttributes(ctx.Context, nil, nil, authUser)
```
ThaminduDilshan, line 145: Previously prerequisites are validated against what's in the context. But with this, can the prerequisites be satisfied even with user profile data?

If so is that correct? Even though prerequisites are satisfied here, if the usages are not updated to check authn user attributes when searching for such attributes, this could ended up in runtime error.

On the other hand, if this is not needed, then what's the requirement to pass authn provider to the method signature and perform this operation?

backend/internal/authnprovider/manager/interface.go
```diff
@@ -31,7 +31,9 @@ type AuthnProviderManagerInterface interface {
...
 		requestedAttributes *authnprovidercm.RequestedAttributes,
 		metadata *authnprovidercm.AuthnMetadata,
-		authUser AuthUser) (AuthUser, *AuthnBasicResult, *serviceerror.ServiceError)
+		authUser AuthUser) (AuthUser, authnprovidercm.AuthenticatedClaims, *serviceerror.ServiceError)
+	GetEntityReference(ctx context.Context, authUser AuthUser) (
```
ThaminduDilshan, line 35: I'm also not clear on the expectation of this new method

backend/internal/authn/common/model.go
ThaminduDilshan: @coderabbitai can you create a follow up issue for this?

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -138,31 +268,52 @@ func (p *defaultAuthnProvider) resolveCredentials(
...
+	if vpCred, ok := credentials["openid4vp"]; ok {
+		return p.authenticateWithOpenID4VP(ctx, vpCred)
+	}
 	if userID, ok := identifiers["userID"]; ok && userID != "" {
 		return p.authenticateByUserID(ctx, userID, credentials)
 	}
 	return p.authenticateByIdentifiers(ctx, identifiers, credentials)
 }
 
+func (p *defaultAuthnProvider) AuthenticateForProvisioning(
```
ThaminduDilshan: Shall we add a method comment?

backend/internal/authnprovider/provider/default_authn_provider.go
```diff
@@ -279,41 +411,67 @@ func (p *defaultAuthnProvider) authenticateWithMagicLink(
...
-				AttributesResponse:        buildAttributesResponse(authResult.VerifiedIdentifiers),
-			},
-		}, nil
+	return &authnResult{
+		token:               result.Token,
+		authenticatedClaims: result.AuthenticatedClaims,
+	}, nil
+}
+
+func (p *defaultAuthnProvider) authenticateWithOpenID4VP(
```
ThaminduDilshan, line 430: Shall we add method comment? Check other places as well

backend/internal/authnprovider/provider/default_authn_provider_test.go
```diff
@@ -235,106 +335,110 @@ func (suite *DefaultAuthnProviderTestSuite) TestAuthenticate_EmptyUserID_FallsBa
...
 	suite.NotNil(result)
-	suite.True(result.IsExistingUser)
-	suite.Equal("u1", result.UserID)
+	suite.Nil(result.EntityReference)
+	suite.NotNil(result.EntityReferenceToken)
+	suite.NotNil(result.AttributeToken)
 }
 
-func (suite *DefaultAuthnProviderTestSuite) TestAuthenticate_OTP_UserNotFound() {
+//nolint:dupl
```
ThaminduDilshan: Can't we use table driven test pattern rather than skipping lint?

backend/internal/authnprovider/provider/default_authn_provider_test.go
```diff
@@ -532,3 +1170,322 @@ func (suite *DefaultAuthnProviderTestSuite) TestAuthenticate_MagicLink_MissingTo
...
+	result, err := provider.Authenticate(context.Background(), nil, credentials, nil)
+
+	suite.Nil(result)
+	suite.NotNil(err)
+	suite.Equal(authnprovidercm.ErrorCodeInvalidRequest, err.Code)
+}
+
+// --- Passkey success/error tests using inline mock ---
+
+type mockPasskeyService struct {
```
ThaminduDilshan: If possible shall we use mockery to generate mocks rather than writing custom mock implementation

backend/internal/authnprovider/provider/default_authn_provider_test.go
```diff
@@ -532,3 +1170,322 @@ func (suite *DefaultAuthnProviderTestSuite) TestAuthenticate_MagicLink_MissingTo
...
+	result, err := provider.Authenticate(context.Background(), nil, credentials, nil)
+
+	suite.Nil(result)
+	suite.NotNil(err)
+	suite.Equal(serviceerror.InternalServerError.Code, err.Code)
+}
+
+// --- Helper ---
+
+func strPtr(s string) *string {
```
ThaminduDilshan: AFAIR this is something not required with go 1.26+ syntax

backend/internal/flow/executor/consent_executor.go
```diff
@@ -95,43 +95,69 @@ func (e *consentExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorRespon
...
+	if !execResp.AuthUser.IsAuthenticated() {
+		execResp.Status = common.ExecFailure
+		execResp.Error = &ErrUserNotAuthenticated
+		return execResp, nil
+	}
+
+	authUser, entityRef, svcErr := e.authnProvider.GetEntityReference(ctx.Context, execResp.AuthUser)
+	execResp.AuthUser = authUser
+	if svcErr != nil {
+		logger.Error(ctx.Context, "Failed to get entity reference from AuthUser", log.Any("error", svcErr))
```
ThaminduDilshan: No need to both log and return. We should return go error if this needs to be treated as server error. Logging is done at task execution node.
Should set exec resp error field only if this is a client error.

backend/internal/flow/executor/magic_link_executor.go
```diff
@@ -355,7 +338,9 @@ func (m *magicLinkExecutor) executeVerify(ctx *core.NodeContext) (*common.Execut
 		}
 		return execResp, fmt.Errorf("failed to verify magic link: %s", svcErr.ErrorDescription.DefaultValue)
 	}
-	execResp.AuthUser = newAuthUser
+	for key, value := range authenticatedClaims {
```
ThaminduDilshan, line 341: Okay now we're writing all the authenticated user claims to runtime data right? This will bloat the runtime data unnecessarily.

What is the requirement to do this? Is this something actually required?

backend/internal/flow/executor/oauth_executor.go
```diff
@@ -249,53 +255,34 @@ func (o *oAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
...
-	if err != nil {
-		return err
-	}
-	if execResp.Status == common.ExecFailure {
-		return nil
-	}
-	if contextUser == nil {
-		logger.Error(ctx.Context, "Failed to resolve context user after OAuth authentication")
-		return errors.New("unexpected error occurred while resolving user")
+	if ctx.FlowType == common.FlowTypeAuthentication {
```
ThaminduDilshan: Shall we use switch-case instead?

backend/internal/flow/executor/magic_link_executor.go
ThaminduDilshan: This implementation is there in multiple executors. Better if we can look for an alternative solution

backend/internal/flow/executor/oidc_auth_executor.go
```diff
@@ -178,62 +188,44 @@ func (o *oidcAuthExecutor) ProcessAuthFlowResponse(ctx *core.NodeContext,
...
-	if err != nil {
-		return err
-	}
-	if execResp.Status == common.ExecFailure {
-		return nil
-	}
-	if contextUser == nil {
-		logger.Error(ctx.Context, "Failed to resolve context user after OIDC authentication")
-		return errors.New("unexpected error occurred while resolving user")
+	if ctx.FlowType == common.FlowTypeAuthentication {
```
ThaminduDilshan: Shall we use switch-case instead?

backend/internal/flow/executor/ou_executor.go
```diff
@@ -82,15 +90,39 @@ func (o *ouExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse, e
...
 	}
 
+	if execResp.AuthUser.IsAuthenticated() {
+		// Check if the user already has an entity reference (existing user).
+		// If so, skip OU creation as the user already belongs to an OU.
+		authUser, entityRef, svcErr := o.authnProvider.GetEntityReference(ctx.Context, execResp.AuthUser)
+		if svcErr != nil {
+			if svcErr.Code != authnprovidermgr.ErrorUserNotFound.Code &&
+				svcErr.Code != authnprovidermgr.ErrorAmbiguousUser.Code {
+				logger.Error(ctx.Context, "Failed to get entity reference", log.String("errorCode", svcErr.Code),
```
ThaminduDilshan: Same as previous comment. No need to log here and currently this is also handled as a client error

backend/internal/flow/executor/ou_executor.go
```diff
@@ -163,7 +195,56 @@ func (o *ouExecutor) getOrganizationUnitRequest(ctx *core.NodeContext) (ou.Organ
...
+		return "", nil
+	}
+
+	selfRegEnabledSchemas := make([]entitytype.EntityType, 0)
+	for _, userType := range ctx.Application.AllowedUserTypes {
+		et, svcErr := o.entityTypeService.GetEntityTypeByName(ctx.Context,
+			entitytype.TypeCategoryUser, userType)
+		if svcErr != nil {
+			logger.Error(ctx.Context,
+				"Failed to retrieve entity type for allowed user type",
```
ThaminduDilshan: No need to log at executor level

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -186,48 +200,109 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
...
+
+	// Set the auto-provisioned flag if it's a user auto provisioning scenario
+	if ctx.FlowType == common.FlowTypeAuthentication {
+		execResp.RuntimeData[common.RuntimeKeyUserAutoProvisioned] = dataValueTrue
 	}
 
 	return execResp, nil
 }
 
+func (p *provisioningExecutor) authenticateProvisionedUser(ctx *core.NodeContext, userID string,
```
ThaminduDilshan, line 219: Shall we add a method comment?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -186,48 +200,109 @@ func (p *provisioningExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorR
...
+	execResp *common.ExecutorResponse, logger *log.Logger) {
+	credential := map[string]interface{}{
+		"provisionedEntityID": userID,
+	}
+	authUser, authenticatedClaims, err := p.authnProvider.AuthenticateUser(ctx.Context, nil, credential,
+		nil, nil, execResp.AuthUser)
+	if !authUser.IsAuthenticated() {
+		if err != nil {
+			logger.Error(ctx.Context, "Failed to authenticate user for provisioning",
```
ThaminduDilshan: No need to log and return error

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -613,6 +662,31 @@ func (p *provisioningExecutor) createUserInStore(nodeCtx *core.NodeContext,
 	return retEntity, nil
 }
 
+func (p *provisioningExecutor) getTargetEntityRef(ctx *core.NodeContext) (*entityRef, error) {
```
ThaminduDilshan, line 660: Shall we add method comment?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -723,3 +797,50 @@ func splitTrimmed(s string) []string {
...
+	}
+
+	// Filter allowed user types to only those with self-registration enabled
+	selfRegEnabledSchemas := make([]entitytype.EntityType, 0)
+	for _, userType := range ctx.Application.AllowedUserTypes {
+		entityType, svcErr := p.entityTypeService.GetEntityTypeByName(ctx.Context,
+			entitytype.TypeCategoryUser, userType)
+		if svcErr != nil {
+			logger.Error(ctx.Context,
+				"Failed to retrieve entity type for allowed user type",
```
ThaminduDilshan: Same previous comment about logging applies here too. Check other places as well

backend/internal/flow/executor/sms_auth_executor.go
```diff
@@ -100,16 +99,32 @@ func (s *smsOTPAuthExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorRes
...
 
 	// Determine the executor mode
 	switch ctx.ExecutorMode {
 	case ExecutorModeSend:
-		if !s.ValidatePrerequisites(ctx, execResp) {
-			logger.Debug(ctx.Context, "Prerequisites not met for SMS OTP authentication executor")
-			return execResp, nil
+		mobileNumber, err := s.getUserMobileFromContext(ctx, execResp)
+		if err != nil {
+			logger.Error(ctx.Context, "Failed to retrieve user mobile number from context",
```
ThaminduDilshan: Verify same logging comment in this file too. No need to log at executor level

backend/internal/flow/executor/utils.go
```diff
@@ -127,26 +126,28 @@ func isCrossOUProvisioningAllowed(ctx *core.NodeContext) bool {
 	return false
 }
 
+func isAllowRegistrationWithExistingUserRuntimeFlagSet(ctx *core.NodeContext) bool {
```
ThaminduDilshan, line 131: Add method comment

backend/internal/authnprovider/manager/manager.go
ThumulaPerera: Most cases were intensional to keep the packages decoupled. But few occurrences were updated to reuse the same constant (eg: sms_otp, magiclink)

backend/internal/authnprovider/manager/interface.go
ThumulaPerera: AuthUser maintains the complete authentication state. Fields in AuthUser are unexported (private). Only authnprovider.manager is expected to modify those. But in a given flow, AuthUser needs to be persisted in the flow context. 

Executor passes the current AuthUser from flow context -> Manager updates the AuthUser upon authentication -> Manager passes back the updated AuthUser to be persisted in the flow context

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: Yes. This is the contract exposed by the entity service. GetEntity needs the entityID. When we don't know the entityID, we need to get the entityID by calling IdentityEntity

backend/internal/authn/service.go
ThumulaPerera: Here also, the requirement to do this with 2 invocations because of the contract exposed by authnprovider manager. 

The need for the contract to expose this separately is,
- Entity reference does not expose any PII. Hence authn providers must resolve this when requested.
- Attributes are PII. Authn providers can opt to not send attributes until consent is obtained

backend/internal/authnprovider/common/model.go
ThumulaPerera: This was meant to indicate, MetadataForGetAttributes. 

I'm +1 to rename but this is not a new addition in this PR. just moved the location. 

https://github.com/thunder-id/thunderid/discussions/3192 also discusses some changes to this. Might cause a confusion if we change the name now. Shall we discuss further and do the name change in a separate PR ?

backend/internal/flow/core/executor.go
ThumulaPerera: Previously, executors fetched the user profile and added user profile data to the context after authentication. Behaviourally, there is no difference between this approach and that.

Usages of ctx.AuthenticatedUser.Attributes are updated to call `authnProvider.GetUserAttributes`.

backend/internal/authnprovider/manager/interface.go
ThumulaPerera: - Entity reference does not expose any PII. Hence authn providers must resolve this when requested. i.e. this can be called anywhere in the flow (irrespective of the placement of a consent execuotr) and the providers are bound to resolve

- Attributes are PII. Authn providers can opt to not send attributes until consent is obtained. i.e. provider may impose the condition that this can only be called after consent is obtained.

backend/internal/flow/executor/consent_executor.go
ThumulaPerera: In server error case, logging and returning the error seems to be the approach followed in surrounding places. Logging is done to preserve the svcErr details it seems. Updated this to the same pattern. We can discuss and fix all places separately if needed

backend/internal/flow/executor/magic_link_executor.go
ThumulaPerera: authenticatedClaims are not the same as user profile data. 

Each authn service will return a set of claims asserted about the user by means of authentication.

Eg:
SMS OTP -> mobile number
Federated -> all federated atributes
Most others -> userID

backend/internal/flow/executor/sms_auth_executor.go
ThumulaPerera: fixed

backend/internal/authnprovider/manager/error_constants.go
ThumulaPerera: Fixed

backend/internal/authnprovider/manager/manager.go
ThumulaPerera: After disambiguation, federated_auth_resolver needs to update the Authentication State of the User. Previously this was done by setting ctx.AuthenticatedUser.IsAuthenticated = true. But with the new architecture, executors cannot update the authentication state. They can only call AuthenticateUser method in manager. Manager will update the state.

backend/internal/flow/core/executor.go
ThumulaPerera: Not intensional. Updated to debug

backend/internal/flow/core/utils.go
ThumulaPerera: Added

backend/internal/flow/core/executor.go
ThumulaPerera: Fixed

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: Added

backend/internal/authnprovider/provider/default_authn_provider.go
ThumulaPerera: Added method level comments for all similar helpers

backend/internal/authnprovider/provider/default_authn_provider_test.go
ThumulaPerera: Updated

backend/internal/authnprovider/provider/default_authn_provider_test.go
ThumulaPerera: Fixed

backend/internal/authnprovider/provider/default_authn_provider_test.go
ThumulaPerera: Updated

backend/internal/flow/executor/oauth_executor.go
ThumulaPerera: Done

backend/internal/flow/executor/oidc_auth_executor.go
ThumulaPerera: Done

backend/internal/flow/executor/ou_executor.go
ThumulaPerera: fixed

backend/internal/flow/executor/ou_executor.go
ThumulaPerera: fixed

backend/internal/flow/executor/provisioning_executor.go
ThumulaPerera: added

backend/internal/flow/executor/provisioning_executor.go
ThumulaPerera: fixed

backend/internal/flow/executor/provisioning_executor.go
ThumulaPerera: added

backend/internal/flow/executor/provisioning_executor.go
ThumulaPerera: fixed

backend/internal/flow/executor/utils.go
ThumulaPerera: Added

ThaminduDilshan (dismissed):

ThaminduDilshan (dismissed):

sadilchamishka (approved):

PR #3224, opened 2026-06-09
[WIP] Block direct HTTP flow initiation for authorization_code grant type apps

### LINKED ISSUE #3215: Implement HTTP guard in loadNewContext() to block direct flow initiation for authorization_code grant type applications
### Description
Currently, `POST /flow/execute` is a fully open endpoint — any caller can initiate a new flow for any application. As part of the secured-by-default initiation model, applications configured with the `authorization_code` grant type (including browser SPAs) must have direct HTTP new-flow initiation blocked.

## Implementation Notes
- The guard must be placed in `loadNewContext()` in the flow execution service, not in `initContext()`. Placing it in `initContext()` would incorrectly block the OAuth component's internal `InitiateFlow()` service call.
- The guard should check the application's configured grant type and reject requests where the grant type is `authorization_code` and no existing `executionId` is present in the request (i.e., it is a new-flow initiation, not a continuation).
- Continuation requests carrying a valid `executionId` and challenge token must be unaffected.
- Return a clear error response (e.g., `403`) with a message indicating that direct flow initiation is not permitted for this application type.

## Acceptance Criteria
- A new `POST /flow/execute` request for an `authorization_code` application is rejected with a `403`.
- A continuation request (with `executionId`) for the same application proceeds normally.
- The OAuth component's internal `InitiateFlow()` call is unaffected.

Malith-19:
### Purpose

Applications configured with the `authorization_code` grant type (browser SPAs, gate) must have their flows initiated server-side by the OAuth component via `/authorize`, not via a direct `POST /flow/execute` call. Currently the endpoint is fully open — any caller can initiate a new flow for any application. This PR adds a guard that blocks new-flow initiation for such applications while leaving continuation requests and the OAuth component's internal `InitiateFlow()` path unaffected.

Closes #3215
Design discussion: #2744

---

`POST /flow/execute` now rejects new-flow initiation requests for applications configured with the `authorization_code` grant type. Previously, any caller could initiate a flow for any application regardless of its grant type.

#### 💥 Impact

Any client that was directly calling `POST /flow/execute` to start a new flow for an `authorization_code` application (i.e. submitting a request with no `executionId`) will now receive a `403 Forbidden` response with error code `FES-1011`.

This affects:
- Browser SPAs calling `/flow/execute` directly without going through the OAuth `/authorize` endpoint first.
- Any integration test or tooling that initiates flows for redirect-based apps via direct API calls.

Continuation requests (those that include a valid `executionId`) are **not affected** — they continue to work as before.

#### 🔄 Migration Guide

For `authorization_code` applications, flow execution must be initiated through the OAuth authorization endpoint:

1. Direct the user to `GET /oauth2/authorize` with the appropriate parameters (client_id, redirect_uri, response_type=code, etc.).
2. The OAuth component will internally call `InitiateFlow()` and redirect the browser to the gate/SPA with an `executionId` in the query parameters.
3. The SPA then calls `POST /flow/execute` with the provided `executionId` to continue the flow — this continuation path is unaffected by this change.

Apps using grant types other than `authorization_code` (e.g. `client_credentials`) are not affected.

---

### Approach

**Where the guard lives — `loadNewContext()`, not `initContext()`.**

The `Execute()` entry point branches on whether the request is a new flow or a continuation:

```
Execute()
  ├── empty executionId  →  loadNewContext()   ← guard added here
  │                              └── initContext()
  └── non-empty executionId  →  loadPrevContext()   (unaffected)

InitiateFlow() [OAuth component]  →  initContext()   (unaffected)
```

Placing the guard in `initContext()` would also fire when the OAuth component calls `InitiateFlow()` internally, which is the *legitimate* initiation path for auth-code apps and must not be blocked. `loadNewContext()` is only reached via direct HTTP calls with no existing `executionId`, so it is the correct and minimal interception point.

**The guard — `checkDirectFlowInitiationAllowed()`**

Calls `GetOAuthProfileByEntityID` to retrieve the app's OAuth profile and checks whether `authorization_code` is in its `GrantTypes` slice. Returns `ErrorDirectFlowInitiationNotPermitted` (new code `FES-1011`) if so.

Fail-open cases (guard passes silently):
- `appID` is empty (app-independent flows like user onboarding).
- `ErrInboundClientNotFound` — the app has no OAuth profile, so no grant type to enforce.
- `nil` OAuth profile returned — same reasoning.

Fail-closed case:
- Any unexpected store/infra error returns a 500 rather than accidentally allowing the request through.

**New `ForbiddenErrorType` + HTTP 403 mapping**

The existing error type system only mapped `ClientErrorType → 400` and `ServerErrorType → 500`. A new `ForbiddenErrorType` was added to `serviceerror` and the flow handler updated to map it to HTTP 403, keeping the status code concern out of the guard itself.

**Test fixes**

The guard exposed two categories of tests that were unintentionally using `authorization_code` apps for direct flow execution:

_Integration tests:_ `CreateApplication` in `testutils/api_utils.go` omitted `grantTypes` from its default OAuth config, causing the server to apply its default of `["authorization_code"]` to every test app. Fixed by explicitly setting `"grantTypes": []` in the default config. `acr_options_flow_test.go` also hardcoded `["authorization_code"]` and a paired `["code"]` response type for an app used only to test ACR flow-graph behaviour (not the OAuth redirect path) — both were removed.

_E2E tests:_ The test infrastructure utilities (`admin-api-auth.ts`, `mfa-setup.ts`) called `/flow/execute` directly using `SAMPLE_APP_ID`, which points to the "Sample App" — an `authorization_code` OAuth app. Fixed by adding a dedicated native app (`E2E Admin Native App`, `client_id: thunder_e2e_admin_client`, `grant_types: []`) to the vanilla sample's `thunderid-config.yaml`. Both utilities now resolve this app at runtime by calling `GET /applications` (a public endpoint) and looking up by the well-known `client_id`, falling back to `SAMPLE_APP_ID` for backward compatibility. No CI workflow changes were required.

### Related Issues
- Fixes #3215
- Design: https://github.com/thunder-id/thunderid/discussions/2744

### Related PRs
- N/A

### Checklist
- [ ] Documentation provided.
- [x] Tests provided.
    - [x] Unit Tests
    - [ ] Integration Tests
- [x] Breaking changes.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Direct HTTP initiation of authentication flows is blocked for applications using the OAuth2 authorization_code grant; attempts return HTTP 403 with a localized error message.

* **Bug Fixes**
  * Service error handling now maps forbidden errors to HTTP 403 consistently.

* **Samples**
  * Sample config adds a native admin app (client_credentials) for direct execution in E2E scenarios.

* **Tests**
  * Expanded tests and helpers to validate direct-init rules and use the dedicated native admin client.

### DIFF
--- a/backend/internal/flow/executor/register.go
+++ b/backend/internal/flow/executor/register.go
@@ -287,7 +287,7 @@ func registerBuiltInExecutors(reg ExecutorRegistryInterface, deps ExecutorDepend
 		}
 	}
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "ExecutorRegistry"))
-	logger.Info(context.Background(), "Registered built-in flow executors",
+	logger.Debug(context.Background(), "Registered built-in flow executors",
 		log.Int("count", len(resolved)),
 		log.Any("executors", resolved))
 	return nil


backend/internal/flow/flowexec/service.go
```diff
@@ -199,6 +204,36 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
 	return engineCtx, nil
 }
 
+// checkDirectFlowInitiationAllowed returns an error if the application's grant type does not
+// permit direct HTTP flow initiation. Applications configured with the authorization_code grant
+// type must have their flows initiated by the OAuth component, not via a direct HTTP call.
+func (s *flowExecService) checkDirectFlowInitiationAllowed(ctx context.Context, appID string,
```
rajithacharith: Shouldn't we have a separate attribute for this? We are allowing accessing if the supported grant types does not contains GrantTypeAuthorizationCode. Is it the correct behaviour we need to get?

backend/internal/flow/flowexec/service.go
```diff
@@ -199,6 +204,36 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
...
+		}
+		logger.ErrorWithContext(ctx, "Failed to retrieve OAuth profile for flow initiation guard",
+			log.String("appID", appID), log.Error(err))
+		return &serviceerror.InternalServerError
+	}
+	if oauthProfile == nil {
+		return nil
+	}
+
+	for _, gt := range oauthProfile.GrantTypes {
```
Malith-19: Check if we have capability to directly check the grant type rather than iterating.

backend/internal/flow/flowexec/service.go
rajithacharith: I propose 

executeReqHandleMethod, err := getExecuteRequestHandleMethod(appId)

if executeReqHandleMethod == BREAK {
   breakTheRequest
}
if executeReqHandleMethod == CHECK_FOR_APIKEY {
   validateAPIKey
}

rajithacharith:
Seems this logic will be breaking the experience in refreshing the gate app on login prompt

backend/internal/flow/flowexec/service.go
Malith-19: > Shouldn't we have a separate attribute for this? We are allowing accessing if the supported grant types does not contains GrantTypeAuthorizationCode. Is it the correct behaviour we need to get?

We only have the `template` attribute at the moment which have values like, `spa`, `mobile`, `server` or `m2m` but along with that value we can't identify the redirect based flows. But may be we can introduce a new property for this. WDYT?

backend/internal/flow/flowexec/service.go
Malith-19: Fixed.

Malith-19:
> Seems this logic will be breaking the experience in refreshing the gate app on login prompt

Hi @rajithacharith, good catch, Yes this will be happening as gate app tries to initiate the flow from the flow API with the refresh. 

To solve this we can check the session storage for the execution id and challenge token before starting a new flow.

PR #3238, opened 2026-06-09
Migrate backend log calls to context-aware logging method

### LINKED ISSUE #2038: Add correlation Ids to the logs
### Current Limitation

Currently when there is an specific error in thunder, there is no way to filter out the logs only for that path. An id needs to be added to the logs in a single routine.

### Suggested Improvement

An id needs to be added to the logs in a single routine.

### Version

_No response_

UdeshAthukorala:
### Purpose

Add correlation IDs (request trace IDs) to backend application logs by migrating the remaining deprecated, non-context log calls to the context-aware `*WithContext` logging methods. This is the follow-up to the foundation PR (#3196), which introduced the `contextHandler` and the `InfoWithContext` / `DebugWithContext` / `WarnWithContext` / `ErrorWithContext` methods. With this PR, **every** application log statement in the backend now propagates the request context, so each log entry is stamped with the `trace_id` when one is present. Refs #2038.

This change spans **186 files** (123 non-test source files, 56 test files, 6 auto-generated mocks) across effectively every backend subsystem (`system`, `oauth`, `flow`, `authn`, `design`, `role`, `group`, `application`, `resource`, `notification`, `inboundclient`, `agent`, `user`, `ou`, `idp`, `entitytype`, `openid4vp`, `consent`, `cmd/server`).

There are **no functional or external API changes** — only logging propagation and internal signature additions of `context.Context`. The previously deprecated methods are **not** deleted in this PR (that is reserved for a later lock-in PR); after this change there are simply zero call sites using them.

### Approach

The migration was applied per call site according to the context source available, and verified against the compiler rather than text search (see Verification).

**1. Method-for-method migration.** Deprecated `Info`/`Debug`/`Warn`/`Error` calls were replaced with their `*WithContext` equivalents. To complete the method set, this PR also adds `FatalWithContext` and deprecates `Fatal` (it was the only logging method without a context-aware variant).

**2. Context source selection (per call site).**
- **Request-scoped code** — used the in-scope `ctx context.Context`, `r.Context()` for HTTP handlers, or `NodeContext.Context` / `EngineContext.Context` for flow execution.
- **Private helpers whose callers already had a context** — threaded `context.Context` as a parameter (e.g. `http_util` response writers `WriteJSONError` / `WriteSuccessResponse` / `WriteErrorResponse`, email `sendViaSMTP`, importer `normalize*ConfigForImport`, application `generateAndAssignClientID` / `resolveClientSecret`, inboundclient `buildInboundClientFromRow`).
- **Startup / lifecycle / infrastructure code** — there is no request scope, so `context.Background()` is used, each with a short rationale comment explaining why (service construction, subscriber `Initialize`/`Close`, declarative-resource loading, cache management, file-adapter I/O and its detached goroutines, server bootstrap, etc.).

**3. Threading context into the observability package.** `observability.PublishEvent` and the publisher's `Publish` now accept `context.Context`, so the synchronous publish path logs with the request trace. Because event dispatch and subscribers run in **detached goroutines** (after the request context may be cancelled), the event already carries its `TraceID`, and the async/subscriber side derives a logging context from it (`sysContext.WithTraceID(context.Background(), evt.TraceID)`). This keeps subscriber and adapter logs correlated without binding them to a possibly-cancelled request context. `event.NewEvent` intentionally keeps its `traceID string` argument because the flow engine deliberately uses the `ExecutionID` as the trace for node-level events.

**4. Mocks.** Interface signature changes (e.g. `PublishEvent`, `http_util` writers, several service interfaces) were followed by regenerating the affected mocks with `make mockery`. This also fixes a stale path in `.mockery.public.yml` that still pointed at `internal/observability` after the package was moved to `internal/system/observability`, which had been silently skipping regeneration of those mocks.

### Related Issues
- Fixes #2038

### Related PRs
- Follow-up to #3196 (introduced the `contextHandler` and the base `*WithContext` methods)

### Checklist
- [x] Manual test round performed and verified. (Confirmed `trace_id` is stamped on application logs end-to-end via a `/flow/execute` request carrying a correlation ID.)
    - [x] Unit Tests (existing unit tests updated for the new signatures; all pass)
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/cmd/server/main.go
+++ b/backend/cmd/server/main.go
@@ -55,22 +55,24 @@ var (
 )
 
 func main() {
+	// Server bootstrap/shutdown logging has no request scope, so context.Background() is used.
+	ctx := context.Background()
 	startupStartedAt := time.Now()
 	logger := log.GetLogger()
 
 	flag.String("resources", "", "Path to declarative resources YAML file")
-	serverHome := getThunderHome(logger)
+	serverHome := getThunderHome(ctx, logger)
 
-	cfg := initThunderConfigurations(logger, serverHome)
+	cfg := initThunderConfigurations(ctx, logger, serverHome)
 	if cfg == nil {
-		logger.Fatal("Failed to initialize configurations")
+		logger.FatalWithContext(ctx, "Failed to initialize configurations")
 	}
 
 	// Install the CORS allowed-origins matcher used by the HTTP middleware.
 	// Compilation errors are already surfaced by config validation; this call
 	// rebuilds the rules and installs them as the cors package singleton.
 	if err := cors.InitializeMatcher(cfg.CORS.AllowedOrigins); err != nil {
-		logger.Fatal("Failed to initialize CORS matcher", log.Error(err))
+		logger.FatalWithContext(ctx, "Failed to initialize CORS matcher", log.Error(err))
 	}
 
 	// Initialize the cache manager.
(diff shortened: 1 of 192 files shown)


backend/cmd/server/main.go
rajithacharith: All these methods are called from main() method, We can use the context initialized at that point passed throughout

backend/internal/agent/handler.go
```diff
@@ -48,25 +49,25 @@ func (h *agentHandler) HandleAgentListRequest(w http.ResponseWriter, r *http.Req
 
 	limit, offset, svcErr := parsePaginationParams(r.URL.Query())
 	if svcErr != nil {
-		writeServiceError(w, svcErr)
+		writeServiceError(r.Context(), w, svcErr)
```
rajithacharith: Let's use the context initialized in line 47

rajithacharith: Check other places as well

UdeshAthukorala: Addressed with: https://github.com/thunder-id/thunderid/compare/e866c18854948314eb196cffac7bb5ec35c328aa..634aca5d67a05acebd494338b766129a8b2c7d2c

backend/cmd/server/main.go
UdeshAthukorala: Addressed with: https://github.com/thunder-id/thunderid/compare/e866c18854948314eb196cffac7bb5ec35c328aa..634aca5d67a05acebd494338b766129a8b2c7d2c

rajithacharith (approved):

backend/internal/openid4vp/handler_test.go
```diff
@@ -352,10 +353,173 @@ func TestAPIRegisterRoutesAddsInitiateAndStatus(t *testing.T) {
...
+func (w *failingResponseWriter) Write([]byte) (int, error) {
+	w.writeCalled = true
+	return 0, errors.New("write failed")
+}
+
+func (w *failingResponseWriter) WriteHeader(code int) { w.status = code }
+
+// stubService implements OpenID4VPServiceInterface to drive handler error
+// branches that a real *service does not reach in unit tests.
+type stubService struct {
```
ThumulaPerera, line 486: Is this a pattern we usually follow?

UdeshAthukorala: You're right that the hand-written stubService doesn't match our usual convention. I've addressed this in a follow-up PR: https://github.com/thunder-id/thunderid/pull/3282

PR #3286, opened 2026-06-12
Add CIBA id_token_hint support and demo scripts for wayfinder upgrade…

### LINKED ISSUE #3212: Implement `id_token_hint` support for CIBA backchannel authentication
### Current Limitation

The CIBA implementation currently only accepts `login_hint` for user resolution. `id_token_hint` is recognized as a valid parameter by the handler but is immediately rejected with `invalid_request`. This limits clients that already hold a valid ID token for the user — for example, a client re-authenticating an already-known user — from leveraging that token as a hint, and forces them to use a plain identifier instead.

### Suggested Improvement

Implement `id_token_hint` resolution following the CIBA Core 1.0 and OIDC Core guidance:

- **User resolution** — decode the provided ID token, extract the `sub` claim, and use it for a direct entity ID lookup (bypassing attribute search since `sub` is the OP-issued stable identifier). This is both spec-aligned and more efficient than attribute-based lookup.

- **`iss` validation** — confirm the OP itself is the issuer of the ID token. Hints from a foreign issuer must be rejected.

- **`aud` validation** — confirm the ID token's audience matches the requesting client, or a client sharing the same Sector Identifier if pairwise subject identifiers are in use. This prevents cross-client hint reuse.

- **Signature verification** — make this best-effort. The CIBA spec explicitly acknowledges that key rotation may make signature verification impractical and states that implementers may skip it, relying on `iss` + `aud` as the trust path instead. Hard-failing on unverifiable signatures is more restrictive than the spec requires.

- **Expiry** — do not hard-reject on `exp`. OIDC Core guidance says the OP SHOULD accept ID tokens even after expiry when the identified user has or recently had a session. `exp` is not a reliable staleness signal for hints. A configurable staleness window (e.g. reject tokens older than N days) is preferable to a hard `exp` block.

The core user lookup path (extract `sub` → direct entity ID lookup → resolve user) is correct and spec-aligned. The main divergence from a naive implementation is relaxing signature verification and expiry handling to match what the spec actually requires rather than what a standard token validation flow would apply.

Dilusha-Madushan:
### Purpose

The CIBA bc-authorize endpoint only accepted login_hint (plain email/username string). This blocked scenarios where the caller holds an OIDC ID token for the user — common in agentic flows where an upstream OBO exchange already produced an id_token. Additionally, Thunder's validator incorrectly enforced that the id_token_hint's aud claim must contain the requesting client ID, which violates CIBA Core 1.0 §7.1 and OIDC Core — the OP must not apply client-audience enforcement when receiving a hint.

On the Wayfinder sample side, the upgrade scheduler lacked a way to lock/unlock Business class flights for demo purposes, making it difficult to reliably trigger the CIBA flow on demand.

### Approach

CIBA id_token_hint support :

Added IDTokenHint field to BackchannelAuthRequest. The handler enforces exactly one hint is provided.
Added validateIDTokenHint which checks issuer, sub presence, JWT signature, and a 30-day staleness window. aud is intentionally not checked — per spec the OP validates the token as a hint, not as a token issued to the client.
When id_token_hint is present, sub is extracted and forwarded to the flow as login_hint so the rest of the flow (including the identifying executor's loginHintAttribute lookup) works unchanged.
Identifying executor fix (identifying_executor.go):

When loginHintAttribute is set on the node, HasRequiredInputs was checking UserInputs against the node's configured input identifier (which could be anything). But buildSearchAttributes always reads from UserInputs["login_hint"]. Made the executor check login_hint presence directly in the loginHintAttribute path, making the two consistent regardless of how the node input identifier is named in the flow definition.
Added a fast path in IdentifyUser: when the search key is userID, skip the DB search and call GetEntity directly by ID.
Wayfinder sample (ai-agent/, backend/, scripts/):

Agent collects and forwards the OIDC id_token via x-id-token header so the scheduler can use it as id_token_hint.
Added lock-business-class.sh, unlock-business-class.sh, and reset-db.sh scripts for demo control, with a Demo Scripts section in the README explaining each one and the typical try-out flow.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2739
- https://github.com/thunder-id/thunderid/issues/3212

### Related PRs
- https://github.com/thunder-id/thunderid/pull/3171

## Summary by CodeRabbit

* **New Features**
  * CIBA now accepts id_token_hint as a login hint and extracts/validates user hints; backchannel flows may use id_token_hint or login_hint.
  * Flight upgrade: availability-aware, booking-driven upgrade tools, demo endpoint to trigger processing, and demo scripts to lock/unlock and reset DB.
  * Optional upgrade scheduler for automated processing.

* **Bug Fixes**
  * More tolerant JWT/ID token validation (clock skew, hint staleness) and improved user-identification handling when explicit user IDs are provided.

* **Documentation**
  * README and env examples updated for CIBA, scheduler, and demo setup.

* **Tests**
  * Added unit tests covering id_token_hint handling and user-identification paths.


backend/internal/oauth/oauth2/ciba/service.go
```diff
@@ -46,6 +47,8 @@ import (
 // cibaMaxBindingMessageLength is the maximum number of characters allowed in a binding_message.
 const cibaMaxBindingMessageLength = 256
 
+const cibaIDTokenHintStalenessSeconds = 30 * 24 * 60 * 60 // 30 days
```
sahandilshan: Is is possible for user to change this value from the deployment.toml or from the applicatino level?

backend/internal/oauth/oauth2/ciba/service.go
```diff
@@ -348,12 +354,52 @@ func mapFlowErrorToCIBAError(failureReason string) *CIBAError {
...
+	}
+
+	if exp, ok := payload[oauth2const.ClaimExp].(float64); ok {
+		if time.Now().Unix() > int64(exp)+cibaIDTokenHintStalenessSeconds {
+			return "", &CIBAError{
+				Code:    oauth2const.ErrorInvalidRequest,
+				Message: "id_token_hint has exceeded the maximum allowed staleness period",
+			}
+		}
+	}
```
sahandilshan: As mentioned above, this `cibaIDTokenHintStalenessSeconds` should be a configurable value. If not let's make it configurable

backend/internal/oauth/oauth2/ciba/service.go
Dilusha-Madushan: Done. Added IDTokenHintMaxAgeDays under CIBAConfig in OAuthConfig, so it can now be set via deployment.yaml:

``
oauth:
  ciba:
    id_token_hint_max_age_days: 30
``

If not set, it defaults to 30 days (the current behaviour). The config is read in validateIDTokenHint at runtime, so no restart-beyond-normal-config-reload is needed.

backend/internal/oauth/oauth2/ciba/service.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/identifying_executor.go
```diff
@@ -79,7 +79,21 @@ func newIdentifyingExecutor(
 func (i *identifyingExecutor) IdentifyUser(ctx context.Context, filters map[string]interface{},
 	execResp *common.ExecutorResponse) (*string, error) {
 	logger := i.logger
-	logger.Debug(ctx, "Identifying user with filters")
+	logger.DebugWithContext(ctx, "Identifying user with filters")
```
ThaminduDilshan: These `WithContext` log methods are removed now

backend/internal/oauth/oauth2/ciba/service.go
Dilusha-Madushan: cc: @ThaminduDilshan

backend/internal/oauth/oauth2/ciba/service.go
```diff
@@ -348,12 +354,52 @@ func mapFlowErrorToCIBAError(failureReason string) *CIBAError {
 	}
 }
 
+func (s *cibaService) validateIDTokenHint(ctx context.Context, idTokenHint string) (string, *CIBAError) {
```
ThaminduDilshan, line 358: Shall we add a method comment?

samples/apps/wayfinder-sample/README.md
```diff
@@ -157,7 +157,8 @@ The agent's client secret defaults to `wayfinder-agent-secret` (set in `thunderi
 | `JOHN_DOE_PASSWORD` | Yes | Password for the `john.doe` demo user |
 | `JANE_SMITH_PASSWORD` | Yes | Password for the `jane.smith` demo user |
 | `ALEX_CARTER_PASSWORD` | Yes | Password for the `alex.carter` demo user |
-| `CIBA_SMS_SENDER_ID` | **SMS only** | Notification sender ID for the CIBA SMS flow. Required only if you switch the upgrade agent to `wayfinder-ciba-sms-flow`. Obtain the sender ID after running the SMS setup script (see [CIBA Upgrade — SMS channel](#ciba-upgrade--sms-channel) below). |
+| `CIBA_LOGIN_HINT_ATTRIBUTE` | Yes | Attribute used by the CIBA flow's `IdentifyingExecutor` to resolve the login hint to a user. Set to `email` (default) when using `login_hint`, or `userID` when using `id_token_hint`. |
+| `SMS_SENDER_ID` | **SMS only** | Notification sender ID for the CIBA SMS flow. Required only if you switch the upgrade agent to `wayfinder-ciba-sms-flow`. Obtain the sender ID after running the SMS setup script (see [CIBA Upgrade — SMS channel](#ciba-upgrade--sms-channel) below). |
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai
The anchor is actually correct — GitHub converts —  (em dash flanked by spaces) to -- (double hyphen), so #ciba-upgrade--sms-channel maps correctly to the heading.

samples/apps/wayfinder-sample/scripts/lock-business-class.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
ThaminduDilshan, line 1: Shall we add license headers to all the new files added?

Dilusha-Madushan: Addressed

backend/internal/oauth/oauth2/ciba/service.go
Dilusha-Madushan: Addressed

backend/internal/flow/executor/identifying_executor.go
Dilusha-Madushan: Addressed

samples/apps/wayfinder-sample/scripts/reset-db.sh
```diff
@@ -0,0 +1,18 @@
...
+for f in "$DB_PATH" "${DB_PATH}-shm" "${DB_PATH}-wal"; do
+  if [ -f "$f" ]; then
+    rm "$f"
+    echo "Deleted $f"
+  fi
+done
+
+echo "Re-seeding database..."
+node "$BACKEND_DIR/scripts/seed.js"
+echo "Database reset complete."
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai Fixed this.

samples/apps/wayfinder-sample/backend/src/mcp.js
```diff
@@ -513,11 +549,33 @@ export async function handleMcpRequest(request, response, body) {
...
+
+      if (claims.sub && claims.sub === user.id) {
+        idToken = rawIdToken;
+      } else {
+        console.warn(`[mcp] x-id-token sub (${claims.sub}) does not match authenticated user (${user.id}) — ignoring`);
+      }
+    } catch {
+      console.warn("[mcp] x-id-token header could not be decoded — ignoring");
+    }
+  }
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check

samples/apps/wayfinder-sample/backend/src/mcp.js
```diff
@@ -295,37 +297,49 @@ function createTravelMcpServer(user) {
...
+      }
+
+      if (existingBooking.flight.cabin?.toLowerCase() !== "economy") {
+        throw new Error("Only Economy bookings can be upgraded.");
+      }
+
+      const bizFlight = findMatchingBusinessFlight(existingBooking.flight.id);
+
+      if (!bizFlight) {
+        return toToolContent({ available: false, message: "No Business class option exists for this flight." });
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check

samples/apps/wayfinder-sample/backend/src/db.js
```diff
@@ -62,6 +63,52 @@ function ensureSchema(database) {
...
+        updated_at TEXT NOT NULL
+      );
+      INSERT INTO upgrade_requests_new (id, user_id, username, booking_id, from_flight_id, to_flight_id, price_difference, id_token, status, created_at, updated_at)
+        SELECT id, user_id, username, booking_id, from_flight_id, to_flight_id, ${priceSelect}, ${idTokenSelect}, status, created_at, updated_at
+        FROM upgrade_requests;
+      DROP TABLE upgrade_requests;
+      ALTER TABLE upgrade_requests_new RENAME TO upgrade_requests;
+    `);
+  } else if (!hasIdToken) {
+    database.exec("ALTER TABLE upgrade_requests ADD COLUMN id_token TEXT;");
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check

backend/internal/oauth/oauth2/ciba/service.go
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai
Fixed. Inverted the guard — tokens without an exp claim now return invalid_request immediately before the staleness check. Added a dedicated test case TestInitiate_WithIDTokenHint_ExpMissing.

samples/apps/wayfinder-sample/README.md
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check

thiva-k:
Hope we have verified that the existing agent use cases from tryout are working as expected with these changes.

Dilusha-Madushan:
> Hope we have verified that the existing agent use cases from tryout are working as expected with these changes.

Yes, those use cases not break with new changes, only added story improvement like booking upgrade option. There can slight changes in agent responses with prompt changes but those changes not break the previous user stories.

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #3289, opened 2026-06-12
Block direct HTTP flow initiation for authorization_code grant type apps

### LINKED ISSUE #3215: Implement HTTP guard in loadNewContext() to block direct flow initiation for authorization_code grant type applications

Malith-19:
### Purpose

Applications configured with the `authorization_code` grant type (browser SPAs) must have their flows initiated by the OAuth component via `/authorize`, not via a direct `POST /flow/execute` call. Previously the endpoint was fully open — any caller could start a new flow for any application. This PR adds a guard that blocks new-flow initiation for such applications, while leaving continuation requests (those carrying an `executionId`) and the OAuth component's internal `InitiateFlow()` path unaffected.

It also fixes a UX regression the guard introduces: when a user refreshes the gate app mid-flow, the React SDK was attempting to start a new flow directly (now blocked), causing an error page. The SDK is updated to rehydrate and resume the in-progress flow on refresh instead.

Closes #3215
Design discussion: #2744

---

`POST /flow/execute` now rejects new-flow initiation for applications configured with the `authorization_code` grant type.

#### 💥 Impact

Callers that send a request with no `executionId` for an `authorization_code` application now receive `403 Forbidden` with error code `FES-1011`. Continuation requests (with a valid `executionId`) are unaffected.

#### 🔄 Migration Guide

Direct the user to `GET /oauth2/authorize`. The OAuth component will call `InitiateFlow()` internally and redirect the browser to the SPA with an `executionId`. The SPA then calls `POST /flow/execute` with that `executionId` to continue — this path is unaffected.

---

### Approach

**Guard placement — `loadNewContext()`, not `initContext()`**

`Execute()` branches on the presence of an `executionId`:

```
Execute()
  ├── empty executionId  →  loadNewContext()   ← guard lives here
  └── non-empty executionId  →  loadPrevContext()   (unaffected)

InitiateFlow() [OAuth component]  →  initContext()   (unaffected)
```

Placing the guard in `loadNewContext()` avoids blocking the OAuth component's legitimate `InitiateFlow()` path, which calls `initContext()` directly.

**`checkDirectFlowInitiationAllowed()` — the guard**

Resolves the app's OAuth profile via `actorProvider.GetOAuthProfileByEntityID` (routed through the existing `actorProvider` dependency — no new service dependencies added to `flowexec`) and checks whether `authorization_code` is in `GrantTypes`. Returns `ErrorDirectFlowInitiationNotPermitted` (error code `FES-1011`, type `ForbiddenErrorType`) if so.

`GetOAuthProfileByEntityID` was added to `ActorProviderInterface` and implemented in `actorprovider.service` to keep the OAuth profile lookup consolidated behind `actorProvider`, consistent with the intent of #3306.

Fail-open cases (guard passes silently):
- `appID` is empty (app-independent flows such as user onboarding).
- App has no OAuth profile registered — `actorProvider` returns `ErrorActorNotFound`, nothing to enforce.
- `nil` OAuth profile returned.

Fail-closed: any unexpected store/infrastructure error returns HTTP 500 rather than accidentally allowing the request through.

**New `ForbiddenErrorType` + HTTP 403 mapping**

The error type system previously only mapped `ClientErrorType → 400` and `ServerErrorType → 500`. A new `ForbiddenErrorType` was added to the `serviceerror` package and `handleFlowError` updated to map it to HTTP 403.

**React SDK — refresh fix (`SignIn` v2)**

On page refresh, the SDK lost its React state and `executionId` was already cleaned from the URL. The component now:
1. Falls back to `sessionStorage.thunderid_execution_id` to rehydrate the `executionId`.
2. Resumes the in-progress flow via `signIn({ executionId, ...challengeToken })` instead of initiating a new flow.
3. Shows a user-friendly "session expired" message if the stored `executionId` is stale. A stale session is detected when the resume request fails with HTTP `400` (a `ThunderIDAPIError` with `statusCode === 400`); transient failures (5xx / network errors) are rethrown so they surface normally.

**Test fixes**

- *Integration tests:* `CreateApplication` in `testutils/api_utils.go` was omitting `grantTypes`, causing the server to default to `["authorization_code"]` for every test app. Fixed by explicitly setting `"grantTypes": ["client_credentials"]` so test apps are not subject to the guard. `acr_options_flow_test.go` previously used a single `authorization_code` app for all ACR tests; it now adds a dedicated native app (`acrOptionsNativeTestApp`, `grant_types: ["client_credentials"]`) for the two tests that call `/flow/execute` directly, while keeping the `authorization_code` app for the tests that go through `/oauth2/authorize`.
- *E2E tests:* `admin-api-auth.ts` and `mfa-setup.ts` were calling `/flow/execute` directly using `SAMPLE_APP_ID` (an `authorization_code` app). Fixed by adding a dedicated native app (`thunder_e2e_admin_client`, `grant_types: ["client_credentials"]`) with a fixed UUID to `tests/e2e/thunderid-config.yaml`, and referencing that UUID in the E2E helpers.

**Test coverage**

Additional tests were added to bring `flowexec` package coverage from 61% to 80%+, covering:
- All handler paths, including the new 403 mapping (converted to `HandlerTestSuite`).
- Previously uncovered engine internals (converted to `EngineTestSuite`).
- `checkDirectFlowInitiationAllowed` and related service paths (in `ServiceTestSuite`).

### Related Issues
- Fixes #3215
- Design: https://github.com/thunder-id/thunderid/discussions/2744

### Related PRs
- N/A

### Checklist
- [ ] Documentation provided.
- [x] Tests provided.
    - [x] Unit Tests
    - [x] Integration Tests
- [x] Breaking changes.

### Security checks

### DIFF
--- a/.github/workflows/pr-builder.yml
+++ b/.github/workflows/pr-builder.yml
@@ -731,6 +731,23 @@ jobs:
           fi
           echo "✅ React SDK Sample resources imported successfully"
 
+          echo "Importing E2E test infrastructure resources..."
+          YAML_CONTENT=$(cat tests/e2e/thunderid-config.yaml)
+          HTTP_STATUS=$(curl -k -s -o /tmp/import_response.json -w "%{http_code}" \
+            -X POST https://localhost:8090/import \
+            -H 'Content-Type: application/json' \
+            -d "{
+              \"content\": $(echo "$YAML_CONTENT" | jq -Rs .),
+              \"options\": {\"upsert\": true}
+            }")
+          echo "Import HTTP status: $HTTP_STATUS"
+          cat /tmp/import_response.json | jq . || cat /tmp/import_response.json
+          if [ "$HTTP_STATUS" != "200" ]; then
+            echo "❌ Failed to import E2E test infrastructure resources (HTTP $HTTP_STATUS)"
+            exit 1
+          fi
+          echo "✅ E2E test infrastructure resources imported successfully"
+
       - name: 🔍 Extract Sample App ID
         id: extract-app-id
         run: |
--- a/backend/internal/actorprovider/interface.go
+++ b/backend/internal/actorprovider/interface.go
@@ -33,6 +33,9 @@ type ActorProviderInterface interface {
 	GetOAuthClientByID(
 		ctx context.Context, id string,
 	) (*inboundmodel.OAuthClient, *serviceerror.ServiceError)
+	GetOAuthProfileByID(
+		ctx context.Context, id string,
+	) (*inboundmodel.OAuthProfile, *serviceerror.ServiceError)
 	GetInboundClientByID(
 		ctx context.Context, id string,
 	) (*inboundmodel.InboundClient, *serviceerror.ServiceError)
--- a/backend/internal/actorprovider/interface.go
+++ b/backend/internal/actorprovider/interface.go
@@ -30,8 +30,8 @@ import (
 
 // ActorProviderInterface resolves inbound actors and exposes their OAuth and membership data.
 type ActorProviderInterface interface {
-	GetOAuthClientByID(
-		ctx context.Context, id string,
+	GetOAuthClientByClientID(
+		ctx context.Context, clientID string,
 	) (*inboundmodel.OAuthClient, *serviceerror.ServiceError)
 	GetOAuthProfileByID(
 		ctx context.Context, id string,
--- a/backend/internal/flow/flowexec/error_constants.go
+++ b/backend/internal/flow/flowexec/error_constants.go
@@ -164,3 +164,18 @@ var ErrorInvalidChallengeToken = serviceerror.ServiceError{
 		DefaultValue: "The challenge token is missing or invalid",
 	},
 }
+
+// ErrorDirectFlowInitiationNotPermitted defines the error for applications that do not allow
+// direct flow initiation via the HTTP endpoint (e.g. authorization_code grant type apps).
+var ErrorDirectFlowInitiationNotPermitted = serviceerror.ServiceError{
+	Code: "FES-1011",
+	Type: serviceerror.ClientErrorType,
+	Error: core.I18nMessage{
+		Key:          "error.flowexecservice.direct_flow_initiation_not_permitted",
+		DefaultValue: "Direct flow initiation not permitted",
+	},
+	ErrorDescription: core.I18nMessage{
+		Key:          "error.flowexecservice.direct_flow_initiation_not_permitted_description",
+		DefaultValue: "Direct flow initiation is not permitted for this application type",
+	},
+}
--- a/backend/internal/flow/flowexec/handler_test.go
+++ b/backend/internal/flow/flowexec/handler_test.go
@@ -1,0 +1,178 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package flowexec
+
+import (
+	"bytes"
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/suite"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	i18ncore "github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+const testFlowExecRequestBody = `{"applicationId":"app-1","flowType":"AUTHENTICATION","action":"submit"}`
+
+type HandlerTestSuite struct {
+	suite.Suite
+}
+
+func TestHandlerTestSuite(t *testing.T) {
+	suite.Run(t, new(HandlerTestSuite))
+}
+
+func (s *HandlerTestSuite) TestNewFlowExecutionHandler() {
+	t := s.T()
+	mockSvc := NewFlowExecServiceInterfaceMock(t)
+	h := newFlowExecutionHandler(mockSvc)
+	s.NotNil(h)
+	s.Equal(mockSvc, h.flowExecService)
+}
+
+func (s *HandlerTestSuite) TestHandleFlowError_ClientError_Returns400() {
+	w := httptest.NewRecorder()
+	svcErr := &serviceerror.ServiceError{
+		Code: "FES-4001",
+		Type: serviceerror.ClientErrorType,
+		Error: i18ncore.I18nMessage{
+			Key:          "client.error",
+			DefaultValue: "bad request",
+		},
+	}
+	handleFlowError(context.Background(), w, svcErr)
+	s.Equal(http.StatusBadRequest, w.Code)
+}
+
+func (s *HandlerTestSuite) TestHandleFlowError_ForbiddenError_Returns403() {
+	w := httptest.NewRecorder()
+	handleFlowError(context.Background(), w, &ErrorDirectFlowInitiationNotPermitted)
+	s.Equal(http.StatusForbidden, w.Code)
+}
+
+func (s *HandlerTestSuite) TestHandleFlowError_ServerError_Returns500() {
+	w := httptest.NewRecorder()
+	svcErr := &serviceerror.ServiceError{
+		Code: "FES-5001",
+		Type: serviceerror.ServerErrorType,
+		Error: i18ncore.I18nMessage{
+			Key:          "server.error",
+			DefaultValue: "internal error",
+		},
+	}
+	handleFlowError(context.Background(), w, svcErr)
+	s.Equal(http.StatusInternalServerError, w.Code)
+}
+
+func (s *HandlerTestSuite) TestConvertToAPIError() {
+	svcErr := &serviceerror.ServiceError{
+		Code: "FES-1234",
+		Error: i18ncore.I18nMessage{
+			Key:          "test.error",
+			DefaultValue: "test message",
+		},
+		ErrorDescription: i18ncore.I18nMessage{
+			Key:          "test.error.desc",
+			DefaultValue: "test description",
+		},
+	}
+	resp := convertToAPIError(svcErr)
+	s.Equal("FES-1234", resp.Code)
+}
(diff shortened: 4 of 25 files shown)


Malith-19:
@coderabbitai review

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -164,3 +164,18 @@ var ErrorInvalidChallengeToken = serviceerror.ServiceError{
 		DefaultValue: "The challenge token is missing or invalid",
 	},
 }
+
+// ErrorDirectFlowInitiationNotPermitted defines the error for applications that do not allow
+// direct flow initiation via the HTTP endpoint (e.g. authorization_code grant type apps).
+var ErrorDirectFlowInitiationNotPermitted = serviceerror.ServiceError{
+	Code: "FES-1011",
```
ThaminduDilshan, line 171: Do we have a error code for `FES-1010`?

sdks/react/src/components/presentation/auth/SignIn/v2/SignIn.tsx
```diff
@@ -448,11 +457,37 @@ const SignIn: FC<SignInProps> = ({
...
+                'react',
+              );
+              setError(expiredError);
+              return;
+            }
+            response = (await signIn({
+              applicationId: effectiveApplicationId,
+              flowType: EmbeddedFlowType.Authentication,
+              ...(scopes && {scopes}),
+            })) as EmbeddedSignInFlowResponseV2;
```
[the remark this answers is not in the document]
Malith-19: @coderabbitai please verify the changes now.

Malith-19:
@coderabbitai full review

backend/cmd/server/servicemanager.go
```diff
@@ -388,7 +388,7 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	)
 
 	flowExecService, err := flowexec.Initialize(mux, flowMgtService, actorProvider,
-		execRegistry, observabilitySvc, runtimeCryptoSvc, flowconfig.FromServerRuntime())
+		inboundClientService, execRegistry, observabilitySvc, runtimeCryptoSvc, flowconfig.FromServerRuntime())
```
ThaminduDilshan: Adding inbound client dependency to the flow execution service deviates the intention of https://github.com/thunder-id/thunderid/pull/3306.

backend/internal/flow/flowexec/engine_extra_test.go
```diff
@@ -0,0 +1,763 @@
...
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/tests/mocks/flow/coremock"
+	"github.com/thunder-id/thunderid/tests/mocks/flow/executormock"
+	"github.com/thunder-id/thunderid/tests/mocks/observability/observabilitymock"
+)
+
+// --- newFlowEngine ---
+
+func TestNewFlowEngine(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern instead? 
Also can't we have these tests in the existing engine_test file and suite? These are tests on engine methods right?
If it's getting lengthy, fine to split to a separate file and test suite

backend/internal/flow/flowexec/handler_test.go
```diff
@@ -0,0 +1,184 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	i18ncore "github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+func TestNewFlowExecutionHandler(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/flow/flowexec/error_constants.go
Malith-19: Yes it's present [1] but not added in the order I guess. 

[1] https://github.com/thunder-id/thunderid/blob/f8324296c9086cc6beac48f7aefa59e0a952ce0e/backend/internal/flow/flowexec/error_constants.go#L114

samples/apps/react-vanilla-sample/thunderid-config/basic/thunderid-config.yaml
```diff
@@ -128,3 +128,23 @@ inboundAuthConfig:
           - given_name
           - family_name
           - picture
+
+---
+# resource_type: application
+# Native app used by E2E test infrastructure for direct flow execution (admin API auth).
+# Does not use the authorization_code grant type so it is not subject to the
+# redirect-based flow initiation guard.
+id: 019e3a5c-0501-7f3e-a66e-66fc7918c3a7
```
ThaminduDilshan: Do we need to add a specific app for the tests? This is supposed to be imported during sample app setup guide. That experience will affect if we have a e2e specific app

backend/cmd/server/servicemanager.go
ThaminduDilshan: Should be able to use actorProvider

backend/internal/flow/flowexec/handler_test.go
Malith-19: Updated.

backend/internal/flow/flowexec/engine_extra_test.go
Malith-19: Updated to use the test suite pattern and seems like this file was there in a earlier commit which is not present now. I have already moved this tests to existing test file.

samples/apps/react-vanilla-sample/thunderid-config/basic/thunderid-config.yaml
Malith-19: Created a resource file in e2e test suite and imported in the pr builder.

backend/cmd/server/servicemanager.go
Malith-19: Yes, added a new method to retrieve the `GetOAuthProfileByEntityID` and used it in the service.

.github/workflows/pr-builder.yml
```diff
@@ -735,6 +731,23 @@ jobs:
           fi
           echo "✅ React SDK Sample resources imported successfully"
 
+          echo "Importing E2E test infrastructure resources..."
```
ThaminduDilshan, line 734: With this change, can we run e2e tests locally?

backend/internal/system/error/serviceerror/error.go
```diff
@@ -31,6 +31,8 @@ const (
 	ClientErrorType ServiceErrorType = "client_error"
 	// ServerErrorType denotes the server error type.
 	ServerErrorType ServiceErrorType = "server_error"
+	// ForbiddenErrorType denotes a forbidden/access-denied error type.
+	ForbiddenErrorType ServiceErrorType = "forbidden_error"
```
ThaminduDilshan: Let's not define another error type. This is also a client error right?

Malith-19: We need to return 403 in this scenario right? The request is valid but the client lacks permission for this operation on this resource. Mapping it to ClientErrorType would return 400, which misguides API consumers into thinking their request is broken rather than their access path being wrong. IMO ForbiddenErrorType keeps the HTTP semantics accurate.

.github/workflows/pr-builder.yml
Malith-19: Yes updated the e2e script to reflect the same and verified running e2e test suite locally with the latest changes.

backend/internal/actorprovider/interface.go
```diff
@@ -33,6 +33,9 @@ type ActorProviderInterface interface {
 	GetOAuthClientByID(
 		ctx context.Context, id string,
 	) (*inboundmodel.OAuthClient, *serviceerror.ServiceError)
+	GetOAuthProfileByEntityID(
+		ctx context.Context, entityID string,
+	) (*inboundmodel.OAuthProfile, *serviceerror.ServiceError)
```
ThaminduDilshan:
```suggestion
	GetOAuthProfileByID(
		ctx context.Context, id string,
	) (*inboundmodel.OAuthProfile, *serviceerror.ServiceError)
```

We intentionally removed entity part from other methods. Let's rename this to just `GetOAuthProfileByID`

ThaminduDilshan: But do we really need to introduce such a new method? We just need to get grant types right? That's already available in `OAuthClient`

backend/internal/system/error/serviceerror/error.go
ThaminduDilshan: Can't we catch the service error with specific code in the handler and set appropriate response status? That should be the way we're already handling 400 and 404 for management APIs

backend/internal/actorprovider/interface.go
Malith-19: To get the oauthclient we need the clientID of the application right? So when we checking the flow initiation we only had the appID.

backend/internal/system/error/serviceerror/error.go
Malith-19: +1, updated.

backend/internal/actorprovider/interface.go
ThaminduDilshan: We should anyway remove entity term from this layer. It's confusing to have entity in the actor provider interface.

ThaminduDilshan: I didn't notice it's the client id initially. That's why suggested to rename that to id in the first place for the existing method.

Shall we rename existing method as below.
```go
GetOAuthClientByClientID(ctx context.Context, clientId string) (*inboundmodel.OAuthClient, *serviceerror.ServiceError)
```

Then we can keep new method like this.
```go
GetOAuthProfileByID(ctx context.Context, id string) (*inboundmodel.OAuthProfile, *serviceerror.ServiceError)
```

Malith-19: Updated from https://github.com/thunder-id/thunderid/pull/3289/commits/aca4bf927d4eca06b6d32ffb08fef596076b644d

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #3295, opened 2026-06-12
Support multi-group and multi-role assignment in provisioning executor

Dilusha-Madushan:
### Purpose

The provisioning executor previously supported assigning only a single group and single role to a newly provisioned user. This change adds support for comma-separated group and role IDs in the assignGroup and assignRole node properties, and ensures all assignments are executed within a single database transaction so a partial failure rolls back all assignments atomically. Single-value properties continue to work unchanged.

### Approach

Added AddUserToGroups to GroupServiceInterface (group/service.go) and AddUserToRoles to RoleAssignmentServiceInterface (role/assignment_service.go). Each method wraps all assignments in a single transactioner.Transact call — a failure on any individual assignment triggers a full rollback via the capturedSvcErr pattern used elsewhere in the service layer.
Updated the provisioning executor to parse the assignGroup/assignRole node properties with a splitTrimmed helper that splits on commas and strips whitespace, producing a []string. A single ID produces a one-element slice, preserving backward compatibility.
Replaced the old per-ID assignToGroup/assignToRole helpers with a unified assignGroupsAndRoles method that delegates to the new bulk service calls.
Regenerated mocks via make mockery to include the new interface methods.

## Release Notes

**New Features**
- Provisioning workflows can now assign multiple groups and roles simultaneously through comma-separated values in configuration properties
- Enhanced error handling and validation for bulk group and role assignments

**Tests**
- Added comprehensive test coverage for multiple group and role assignment scenarios
- Updated tests to verify new bulk assignment behavior and error handling


backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -638,161 +638,88 @@ func (p *provisioningExecutor) getUserType(ctx *core.NodeContext) string {
...
-			return fmt.Errorf("group assignment error: %w; role assignment error: %s", groupErr, roleErr.Error())
-		}
-		if groupErr != nil {
-			return groupErr
+
+	if len(roleIDs) > 0 {
+		if svcErr := p.roleAssignmentService.AddUserToRoles(
+			ctx.Context, userID, role.AssigneeTypeUser, roleIDs); svcErr != nil {
+			return fmt.Errorf("role assignment failed: %s", svcErr.Error.DefaultValue)
 		}
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai
This is a pre-existing condition, not a regression introduced by this PR. The original implementation also ran assignToGroup and assignToRole as two independent service calls with no shared transaction — and it was worse: it attempted both assignments even when the first had already failed, then reported combined errors.

All group assignments are now wrapped in a single transaction (previously each was a separate call).
All role assignments are now wrapped in a single transaction (same).
Execution fails fast — if the group assignment fails, the role assignment is never attempted.
True cross-service atomicity (groups + roles in one transaction) would require injecting the transactioner into the executor, which is a separate concern and out of scope for this PR.

cc: @ThaminduDilshan

backend/internal/group/service.go
```diff
@@ -1168,6 +1170,44 @@ func (gs *groupService) GetGroupsByIDs(
 	return result, nil
 }
 
+// AddUserToGroups adds a user to multiple groups in a single transaction. All group IDs are
+// validated before the transaction begins; a single failure rolls back all assignments.
+func (gs *groupService) AddUserToGroups(
+	ctx context.Context,
+	userID string,
+	memberType MemberType,
```
rajithacharith: As this function is AddUserToGroups, we can avoid having this as a method parameter.

Dilusha-Madushan: Addressed this

Dilusha-Madushan: @coderabbitai check

backend/internal/group/service.go
```diff
@@ -59,6 +59,8 @@ type GroupServiceInterface interface {
 	GetGroupsByIDs(ctx context.Context, groupIDs []string) (map[string]*Group, *serviceerror.ServiceError)
 	AddGroupMembers(ctx context.Context, groupID string, members []Member) (*Group, *serviceerror.ServiceError)
 	RemoveGroupMembers(ctx context.Context, groupID string, members []Member) (*Group, *serviceerror.ServiceError)
+	AddUserToGroups(ctx context.Context, userID string,
```
ThaminduDilshan: Shall we keep this generic for the entity? I.e. let's not make the new method specific to users.

Suggestion: `AddMemberToGroups`

Also since we're using the existing AddGroupMembers method internally, what if we provide flexibility to add a set of users to the groups as well. WDYT?

Suggestion: `AddMembersToGroups`

ThaminduDilshan: Modify parameter names accordingly

backend/internal/group/service_test.go
```diff
@@ -50,6 +50,12 @@ func (s *stubTransactioner) Transact(ctx context.Context, txFunc func(context.Co
 	return txFunc(ctx)
 }
 
+type errTransactioner struct{ err error }
```
ThaminduDilshan: Can't we use mockery for this?

backend/internal/role/assignment_service.go
```diff
@@ -43,6 +43,8 @@ type RoleAssignmentServiceInterface interface {
 		includeDisplay bool, assigneeType string) (*AssignmentList, *serviceerror.ServiceError)
 	AddAssignments(ctx context.Context, id string, assignments []RoleAssignment) *serviceerror.ServiceError
 	RemoveAssignments(ctx context.Context, id string, assignments []RoleAssignment) *serviceerror.ServiceError
+	AddUserToRoles(ctx context.Context, userID string,
```
ThaminduDilshan: Same comment applies for this too

 https://github.com/thunder-id/thunderid/pull/3295/changes#r3402391011

ThaminduDilshan:
Hope we don't need any changes to the flow builder UI, at least for the executor property descriptions

backend/internal/group/service.go
Dilusha-Madushan: @coderabbitai check with new updates

backend/internal/group/service_test.go
Dilusha-Madushan: @coderabbitai check with new updates

backend/internal/role/assignment_service.go
Dilusha-Madushan: @coderabbitai check with new updates

Dilusha-Madushan:
> Hope we don't need any changes to the flow builder UI, at least for the executor property descriptions

Update the i18n keys used in flow builder.

backend/internal/role/assignment_service.go
Dilusha-Madushan: @coderabbitai
AddUserToRoles has already been removed in this branch — the interface now only exposes AddAssigneesToRoles(ctx context.Context, assignments []RoleAssignment, roleIDs []string), which is exactly Option 2 you described. The provisioning executor builds []RoleAssignment{{ID: userID, Type: AssigneeTypeUser}} and calls it directly. There is no separate AddUserToRoles method.

backend/internal/group/service_test.go
Dilusha-Madushan: @coderabbitai Is this now fixed?

ThaminduDilshan (approved):

PR #3298, opened 2026-06-12
feat(scim): implement SCIM discovery endpoints per RFC 7643 and RFC 7644

ravindu439:
### Purpose
Implements the foundational SCIM 2.0 package structure for ThunderID and three read-only discovery endpoints as defined in RFC 7643 and RFC 7644:

- `GET /scim/v2/ServiceProviderConfig` — allows SCIM clients to discover server capabilities (patch, bulk, filter, ETag, sort, change-password, authentication schemes) before making any provisioning calls.
- `GET /scim/v2/Schemas` and `GET /scim/v2/Schemas/{id}` — allows SCIM clients to introspect the attribute definitions of all supported resource types, including a minimalist core User schema and ThunderID extension schemas dynamically built from EntityType definitions.
- `GET /scim/v2/ResourceTypes` and `GET /scim/v2/ResourceTypes/{id}` — allows SCIM clients to discover the resource types supported by the server and their associated schemas and schema extensions.

All three endpoints are required by RFC 7644 §4 for a compliant SCIM server and form the complete discovery layer that all future provisioning endpoints (Users, Groups, Bulk) will build upon.

#### Core Package Structure
Introduced a new `backend/internal/scim` package with a clean layered architecture:

- `model.go` — SCIM data models (`ServiceProviderConfig`, `SCIMSchema`, `SCIMSchemaAttribute`, `SCIMResourceType`, `SCIMListResponse`, `SCIMMeta`, `AuthScheme`, etc.)
- `constants.go` — Schema URNs, attribute mutability/returned/uniqueness constants, and package-level configuration constants
- `error_constants.go` — All SCIM error codes mapped to ThunderID `ServiceError` types; internal codes are never exposed to clients
- `response.go` — Shared `writeSCIMSuccessResponse` and `writeSCIMErrorResponse` helpers using a buffer-first pattern to prevent half-written HTTP responses
- `schema_mapper.go` — Converts ThunderID `EntityType` definitions into RFC 7643-compliant `SCIMSchema` resources; credential fields are mapped to `returned=never` so secrets are never disclosed
- `service.go` — `SCIMServiceInterface` and `scimService` implementation
- `handler.go` — HTTP handlers delegating to the service layer
- `init.go` — Service initialization, config wiring, and route registration

#### ServiceProviderConfig (RFC 7643 §5)
- Treated as a **static resource** per RFC 7644 — `lastModified` is set once at startup and never changes per request
- A stable weak ETag (`W/"<sha256-prefix>"`) is computed from the `SCIMConfig` struct at startup and only changes when an operator toggles a capability flag
- All capability flags (`patch`, `bulk`, `filter`, `sort`, `changePassword`, `etag`) are driven by the ThunderID config so no code change is needed to enable or disable a feature

#### Schema Endpoint (RFC 7643 §7)
- `GET /scim/v2/Schemas` returns a `ListResponse` containing a minimalist core User schema (only `id` attribute) plus one extension schema per ThunderID user type
- `GET /scim/v2/Schemas/{id}` resolves by URN — accepts the core `urn:ietf:params:scim:schemas:core:2.0:User` URN and ThunderID extension URNs (`urn:thunderid:params:scim:schemas:<userTypeName>:2.0:User`) case-insensitively per the proposal decision
- Extension schemas are built dynamically at request time from `EntityType` definitions via `schema_mapper.go` — no static schema registry required
- ThunderID `credential` fields map to `mutability=writeOnly, returned=never` so sensitive attributes are structurally prevented from leaking in any SCIM response

#### ResourceTypes Endpoint (RFC 7643 §6)
- `GET /scim/v2/ResourceTypes` returns a `ListResponse` with the single `User` resource type supported by ThunderID
- `GET /scim/v2/ResourceTypes/{id}` resolves by ID case-insensitively; returns `404` for any unknown resource type
- The `schemaExtensions` field is built dynamically at request time — one entry per registered ThunderID entity type — so no manual registry updates are needed when new user types are added
- ThunderID only exposes a single `User` resource type per the proposal design decision; `Group` and other resource types are not supported in this phase

#### Error Handling
- All error responses are routed through a single `handleSCIMError` translator that maps internal `ServiceError` codes to RFC 7644 §3.12 compliant HTTP status and `scimType` pairs
- `HandleUnsupportedRequest` returns a consistent `501 notImplemented` SCIM response for all endpoints not yet implemented

### Related Issues
- #3318

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks
- [x] Followed secure coding standards in [[WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)]

## Summary

* **New Features**
  * Added CORS-enabled SCIM v2 discovery endpoints: `ServiceProviderConfig`, `Schemas` (list + get), and `ResourceTypes` (list + get).
  * Dynamic schema and resource-type discovery, including weak ETag support based on configuration (disabled by default).
  * Requests to unsupported SCIM endpoints now return standards-based “not implemented” responses.

* **Bug Fixes**
  * Improved SCIM error handling to consistently map internal failures to correct SCIM/HTTP statuses.

* **Tests**
  * Added unit tests covering discovery responses and SCIM error-to-HTTP mapping.

* **Chores**
  * Updated default configuration with a new SCIM settings block (all features off by default).


backend/cmd/server/repository/conf/deployment.yaml
```diff
@@ -73,3 +73,14 @@ email:
     from_address: "noreply@thunderid.dev"
     enable_start_tls: false
     enable_authentication: true
+
+scim:
```
sadilchamishka: This configuration approach is good. But as we are not providing the support yet, we may not need to allow configure yet. After we provide the support, let's give the configuration for enabling / disabling the required support.

backend/internal/scim/service.go
```diff
@@ -0,0 +1,386 @@
...
+// The schemaExtensions array is built dynamically by paginating through all
+// user-type entity types — identical pagination pattern to ListSchemas.
+// The core User schema URN is always the primary Schema field; each registered
+// user-type contributes one required=false extension entry.
+// RFC 7643 §6
+func (s *scimService) buildUserResourceType(ctx context.Context, baseURL string) (SCIMResourceType, *serviceerror.ServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
+
+	ctx = security.WithRuntimeContext(ctx)
+	location := fmt.Sprintf("%s/%s/ResourceTypes/%s", baseURL, SCIMBasePath, scimResourceTypeUserID)
```
sadilchamishka: **Malformed `location` — double slash.**

`SCIMBasePath` is already `/scim/v2`, so `"%s/%s/ResourceTypes/%s"` produces `https://host//scim/v2/ResourceTypes/User` (note the `//`). Every other location builder in this PR uses `"%s%s"` (e.g. `GetServiceProviderConfig` and `mapEntityTypeToSCIMSchema`). Suggest changing `"%s/%s/ResourceTypes/%s"` → `"%s%s/ResourceTypes/%s"` for consistency.

sadilchamishka (commented):
Nice, clean work — the layered package structure is easy to follow, test coverage is solid, and mapping credential fields to `returned=never` / `writeOnly` is exactly right. Two things I'd like to flag as potentially major; the rest looks good to me.

backend/internal/scim/schema_mapper.go
```diff
@@ -0,0 +1,203 @@
...
+	Unique      bool                      `json:"unique"`
+	Credential  bool                      `json:"credential"`
+	DisplayName string                    `json:"displayName"`
+	Properties  map[string]rawPropertyDef `json:"properties"` // for type=object
+	Items       *rawPropertyDef           `json:"items"`      // for type=array
+}
+
+// buildSchemaURN returns the canonical lowercase SCIM extension URN for a ThunderID user type.
+// Format: urn:thunderid:params:scim:schemas:<userTypeName>:2.0:User
+func buildSchemaURN(userTypeName string) string {
```
sadilchamishka: Let's move the utils to different file and keep the schema mapping in this file.

backend/internal/scim/service.go
ravindu439: will fix it .

backend/cmd/server/config/default.json
```diff
@@ -272,5 +270,16 @@
     "credential_validity_seconds": 2592000,
     "batch_size": 5,
     "enforce_scope": false
+  },
+  "scim": {
```
ThaminduDilshan: Is there a requirement to turn on/off these features? I'm thinking whether we need to have these configs. Having unnecessary configs going to be a pain in the long run.
Open to discussion

backend/cmd/server/servicemanager.go
```diff
@@ -207,6 +208,15 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	}
 	exporters = append(exporters, userExporter)
 
+	// Initialize SCIM service.
+	scim.Initialize(
+		mux,
+		userService,
+		entityTypeService,
+		config.GetServerRuntime().Config.Server.PublicURL,
```
ThaminduDilshan: Recent pattern is to define a config at each package. We can include this public url inside that SCIM config  and pass only that.

Refer: https://github.com/thunder-id/thunderid/blob/main/backend/internal/flow/config/config.go

backend/internal/scim/constants.go
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package scim implements the SCIM v2.0 API endpoints for ThunderID,
```
ThaminduDilshan: Let's define package comment only in a single file. Usually we define it in init or service file

backend/internal/scim/constants.go
```diff
@@ -0,0 +1,81 @@
...
+// following RFC 7643 and RFC 7644.
+package scim
+
+const (
+	loggerComponentName              = "SCIMhandler"
```
ThaminduDilshan: We don't need this constant. We can directly define logger inside struct and initialize in the constructor method.

```go
type handler struct {
    logger *log.Logger
}

func newHandler() *handler {
	return &handler{
        logger: log.GetLogger().With(log.String(log.LoggerKeyComponentName, "Handler")),
	}
}

func (h *handler) HandleRequest(w http.ResponseWriter, r *http.Request) {
	logger := h.logger
}
```

backend/internal/scim/constants.go
```diff
@@ -0,0 +1,81 @@
...
+	scimServiceProviderConfigCreated = "2025-01-01T00:00:00Z"
```
ThaminduDilshan: What's the purpose of this constant? Variable name and the value doesn't tally.
If this is a test only constant, let's remove it from source files

backend/internal/scim/constants.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// ThunderIDURNPrefix is the custom URN prefix for ThunderID SCIM schemas.
+	ThunderIDURNPrefix = "urn:thunderid:params:scim:schemas:"
+
+	// ThunderIDURNSuffix is the custom URN suffix for ThunderID SCIM user schemas.
+	ThunderIDURNSuffix = ":2.0:User"
+
+	// scimCoreUserSchemaCreated is the static creation timestamp for the core User schema.
+	// This resource is static and never mutated by operators.
+	scimCoreUserSchemaCreated = "2025-01-01T00:00:00Z"
```
ThaminduDilshan: What's the purpose to have a hard coded timestamp like this?

backend/internal/scim/constants.go
```diff
@@ -0,0 +1,81 @@
...
+
+	scimAttrTypeString  = "string"
```
ThaminduDilshan: 1. Shall we define go types for the appropriate constants? No need to define for all. Maybe attr type, mutability, return type, resource type, etc.
2. Shall we split constants into multiple constant blocks and group related constants together to increase readability?

backend/internal/scim/error_constants.go
```diff
@@ -0,0 +1,212 @@
...
+
+import (
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// Internal SCIM service errors.
+// These codes are NEVER sent to SCIM clients over the wire.
+// handleSCIMError translates these into the SCIM-standard wire format.
+
+//nolint:gochecknoglobals
```
ThaminduDilshan: Let's avoid skipping go lint rules. Instead fix the issues

backend/internal/scim/handler.go
```diff
@@ -0,0 +1,207 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/thunder-id/thunderid/internal/system/log"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// scimHandler handles SCIM HTTP requests.
+type scimHandler struct {
+	// svc is named differently from the concrete *scimService type to avoid
```
ThaminduDilshan: Let's remove this comment. Not required

backend/internal/scim/handler.go
```diff
@@ -0,0 +1,207 @@
...
+// HandleUnsupportedRequest returns a SCIM-standard 501 for unimplemented endpoints.
+// Delegates to handleSCIMError so that all error paths go through the same translator.
+func (sh *scimHandler) HandleUnsupportedRequest(w http.ResponseWriter, r *http.Request) {
+	handleSCIMError(w, r, &ErrorUnsupportedOperation)
+}
+
+// handleSCIMError translates an internal ThunderID ServiceError into the
+// SCIM-standard wire error response (RFC 7644 §3.12).
+// Internal codes (SCIM-1001 etc.) are NEVER sent to the client.
+func handleSCIMError(w http.ResponseWriter, r *http.Request, svcErr *tidcommon.ServiceError) {
```
ThaminduDilshan: Shall we convert this into a struct receiver function?

backend/internal/scim/handler.go
```diff
@@ -0,0 +1,207 @@
...
+		return
+	}
+
+	writeSCIMSuccessResponse(ctx, w, rt)
+	logger.Debug(ctx, "SCIM ResourceType GET response sent", log.String("id", resourceTypeID))
+}
+
+// HandleUnsupportedRequest returns a SCIM-standard 501 for unimplemented endpoints.
+// Delegates to handleSCIMError so that all error paths go through the same translator.
+func (sh *scimHandler) HandleUnsupportedRequest(w http.ResponseWriter, r *http.Request) {
```
ThaminduDilshan: Shall we make this private?

backend/internal/scim/handler_test.go
```diff
@@ -0,0 +1,328 @@
...
+
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+const testBaseURL = "https://thunder.example.com"
+
+// TestHandleServiceProviderConfigGetRequest_Success verifies that the handler
+// returns HTTP 200 with application/scim+json content type and a well-formed
+// ServiceProviderConfig body.
+func TestHandleServiceProviderConfigGetRequest_Success(t *testing.T) {
```
ThaminduDilshan, line 36: Let's follow test suite pattern. Check all the test files

backend/internal/scim/init.go
```diff
@@ -0,0 +1,128 @@
...
+	scimCfg config.SCIMConfig,
+) {
+	svc := newSCIMService(userService, entityTypeService, scimCfg)
+	h := newSCIMHandler(svc, baseURL)
+	registerSCIMRoutes(mux, h)
+}
+
+// registerSCIMRoutes registers all /scim/v2 routes using the same
+// middleware.WithCORS pattern as all other ThunderID modules.
+func registerSCIMRoutes(mux *http.ServeMux, h *scimHandler) {
```
ThaminduDilshan:
```suggestion
func registerRoutes(mux *http.ServeMux, h *scimHandler) {
```

backend/internal/scim/response.go
```diff
@@ -0,0 +1,83 @@
...
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+// scimContentType is the SCIM-specific content type required by RFC 7644.
+const scimContentType = "application/scim+json"
```
ThaminduDilshan: Shall we move this to the constants.go file?

backend/internal/scim/response.go
```diff
@@ -0,0 +1,83 @@
...
+	Schemas  []string `json:"schemas"`
+	Status   string   `json:"status"`
+	ScimType string   `json:"scimType,omitempty"`
+	Detail   string   `json:"detail,omitempty"`
+}
+
+// writeSCIMSuccessResponse writes a SCIM-compliant success response.
+// Uses application/scim+json as required by RFC 7644, and uses a
+// buffer-first pattern to avoid sending headers before encoding succeeds.
+func writeSCIMSuccessResponse(ctx context.Context, w http.ResponseWriter, data any) {
```
ThaminduDilshan: Do we need this response.go file? These methods are accessed in handler.go right? I feel like we can merge these into handler.go

backend/internal/scim/response.go
```diff
@@ -0,0 +1,83 @@
...
+	logger := log.GetLogger()
+
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(data); err != nil {
+		logger.Error(ctx, "Failed to encode SCIM response", log.Error(err))
+		w.Header().Set("Content-Type", scimContentType)
```
ThaminduDilshan: Shall we use constants from system/constants package? There should be constants already defined there.
Check other places as well

backend/internal/scim/response.go
ThaminduDilshan: In fact this can also be moved to system/constants

backend/internal/scim/schema_builder.go
```diff
@@ -0,0 +1,248 @@
...
+			Required:    false,
+			CaseExact:   true,
+			Mutability:  scimMutabilityReadOnly,
+			Returned:    scimReturnedAlways,
+			Uniqueness:  scimUniquenessServer,
+		},
+	}
+}
+
+type rawPropertyDef struct {
```
ThaminduDilshan, line 187: Add a method comment

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+	"github.com/thunder-id/thunderid/internal/entitytype"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/user"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// SCIMServiceInterface defines the SCIM service operations.
+// SCIMServiceInterface defines the SCIM service operations.
```
ThaminduDilshan, line 37: Duplicated comment

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+type SCIMServiceInterface interface {
+	GetServiceProviderConfig(ctx context.Context, baseURL string) SCIMServiceProviderConfig
+
```
ThaminduDilshan: Shall we remove unnecessary empty lines? Fine to have them if we're trying to increase resadability, maybe if we're having comments.
But feels unnecessary here

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+}
+
+// computeSCIMConfigVersion produces a stable weak ETag value from the SCIM
+// config JSON. The format follows RFC 7232 weak validator convention: W/"<value>".
+// It changes whenever an operator toggles a capability flag, ensuring SCIM
+// clients can detect ServiceProviderConfig changes via conditional GET.
+func computeSCIMConfigVersion(cfg config.SCIMConfig) string {
+	b, err := json.Marshal(cfg)
+	if err != nil {
+		panic(fmt.Sprintf("scim: failed to marshal SCIM config for ETag generation: %v", err))
```
ThaminduDilshan, line 114: Panic inside services will cause the server to stop/ restart disrupting all other operations. Is this the intended behaviour here?

Or should we just log an error log and allow rest of the services to continue. I'm bit unclear on the intention of this dynamically adjustable config version here.

Also how these scim configs get changed dynamically at runtime? Service is initialized and started at server startup. Once started, a singleton service will keep running in the server. I don't see a way for this to get updated in the current implementation.

Open to discussion

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+// operations to existing ThunderID services.
+type scimService struct {
+	userService       user.UserServiceInterface
+	entityTypeService entitytype.EntityTypeServiceInterface
+	cfg               config.SCIMConfig
+
+	// configVersion is a short deterministic hash of the SCIM config used
+	// as the ETag/version value. It changes only when operator toggles a
+	// capability flag — not on every server restart.
+	configVersion string
```
ThaminduDilshan, line 65: Let's also add logger here aligning with one of my previous comment. That way we don't need to initialize logger inside every service method

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+func computeSCIMConfigVersion(cfg config.SCIMConfig) string {
+	b, err := json.Marshal(cfg)
+	if err != nil {
+		panic(fmt.Sprintf("scim: failed to marshal SCIM config for ETag generation: %v", err))
+	}
+	h := sha256.Sum256(b)
+	return fmt.Sprintf("W/%q", hex.EncodeToString(h[:8]))
+}
+
+func (s *scimService) GetServiceProviderConfig(_ context.Context, baseURL string) SCIMServiceProviderConfig {
```
ThaminduDilshan, line 122: Add method comments. Check other places as well

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+		Meta: meta,
+	}
+}
+
+// listUserEntityTypeNames paginates through all user-category entity types and
+// returns a flat slice of their names.
+//
+// This is the single authoritative pagination loop for entity type name discovery.
+// ListSchemas uses it to avoid duplicating pagination logic.
+func (s *scimService) listUserEntityTypeNames(ctx context.Context) ([]string, *tidcommon.ServiceError) {
```
ThaminduDilshan: Shall we re-order service file so that public methods appear first?

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+	return SCIMSchemaListResponse{
+		Schemas:      []string{SCIMListResponseSchemaURN},
+		TotalResults: len(schemas),
+		StartIndex:   1,
+		ItemsPerPage: len(schemas),
+		Resources:    schemas,
+	}, nil
+}
+
+func (s *scimService) GetSchema(
```
ThaminduDilshan, line 218: Add method comments. Check other places as well.

@coderabbitai any idea why this hasn't been detected by the go linter? Provide a short reply. Send a PR if it's bug with linter configs

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+	names := make([]string, 0, 16)
+	offset := 0
+	for {
+		page, svcErr := s.entityTypeService.GetEntityTypeList(
+			ctx, entitytype.TypeCategoryUser, serverconst.MaxPageSize, offset, false,
+		)
+		if svcErr != nil {
+			logger.Error(ctx, "Failed to list entity types",
+				log.Int("offset", offset), log.Any("error", svcErr))
+			return nil, svcErr
```
ThaminduDilshan: We should wrap this with a scim service error and return. It's not appropriate a service to return a different service's error.
Check other places too

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+			log.String("resolvedUserTypeName", userTypeName),
+		)
+		return nil, &ErrorSchemaNotFound
+	}
+
+	et, svcErr := s.entityTypeService.GetEntityTypeByName(
+		ctx, entitytype.TypeCategoryUser, entityTypeName,
+	)
+	if svcErr != nil {
+		if svcErr.Code == tidcommon.ErrorUnauthorized.Code {
```
ThaminduDilshan: Can there be a unauthorized error getting throwed at this level?

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+		logger.Debug(ctx, "Entity type not found for SCIM schema URN",
+			log.String("urn", schemaURN),
+			log.String("resolvedUserTypeName", entityTypeName),
+		)
+		return nil, &ErrorSchemaNotFound
+	}
+
+	scimSchema, err := mapEntityTypeToSCIMSchema(*et, baseURL)
+	if err != nil {
+		// The entity type exists but its schema JSON is malformed — this is a
```
ThaminduDilshan: Shall we remove the comment? We don't need to add unnecessary comments

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+		return nil, &tidcommon.InternalServerError
+	}
+
+	return &scimSchema, nil
+}
+
+// ListResourceTypes returns all SCIM resource types supported by ThunderID.
+// ThunderID only exposes a single "User" resource type. The schemaExtensions
+// array is built dynamically — one entry per registered user-type entity type.
+// RFC 7643 §6, RFC 7644 §4
```
ThaminduDilshan: Shall we recheck these comments? Comments should end with a fullstop. But we have mentioned spec sections afterwards.
Noticed this in multiple places

backend/internal/scim/service.go
```diff
@@ -0,0 +1,421 @@
...
+			Location:     location,
+			// ResourceType definitions are server-managed and never mutated by clients.
+			// Reuse the same stable timestamp constant used by ServiceProviderConfig.
+			Created:      scimServiceProviderConfigCreated,
+			LastModified: scimServiceProviderConfigCreated,
+		},
+	}, nil
+}
+
+func (s *scimService) resolveEntityTypeNameForSchemaURN(
```
ThaminduDilshan, line 400: Add method comment

backend/internal/scim/service_test.go
```diff
@@ -0,0 +1,1287 @@
...
+		testGenericBaseURL,
+	)
+
+	require.Nil(t, svcErr)
+	require.NotNil(t, result)
+	require.Equal(t, "Person", entityTypeSvc.requestedName)
+	require.Equal(t, "urn:thunderid:params:scim:schemas:person:2.0:User", result.ID)
+}
+
+type caseSensitiveEntityTypeService struct {
```
ThaminduDilshan: If possible shall we use mockery to generate mocks rather than writing custom mock implementations in tests?

backend/internal/system/config/config.go
```diff
@@ -433,6 +446,7 @@ type Config struct {
 	Email                EmailConfig                      `yaml:"email"                 json:"email"`
 	Notification         NotificationConfig               `yaml:"notification"          json:"notification"`
 	Consent              engineconfig.ConsentConfig       `yaml:"consent"               json:"consent"`
+	SCIM                 SCIMConfig                       `yaml:"scim" json:"scim"`
```
ThaminduDilshan: If we're adding new configs, we should add these to the helm configs too. But let's first decide whether these are actually required

backend/internal/scim/constants.go
ravindu439: Will fix — the // Package scim implements... doc comment will be moved to init.go only and removed from constants.go.

backend/internal/scim/constants.go
ravindu439: `scimServiceProviderConfigCreated` and `scimCoreUserSchemaCreated` are both set to `"2025-01-01T00:00:00Z"` as static `meta.created` timestamps for resources that are treated as static/immutable per RFC 7644 (`ServiceProviderConfig` and the core User schema). These are never mutated, so a static creation time was used. That said, I agree it's confusing to hardcode an arbitrary date — will replace this with the server's actual startup time captured at Init(), which is more meaningful and still stable for the lifetime of a running instance.

backend/internal/scim/constants.go
ravindu439: Good suggestion. Will refactor `constants.go` to:

- Introduce typed string types: e.g. type SCIMAttrType string, type SCIMMutability string, type SCIMReturned string, type SCIMUniqueness string
- Split into grouped const blocks with clear section headers (Schema URNs, Attribute Types, Mutability, Returned, Uniqueness, Resource Type metadata)

backend/internal/scim/error_constants.go
ravindu439: Will remove the //nolint directive and fix the underlying lint issues instead.

backend/internal/scim/handler.go
ravindu439: Will remove it.

backend/cmd/server/config/default.json
ravindu439: These config flags exist because RFC 7643 §5 requires `ServiceProviderConfig` to advertise server capabilities (patch, bulk, filter, sort, etc.) as concrete boolean values — SCIM clients read these before making provisioning calls. Wiring them to config allows operators to truthfully reflect support without a code change when features are implemented.
Agreed that placing them in default.json this early is premature since all values are false anyway. Will move this into the SCIM package-level config and remove from default.json .

backend/internal/scim/response.go
ravindu439: will do .

backend/internal/scim/service.go
ravindu439: The panic is safe here — `computeSCIMConfigVersion` is called only once at server startup inside `newSCIMService,` not during request handling, so it cannot disrupt running operations. `SCIMConfig` is a flat struct of `booleans` and integers so `json.Marshal` cannot fail in practice; the panic is purely a guard against a future programming error. Also you are correct that `configVersion` does not change at runtime — the struct field comment was misleading, will fix that. We cannot silently fall back on marshal failure because we advertise `ETagSupported: true` and a broken ETag would cause incorrect conditional GET behavior per RFC 7643 §3.1.

backend/cmd/server/config/default.json
ThaminduDilshan: I agree that we need to expose them via scim ServiceProviderConfig, but do we need to allow turning on/off them? We can keep this scim internal and set false to whichever configs we don't support.
Can set inidividual fields to true when implementing the related operations

ravindu439: agree , Will move this into the SCIM package-level config and remove from default.json  .

backend/internal/scim/service.go
```diff
@@ -0,0 +1,423 @@
...
+		FilterSupported:         scimconfig.FilterSupported,
+		FilterMaxResults:        scimconfig.FilterMaxResults,
+		ChangePasswordSupported: scimconfig.ChangePasswordSupported,
+		SortSupported:           scimconfig.SortSupported,
+		ETagSupported:           scimconfig.ETagSupported,
+	}
+	b, err := json.Marshal(state)
+	if err != nil {
+		panic(fmt.Sprintf("scim: failed to marshal SCIM config for ETag generation: %v", err))
+	}
```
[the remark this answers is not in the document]
ravindu439: The panic is safe here — `computeSCIMConfigVersion` is called only once at server startup inside `newSCIMService,` not during request handling, so it cannot disrupt running operations. `SCIMConfig` is a flat struct of `booleans` and integers so `json.Marshal` cannot fail in practice; the panic is purely a guard against a future programming error. Also you are correct that `configVersion` does not change at runtime — the struct field comment was misleading, will fix that. We cannot silently fall back on marshal failure because we advertise `ETagSupported: true` and a broken ETag would cause incorrect conditional GET behavior per RFC 7643 §3.1.

PR #3319, opened 2026-06-15
Add multi-auth mode support to Wayfinder sample

Dilusha-Madushan:
### Purpose

The Wayfinder sample app previously supported only redirect-based authentication (OAuth2 `authorization_code` flow via the Login Gate). This PR extends it to demonstrate two additional embedded authentication modes, making the sample a reference implementation for all three auth patterns ThunderID supports.

### Approach

Three auth modes, selected at build time via environment variables:

| Mode | VITE_AUTH_IS_REDIRECT_BASED | VITE_AUTH_IS_VERBOSE | Description |
|--------|--------|--------|--------|
| Redirect | `true` (default) | — | OAuth2 `authorization_code` flow through the Login Gate |
| App-native custom UI | `false` | `false` | Direct `/flow/execute` calls with a hand-rolled form UI |
| App-native verbose SDK | `false` | `true` | SDK `SignIn`/`SignUp` components render backend-driven component trees |

##### `auth/config.js`

Single source of truth for mode flags and the shared `SCOPES` array used across all three modes.

##### `NativeAuthContext` + `nativeAuthService`

Lightweight token store and raw API helpers for the two non-redirect modes.

The context exposes:

```javascript
{
  token,
  setToken,
  clearToken
}
```

This allows all protected pages to read the access token without knowing which native mode is active.

##### `AuthPage.jsx`

Single route (`/signin`, `/signup`, `/recovery`) that fans out to the correct auth UI based on the mode flags, keeping `App.jsx` routing unchanged.

##### `NativeAuthPage`

Bypasses the SDK entirely and talks directly to:

- `/flow/execute`
- `/oauth2/token` (token-exchange grant)

This demonstrates the minimal integration surface required for native authentication.

##### `NativeVerboseAuthPage`

Uses the SDK's:

- `SignIn`
- `SignUp`
- `Recovery`
- `AcceptInvite`

components.

Registers custom `rich_text_*` renderers via `nativeVerboseAuthExtensions` to convert backend-managed forgot-password and sign-up links into SPA `navigate()` calls instead of full-page navigations.

##### Separate `thunderid-config/` Directories

`redirect/` and `app-native/` each contain a self-contained:

- `thunderid-config.yaml`
- `thunderid.env`

This allows each mode to be bootstrapped independently without modifying a shared configuration file.

The `app-native` configuration additionally enables:

- `urn:ietf:params:oauth:grant-type:token-exchange`
- A registration flow with automatic sign-in

The registration flow uses:

```text
AuthorizationExecutor → AuthAssertExecutor
```

so the returned assertion can be immediately exchanged for an access token after sign-up.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3003

## Summary by CodeRabbit

* **New Features**
  * Added native authentication support with dedicated **/signin**, **/signup**, and **/recovery** pages, plus optional verbose sign-in rendering.
  * Added a native auth experience with JWT-based session storage, token acquisition, and resuming in-progress flows.
  * Updated auth-aware app routing and page chrome (hides header/chat/footer on auth pages).
* **Bug Fixes**
  * Unified sign-in/sign-out behavior across authenticated pages, removing Thunder-specific session-clearing/redirect handling.
* **Style**
  * Added styling for the native authentication UI.
* **Chores**
  * Updated sample environment variables and Thunder/Wayfinder auth configuration; refreshed the ThunderID React dependency override.


samples/apps/wayfinder-sample/frontend/.env.example
```diff
@@ -2,3 +2,13 @@
 # This is the WAYFINDER application — separate from the chat agent's client.
 VITE_THUNDER_CLIENT_ID=WAYFINDER
 VITE_THUNDER_BASE_URL=https://localhost:8090
+
+# Application ID used for embedded /flow/execute calls. Defaults to VITE_THUNDER_CLIENT_ID.
+VITE_THUNDER_APP_ID=wayfinder-app
+
+# Set to "false" to use embedded app-native flows instead of the Login Gate redirect.
+VITE_AUTH_IS_REDIRECT_BASED=true
```
brionmario, line 16: Can't we infer this from some other config? Ex: If `ClientID` is defined, or a `SignInPath` is defined?

Dilusha-Madushan: clientId is shared across all modes so it can't infer this. We could consolidate the two boolean flags into a single VITE_AUTH_MODE=redirect|native|verbose.

Dilusha-Madushan: ClientId is required for the app native flows, because we are exchanging the token with assertion.

samples/apps/wayfinder-sample/thunderid-config/redirect/thunderid-config.yaml
```diff
@@ -2,9 +2,9 @@
 id: wayfinder-customer-type
 category: user
 name: Customer
-ouHandle: default
-allowSelfRegistration: true
-systemAttributes:
+ou_handle: default
```
ThaminduDilshan: This file location change will require changes to the build scripts (sh and ps1) and console try it out config import implementation. Otherwise console try it out will get break. Check b2c, ai and mcp all three.

[1] https://github.com/thunder-id/thunderid/pull/3125/changes#diff-4d2a8eefdf2a9783512a35da4dc7676a66404b6f3826a8af9aad038722da6823R155
[2] https://github.com/thunder-id/thunderid/pull/3125/changes#diff-62bc756f9158a974b56ad92851fcc30e981f0a3ebc83a297ff2e3342058a2368

Dilusha-Madushan: Updated WAYFINDER_BUNDLE_KEY in WayfinderConfigImport.tsx from 'wayfinder' to 'wayfinder/redirect' — this fixes the console try it out for b2c, ai, and mcp, all of which share that component.

The build scripts (build.sh / build.ps1) didn't need changes — both already use a recursive copy of the thunderid-config/ directory, so the new redirect/ subdirectory is picked up automatically.

samples/apps/wayfinder-sample/frontend/package.json
```diff
@@ -13,7 +13,7 @@
     "react-router": "6.30.4"
   },
   "dependencies": {
-    "@thunderid/react": "0.0.5",
+    "@thunderid/react": "^0.3.0",
```
ThaminduDilshan: Is this intentional?

Dilusha-Madushan: Yes, Updated @thunderid/react from workspace:^ to 0.3.0. This branch includes SDK-side changes that require this version. Was previously resolving from the workspace during development, but since the JS SDKs were moved out of the monorepo, it now references the published version directly.

ThaminduDilshan: But AFAIK we are trying to avoid using carrot as much as possible with external dependencies. Better to align the same with internal dependencies too.
WDYT @brionmario @DonOmalVindula ?

Dilusha-Madushan: Changed "^0.3.0" to "0.3.0"

ThaminduDilshan (approved):

samples/apps/wayfinder-sample/thunderid-config/redirect/thunderid-config.yaml
```diff
@@ -226,11 +226,11 @@ nodes:
...
         action:
           ref: action_credentials
-          nextNode: credentials_auth
-  - id: credentials_auth
+          nextNode: basic_auth
+  - id: basic_auth
     type: TASK_EXECUTION
     executor:
-      name: CredentialsAuthExecutor
+      name: BasicAuthExecutor
```
ThaminduDilshan: This change is incorrect right? Shall we fix it?

samples/apps/wayfinder-sample/thunderid-config/app-native/thunderid-config.yaml
```diff
@@ -0,0 +1,1849 @@
...
+            identifier: password
+            type: PASSWORD_INPUT
+            required: true
+        action:
+          ref: action_credentials
+          nextNode: basic_auth
+  - id: basic_auth
+    type: TASK_EXECUTION
+    executor:
+      name: BasicAuthExecutor
```
ThaminduDilshan: Need to update here as well

ThaminduDilshan (changes requested):

samples/apps/wayfinder-sample/thunderid-config/app-native/thunderid-config.yaml
Dilusha-Madushan: Addressed

@coderabbitai check this

samples/apps/wayfinder-sample/thunderid-config/redirect/thunderid-config.yaml
Dilusha-Madushan: Addressed

@coderabbitai check this

ThaminduDilshan (dismissed):

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #3320, opened 2026-06-15
verified_claims support in claims parameter of authorization request

### LINKED ISSUE #3315: Accept and Store `verified_claims` in the Authorize `claims` Parameter
## Summary

Accept the OIDC Identity Assurance (IDA) `verified_claims` member in the `userinfo`
section of the authorize `claims` parameter, store it in the authorization request
store, and ignore it everywhere else.

`ClaimsRequest.UserInfo` (currently `map[string]*IndividualClaimRequest`) cannot hold a
`verified_claims` value, whose request form is the nested IDA structure
(`verification` + `claims`), not the `{essential, value, values}` shape. Widening
`UserInfo` to `map[string]any` lets it carry both normal claims and `verified_claims`.

This is accept-and-store only: Thunder never resolves, verifies, or returns
`verified_claims`.

## Goals

- Change `ClaimsRequest.UserInfo` to `map[string]any`.
- Validate `userinfo.verified_claims` against the IDA request structure; keep existing
  OIDC validation for normal claims.
- Persist `verified_claims` in `auth_req_store` (DB JSONB + cache) and reconstruct it
  faithfully on retrieval.
- Keep all other consumers (essential/optional classification, claim filtering,
  sub-claim constraint, UserInfo/token building) operating only on normal claims.

## Non-goals

- No `verified_claims` in the `id_token` section; `ClaimsRequest.IDToken` is unchanged.
- No resolution, verification, or return of `verified_claims` in any output.
- No new data model, schema change, or client config for verified data
  (`verified_claims` rides inside the existing `REQUEST_DATA` JSONB).

## Acceptance criteria

1. `ClaimsRequest.UserInfo` is `map[string]any`; `IDToken` unchanged; existing tests
   pass after consumers are adapted.
2. A `claims` parameter with `userinfo.verified_claims` is accepted and the value is
   retained in the parsed `ClaimsRequest`.
3. Full IDA validation: malformed `verified_claims` (missing `verification` /
   `trust_framework` / `claims`) is rejected with `invalid_request`; both object and
   array-of-objects forms are accepted when valid. Normal-claim validation is unchanged.
4. Round-trip: after persist + reload (DB and cache), the reconstructed `verified_claims`
   equals the submitted one.
5. `verified_claims` never appears in required-attribute sets, `BuildClaims` output, the
   UserInfo response, ID token, access token, or sub-claim constraint checks. Responses
   for normal claims are identical with or without `verified_claims` present.
6. Unit tests cover parsing, IDA validation (valid/invalid/array), the store round-trip,
   and the "ignored everywhere else" guarantees.

## Known edge cases / risks / dependencies

- **Normalization**: with `UserInfo` as `map[string]any`, every non-`verified_claims`
  entry must be re-decoded into `*IndividualClaimRequest` wherever the old code iterated
  typed values. A `null` value (`"email": null`) still means "requested, no constraint".
  Centralize the `verified_claims` exclusion in one helper to keep the no-leakage
  guarantee robust.
- **Shape variants**: accept single-object and array-of-objects `verified_claims`;
  aggregated/distributed forms are out of scope (stored opaquely).
- **Round-trip fidelity**: `parseClaimsRequestFromJSON` must avoid lossy coercion (e.g.
  numeric handling) so the reloaded value matches the original.
- **Wider consumer surface**: the type change touches every reader of `UserInfo`; a
  missed site either fails to compile or mis-handles normal claims — covered by tests.
- **Dependency**: OIDC IDA spec for the `verified_claims` request structure; existing
  `auth_req_store` and claims parse/validate utilities
  (`oauthutils.go`, `model/parameter.go`).

sacrana0:
### Purpose
Support verified_claims in the claims.userinfo parameter in authorize endpoint

### Related Issues
- #3315 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added support for OIDC Identity Assurance `verified_claims` in both userinfo and ID token requests, including trust framework, evidence/claim constraints, and time constraints.
* **Bug Fixes**
  * Improved `verified_claims` handling so they’re validated and preserved separately from normal claim resolution, and ensured access-token claim inclusion excludes `verified_claims`.
* **Tests**
  * Added extensive parsing, validation, marshal/unmarshal, and serialize/round-trip persistence tests for `verified_claims` (including object and array forms) and related error cases.

### DIFF
--- a/backend/internal/oauth/oauth2/model/parameter.go
+++ b/backend/internal/oauth/oauth2/model/parameter.go
@@ -18,7 +18,14 @@
 
 package model
 
-import "github.com/thunder-id/thunderid/internal/system/utils"
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/system/utils"
+)
 
 // OAuthParameters represents the parameters required for OAuth2 authorization.
 type OAuthParameters struct {
@@ -39,10 +46,26 @@ type OAuthParameters struct {
 	DPoPJkt             string
 }
 
+// VerifiedClaimsMember is the OIDC Identity Assurance member name that may appear in the
+// userinfo and id_token sections of the claims request. Its request form is the nested IDA
+// structure (verification + claims) rather than the {essential, value, values} shape, so each
+// section's verified_claims is held separately in VerifiedUserInfo and VerifiedIDToken and
+// excluded from all normal-claim processing.
+const VerifiedClaimsMember = "verified_claims"
+
+// jsonNull is the JSON wire representation of a null value.
+const jsonNull = "null"
+
 // ClaimsRequest represents the OIDC claims request parameter structure.
+// UserInfo and IDToken carry only normal claims; each section's verified_claims IDA structure is
+// held separately in VerifiedUserInfo and VerifiedIDToken, normalized to an array of typed
+// VerifiedClaimsRequest entries. On the wire verified_claims remains nested under its
+// userinfo/id_token object, handled by the custom MarshalJSON/UnmarshalJSON.
 type ClaimsRequest struct {
-	UserInfo map[string]*IndividualClaimRequest `json:"userinfo,omitempty"`
-	IDToken  map[string]*IndividualClaimRequest `json:"id_token,omitempty"`
+	UserInfo         map[string]*IndividualClaimRequest `json:"-"`
+	VerifiedUserInfo []*VerifiedClaimsRequest           `json:"-"`
+	IDToken          map[string]*IndividualClaimRequest `json:"-"`
+	VerifiedIDToken  []*VerifiedClaimsRequest           `json:"-"`
 }
 
 // IndividualClaimRequest represents a request for an individual claim.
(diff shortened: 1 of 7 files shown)


backend/internal/oauth/oauth2/model/parameter_test.go
```diff
@@ -0,0 +1,111 @@
+/*
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/oauth/oauth2/model/parameter_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestClaimsRequest_UnmarshalJSON_NormalizesUserInfo verifies that a plain json.Unmarshal
+// (the path used by the Redis-backed request/code/PAR stores) normalizes userinfo entries
+// into *IndividualClaimRequest so UserInfoClaims() returns them rather than silently dropping
+// them as raw map[string]interface{} values.
+func TestClaimsRequest_UnmarshalJSON_NormalizesUserInfo(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -52,6 +65,78 @@ type IndividualClaimRequest struct {
...
+		return nil, err
+	}
+	return &icr, nil
+}
+
+// UserInfoClaims returns the normal (non-verified_claims) userinfo claim requests.
+// verified_claims is excluded so downstream claim resolution never sees it. Values are
+// normalized to *IndividualClaimRequest by UnmarshalJSON on every decode path;
+// a requested claim with no constraint ("email": null) maps to a nil entry.
+func (cr *ClaimsRequest) UserInfoClaims() map[string]*IndividualClaimRequest {
```
ThaminduDilshan: Isn't this confusing? Method name is generic and expected usage is only for the normal claims handling. I'd suggest having a specific method to get normal claims and if required have another method to get verified claims. 
I.e. `GetNormalUserInfoClaims()` & `GetVerifiedUserInfoClaims()`. Or else having `GetUserInfoClaims()` is also fine if we have another method to retrieve verified claims too (`GetVerifiedUserInfoClaims`).

Also see whether the existing pattern across code base is `UserInfo` or `Userinfo`

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -39,9 +44,17 @@ type OAuthParameters struct {
...
+// opaquely alongside normal claims and excluded from all normal-claim processing.
+const VerifiedClaimsMember = "verified_claims"
+
 // ClaimsRequest represents the OIDC claims request parameter structure.
+// UserInfo is map[string]any so it can carry both normal claims (each decoded as
+// *IndividualClaimRequest) and the opaque verified_claims IDA structure.
 type ClaimsRequest struct {
-	UserInfo map[string]*IndividualClaimRequest `json:"userinfo,omitempty"`
+	UserInfo map[string]any                     `json:"userinfo,omitempty"`
 	IDToken  map[string]*IndividualClaimRequest `json:"id_token,omitempty"`
```
ThaminduDilshan: Any specific reason to allow this only for the user info field, but not id token?

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -39,9 +44,17 @@ type OAuthParameters struct {
...
+// (verification + claims) rather than the {essential, value, values} shape, so it is carried
+// opaquely alongside normal claims and excluded from all normal-claim processing.
+const VerifiedClaimsMember = "verified_claims"
+
 // ClaimsRequest represents the OIDC claims request parameter structure.
+// UserInfo is map[string]any so it can carry both normal claims (each decoded as
+// *IndividualClaimRequest) and the opaque verified_claims IDA structure.
 type ClaimsRequest struct {
-	UserInfo map[string]*IndividualClaimRequest `json:"userinfo,omitempty"`
+	UserInfo map[string]any                     `json:"userinfo,omitempty"`
```
ThaminduDilshan: With the current implementation, this silently accepts malformed entries — {"email": {"foo": "bar"}} decodes to an empty IndividualClaimRequest{} instead of being rejected. We need a validation layer in ParseClaimsRequest that strictly decodes each entry:

- verified_claims → IDA struct
- other keys → *IndividualClaimRequest (allowing null)
- reject anything else

ThaminduDilshan: Better alternative will be to consider `map[string]json.RawMessage` instead. It lets us decode strictly in one step with json.Decoder + DisallowUnknownFields(), and avoids any lossy intermediate decode (numbers coerced to float64, etc.) and the current marshal→unmarshal round-trip.

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -52,6 +65,78 @@ type IndividualClaimRequest struct {
...
+	if err := json.Unmarshal(data, &alias); err != nil {
+		return err
+	}
+	*cr = ClaimsRequest(alias)
+
+	for name, raw := range cr.UserInfo {
+		if name == VerifiedClaimsMember {
+			continue
+		}
+		claimReq, err := DecodeIndividualClaimRequest(raw)
```
ThaminduDilshan: This also only decodes individual claims requests right? Better to decode it in a general way such that both normal and verified claims are accepted.

Related to: https://github.com/thunder-id/thunderid/pull/3320/changes#r3422673948

backend/internal/oauth/oauth2/model/parameter.go
sacrana0: In eSignet we don't look for any requested claim in idtoken. idtoken is allowed but we don't honor any values in it. So didn't modify the idtoken handling. Should we add it there as well? @ThaminduDilshan

backend/internal/oauth/oauth2/model/parameter.go
sacrana0: The existing impl also didn't validate the fields properly. But yeah let's add complete validation for all fields.

sacrana0: How about we separate out normal and verified claims in separate fields? @ThaminduDilshan

type ClaimsRequest struct {
      UserInfo       map[string]*IndividualClaimRequest // normal claims, strictly decoded
      VerifiedUserInfo json.RawMessage                  // opaque IDA blob
      IDToken        map[string]*IndividualClaimRequest
  }

backend/internal/oauth/oauth2/model/parameter.go
ThaminduDilshan: But for thunderid it's better to align this correctly as it's a separate product. I prefer to do this for the id token as well if possible

backend/internal/oauth/oauth2/model/parameter.go
ThaminduDilshan: But in the existing implementation JSON decoder only decodes it in to the defined format. Since it doesn't have any/ interface{} type fields, it will be decoded correctly discarding any unknown fields.
But with this, it will accept basically anything right?

ThaminduDilshan: > How about we separate out normal and verified claims in separate fields?

This is also one approach I considered during initial discussion. I'm okay with this, but we'll need to override the json decode method to decode it correctly and pass to the defined struct.

sacrana0:
@coderabbitai full-review   check if all comments concerns are resolved

backend/internal/oauth/oauth2/model/parameter.go
```diff
@@ -52,9 +72,315 @@ type IndividualClaimRequest struct {
...
+	MaxAge *int64 `json:"max_age,omitempty"`
+}
+
+// UnmarshalJSON decodes a single verified_claims entry, requiring a verification element and a
+// non-empty claims object. Each claims entry follows the normal-claim constraint grammar.
+func (vcr *VerifiedClaimsRequest) UnmarshalJSON(data []byte) error {
+	var raw struct {
+		Verification *VerificationRequest       `json:"verification"`
+		Claims       map[string]json.RawMessage `json:"claims"`
+	}
```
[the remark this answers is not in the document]
sacrana0: @ThaminduDilshan What is your view on this? There are many more fields possible in verified_claims, do we add everything in the struct and throw on unknown fields or  only keep minimum fields in struct that we actually use and ignore other known/unknown fields?

backend/internal/oauth/oauth2/utils/oauthutils.go
```diff
@@ -198,26 +200,9 @@ func validateClaimsRequest(cr *model.ClaimsRequest) error {
...
-
-	// value and values are mutually exclusive
-	if icr.Value != nil && len(icr.Values) > 0 {
-		return fmt.Errorf(
-			"invalid claims parameter: claim '%s' in %s has both 'value' and 'values' specified "+
-				"(mutually exclusive per OIDC spec)",
-			claimName, location)
+	if err := icr.Validate(); err != nil {
+		return fmt.Errorf("invalid claims parameter: claim '%s' in %s %w", claimName, location, err)
 	}
```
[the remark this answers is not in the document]
sacrana0: It doesn't seem awkward to me. The sentence is fine.

backend/internal/oauth/oauth2/model/parameter.go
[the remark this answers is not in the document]
ThaminduDilshan: We can only scope to what's supported as of now. Still it allows us to introduce new fields without breaking the existing implementation right?
Also the implementation should accept unsupported fields still without failing the request. As long as that's satisified, we can keep definition limited to the supported fields

ThaminduDilshan (approved):

PR #3354, opened 2026-06-17
Preserve sub claim during IDP attribute mapping

sadilchamishka:
### Purpose

Fix IDP external→local attribute mapping so the OIDC `sub` claim is propagated correctly, and align the IDP API examples with the canonical default user type.

Two related mapping problems:

1. **Mapping layer (shared by token exchange and federated authentication):** `ApplyAttributeMappings` consumed any external attribute used as a mapping source, removing it from the resulting claims. For the reserved `sub` claim this meant it could be mapped to a target but was then dropped, so you could never keep `sub` as-is *and* map it.
2. **Token exchange subject-token path:** user attributes were extracted *before* mappings were applied (`ApplyAttributeMappings(ExtractUserAttributes(claims), ...)`). `ExtractUserAttributes` strips reserved JWT claims including `sub`, so by the time mappings ran `sub` was already gone — a `sub → picture` mapping produced nothing, while `org_name → picture` worked.

Additionally, the `api/idp.yaml` examples referenced the default user type as lowercase `"person"`, which does not match the `Person` user type created at bootstrap.

### Approach

- **`idp.ApplyAttributeMappings`** ([backend/internal/idp/utils.go](backend/internal/idp/utils.go)): never treat `sub` as a consumable mapping source. `sub` always passes through under its original key in addition to any attributes it is mapped to, so `sub → username` and `sub → email` yield `sub`, `username`, and `email` all carrying the subject value. Non-`sub` sources keep the existing rename-and-consume behavior.
- **`extractSubjectTokenClaims`** ([backend/internal/oauth/oauth2/tokenservice/validator.go](backend/internal/oauth/oauth2/tokenservice/validator.go)): apply mappings against the full claim set first, then filter reserved claims — `ExtractUserAttributes(idp.ApplyAttributeMappings(claims, attributeMappings))`. This makes reserved claims like `sub` available as mapping sources. The issued token's `sub` continues to come from the subject token's `sub` independently.
- **`api/idp.yaml`**: use `"Person"` instead of `"person"` in all attribute-configuration examples and schema field examples.

Self-issued tokens (which pass `nil` mappings) and the common case where `sub` is not mapped are unaffected.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed OIDC `sub` handling during attribute mapping so the reserved subject claim is preserved and not overwritten when mappings include `sub`.

* **New Features**
  * Reserved identity claims (including `sub`) can now be used as sources for attribute mappings during token claim processing.

* **Documentation**
  * Updated identity provider attribute configuration examples to use capitalized user type values (`"Person"` instead of `"person"`).

* **Tests**
  * Added coverage for preserving `sub` and mapping reserved claims into target attributes.

### DIFF
--- a/api/idp.yaml
+++ b/api/idp.yaml
@@ -90,9 +90,9 @@ paths:
                   isSecret: false
               attributeConfiguration:
                 userTypeResolution:
-                  default: "person"
+                  default: "Person"
                 userTypeAttributeMappings:
-                  - userType: "person"
+                  - userType: "Person"
                     attributes:
                       - externalAttribute: "given_name"
                         localAttribute: "firstName"
@@ -122,9 +122,9 @@ paths:
                     isSecret: false
                 attributeConfiguration:
                   userTypeResolution:
-                    default: "person"
+                    default: "Person"
                   userTypeAttributeMappings:
-                    - userType: "person"
+                    - userType: "Person"
                       attributes:
                         - externalAttribute: "given_name"
                           localAttribute: "firstName"
@@ -231,9 +231,9 @@ paths:
                     isSecret: false
                 attributeConfiguration:
                   userTypeResolution:
-                    default: "person"
+                    default: "Person"
                   userTypeAttributeMappings:
-                    - userType: "person"
+                    - userType: "Person"
                       attributes:
                         - externalAttribute: "given_name"
                           localAttribute: "firstName"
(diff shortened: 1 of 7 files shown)


api/idp.yaml
```diff
@@ -90,9 +90,9 @@ paths:
                   isSecret: false
               attributeConfiguration:
                 userTypeResolution:
-                  default: "person"
```
ThumulaPerera, line 93: This seems to be unrelated to the issue being fixed. Shall we separate this out to a new PR?

sadilchamishka: As the default user type is "Person" and when try out the doc it might fail. Hence fix it along with the PR

ThumulaPerera (approved):

PR #3381, opened 2026-06-18
Remove SKIP_SECURITY and create default resources in-process

### LINKED ISSUE #2860: Remove skip security flag from the product
Current SKIP_SECURITY Flag is used to create default resources in thunderID product. 
This needs to be removed while keeping the ability to create default resources. 

Suggested fix:
- Add a API Key param to the deployment.yaml
- Add a middleware to validate the API Key header for the mentioned value.
- This value needs to be securely handled in deployments

UdeshAthukorala:
### Purpose

Today the default resources (organization unit, user types, admin user, the system resource server, groups, roles, flows, the Console application, themes and translations) are created at install time in a pretty roundabout way: `setup.sh` boots a *temporary* server with `SKIP_SECURITY=true` and then `curl`s ~20 REST endpoints with no auth, after which the server is stopped and the real one is started.

The problem is `SKIP_SECURITY` itself. It's a blunt, server‑wide switch that turns off **both** authentication and authorization for every endpoint. It only exists to get around the bootstrap chicken‑and‑egg problem (you can't authenticate to create the first admin because there's no admin yet), but it's also wired into CI (`SKIP_SECURITY=true ./start.sh`) and exposed as a Helm value — so if it's ever left on, the whole management API is wide open.

This PR removes `SKIP_SECURITY` completely and seeds the default resources **in‑process** instead, through the normal service layer, using the same internal‑privilege path (`security.WithRuntimeContext`) that flow executors already use. No temporary server, no unauthenticated surface, and the server now *always* runs with security enabled.

Resolves the design discussion in #3242.
Fixes: https://github.com/thunder-id/thunderid/issues/2860

---
The `SKIP_SECURITY` environment variable and the Helm `deployment.skipSecurity` value are removed. The server no longer has any way to disable authentication/authorization.

#### 💥 Impact
Anyone setting `SKIP_SECURITY=true` (locally, in CI, or via Helm) to reach management APIs without a token will now get `401/403`. The bundled curl bootstrap scripts (`backend/cmd/server/bootstrap/01-default-resources.sh` / `.ps1`, `common.sh` / `.ps1`) are deleted.

#### 🔄 Migration Guide
- Drop `SKIP_SECURITY` from any scripts/CI and `deployment.skipSecurity` from Helm values — they no longer do anything.
- To seed resources, use the in‑process bootstrap instead of the curl scripts:
  - Packaged distribution: `./setup.sh` (unchanged), or `./start.sh --bootstrap` (seed only) / `./start.sh --bootstrap-and-serve` (seed then serve).
  - From source: `go run -C backend/cmd/server . bootstrap`.
- Custom Helm bootstrap (`bootstrap.scripts` / `bootstrap.configMap`) now takes **resource definition files** (YAML/JSON) mounted into `/opt/thunderid/bootstrap/`, not shell scripts.

---

### Approach

The core idea: add a `thunderid bootstrap` subcommand that loads a templated YAML resource bundle and hands it to the **existing import service** (`internal/system/importer`). I deliberately didn't reimplement per‑resource creation — the import engine already gives us dependency ordering, idempotent upsert, cross‑reference resolution (e.g. remapping the Console app's `authFlowId`/`registrationFlowId`), credential handling and service‑layer validation for free. The orchestrator just reads the bundle, injects a few template variables (admin creds, public URL), and calls `ImportResources` under a runtime context.

`setup.sh` is intentionally **kept** as the entry point so every call site (docker‑compose, both Helm charts, the test harnesses, docs) stays valid — only its internals changed: it now delegates to `start.sh --bootstrap`.

**Before vs after**

```mermaid
flowchart LR
  subgraph before[Before: SKIP_SECURITY=true]
    S1[setup.sh] -->|starts temp server| TS[(thunderid server\nauth DISABLED)]
    S1 -->|runs| BS[01-default-resources.sh\ncurl, no auth]
    BS -->|HTTP POST x20| TS
    S1 -->|stops| TS
  end
  subgraph after[After: no security bypass]
    S2[setup.sh\nkept] -->|start.sh --bootstrap| B[thunderid bootstrap\none-shot, no HTTP]
    B --> ORCH[bootstrap.Run\nload YAML bundle + variables]
    ORCH -->|reads| FILES[/bootstrap/01-default-resources.yaml/]
    ORCH -->|ImportResources, runtime ctx| IMP[import service\norder + upsert + ref-resolve]
    IMP --> SVC[(service layer)]
    SVC --> DB[(database)]
  end
```

**What the bootstrap one-shot does**

```mermaid
sequenceDiagram
  participant CLI as main.go (arg "bootstrap")
  participant Init as service init (DB, cache, services)
  participant Orch as bootstrap.Run
  participant Imp as import service
  participant Svc as service layer

  CLI->>Init: init config + DB + services (no HTTP listener)
  Init-->>CLI: ServiceRegistry{ Import }
  CLI->>Orch: Run(WithRuntimeContext(ctx), registry, opts)
  Orch->>Orch: load YAML bundle from bootstrap dir
  Orch->>Imp: ImportResources(content, variables, upsert=true)
  Imp->>Imp: resolve templates, parse docs, order by dependency
  loop each document (ordered)
    Imp->>Svc: create/upsert (runtime ctx -> authz granted)
  end
  Imp-->>Orch: per-document outcomes
  Orch-->>CLI: exit 0 on success, non-zero on failure
```

**Authorization after the change** — the old `IsSecuritySkipped` short‑circuit in `sysauthz` is gone; internal seeding flows through the existing runtime‑context check, and everything else goes through the real checks:

```mermaid
flowchart TD
  A[sysauthz.IsActionAllowed] --> C{IsRuntimeContext?}
  C -- yes --> G[GRANT]
  C -- no --> D{authenticated, system perm, owner, sufficient perm, policies}
  D -- ok --> G
  D -- no --> Y[DENY]
```

**Key design decisions**

- **Reuse the import service** rather than hand‑porting the 1,588‑line shell script. Less code, and ordering/upsert/reference‑resolution are already battle‑tested.
- **Single bundle file** at `backend/cmd/server/bootstrap/01-default-resources.yaml` — a multi‑document YAML (24 docs) in dependency/creation order: `organization_unit → user_type (Person + agent) → resource_server (+ resources/actions) → flow ×9 → theme ×6 → application (Console) → user (admin) → group → role → translation`. Resource docs use Go‑template placeholders; flow/theme/translation docs carry **fixed ids** so re‑runs upsert idempotently.
- **`setup.sh` kept; new `--bootstrap-and-serve`.** `start.sh` gains `--bootstrap` (seed then exit, used by `setup.sh`) and `--bootstrap-and-serve` (seed then serve — handy for local/dev). The deployment topology (db‑init → setup → server) is unchanged.

```
backend/internal/system/bootstrap/
  registry.go         # ServiceRegistry{ Import importer.ImportServiceInterface }
  orchestrator.go     # Run: load bundle -> ImportResources -> check outcomes
  orchestrator_test.go
backend/cmd/server/
  bootstrap_cmd.go    # `thunderid bootstrap` subcommand (flags/env, teardown)
backend/cmd/server/bootstrap/
  01-default-resources.yaml   # the default-resource bundle
```

| `start.sh` invocation | Behavior |
|---|---|
| `./start.sh` | Serve only |
| `./start.sh --bootstrap` | Seed default resources in‑process, then exit (used by `setup.sh`) |
| `./start.sh --bootstrap-and-serve` | Seed once, then start the server (local/dev) |

The Windows scripts (`setup.ps1` / `start.ps1` / `build.ps1`) were updated to match, and the Helm `bootstrap.scripts` / `configMap` feature was repurposed to mount resource‑definition files into the bootstrap directory.

### Related Issues
- #3242

### Related PRs
- N/A

### Checklist
- [x] Documentation provided. (Updated `docs/.../configure-and-run.mdx`, `development-pipeline.mdx`, `debugging.mdx`, `get-thunderid.mdx`, and `install/helm/README.md`.)
- [x] Tests provided.
    - [x] Unit Tests (new `bootstrap/orchestrator_test.go`; updated `security`/`sysauthz`/`cmd/server` tests)
    - [ ] Integration Tests (covered by the existing extracted‑product setup flow)

### Security checks

---

#### Verification
Backend `build` / `vet` / `gofmt` / `golangci-lint` (0 issues) and the full unit‑test suite pass. End‑to‑end against SQLite the bootstrap creates **24 resources** and is idempotent on re‑run (fixed‑id upsert), with security enabled throughout (authorized only via the runtime context). `helm template` renders cleanly with no `SKIP_SECURITY`. Windows scripts were ported but should be validated by the `windows-build-validation` CI job.

## Summary by CodeRabbit

* **New Features**
  * Added in-process bootstrap subcommand for creating default resources without disabling security.
  * Added `--bootstrap` and `--bootstrap-and-serve` modes to startup scripts for flexible initialization.

* **Bug Fixes**
  * Security enforcement is now always enabled; removed insecure `SKIP_SECURITY` bypass mechanism.

* **Documentation**
  * Updated guides and helm charts to reflect new bootstrap workflow using resource definitions instead of scripts.


.github/workflows/pr-builder.yml
```diff
@@ -661,7 +657,7 @@ jobs:
         run: |
           cd tests/e2e/server
           ./setup.sh
-          SKIP_SECURITY=true ./start.sh &
+          ./start.sh &
```
rajithacharith, line 660: do we need to pass the flag here to load the resources?

backend/cmd/server/main.go
```diff
@@ -88,7 +88,16 @@ func main() {
...
 
 	// Register the services.
-	jwtService, runtimeCryptoSvc := registerServices(mux, cacheManager)
+	jwtService, runtimeCryptoSvc, bootstrapRegistry := registerServices(mux, cacheManager)
+
+	// When invoked as the bootstrap one-shot (`thunderid bootstrap`), create the
+	// default resources in-process and exit without starting the HTTP server.
+	if isBootstrapInvocation() {
+		if !runBootstrap(ctx, logger, serverHome, bootstrapRegistry, cacheManager) {
+			os.Exit(1)
```
rajithacharith, line 90: Proper logs needs to be added to the os.Exit(1) scenario mentioning why this was failed

backend/internal/system/bootstrap/orchestrator.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/system/security"
+)
+
+// Options configures a bootstrap run.
+type Options struct {
+	// DefaultsDir is the directory holding the resource definition bundle.
+	DefaultsDir string
+	// AdminUsername and AdminPassword are the credentials for the default admin user.
+	AdminUsername string
```
rajithacharith: Do we need to handle this specially? 
Can we use {{ .ADMIN_USERNAME}} and {{.ADMIN_PASSWORD}} templating capability in the yaml and get them from the environment variables?

backend/internal/system/bootstrap/registry.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/thunder-id/thunderid/internal/system/importer"
+)
+
+// ServiceRegistry holds the services the bootstrap orchestrator needs. The import
+// service performs ordered, idempotent (upsert) creation of every resource type,
+// including cross-reference resolution (e.g. the Console application's flow IDs),
+// so it is the only dependency required. It is assembled at the composition root
+// (registerServices) so the server and the bootstrap subcommand share the wiring.
+type ServiceRegistry struct {
+	Import importer.ImportServiceInterface
```
rajithacharith: Do we need this struct? Can't we directly use importer service?

.github/workflows/pr-builder.yml
UdeshAthukorala: No, the flag isn't needed here. The previous step (`./setup.sh`) already creates the default resources via the in-process bootstrap, so the server doesn't need security disabled at startup for that.

`SKIP_SECURITY=true` was only being used so the **sample app** `/import` calls in the next step could run without authentication. Now that `SKIP_SECURITY` is removed, those imports require an `Authorization` header — so I added the `🔑 Obtain Admin Access Token` step that fetches an admin token (OAuth2 auth code + PKCE via the CONSOLE app, authenticating the bootstrapped admin user), and the `/import` and `/applications` calls now pass it as a `Bearer` token.

backend/cmd/server/main.go
UdeshAthukorala: Fixed with https://github.com/thunder-id/thunderid/compare/07b10b2e68616ef9bbca663a7d0afadae6dafe43..b068473ab0177e9527426505904aedd38e4e93b1. `runBootstrap` now returns an `error`, and `main.go` logs it before exiting: `In-process bootstrap failed; exiting` with the underlying reason (and logs success on the happy path). So the `os.Exit(1)` path always explains why.

backend/internal/system/bootstrap/orchestrator.go
UdeshAthukorala: Fixed with https://github.com/thunder-id/thunderid/compare/07b10b2e68616ef9bbca663a7d0afadae6dafe43..b068473ab0177e9527426505904aedd38e4e93b1. The orchestrator now runs the bundle through `utils.SubstituteEnvironmentVariables` (the same helper the server config and declarative resources already use), so the YAML resolves `{{ .ADMIN_USERNAME }}`, `{{ .ADMIN_PASSWORD }}` and `{{ .PUBLIC_URL }}` from the environment. The special `AdminUsername`/`AdminPassword`/`PublicURL` fields are gone — `Options` is just `DefaultsDir` now. The subcommand exports those env vars before running (flags > env > defaults; `PUBLIC_URL` defaults to the configured server URL).

backend/internal/system/bootstrap/registry.go
UdeshAthukorala: Removed with https://github.com/thunder-id/thunderid/compare/07b10b2e68616ef9bbca663a7d0afadae6dafe43..b068473ab0177e9527426505904aedd38e4e93b1. `registerServices` now returns `importer.ImportServiceInterface` directly and `bootstrap.Run(ctx, importSvc, opts)` takes it, so the single-field `ServiceRegistry` struct (and `registry.go`) is gone.

start.sh
```diff
@@ -165,10 +186,13 @@ load_env_file() {
...
-check_port $BACKEND_PORT "${PRODUCT_NAME} server"
-if [ "$DEBUG_MODE" = "true" ]; then
-    check_port $DEBUG_PORT "Debug server"
+# Check if ports are available. The bootstrap one-shot does not bind the server
+# port, so the check is skipped in that mode.
+if [ "$BOOTSTRAP_MODE" != "true" ]; then
+    check_port $BACKEND_PORT "${PRODUCT_NAME} server"
+    if [ "$DEBUG_MODE" = "true" ]; then
+        check_port $DEBUG_PORT "Debug server"
+    fi
```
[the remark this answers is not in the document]
UdeshAthukorala: Good catch — quoted `"$BACKEND_PORT"` and `"$DEBUG_PORT"` in the `check_port` calls. Fixed.

build.ps1
```diff
@@ -1580,73 +1580,35 @@ function Run {
...
+    Write-Host "⚙️  Creating default resources..."
     Write-Host ""
-    
-    # Wait for server to be ready
-    $MAX_RETRIES = 30
-    $RETRY_INTERVAL = 2
-    $retries = 0
-    
-    # Configure TLS to use modern protocols (required for HTTPS requests on Windows)
+    $env:PUBLIC_URL = $BASE_URL
```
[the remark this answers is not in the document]
UdeshAthukorala: Valid — bootstrap seeds persisted external URLs, so it should use the resolved `$PUBLIC_URL` (which respects `public_hostname`), not `$BASE_URL`. Fixed here, and applied the same fix to the equivalent line in `build.sh` to keep both scripts consistent.

docs/content/guides/getting-started/get-thunderid.mdx
```diff
@@ -85,6 +85,33 @@ Launch the <ProductName /> server:
 </TabItem>
 </Tabs>
 
+:::tip Seed and start in one step
+For local and development setups you can combine both steps with the
+`--bootstrap-and-serve` flag — it creates the default resources (only if they
+don't already exist) and then starts the server:
```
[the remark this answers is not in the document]
UdeshAthukorala: Agreed — the bootstrap upserts (create-or-update), so the create-only wording was inaccurate. Reworded to: "creates or updates the default resources as needed (idempotent, so it's safe to re-run)".

start.ps1
```diff
@@ -301,54 +336,76 @@ try {
...
+            $proc = Start-Process -FilePath $serverExecPath -ArgumentList $resourcesArgs -WorkingDirectory $scriptDir -NoNewWindow -PassThru
+        }
 
-    Write-Host "Press Ctrl+C to stop the server."
+        Write-Host ""
+        Write-Host "[INFO] Server running. PID: $($proc.Id)"
+        Write-Host ""
+        Write-Host "[INFO] Frontend Apps:"
+        Write-Host "   [GATE] Gate (Login/Register): $BACKEND_PORT/gate"
+        Write-Host "   [DEV]  Console (System Management): $BACKEND_PORT/console"
```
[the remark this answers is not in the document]
UdeshAthukorala: Fixed — the URLs now include scheme and host (`https://localhost:$BACKEND_PORT/gate` and `/console`), matching the convention already used in `start.sh`.

start.sh
```diff
@@ -47,6 +50,22 @@ while [[ $# -gt 0 ]]; do
...
+        --bootstrap-and-serve)
+            # Run the in-process bootstrap one-shot first, then start the long-running
+            # server (convenient for local/dev: seed once, then serve).
+            BOOTSTRAP_AND_SERVE=true
+            shift
+            ;;
+        --console-redirect-uris)
+            BOOTSTRAP_EXTRA_ARGS+=(--console-redirect-uris "$2")
+            shift 2
+            ;;
```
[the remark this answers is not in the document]
UdeshAthukorala: Skipping for now to stay consistent with the rest of the parser — the other value-taking flags in this script (`--debug-port`, `--port`, `--env`) don't validate `$2` either. If we want this guard, it'd be better added uniformly across all of them rather than just this one.

rajithacharith (dismissed):

rajithacharith (approved):

Issue #3260, opened 2026-06-10
Support per-application CORS allowed origins in addition to deployment.yaml


rajithacharith:
## Summary

Currently CORS allowed origins are configured globally in `deployment.yaml` only. This change adds support for per-application CORS origin configuration, merged as a **union** with the global list (a request is allowed if it matches either the global config or the application's own origins).

## Motivation

Different applications (SPAs, mobile clients, partner integrations) are hosted on different origins. Requiring all origins to be listed in `deployment.yaml` is operationally inflexible — operators need a config reload to add an origin for a single application.

### Data storage
`corsAllowedOrigins []string` stored in `InboundClient.Properties` under key `cors_allowed_origins` (same pattern as `contacts`). No new DB columns or migrations needed.

### In-memory registry (`internal/system/cors/registry.go`)
A `sync.RWMutex`-guarded `map[clientID]*Matcher` populated at server start from all persisted applications and kept in sync by the application service on create/update/delete.

```
SetAppMatcher(clientID string, origins []string) error
RemoveAppMatcher(clientID string)
GetAppMatcher(clientID string) *Matcher
```

Uses the existing `compileAll` + `newMatcher` in the `cors` package — no new compilation logic.

### Middleware (`internal/system/middleware/cors.go`)
After the global matcher check, fall through to a per-app check using `client_id` from the request query string:

```go
allow, echo := globalMatcher.Match(parsed)
if !allow {
    if clientID := r.URL.Query().Get("client_id"); clientID != "" {
        if appMatcher := cors.GetAppMatcher(clientID); appMatcher != nil {
            allow, echo = appMatcher.Match(parsed)
        }
    }
}
```

**Scope:** per-app CORS applies to requests carrying `client_id` as a query param (authorization endpoint always satisfies this). Token endpoint POST-body requests fall back to global config only.

## Files to change

| File | Change |
|---|---|
| `internal/system/cors/registry.go` | **New** — registry with Set/Get/Remove |
| `internal/application/model/application.go` | Add `CORSAllowedOrigins []string` to 6 structs |
| `internal/application/constants.go` | Add `propCORSAllowedOrigins = "cors_allowed_origins"` |
| `internal/application/service.go` | Wire field through toInboundClient, toProcessedDTO, all builders, and CRUD hooks; add `loadCORSRegistry` |
| `internal/application/init.go` | Call `loadCORSRegistry` at startup |
| `internal/system/middleware/cors.go` | Union check: global OR per-app matcher |

[...]

PR #3382, opened 2026-06-18
Add runtime-configurable server-config store with dynamic CORS

ImalshaD:
### Purpose

CORS allowed origins can currently only be set in `deployment.yaml`, which requires a server restart/reload to change. This PR introduces a **generic, runtime-mutable server-wide configuration store** — a `SERVER_CONFIG` table in `configdb`, a service, and an authorized REST API — with **dynamic CORS as its first consumer**.

- Effective allowed origins = **union** of the `deployment.yaml` baseline (boot-time) and the runtime `cors` config. The API only ever *adds* origins; existing behaviour is unchanged.
- Origins added via the API take effect **immediately, with no restart**.
- The store is generic (one row per config section, value stored as opaque JSON), so a future server-wide config is a **new row, never a schema change**.

**Note on approach (differs from the issue).** Issue #3260 proposes binding origins to individual **applications** (`InboundClient.Properties` + a per-app registry). After design discussion, this PR deliberately takes a different approach: the underlying need is a **server-wide** setting that is changeable at runtime — CORS allowed origins behave like a `deployment.yaml` setting, not an application attribute — so it is modeled as a generic server-config store rather than per-application/per-OU scoping. The rationale is captured in the design discussion on the issue.

### Approach

**Layering.** The **service** is the single, reusable entry point: the REST API and every in-process consumer (the CORS check today, future ones later) go through it. The **store** beneath it owns DB CRUD + read-through caching, and consumers never touch it directly. The generic service/store know nothing about CORS.

**Data model** (`SERVER_CONFIG` in `configdb`):
- Composite PK `(DEPLOYMENT_ID, NAME)`, no surrogate `ID` (config-table convention). Server-global — no `OU_ID`.
- Generic key → JSON value (`JSONB` on Postgres / `TEXT` on SQLite). Writes are upserts.

**Validation (the service is the policy gate), in two layers:**
- **Name gate** — a static allow-list of supported config names; unknown names are rejected.
- **Value validation (pluggable)** — a per-config validator registry. Concrete validators live in their **domain** package (CORS origin validation stays in the `cors` package) and are **registered at the composition root** (`cmd/server/servicemanager.go`). This is required, not just tidy: `cors` is already an indirect dependency of the config/cache layer, so binding the validator at the composition root keeps the generic service free of `cors` and avoids an import cycle. **Adding a new config = add an allowed name + register its validator.**
- **Atomic batch writes** — one `PUT` may carry several sections; validate *all* present sections first (one bad section ⇒ whole request rejected, no partial write), then persist in a single transaction, invalidating cache only after commit.

**Dynamic CORS (consumer of the store).** The middleware checks the boot-time global matcher (`deployment.yaml`) first; on a miss it falls through to a **nil-safe dynamic matcher** (`matchOrigin` → global, then dynamic, first hit wins). The dynamic matcher reads the cached `cors` config and **memoizes** the compiled matcher, recompiling only when the stored value changes (no regex compilation on the hot path). **Fail-safe:** on missing/invalid/uncompilable config it keeps the last-good matcher and logs, rather than dropping CORS for everyone. The middleware cannot import the config store (import cycle via route registration), so the config reader is injected as a hook from the composition root. Per-route CORS options (methods, credentials, max-age) are unchanged — only whether an origin is allowed.

**API surface** (`/server-config`, OpenAPI in `api/server-config.yaml`):
- `PUT /server-config` — upsert, multi-section, authorized.
- `GET /server-config` — list; supported sections are always present with their empty default (returns `{"cors": []}` when unset).
- `GET /server-config/{name}` — read-only single section; unsupported name → **400 SVRCFG-1001**. (Empty defaults apply here too: unset → **200 `{"cors": []}`**.)
- Empty defaults are applied at the **HTTP layer only**; the service's `GetConfig` still returns "not set", so the dynamic matcher's fail-safe is unchanged.
- Error codes: `SVRCFG-1001` unsupported name, `SVRCFG-1002` not found, `SVRCFG-1003` invalid value, `SVRCFG-1004` malformed request.

**Access control.** No new permissions. `/server-config` is not a public path, so it requires the existing root **`system`** permission, scoping management to system administrators.

**Change surface:**

| Area | Files |
|---|---|
| New module | `internal/serverconfig/` — store, cache-backed store, service, handler, init, model, validator, constants, error_constants (+ unit tests + generated mocks) |
| CORS package | `internal/system/cors/` — `compiler.go` (compile-only `CompileMatcher` + JSON unmarshal), `dynamic.go` (memoized runtime matcher + fail-safe), `validator.go` (`OriginValidator`) (+ tests) |
| Middleware | `internal/system/middleware/cors.go` — nil-safe union fall-through `matchOrigin` (+ test) |
| Composition root | `cmd/server/servicemanager.go` — register CORS validator + install dynamic matcher reader |
| Schema | `dbscripts/configdb/{postgres,sqlite}.sql` — `SERVER_CONFIG` table |
| Spec / i18n | `api/server-config.yaml` (OpenAPI); `internal/system/i18n/core/defaults.go` (SVRCFG message defaults) |
| Mocks | `.mockery.private.yml` — `serverconfig` entry (mocks generated via `make mockery`) |
| Integration tests | `tests/integration/serverconfig/` + `tests/integration/cors/` (+ static CORS baseline in `tests/integration/resources/deployment.yaml`) |

### Testing

**Unit tests** — `serverconfig` **99.1%**, `middleware` **100%**, `cors` new code **100%**.
```
cd backend
go test ./internal/serverconfig/... ./internal/system/cors/... ./internal/system/middleware/... -cover
```

**Integration tests** — split into two packages by concern, run via the harness against a live server on `:8095` with a real admin OAuth token; **13 tests, all passing**:
- `tests/integration/serverconfig/` — `ServerConfigAPITestSuite` (7): list/empty defaults; `GET /{name}` section / empty / unsupported → 1001; `PUT` invalid (`["*"]`) → 1003 (not persisted); `PUT` malformed → 1004; auth gate → 401 `AUTH-4010`.
- `tests/integration/cors/` — `DynamicCORSTestSuite` (6): dynamic origin allowed after `PUT` with no restart; union with the static baseline; regex origin; multiple origins; replace (de-registers old origin); clear (`PUT []`) revokes while baseline survives.
```
cd thunderid
make build_backend   # package the pack (only when backend code changes)
make test_integration PACKAGE=./... RUN='TestServerConfigAPITestSuite|TestDynamicCORSTestSuite'
```

**Manual runtime smoke test** — verified end-to-end on a live server: dynamic origin persisted at runtime is allowed cross-origin with no restart; static baseline preserved; `/server-config` requires auth.

### Related Issues
- Refs #3260 (see "Note on approach" above — implemented as a system-wide server-config store rather than per-application scoping).

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests
- [ ] Breaking changes. (Fill if applicable) — none.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added Server Configuration API for managing deployment settings at runtime.
  * Enabled dynamic CORS origin configuration without requiring server restart.
  * Added support for both literal and regex-pattern origin matching in CORS policies.

* **Documentation**
  * Added OpenAPI 3.0.3 specification for the Server Configuration API.


backend/internal/serverconfig/error_constants.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/thunder-id/thunderid/internal/system/error/serviceerror"
+	"github.com/thunder-id/thunderid/internal/system/i18n/core"
+)
+
+// Client errors for server-config operations.
+var (
+	// ErrorUnsupportedConfigName is the error returned when the config name is not supported.
+	ErrorUnsupportedConfigName = serviceerror.ServiceError{
+		Type: serviceerror.ClientErrorType,
+		Code: "SVRCFG-1001",
```
rajithacharith: We are following XXX-0000 pattern when defining errors. Let's follow that

backend/cmd/server/deployment.yaml
rajithacharith, line 1: Let's remove this from the commit

backend/internal/system/cors/dynamic.go
```diff
@@ -0,0 +1,103 @@
...
+// dynamicInstance holds the runtime CORS matcher. It is installed once at server start by
+// InitializeDynamicMatcher and read on every CORS request via GetDynamicMatcher.
+//
+// Like instance, it is a plain var rather than an atomic — installation runs on the main goroutine
+// before the HTTP server starts, so there is no concurrent reader/writer overlap to guard against.
+var dynamicInstance *dynamicMatcher
+
+// OriginConfigReader returns the current raw JSON origin config and whether it is set. It is supplied
+// by the configuration consumer so the cors package never imports the configuration store.
+type OriginConfigReader func() (raw []byte, ok bool)
```
rajithacharith, line 39: We can change this to 
```
type OriginConfigReader interface {
	GetConfig(ctx context.Context, name ConfigName) (json.RawMessage, *serviceerror.ServiceError)
}
```
And pass the server config service from the servicemanager.go

This way it will keep the same pattern

backend/internal/serverconfig/cache_backed_store.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/thunder-id/thunderid/internal/system/cache"
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+const serverConfigCacheName = "ServerConfigByNameCache"
+
+// cachedBackStore wraps a serverConfigStoreInterface with an in-memory cache for
+// GetServerConfigByName. Writes invalidate the affected cache entries.
+type cachedBackStore struct {
+	configCache cache.CacheInterface[*ServerConfig]
```
rajithacharith, line 33: What is the possibility of having cors specific cache?
```
	corsCache cache.CacheInterface[bool]
```

api/server-config.yaml
```diff
@@ -0,0 +1,322 @@
...
+    OAuth2:
+      type: oauth2
+      flows:
+        authorizationCode:
+          authorizationUrl: https://localhost:8090/oauth2/authorize
+          tokenUrl: https://localhost:8090/oauth2/token
+          scopes:
+            system: Access to system management APIs
+        clientCredentials:
+          tokenUrl: https://localhost:8090/oauth2/token
```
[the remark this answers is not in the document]
ImalshaD: This is fixed.

api/server-config.yaml
```diff
@@ -0,0 +1,322 @@
...
+          scopes:
+            system: Access to system management APIs
+
+  schemas:
+    CORSOriginEntries:
+      type: array
+      description: |
+        Allowed CORS origins. Each entry is either a literal origin string (or the special value
+        `null`) or a regex object. These are added to the origins configured in `deployment.yaml`.
+      items:
```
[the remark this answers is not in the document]
ImalshaD: This is fixed.

tests/integration/cors/cors_dynamic_test.go
```diff
@@ -0,0 +1,169 @@
...
+	suite.Run(t, new(DynamicCORSTestSuite))
+}
+
+func (suite *DynamicCORSTestSuite) SetupSuite() {
+	suite.adminClient = testutils.GetHTTPClient()
+	suite.plainClient = &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+		},
+	}
```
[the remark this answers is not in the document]
ImalshaD: This is fixed.

ThumulaPerera:
Added a comment regarding the general direction in the issue https://github.com/thunder-id/thunderid/issues/3260#issuecomment-4748864301

ImalshaD:
Closing the PR as Issue is closed with a Separate PR.

PR #3411, opened 2026-06-19
Resolve favicon by OS color scheme and Vite base URL

KD23243:
## Purpose
The `Head` component in the console and gate apps resolved the favicon in JavaScript from the in-app OxygenUI color scheme, coupling the tab icon to the in-app theme toggle instead of the OS / browser color scheme. It also emitted favicon paths verbatim, so relative paths failed to load when the app is served under a sub-path such as `/console` or `/gate`.

## Related Issue
- https://github.com/thunder-id/thunderid/issues/3410

### Gate
Screenshot 2026-06-19 at 16 41 44

### Console
Screenshot 2026-06-19 at 16 41 58

## Implementation
- Replaced the JS color-scheme resolution (`useColorScheme`) with two `<link rel="icon">` tags scoped to `prefers-color-scheme: light` and `prefers-color-scheme: dark`, so the browser selects the matching favicon from the OS theme and switches live when it changes.
- Added `resolveFaviconHref`, which prefixes relative favicon paths with the Vite base URL while leaving already-resolved values (absolute, protocol-relative, root-relative, and `data:` URIs) untouched.
- Applied the same change to both `frontend/apps/console/src/components/Head.tsx` and `frontend/apps/gate/src/components/Head.tsx`.
- Updated the `Head` tests in both apps to assert the two media-scoped links, the base-URL prefixing for relative paths, and that absolute URLs pass through unchanged.

## Release Notes

* **New Features**
  * Favicons now automatically switch between light and dark variants based on your system theme preference (browser `prefers-color-scheme`).

* **Bug Fixes**
  * Favicon URLs are now resolved correctly when the app is deployed under a configured base path (relative paths are base-prefixed; absolute/protocol/data URLs remain unchanged).
  * Updated and expanded favicon behavior coverage in tests for both light and dark variants.

### DIFF
--- a/frontend/apps/console/src/components/Head.tsx
+++ b/frontend/apps/console/src/components/Head.tsx
@@ -18,27 +18,37 @@
 
 import {Helmet} from '@thunderid/components';
 import {useConfig} from '@thunderid/contexts';
-import {useColorScheme} from '@wso2/oxygen-ui';
+
+/**
+ * Resolves a configured favicon path against the Vite base URL so it loads
+ * correctly when the console is served under a sub-path such as `/console`.
+ * Already-resolved values (absolute, protocol-relative or root-relative URLs,
+ * and `data:` URIs) are returned untouched.
+ */
+function resolveFaviconHref(path: string): string {
+  if (/^([a-z]+:|\/)/i.test(path)) {
+    return path;
+  }
+  return `${import.meta.env.BASE_URL.replace(/\/$/, '')}/${path}`;
+}
 
 /**
  * Manages document head metadata, specifically the favicon, with support for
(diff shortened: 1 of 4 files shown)


brionmario (approved):

brionmario:
With this change, changing the favicon when the Theme is switched has been broken.

**Expected Behaviour**

https://github.com/user-attachments/assets/782c5c39-f50c-41bd-804a-6293d346cddb

brionmario:
> With this change, changing the favicon when the Theme is switched has been broken.
> 
> **Expected Behaviour**
> 
>  Aug-11-2026.14-12-24.mp4

@coderabbitai Can you create an issue for this?

PR #3422, opened 2026-06-21
Propagate correlation ID to outbound HTTP requests

UdeshAthukorala:
### Purpose

We already generate a correlation ID (trace ID) per request and attach it to every log line via the context-aware logging work (#2038). The problem is that the ID never left the process: any time Thunder made an *outbound* HTTP call, the correlation ID sitting in the context was dropped on the floor. So the moment a request crossed into an external service (a REST authn provider, the consent service, a flow's HTTP-request node, or an SMS provider), the trace was broken and you couldn't follow a single request end to end.

I actually confirmed this by hand before fixing it. With a REST authn provider pointed at a small listener that echoes back the headers it receives, a request sent in with `X-Correlation-ID: manual-test-123` showed up in our logs and in the response header, but the outbound call to the provider arrived with **no** correlation header at all.

This PR closes that gap. Every outbound HTTP client now forwards the request's correlation ID as the `X-Correlation-ID` header, so the trace continues across the service boundary.

This addresses the outbound-propagation part of #1484. See the note under *Approach* about why the user/entity provider side is intentionally not in here.

### Approach

The trace ID is already in `context.Context` by the time any handler runs (the `CorrelationIDMiddleware` puts it there on the way in). So the fix is simply to read it back out where we build an outbound request and set it as a header — no new plumbing on the request path, since the context was already threaded through.

Concretely:

- Added a single `CorrelationIDHeaderName = "X-Correlation-ID"` constant in `system/constants` and switched the middleware's response header over to it, so the inbound and outbound sides agree on one name.
- Each outbound request builder now sets that header from `sysContext.GetTraceID(ctx)`:
  - REST authn provider (`doRequest`) — this one spot covers `Authenticate`, `GetEntityReference` and `GetAttributes`, since they all go through it.
  - Consent client (`doRequest`).
  - Flow HTTP-request executor (`executeRequest`) — only when the flow hasn't already set a header of that name, so an explicit node-configured value still wins.
  - Vonage SMS client (`sendSMS`).
- The REST authn provider accepts an optional per-provider override, `authn_provider.rest.correlation_id_header`, which defaults to the constant when empty. This lets a downstream service that expects a different header name receive it. (We considered a server-wide setting but decided against it — only the pluggable authn provider needs the knob.)

`GetTraceID` generates a fresh UUID if the context somehow doesn't have one, so the header is always populated; in a normal request it's just the same ID that came in.

Here's how the ID flows after this change:

```mermaid
sequenceDiagram
    participant C as Client
    participant MW as CorrelationIDMiddleware
    participant Ctx as request context
    participant H as Handler / Service
    participant OUT as Outbound HTTP client
    participant EXT as External service

    C->>MW: request (maybe with X-Correlation-ID)
    MW->>Ctx: WithTraceID(extracted or generated)
    MW-->>C: response header X-Correlation-ID
    MW->>H: ServeHTTP (ctx carries trace ID)
    H->>OUT: call with ctx
    Note over OUT: GetTraceID(ctx) → set X-Correlation-ID
    OUT->>EXT: outbound request + X-Correlation-ID
    EXT-->>OUT: response
```

And the four outbound paths this touches, all reading from the same context:

```mermaid
flowchart LR
    Ctx["context.Context<br/>(trace ID)"]
    Ctx --> A["REST authn provider<br/>doRequest*"]
    Ctx --> B["Consent client<br/>doRequest"]
    Ctx --> D["Flow HTTP executor<br/>executeRequest**"]
    Ctx --> E["Vonage client<br/>sendSMS"]
    A --> X1["external authn service"]
    B --> X2["consent service"]
    D --> X3["flow-configured endpoint"]
    E --> X4["Vonage API"]
    classDef ext fill:#eee,stroke:#999;
    class X1,X2,X3,X4 ext;
```
<sub>* the REST authn provider can override the header name via `authn_provider.rest.correlation_id_header`.<br/>** the flow executor does not overwrite a correlation header the flow author set explicitly.</sub>

**Why the entity provider isn't here.** #1484 was originally written against the `UserProviderInterface` from #1420. That package was later replaced by the entity provider. As it stands today the entity provider has no REST implementation (only the in-process default and a disabled variant) and its interface methods don't take a `context.Context` at all, so there's nothing outbound to tag and no context to read from. Propagating the ID there is a separate follow-up that first needs `context.Context` threaded through `EntityProviderInterface` and a REST entity provider to exist. Calling it out here so this PR isn't mistaken for fully closing #1484.

### Related Issues
- Refs thunder-id/thunderid#1484
- Background: thunder-id/thunderid#2038 (correlation IDs in logs)

### Related PRs
- Builds on the context-aware logging work (#3141, #3238, #3254)

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Outbound REST authentication, consent, flow execution, and notification requests now propagate correlation/trace IDs via an HTTP header.
  * The correlation header name is configurable (defaults to `X-Correlation-ID`) and can be customized.
* **Bug Fixes**
  * Correlation ID middleware now uses the configured header name consistently for response headers.
  * If an outbound request already specifies a correlation header, it will not be overwritten.
* **Chores**
  * Expanded automated tests and documentation to cover correlation ID propagation end-to-end.

### DIFF
--- a/backend/internal/authnprovider/provider/rest_authn_provider.go
+++ b/backend/internal/authnprovider/provider/rest_authn_provider.go
@@ -29,16 +29,18 @@ import (
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 
 	authnprovidercm "github.com/thunder-id/thunderid/internal/authnprovider/common"
+	sysContext "github.com/thunder-id/thunderid/internal/system/context"
 	systemhttp "github.com/thunder-id/thunderid/internal/system/http"
 	"github.com/thunder-id/thunderid/internal/system/log"
 )
 
 // restAuthnProvider is an authentication provider that communicates with an external service via REST.
 type restAuthnProvider struct {
-	baseURL    string
-	apiKey     string
-	httpClient systemhttp.HTTPClientInterface
-	logger     *log.Logger
+	baseURL             string
+	apiKey              string
+	correlationIDHeader string
+	httpClient          systemhttp.HTTPClientInterface
+	logger              *log.Logger
 }
 
 // AuthenticateRequest is the request body for the authentication endpoint.
@@ -62,12 +64,14 @@ type apiErrorResponse struct {
 }
 
 // newRestAuthnProvider creates a new REST authentication provider.
-func newRestAuthnProvider(baseURL, apiKey string, httpClient systemhttp.HTTPClientInterface) AuthnProviderInterface {
+func newRestAuthnProvider(baseURL, apiKey, correlationIDHeader string,
+	httpClient systemhttp.HTTPClientInterface) AuthnProviderInterface {
 	return &restAuthnProvider{
-		baseURL:    baseURL,
-		apiKey:     apiKey,
-		httpClient: httpClient,
-		logger:     log.GetLogger().With(log.String(log.LoggerKeyComponentName, "RestAuthnProvider")),
+		baseURL:             baseURL,
+		apiKey:              apiKey,
+		correlationIDHeader: correlationIDHeader,
+		httpClient:          httpClient,
+		logger:              log.GetLogger().With(log.String(log.LoggerKeyComponentName, "RestAuthnProvider")),
 	}
 }
 
@@ -183,6 +187,7 @@ func (p *restAuthnProvider) doRequest(ctx context.Context, url string, body io.R
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set(p.correlationIDHeader, sysContext.GetTraceID(ctx))
 	if p.apiKey != "" {
 		req.Header.Set("API-KEY", p.apiKey)
 	}
--- a/backend/internal/system/config/config.go
+++ b/backend/internal/system/config/config.go
@@ -256,9 +256,10 @@ type EntityProviderConfig struct {
 
 // RestConfig holds the REST authentication provider configuration details.
 type RestConfig struct {
-	BaseURL  string             `yaml:"base_url" json:"base_url"`
-	Timeout  int                `yaml:"timeout"  json:"timeout"`
-	Security RestSecurityConfig `yaml:"security" json:"security"`
+	BaseURL             string             `yaml:"base_url" json:"base_url"`
+	Timeout             int                `yaml:"timeout" json:"timeout"`
+	CorrelationIDHeader string             `yaml:"correlation_id_header" json:"correlation_id_header"`
+	Security            RestSecurityConfig `yaml:"security" json:"security"`
 }
 
 // RestSecurityConfig holds the REST authentication provider security configuration details.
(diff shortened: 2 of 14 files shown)


backend/internal/authnprovider/provider/rest_authn_provider.go
```diff
@@ -182,6 +184,7 @@ func (p *restAuthnProvider) doRequest(ctx context.Context, url string, body io.R
 		return nil, err
 	}
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set(serverconst.CorrelationIDHeaderName, sysContext.GetTraceID(ctx))
```
rajithacharith: Can we make this header name configurable and default to our correlation Id. 
Some services might have a separate header name for this? 

WDYT? @ThumulaPerera

ThumulaPerera: +1. Sounds good

UdeshAthukorala: Made the correlation ID header name configurable on the REST authn provider via `authn_provider.rest.correlation_id_header`, which defaults to `X-Correlation-ID` when left empty:

```yaml
authn_provider:
  type: rest
  rest:
    base_url: "https://my-authn-service"
    correlation_id_header: "X-Trace-Token"
```

The default is wired through the `CorrelationIDHeaderName` constant, so empty config behaves exactly as before. Added a unit test (`TestAuthenticate_PropagatesConfiguredCorrelationIDHeader`) asserting a custom header name is honoured and the default one isn't set.

We briefly explored extending this to a server-wide setting, but after discussing it we decided against it — the header name stays configurable on the REST authn provider only, since that's the pluggable external integration this thread is about. The other outbound callers keep the standard header: Vonage doesn't consume it, consent is our own service, and the flow HTTP node already lets the author set any header explicitly.

backend/internal/system/config/config.go
```diff
@@ -644,9 +644,10 @@ type EntityProviderConfig struct {
 
 // RestConfig holds the REST authentication provider configuration details.
 type RestConfig struct {
-	BaseURL  string             `yaml:"base_url" json:"base_url"`
-	Timeout  int                `yaml:"timeout" json:"timeout"`
-	Security RestSecurityConfig `yaml:"security" json:"security"`
+	BaseURL             string             `yaml:"base_url" json:"base_url"`
+	Timeout             int                `yaml:"timeout" json:"timeout"`
+	CorrelationIDHeader string             `yaml:"correlation_id_header" json:"correlation_id_header"`
```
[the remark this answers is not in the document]
UdeshAthukorala: Added documentation under `docs/content/guides/getting-started/configuration.mdx`:

- New row for `authn_provider.rest.correlation_id_header` in the Authentication Provider Configuration table, with its default (`X-Correlation-ID`).
- A **Correlation ID Propagation** subsection explaining that the REST authn provider, consent service, flow HTTP request executor, and Vonage notification provider all set `X-Correlation-ID` on outbound calls when it isn't already present, plus a YAML example for overriding the header name on the REST authn provider.

rajithacharith (approved):

PR #3470, opened 2026-06-23
Validate logo URL input in the resource logo dialog

### LINKED ISSUE #2926: Logo URL not validated
Two issues observed in the logo URL

1. Logo URL is not validated at the point of input. It returns invalid URL upon creation from backend

Image

2. Logo URL input field has a delay in reflecting input

KD23243:
## Purpose
Custom logo URLs entered through the resource logo dialog (used in both the application create design step and the application edit header) were not checked at the point of input. Any non-empty string could be confirmed, and the only check happened on the backend at create/update time, so the failure was only reported after the user committed the change. The frontend's own URL check was also inconsistent with the backend's accepted set, and typing in the URL field felt laggy.

## Behaviour
Screenshot 2026-06-23 at 11 57 38

## Related Issue
- Fixes https://github.com/thunder-id/thunderid/issues/2926

## Implementation
- Added `isValidLogoUri` (`frontend/packages/components/src/lab/utils/isValidLogoUri.ts`), a single shared helper that mirrors the backend `IsValidLogoURI` allowlist: `http`/`https` require a non-empty host, `data`/`blob`/`emoji` are accepted, no-scheme values must be absolute paths, and all other schemes are rejected.
- `ResourceLogoDialog` now validates the URL field with this helper: invalid input shows an inline error and disables the select action, and the emoji-vs-URL routing uses the same predicate so relative and `data:` URLs route to the URL field instead of the emoji branch.
- `ResourceAvatar` renders an image for any valid non-emoji URI (http/https/data/blob/relative), not just `http(s)`.
- Memoized `EmojiPicker` so typing in the URL field no longer re-renders the full emoji grid, removing the input lag.
- Added the `resource_logo_dialog.url_section.error_text` translation key.
- Added tests for the new helper (mirroring the backend test matrix) and new URL validation cases in the dialog tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for custom logo URLs in the Resource logo dialog, including clearer inline error feedback and a dedicated invalid-URL message.
  * The Select action is now blocked until the entered value is valid, while valid absolute paths and additional URI formats (such as data/blob/emoji and approved http/https forms) are accepted.
  * Resource avatar rendering now follows the same validation rules for logo URL vs non-URL values.
* **Tests**
  * Added client-side and interaction test coverage for logo URI validation.

### DIFF
--- a/frontend/packages/components/src/lab/utils/isValidLogoUri.ts
+++ b/frontend/packages/components/src/lab/utils/isValidLogoUri.ts
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+const SCHEME_PATTERN = /^([a-zA-Z][a-zA-Z0-9+.-]*):/;
+const HTTP_AUTHORITY_PATTERN = /^https?:\/\/([^/?#]*)/i;
+// A non-empty host (bracketed IPv6 or any run of non-space, non-colon chars)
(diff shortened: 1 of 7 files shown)


frontend/packages/components/src/lab/utils/isValidLogoUri.ts
```diff
@@ -0,0 +1,67 @@
...
+ * - A value with no scheme is accepted only when it is an absolute path
+ *   (starts with `/`).
+ * - Every other scheme (e.g. `javascript`, `file`, `ftp`) is rejected.
+ *
+ * @param value - The candidate logo URI.
+ * @returns `true` when the value is an acceptable logo URI.
+ *
+ * @public
+ */
+export default function isValidLogoUri(value: string): boolean {
```
brionmario, line 43: Does this allow relative paths?
@coderabbitai

PR #3471, opened 2026-06-23
Fix stale in-memory graph cache issue in multi-node deployments

### LINKED ISSUE #1998: Finalize FlowGraphCache Strategy
### Description

The current FlowGraphCache is implemented as an in-memory cache to avoid the high cost of repeatedly constructing flow execution graphs during runtime. 

### Problem

- Flow execution graphs are expensive to construct, so caching is required
- Current implementation uses in-memory cache only
- Redis (or other distributed caches) cannot directly store Go objects

### Objective

- Define and implement a proper caching strategy for FlowGraphCache that works with the overall caching architecture.
- Removes the temporary fix
  - https://github.com/asgardeo/thunder/pull/1997

### LINKED ISSUE #3064: Bug: Graph cache is in-memory only, causing stale graphs in multi-node deployments
## Summary

The compiled flow graph cache is hardcoded to in-memory. In a multi-node deployment, invalidating the cache on one node after a flow update has no effect on the other nodes — they continue serving the stale compiled graph until their process restarts or the entry naturally expires.

## Root cause

In `backend/internal/flow/core/init.go`, the graph cache is always created using `GetInMemoryCache`:

```go
graphCacheInst := cache.GetInMemoryCache[*graph](cacheManager, "FlowGraphCache")
```

Unlike `GetCache`, which respects the configured cache backend (in-memory or Redis), `GetInMemoryCache` always creates a local in-memory store. Each node has its own isolated cache with no awareness of invalidations happening on other nodes.

## Impact

- Node A receives a `PUT /flows/{flowId}` request. `flowmgt` calls
  `graphBuilder.InvalidateCache(ctx, flowID)`, which removes the entry from **Node A's local cache only**.
- Nodes B and C still hold the old compiled graph in their own in-memory caches.
- All subsequent flow executions on B and C use the stale graph until their process restarts.

This is a silent correctness issue — no error is returned, but users on different nodes can observedifferent flow behaviour after a flow update.

## Why a Redis-backed graph cache is not straightforward

The compiled `*graph` struct holds live Go objects — nodes with interface references, executor pointers, segment data. It cannot be serialised to Redis as-is, so simply swapping `GetInMemoryCache` for `GetCache` is not sufficient without a dedicated serialisation layer.

## Proposed fix

Remove the graph cache entirely. Flow definitions are already cached in a Redis-backed `FlowByIDCache` inside `flowmgt`, so graph recompilation — iterating node definitions and wiring up Go objects — is cheap. The graph cache is a premature optimisation on top of an already-cached store lookup, and keeping it means accepting this multi-node inconsistency or taking on the complexity of a serialisable distributed cache for live Go objects.

NutharaNR:
### Purpose

- The compiled flow graph cache is hardcoded to in-memory. In a multi-node deployment, invalidating the cache on one node after a flow update has no effect on the other nodes — they continue serving the stale compiled graph until their process restarts or the entry naturally expires. 

### Approach

- Compiled graph objects cannot be serialized into the distributed (Redis) cache, so the graph cache is kept in-memory on each node. To detect staleness, the flow's `ActiveVersion` is stored in the compiled graph at build time. On retrieval, the cached graph's version is compared against the flow definition's current `ActiveVersion`. If there is a mismatch, the stale graph is invalidated and rebuilt.

#### Changes

**1. Added `version` field to the `graph` struct** (`backend/internal/flow/core/graph.go`)

```go
type graph struct {
    // ... existing fields
    version      int
}
```

With corresponding `GetVersion()` / `SetVersion()` methods on `GraphInterface`.

**2. Version stamped at build time** (`backend/internal/flow/mgt/graph_builder.go`)

```go
graph := b.flowFactory.CreateGraph(flow.ID, flow.FlowType)
graph.SetVersion(flow.ActiveVersion)
```

**3. Version-based staleness detection on cache retrieval** (`backend/internal/flow/mgt/graph_builder.go`)

```go
if cachedGraph, ok := b.graphCache.Get(ctx, flow.ID); ok {
    if cachedGraph.GetVersion() == flow.ActiveVersion {
        return cachedGraph, nil
    }
    // Version mismatch — invalidate and rebuild
    b.graphCache.Invalidate(ctx, flow.ID)
}
```

- The default global TTL is set to `3600s` = `1hr`. The same TTL has been used for graph cache as well.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3064
- https://github.com/thunder-id/thunderid/issues/1998

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added graph version tracking so cached graphs are only reused when they match the currently active version; otherwise the stale cache entry is discarded and the graph is rebuilt.

* **Tests**
  * Added unit tests for default version behavior plus set/get version updates.
  * Added coverage for cache hit vs version mismatch handling, including cache invalidation and rebuild behavior.
  * Updated existing test fixtures and mock expectations to include version-aware graph creation and retrieval.

### DIFF
--- a/backend/internal/flow/core/factory.go
+++ b/backend/internal/flow/core/factory.go
@@ -30,7 +30,7 @@ import (
 type FlowFactoryInterface interface {
 	CreateNode(id, _type string, properties map[string]interface{}, isStartNode, isFinalNode bool) (
 		NodeInterface, error)
-	CreateGraph(id string, _type common.FlowType) GraphInterface
+	CreateGraph(id string, _type common.FlowType, version int) GraphInterface
 	CreateExecutor(name string, executorType common.ExecutorType,
 		defaultInputs, prerequisites []common.Input) ExecutorInterface
 	CreateInterceptor(name string, isDefault bool, priority int) InterceptorInterface
@@ -74,19 +74,23 @@ func (f *flowFactory) CreateNode(id, _type string, properties map[string]interfa
 }
 
 // CreateGraph creates a new graph with the given ID and type
-func (f *flowFactory) CreateGraph(id string, _type common.FlowType) GraphInterface {
+func (f *flowFactory) CreateGraph(id string, _type common.FlowType, version int) GraphInterface {
 	if id == "" {
 		id = sysutils.GenerateUUID()
 	}
 	if _type == "" {
 		_type = common.FlowTypeAuthentication
 	}
+	if version <= 0 {
+		version = 1
+	}
 
 	return &graph{
-		id:    id,
-		_type: _type,
-		nodes: make(map[string]NodeInterface),
-		edges: make(map[string][]string),
+		id:      id,
+		_type:   _type,
+		version: version,
+		nodes:   make(map[string]NodeInterface),
+		edges:   make(map[string][]string),
 	}
 }
 
--- a/backend/internal/flow/mgt/graph_builder.go
+++ b/backend/internal/flow/mgt/graph_builder.go
@@ -75,10 +75,20 @@ func (b *graphBuilder) GetGraph(ctx context.Context, flow *CompleteFlowDefinitio
 	}
 
 	logger := b.logger.With(log.String("flowID", flow.ID))
-	// Check cache first
+	// Check cache first, with version-based staleness detection.
+	// The flow definition (with ActiveVersion) comes from the Redis-backed store,
+	// so all nodes see the latest version even though the graph cache is local.
 	if cachedGraph, ok := b.graphCache.Get(ctx, flow.ID); ok {
-		logger.Debug(ctx, "Graph retrieved from cache")
-		return cachedGraph, nil
+		if cachedGraph.GetVersion() == flow.ActiveVersion {
+			logger.Debug(ctx, "Graph retrieved from cache")
+			return cachedGraph, nil
+		}
+		logger.Debug(ctx, "Cached graph version mismatch, rebuilding",
+			log.Int("cachedVersion", cachedGraph.GetVersion()),
+			log.Int("activeVersion", flow.ActiveVersion))
+		if err := b.graphCache.Invalidate(ctx, flow.ID); err != nil {
+			logger.Error(ctx, "Failed to invalidate stale graph from cache", log.Error(err))
+		}
 	}
 
 	graph, err := b.buildGraph(ctx, flow)
@@ -119,7 +129,7 @@ func (b *graphBuilder) buildGraph(ctx context.Context, flow *CompleteFlowDefinit
 	}
 
 	// Create a graph
-	graph := b.flowFactory.CreateGraph(flow.ID, flow.FlowType)
+	graph := b.flowFactory.CreateGraph(flow.ID, flow.FlowType, flow.ActiveVersion)
 
 	// Process all nodes and build the graph structure
 	edges := make(map[string][]string)
--- a/backend/internal/flow/core/FlowFactoryInterface_mock_test.go
+++ b/backend/internal/flow/core/FlowFactoryInterface_mock_test.go
@@ -232,16 +232,16 @@ func (_c *FlowFactoryInterfaceMock_CreateExecutor_Call) RunAndReturn(run func(na
 }
 
 // CreateGraph provides a mock function for the type FlowFactoryInterfaceMock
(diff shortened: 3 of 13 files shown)


backend/internal/flow/core/init.go
```diff
@@ -21,10 +21,12 @@ package core
...
-	graphCacheInst := cache.GetInMemoryCache[*graph](cacheManager, "FlowGraphCache")
-	graphCache := newGraphCache(graphCacheInst)
-	return flowFactory, graphCache
+// Initialize initializes the core flow package.
+func Initialize() FlowFactoryInterface {
+	return newFlowFactory()
+}
+
+// NewGraphCache creates a new in-memory graph cache instance.
+func NewGraphCache(cacheManager cache.CacheManagerInterface) GraphCacheInterface {
```
ThaminduDilshan: This doesn't align with the dependency injection pattern. Having public constructor methods like this allows unintended service object creation. Also this will not resolve the concern raised in https://github.com/thunder-id/thunderid/issues/3064#issuecomment-4570526441. 

This graph cache is utilized by the graph builder which is the implementation exposed to the other services. Graph cache can be kept as an internal implementation within the graph builder. Ideal solution is to extract graph builder out from the flow mgt package and move to the core or a separate package.

The same concern is already addressed in the WIP PR: https://github.com/thunder-id/thunderid/pull/3054. Hence shall we take that task out from this PR?

NutharaNR: Removed changes related to graph cache initilization.

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -75,10 +75,20 @@ func (b *graphBuilder) GetGraph(ctx context.Context, flow *CompleteFlowDefinitio
...
+	// The flow definition (with ActiveVersion) comes from the Redis-backed store,
+	// so all nodes see the latest version even though the graph cache is local.
 	if cachedGraph, ok := b.graphCache.Get(ctx, flow.ID); ok {
-		logger.Debug(ctx, "Graph retrieved from cache")
-		return cachedGraph, nil
+		if cachedGraph.GetVersion() == flow.ActiveVersion {
+			logger.Debug(ctx, "Graph retrieved from cache")
+			return cachedGraph, nil
+		}
+		logger.Info(ctx, "Cached graph version mismatch, rebuilding",
```
ThaminduDilshan:
```suggestion
		logger.Debug(ctx, "Cached graph version mismatch, rebuilding",
```

General practice is to keep info logs minimal

backend/internal/flow/mgt/graph_builder.go
```diff
@@ -120,6 +130,7 @@ func (b *graphBuilder) buildGraph(ctx context.Context, flow *CompleteFlowDefinit
 
 	// Create a graph
 	graph := b.flowFactory.CreateGraph(flow.ID, flow.FlowType)
+	graph.SetVersion(flow.ActiveVersion)
```
ThaminduDilshan: Instead isn't it better to set the version from the flow factory itself? WDYT?

backend/internal/flow/mgt/graph_builder.go
NutharaNR: Updated to debug logs.

backend/internal/flow/mgt/graph_builder.go
NutharaNR: Yes it would be better, as it is only a property and doesn't require a building process moved it to flow factory.

backend/internal/flow/core/factory.go
```diff
@@ -74,19 +74,23 @@ func (f *flowFactory) CreateNode(id, _type string, properties map[string]interfa
...
+func (f *flowFactory) CreateGraph(id string, _type common.FlowType, version int) GraphInterface {
 	if id == "" {
 		id = sysutils.GenerateUUID()
 	}
 	if _type == "" {
 		_type = common.FlowTypeAuthentication
 	}
+	if version <= 0 {
+		version = 1
+	}
```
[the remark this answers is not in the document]
NutharaNR: the counter start with 1

ThaminduDilshan (approved):

Issue #2380, opened 2026-04-21
Implement solution for app native flow client verification


ThaminduDilshan:
This is to track the implementation work of https://github.com/asgardeo/thunder/issues/2009 based on the initial research outcome

PR #3476, opened 2026-06-23
Verify Flow Secret for backend applications at flow initiation

### LINKED ISSUE #3451: Add App Secret verification for backend applications at flow initiation
## Description
As part of the secured-by-default flow initiation model (parent: #2380), backend / server-side clients calling `POST /flow/execute` to initiate a new flow must authenticate themselves. Unlike browser SPAs (which are restricted to redirect-based flows) and mobile clients (platform attestation), backend applications are confidential clients and can safely hold a secret. Thunder should issue an **App Secret** to backend application types and verify it when a new flow is initiated over HTTP.

This was agreed in the design discussion #2744:
- "Backend apps will use API keys or secrets."
- Implementation Plan, step 2: "Backend / server-side clients — API key or client secret verification at flow initiation."

## Implementation Notes
- Issue an **App Secret** for backend application types as part of application registration.
- Verify the presented App Secret at flow initiation (the HTTP new-flow path in `loadNewContext()`), consistent with where the SPA guard from #3215 lives.
- Reject new-flow initiation for backend apps when the App Secret is missing or invalid, with a descriptive error.
- Continuing an existing flow with a known `executionId` is unaffected.
- Ensure the App Secret is stored securely (hashed) following existing credential-handling conventions.

## Acceptance Criteria
- A backend application is provisioned with an App Secret on creation.
- A `POST /flow/execute` new-flow initiation from a backend app succeeds only with a valid App Secret and is rejected otherwise with a descriptive error.
- Existing flow continuation via `executionId` is unaffected.
- The check is enforced at the service layer, not only at the API handler.

Malith-19:
### Purpose

Resolves #3451 — part of the secured-by-default flow initiation model (#2380).

Introduces a **Flow Secret**, a credential distinct from the OAuth client secret. Backend / server-side applications authenticate with it when they initiate a sign-in flow directly via `POST /flow/execute`. Redirect-based applications (which initiate through the OAuth component) are blocked from direct initiation and are not issued a Flow Secret.

The flow-execution layer stays protocol-agnostic: it derives "may this app initiate, and how" from neutral actor data and delegates the credential check to the actor layer, so it never inspects OAuth-specific data beyond a generic profile lookup.

> **Naming:** this credential was originally called *App Secret*. Following the design discussion (thunder-id/thunderid#2744), it is renamed to **Flow Secret** since it is scoped to flow initiation and to avoid confusion with the OAuth Client Secret.

---
Direct initiation of a **new** authentication flow via `POST /flow/execute` is now authenticated:

- Backend / server-side apps (non-public, non-redirect — including embedded apps with no OAuth profile) **must present a valid `flowSecret`** in the INIT request.
- Apps using the redirect-based `authorization_code` grant are **blocked** from direct initiation and must use the OAuth redirect flow.

#### 💥 Impact
- **Backend / embedded integrations** that call `POST /flow/execute` (action `INIT`) without a `flowSecret` now receive **`401` (`FES-1011`)**; an invalid secret returns **`401` (`FES-1012`)**.
- **Applications created before this change have no Flow Secret.** They must be issued one (Console → application → **Danger Zone → Regenerate Flow Secret**) before they can initiate flows directly.
- **Confidential Full-stack (`authorization_code`) apps** that were initiating flows directly now receive **`403` (`FES-1010`)**. (Public redirect apps — SPA/mobile — were already blocked.)
- **Machine-to-machine Backend Services (`client_credentials` only)** are not issued a Flow Secret and are blocked from direct initiation with **`403` (`FES-1010`)** — they obtain tokens directly at the token endpoint. Adding the `token-exchange` grant makes such an app flow-native (eligible for a Flow Secret).
- **Not affected:** flow **continuation** requests (those carrying an `executionId`), and registration / recovery flows.

#### 🔄 Migration Guide
1. **Backend / server-side apps:** obtain the Flow Secret — it is shown once at creation, or can be regenerated from **Console → application → Danger Zone → Regenerate Flow Secret** — and pass it as `flowSecret` when starting a new flow:

   ```json
   {
     "applicationId": "<clientId>",
     "flowSecret": "<flowSecret>",
     "flowType": "AUTHENTICATION",
     "action": "INIT",
     "inputs": {}
   }
   ```

2. **Declaratively-managed apps (YAML resources):** add an explicit `flowSecret` field to the application definition — auto-generated values are not surfaced back into the definition, so without one the app cannot initiate a flow.

3. **Redirect-based apps (SPA / mobile / full-stack):** no change — continue using the OAuth redirect sign-in flow. They are not issued a Flow Secret.

---

### Approach

Application "type" is a console-only concept — the backend only sees OAuth attributes (`publicClient`, `grantTypes`). Apps created with **embedded sign-in** carry no OAuth profile and no client secret, so the client secret alone could not cover the primary use case. The Flow Secret is therefore an independent credential.

**Issuance** (at registration): a Flow Secret is issued only when the app can actually initiate a flow directly and consume the resulting assertion — i.e. an **embedded app (no OAuth profile)**, or a **non-public, non-redirect app that has the `token-exchange` grant**. A machine-to-machine app that uses only `client_credentials` cannot redeem a flow assertion (the token endpoint rejects `token-exchange` when it isn't registered), so it is treated as M2M and issued **no** Flow Secret:

| App | Client Secret | Flow Secret |
|---|---|---|
| Browser SPA / Mobile (public, redirect) | — | no |
| Full-stack redirect (confidential, `authorization_code`) | yes | no |
| Backend / M2M (`client_credentials` only) | yes | no |
| Confidential non-redirect with `token-exchange` | yes | yes |
| Embedded server-side (no OAuth profile) | — | yes |

- Stored hashed in `SYSTEM_CREDENTIALS` under the `flowSecret` key, returned once in the create response.
- A top-level `flowSecret` field on the application request/response models; accepted on update to support rotation. Only eligible apps ever store one. Declarative resources only get an explicitly declared value.

A confidential flow-native app can legitimately hold **both** secrets: the Client Secret authenticates it at the OAuth **token endpoint** (to exchange the flow assertion for a token via `token-exchange`), while the Flow Secret authenticates it when initiating the flow.

**Flow-initiation guard** (`checkDirectFlowInitiationAllowed`, in the flow-execution layer): classifies the app from neutral actor data and enforces the rule:
- **redirect app** (`authorization_code`) — must initiate via the OAuth component, not a direct HTTP call → `FES-1010` (403);
- **M2M app** (`client_credentials` as its only grant) — obtains tokens directly and does not initiate flows → `FES-1010` (403);
- **flow-native app** (embedded, or non-redirect with an additional grant such as `token-exchange`) — must present a valid Flow Secret → `FES-1011` (missing) / `FES-1012` (invalid), both 401;
- an unknown application id surfaces as `FES-1003` (invalid app id, 400);
- flow **continuation** (requests carrying an `executionId`) is unaffected.

**Layering** — the flow executor depends only on the actor provider for this, and takes no direct `entityService` dependency:
- `resolveFlowInitiationMode(...)` (in the flow layer) derives the mode at runtime from the inbound protocol configuration via `actorProvider.GetOAuthProfileByID(...)` / `GetInboundClientByID(...)` — nothing is stored in the DB.
- `actorProvider.AuthenticateActor(identifiers, credentials)` verifies the Flow Secret, delegating to the authentication provider (the same hashed, constant-time check used elsewhere). No OAuth-specific data leaks into the flow executor, keeping it protocol-agnostic and ready for future protocols (e.g. SAML, mobile attestation).

Flow Secret **generation** stays in the application domain layer (mirroring how the OAuth client secret is generated), while storage and verification happen generically through the entity layer.

**Console**: the Flow Secret is shown once on creation alongside the client secret with distinct purpose labels, and a **Regenerate Flow Secret** action is available in the danger zone for backend (non-public, non-redirect) clients.

### Related Issues
- Fixes #3451
- Refs #2380

### Related PRs
- N/A

### Checklist
- [x] Documentation provided.
- [x] Tests provided.
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks


backend/internal/application/declarative_resource.go
```diff
@@ -310,7 +311,9 @@ func makeAppEntityParser(
 			return nil, nil, nil, fmt.Errorf("failed to build system attributes: %w", err)
 		}
 
-		sysCredsJSON, err := buildSystemCredentials(clientSecret)
+		// Declarative resources do not get an auto-generated App Secret; an explicitly declared
+		// value is honored to keep declarative definitions deterministic.
+		sysCredsJSON, err := buildSystemCredentials(clientSecret, appDTO.AppSecret)
```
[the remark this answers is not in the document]
Malith-19: Fixed.

backend/internal/application/model/application.go
```diff
@@ -110,6 +111,7 @@ type ApplicationRequest struct {
 	Name        string   `json:"name" yaml:"name" native:"required,min=3,max=100"`
 	Description string   `json:"description" yaml:"description"`
 	Template    string   `json:"template,omitempty" yaml:"template,omitempty"`
+	AppSecret   string   `json:"appSecret,omitempty" yaml:"appSecret,omitempty"`
```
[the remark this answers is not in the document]
Malith-19: I verified the field is load-bearing in two places, so removing it would break working functionality:

1. Declarative determinism (ApplicationRequestWithID): declarative/import resources must declare credentials deterministically — exactly as they already declare clientSecret (e.g. decl-conf-secret-1 in the test configs). An auto-generated secret is never written back into the YAML, so a declaratively-managed backend app would have no usable secret. The E2E admin app and the whole E2E suite depend on this — it's the fix that just turned the E2E checks green.
2. Rotation (ApplicationRequest update): the console's "Regenerate App Secret" submits the new secret via the update endpoint (frontend/apps/console/src/features/applications/api/useRegenerateAppSecret.ts). There's no dedicated regenerate endpoint yet (documented as future work in the hook). Removing the field breaks rotation.

Why it isn't the security weakness described:

1. Parity with clientSecret: the REST API already accepts a caller-supplied clientSecret via inboundAuthConfig, on both create and update. Accepting appSecret is consistent with the established credential model, not a new weakness class.
2. The real risk is already closed by gating: an App Secret is only ever persisted for eligible app types (non-public, non-redirect). Public/redirect apps never store one regardless of caller input. A caller choosing a weak secret only weakens their own app — identical to clientSecret today.

backend/internal/application/model/application.go
```diff
@@ -33,6 +33,7 @@ type ApplicationDTO struct {
 	Name        string `json:"name" jsonschema:"Application name."`
 	Description string `json:"description,omitempty" jsonschema:"Optional description of the application's purpose or functionality."`
 	Template    string `json:"template,omitempty" jsonschema:"Application template. Optional. Pre-configured application type template."`
+	AppSecret   string `json:"appSecret,omitempty" jsonschema:"App Secret. Returned once on creation for backend/server-side apps — non-public clients that do not use the redirect-based authorization_code flow. Used to authenticate when initiating a flow directly via the Flow Execution API."`
```
ThaminduDilshan: We need to update the api swagger definition as well right?

ThaminduDilshan: How is this app secret generated? Is it generated for all the apps or just native backend apps?

backend/internal/application/declarative_resource.go
```diff
@@ -310,7 +311,9 @@ func makeAppEntityParser(
 			return nil, nil, nil, fmt.Errorf("failed to build system attributes: %w", err)
 		}
 
-		sysCredsJSON, err := buildSystemCredentials(clientSecret)
+		// Declarative resources do not get an auto-generated App Secret; an explicitly declared
```
ThaminduDilshan: Hope this is the same pattern used for OAuth client secret as well. If not better to align implementation with that

backend/internal/application/service.go
```diff
@@ -121,12 +121,36 @@ func (as *applicationService) CreateApplication(ctx context.Context, app *model.
...
+	// Execution API — i.e. backend / server-side apps. Public clients (browser SPAs, mobile apps)
+	// and redirect-based (authorization_code) clients initiate their flows through the OAuth
+	// component, so they have no use for an App Secret and never receive one — any caller-supplied
+	// value for such apps is ignored. For eligible apps, an explicitly provided value (e.g.
+	// declarative resources) is preserved; otherwise one is generated.
+	appSecret := ""
+	if !isPublicClient && !isRedirectClient {
+		appSecret = app.AppSecret
+		if appSecret == "" {
+			generatedAppSecret, secretErr := oauthutils.GenerateOAuth2ClientSecret()
```
ThaminduDilshan: @thiva-k is this the correct place to generate the app secret? From what I understood, application is a wrapper layer built encapsulating inbound client and entity provider. Hence this should not have any business logic being implemented.

On a related note, the flow execution requirement can come for a agent as well in future right? Hence the same app secret guard should be applied there too. With that regard, this might go into entity side.

 Better to validate with @thiva-k before changing the implementation.

backend/internal/application/service.go
```diff
@@ -447,27 +474,47 @@ func (as *applicationService) updateEntityDataForApplicationUpdate(ctx context.C
...
-	// - OAuth method requires a secret + no new secret supplied → leave existing secret intact (no rotation).
-	if inboundAuthConfig == nil || inboundAuthConfig.OAuthConfig == nil ||
-		!appRequiresClientSecret(inboundAuthConfig.OAuthConfig) {
-		if epErr := as.entityProvider.UpdateSystemCredentials(appID, nil); epErr != nil {
+	// Rotate the App Secret when a new value is supplied (e.g. a regenerate request). Only
+	// applications eligible to hold an App Secret — backend / server-side apps that are neither
+	// public nor redirect-based — may have one set; a value supplied for an ineligible app is
+	// ignored. Credential updates merge, so this preserves the stored client secret, and an empty
+	// value leaves the existing App Secret intact.
+	isPublicClient := false
```
ThaminduDilshan: Shall we write a private helper method as this logic is being repeated twise?

backend/cmd/server/servicemanager.go
```diff
@@ -382,7 +382,7 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 		agentService,
 	)
 
-	flowExecService, err := flowexec.Initialize(mux, flowMgtService, actorProvider,
+	flowExecService, err := flowexec.Initialize(mux, flowMgtService, actorProvider, entityService,
```
ThaminduDilshan: I think plugging entity service dependency to the flow execution service deviates the intention of https://github.com/thunder-id/thunderid/pull/3306. We should be able to utilize `ActorProviderInterface`, maybe with some improvement. But keep adding new methods to ActorProviderInterface is also not nice and seems to be getting out of control.

@senthalan @anushasunkada any suggestion on this?

backend/internal/flow/flowexec/service.go
```diff
@@ -190,36 +196,68 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
...
 
-// checkDirectFlowInitiationAllowed returns an error if the application's grant type does not
-// permit direct HTTP initiation of an authentication flow. Applications configured with the
-// authorization_code grant type must have their authentication flows initiated by the OAuth
-// component, not via a direct HTTP call. Other flow types (registration, recovery, user
-// onboarding) are not subject to this restriction.
+// checkDirectFlowInitiationAllowed governs which applications may initiate an authentication flow
+// directly over HTTP. Any application whose OAuth profile uses the redirect-based
+// authorization_code grant — public or confidential — must initiate its flows through the OAuth
+// component, not via a direct HTTP call. Every other application — one whose OAuth profile does not
```
ThaminduDilshan: Shall we generalize this comment not to have oauth terms? Flow is a internal implementation that is used by oaht, but this is not coupled to oauth. Even protocols like SAML will use the same flow execution layer.
(It's okay to mention oauth as one usage, maybe within brackets)

backend/internal/flow/flowexec/service.go
```diff
@@ -190,36 +196,68 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
...
 		return nil
 	}
 	if appID == "" {
 		return nil
 	}
 
+	// A missing OAuth profile (ErrorActorNotFound) is expected for server-side embedded apps,
+	// which carry no OAuth configuration. Treat it as a nil profile and fall through to the
+	// App Secret check rather than allowing the flow unauthenticated.
 	oauthProfile, svcErr := s.actorProvider.GetOAuthProfileByID(ctx, appID)
```
ThaminduDilshan: On a separate thought (related to previous PR as well), maybe instead of this, we can improve the underline data structures to add a field which indicates the app type or somewhat similar flags. This way flow execution layer doesn't need to retrieve oauth profile or perform any other protocol specific validations. In a way flow exec service calling `GetOAuthProfileByID()` is wrong.

We don't need to store such values in DB too. Underline service can compute their values at runtime and set in the returned model.

Also this simplifies the complex logic added to the flow exec service as well. With that improvement, flow exec service just can rely on one or few flags to determine the behaviour. No need to call multiple services as well.

Shall we revalidate the implementation on that angle? This might change some implementations introduced with the previous PR as well

ThaminduDilshan: Maybe appSecret can also be returned in the same data structure/ model. Need to evaluate pros and cons

backend/internal/application/service.go
```diff
@@ -121,12 +121,36 @@ func (as *applicationService) CreateApplication(ctx context.Context, app *model.
 	// Create entity.
 	var clientID string
 	var clientSecret string
+	isPublicClient := false
+	isRedirectClient := false
 	if inboundAuthConfig != nil && inboundAuthConfig.OAuthConfig != nil {
 		clientID = inboundAuthConfig.OAuthConfig.ClientID
 		clientSecret = inboundAuthConfig.OAuthConfig.ClientSecret
+		isPublicClient = inboundAuthConfig.OAuthConfig.PublicClient
```
ThaminduDilshan: Btw do we need to issue app secrets for the apps with client credentials and other grant types as well? If I'm not mistaken, we currently check for the authorization code grant type specifically. Thinking whether we need that validation at all (I.e. `isPublicClient` and `isRedirectClient` validation).

This is only required for native apps right? Maybe having a type aligning with my previous comment simplifies this implementation too

ThaminduDilshan: But we need to think on the possibility of having an app that can execute both flow native and some oauth authentication protocol

ThaminduDilshan:
Shall we also add some flow/exec integrations tests to cover your scenarios? Both success and failure. Success maybe already covered in existing tests, we need tests for the negative cases

backend/internal/application/model/application.go
Malith-19: Yes this is only for the non public client applications with embedded flows.

backend/internal/application/declarative_resource.go
Malith-19: Yes this follows same approach as from the declarative resource we can provide client secret or the app secret. Only when it's not present we will be generating a new secret. eg: https://github.com/thunder-id/thunderid/blob/3971cc1917f3bd15284f051fba34211563a4593a/samples/apps/wayfinder-sample/thunderid-config/thunderid-config.yaml#L1459

backend/internal/application/service.go
Malith-19: Updated.

backend/internal/flow/flowexec/service.go
Malith-19: Update as below,

```
// directly over HTTP. Applications that sign users in through a redirect-based protocol component
// (currently OAuth 2.0 apps using the authorization_code grant — public or confidential) must have
// their flows initiated by that component, not via a direct HTTP call. Every other application —
// one that does not sign in by redirect, or a server-side embedded app with no protocol profile at
// all — is treated as a backend application and must authenticate by presenting its App Secret.
// Other flow types (registration, recovery, user onboarding) are not restricted.
//
// The detection is currently expressed in OAuth terms because OAuth is the only redirect-based
// inbound protocol today; the flow-execution layer itself is protocol-agnostic and other protocols
// (e.g. SAML) would slot into the same guard.
```

backend/internal/flow/flowexec/service.go
Malith-19: proposed this in the [discussion](https://github.com/thunder-id/thunderid/discussions/2744#discussioncomment-17415644) and resolved

backend/internal/application/service.go
Malith-19: when we come to the other grant types (client credentials) we don't need to engage the flows APIs right? They can call the token endpoint directly with the client credentials.

backend/internal/actorprovider/interface.go
```diff
@@ -39,6 +39,9 @@ type ActorProviderInterface interface {
 	GetInboundClientByID(
 		ctx context.Context, id string,
 	) (*inboundmodel.InboundClient, *serviceerror.ServiceError)
+	GetFlowInitiationMode(
```
ThaminduDilshan: I don't think this is a correct method to be added here

backend/internal/application/service.go
ThaminduDilshan: No what I meant is an app that can be used for both protocol based and flow native execution.

backend/internal/application/service.go
thiva-k: Validations specific to an entity type are done in the domain layer(app/agent/user). As per my understanding this needs to be handled similarly to OAuth client secret right? If so secret generation which are entity-type specific(since users won't have this logic) happens in the domain layer. And the storage, secret verification at runtime, authentication etc. should happen generically through entity layer.

We are current are using flows for agent use cases too and the native way should also be supported. So we need to have this logic in agent side too. With this change if a login is initiated through native way to agents, it will fail right? The runtime secret check from flow-side will anyway happen through entity layer which is shared by apps and agents.

thiva-k: And if we are exposing this app-secret term to end users, better have a generic name since it will be shared by all current and future entity types that needs to use flows.

backend/cmd/server/servicemanager.go
Malith-19: I had a chat with @senthalan regarding this and his suggestion is to have a method in the actorProvider. Also we noticed that authnService is also using entityProvider to do the clientId:clientSecret validation. Ideally that should also go through the actorProvider.

So suggestion is to have a generic method in the actorProvider and use it for both cases mentioned above.

senthalan: We need to bring the authenticate generic function to the actorProvider and use it here and in the oauth package where we do client authentication. 

In our default actorProvider implementation, it can directly talk to our authnProvider, if someone is bringing in their own provider, they can handle it on their own

backend/internal/application/service.go
Malith-19: - Generation stays in the application/domain layer — issuing a credential is a domain responsibility, not something the entity layer should own. The application layer only decides whether an app is eligible for an App Secret and generates one if so.
- Storage, verification, and authentication are kept generic through the entity layer. The secret is persisted in SYSTEM_CREDENTIALS (hashed), and verification happens via the generic authn provider — no OAuth-specific logic.
- On the agent point: The flow-initiation guard no longer lives in the flow service coupled to OAuth. It now delegates to the actor layer — actorProvider.GetFlowInitiationMode(...) to derive the required mode and actorProvider.AuthenticateActor(...) to verify the secret. Since this is keyed on the actor, a future agent actor would be subject to the same guard without any change to the flow executor. The flow layer no longer takes a direct entityService dependency.

backend/internal/application/model/application.go
Malith-19: Updated.

backend/internal/application/service.go
Malith-19: Had a chat with @senthalan regarding this and summary is we are not considering single application to support both native and redirect based logins. If a user wants that, they need to create two applications use the same login flow to achieve it. Hence we are sticking with the current implementation.

backend/internal/actorprovider/interface.go
Malith-19: Is it due to, method is specific to flows? Or do you think this method should move to flows package?

ThaminduDilshan: > Is it due to, method is specific to flows?

Yes. That's my initial thinking. Actor provider serves a different purpose. We need to think whether it makes sense to have a flow initiation for the actors. Thinking again, maybe it's fine to have it here as well. Shall we think through and decide?

@senthalan any suggestions here?

senthalan: We don't need to have it in the ActorProvider. 

The `GetFlowInitiationMode` implementation can be in the backend/internal/flow/flowexec/service.go. It doesn't make sense to be in the Provider.

Malith-19: Moved `GetFlowInitiationMode` out of `ActorProvider `and into `flowexec/service.go` as an internal `resolveFlowInitiationMode` helper. It derives the mode from the actor provider's neutral data (GetOAuthProfileByID / GetInboundClientByID), so the actor layer no longer knows about flow concepts and the flow executor owns the policy. The FlowInitiationMode enum moved into the flow package too.

backend/internal/actorprovider/service.go
```diff
@@ -30,21 +30,30 @@ import (
 	"github.com/thunder-id/thunderid/internal/system/log"
 )
 
-// actorProvider delegates actor resolution to inbound-client and entity-provider services.
+// fieldUserID is the identifier key the authentication provider uses to authenticate an actor by
+// its entity ID.
+const fieldUserID = "userID"
```
ThaminduDilshan: Rather than hard coding to the user id, shall we modify the `AuthenticateActor()` to accept map of identifiers similar to authn provider?

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -164,3 +164,33 @@ var ErrorDirectFlowInitiationNotPermitted = tidcommon.ServiceError{
...
+var ErrorAppSecretRequired = tidcommon.ServiceError{
+	Code: "FES-1012",
+	Type: tidcommon.ClientErrorType,
+	Error: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.app_secret_required",
+		DefaultValue: "Authentication required",
+	},
+	ErrorDescription: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.app_secret_required_description",
+		DefaultValue: "Backend applications must provide a valid App Secret to initiate a new flow",
```
ThaminduDilshan:
```suggestion
		DefaultValue: "Backend applications must provide a valid app secret to initiate a new flow",
```

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -164,3 +164,33 @@ var ErrorDirectFlowInitiationNotPermitted = tidcommon.ServiceError{
...
+var ErrorAppSecretInvalid = tidcommon.ServiceError{
+	Code: "FES-1013",
+	Type: tidcommon.ClientErrorType,
+	Error: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.app_secret_invalid",
+		DefaultValue: "Authentication failed",
+	},
+	ErrorDescription: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.app_secret_invalid_description",
+		DefaultValue: "The provided App Secret is invalid",
```
ThaminduDilshan:
```suggestion
		DefaultValue: "The provided app secret is invalid",
```

backend/internal/flow/flowexec/service.go
```diff
@@ -195,37 +197,102 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
...
 
-// checkDirectFlowInitiationAllowed returns an error if the application's grant type does not
-// permit direct HTTP initiation of an authentication flow. Applications configured with the
-// authorization_code grant type must have their authentication flows initiated by the OAuth
-// component, not via a direct HTTP call. Other flow types (registration, recovery, user
-// onboarding) are not subject to this restriction.
+// flowInitiationMode classifies how an application is permitted to initiate a new authentication
+// flow directly over HTTP. It is derived at runtime from the application's inbound protocol
+// configuration.
+type flowInitiationMode int
```
ThaminduDilshan: Shall we move these to a constants.go file?

backend/internal/flow/flowexec/service.go
```diff
@@ -195,37 +197,102 @@ func (s *flowExecService) loadNewContext(ctx context.Context, appID, flowTypeStr
...
+	switch mode {
+	case flowInitiationRedirectOnly:
+		return &ErrorDirectFlowInitiationNotPermitted
+	case flowInitiationAppSecret:
+		if appSecret == "" {
+			return &ErrorAppSecretRequired
+		}
+		if authErr := s.actorProvider.AuthenticateActor(ctx, appID,
+			map[string]interface{}{fieldAppSecret: appSecret}); authErr != nil {
+			logger.Warn(ctx, "Backend application provided an invalid App Secret",
```
ThaminduDilshan:
```suggestion
			logger.Debug(ctx, "Backend application provided an invalid App Secret",
```

docs/content/guides/guides/applications.mdx
```diff
@@ -150,6 +150,35 @@ You can select multiple user types. If no user types are configured, all user ty
...
+### App Secret
+
+The **App Secret** is a credential separate from the Client Secret. <ProductName /> issues an App Secret only to applications that can initiate a sign-in flow directly. These are Backend Services and server-side apps using embedded sign-in — any non-public app that does **not** use the redirect-based `authorization_code` flow. Applications that sign in by redirect (Browser Apps, Mobile Apps, and redirect-based Full-stack Apps) initiate their flows through the OAuth component. They are not issued an App Secret.
+
+The two credentials serve different purposes:
+
+- The **Client Secret** authenticates your application at the OAuth 2.0 **token endpoint**.
+- The **App Secret** authenticates your application when it **initiates a sign-in flow directly** via the Flow Execution API (`POST /flow/execute`) — the embedded sign-in path used by server-side applications.
+
+An application may hold both. Copy the App Secret from the application details page immediately after creation — like the Client Secret, it is shown only once. To rotate it later, open the application's settings and click **Regenerate App Secret**.
```
ThaminduDilshan: Is this a valid scenario if we're not permitting an app to use both oauth and flow?

docs/content/guides/guides/applications.mdx
```diff
@@ -150,6 +150,35 @@ You can select multiple user types. If no user types are configured, all user ty
...
+
+Pass the App Secret in the `appSecret` field when starting a new flow:
+
+```json
+{
```
ThaminduDilshan: This should go into the flow docs with some details. Not all, at least flow execution payload doesn't seems valid to be here

backend/internal/actorprovider/service.go
Malith-19: Fixed.

backend/internal/flow/flowexec/error_constants.go
Malith-19: Fixed.

backend/internal/flow/flowexec/error_constants.go
Malith-19: Fixed.

backend/internal/flow/flowexec/service.go
Malith-19: Moved along with the other constants declared in here.

backend/internal/flow/flowexec/service.go
Malith-19: Updated.

docs/content/guides/guides/applications.mdx
Malith-19: Removed from this doc as flows-execution.yaml already contains this information.

docs/content/guides/guides/applications.mdx
Malith-19: Removed.

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -164,3 +164,33 @@ var ErrorDirectFlowInitiationNotPermitted = tidcommon.ServiceError{
 		DefaultValue: "Direct flow initiation is not permitted for this application type",
 	},
 }
+
+// ErrorFlowSecretRequired defines the error when a backend application initiates a new flow
+// without providing its Flow Secret.
+var ErrorFlowSecretRequired = tidcommon.ServiceError{
+	Code: "FES-1012",
```
ThaminduDilshan:
```suggestion
	Code: "FES-1011",
```

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -164,3 +164,33 @@ var ErrorDirectFlowInitiationNotPermitted = tidcommon.ServiceError{
...
+	ErrorDescription: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.flow_secret_required_description",
+		DefaultValue: "Backend applications must provide a valid flow secret to initiate a new flow",
+	},
+}
+
+// ErrorFlowSecretInvalid defines the error when a backend application provides an incorrect
+// Flow Secret when initiating a new flow.
+var ErrorFlowSecretInvalid = tidcommon.ServiceError{
+	Code: "FES-1013",
```
ThaminduDilshan:
```suggestion
	Code: "FES-1012",
```

backend/internal/flow/flowexec/constants.go
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2025-2026, WSO2 LLC. (https://www.wso2.com).
```
ThaminduDilshan:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

backend/internal/flow/flowexec/constants.go
```diff
@@ -0,0 +1,49 @@
...
+	// via a direct HTTP call.
+	flowInitiationRedirectOnly flowInitiationMode = iota
+	// flowInitiationFlowSecret indicates a backend / server-side application — one that does not sign
+	// in by redirect, or an embedded app with no protocol profile at all — that may initiate a flow
+	// directly by presenting a valid Flow Secret.
+	flowInitiationFlowSecret
+	// flowInitiationM2M indicates a machine-to-machine application (client_credentials as its only
+	// grant type). It obtains tokens directly at the token endpoint and does not initiate flows, so
+	// it may not initiate a flow directly and is not issued a Flow Secret.
+	flowInitiationM2M
```
ThaminduDilshan: Rather than saying m2m, shall we say not allowed/ prohibited or something similar?

backend/internal/flow/flowexec/error_constants.go
Malith-19: Updated.

backend/internal/flow/flowexec/constants.go
Malith-19: Updated.

backend/internal/flow/flowexec/error_constants.go
Malith-19: Updated.

backend/internal/flow/flowexec/constants.go
Malith-19: flowInitiationRedirectOnly and flowInitiationM2M both produced the identical ErrorDirectFlowInitiationNotPermitted outcome, hence merged them into a single outcome-named flowInitiationNotPermitted. resolveFlowInitiationMode now returns it for both authorization_code and client_credentials-only apps, and the guard switch collapses to one case.  WDYT?

ThaminduDilshan (approved):

PR #3477, opened 2026-06-23
Introduce generic OTPExecutor with channel-agnostic OTP generation

### LINKED ISSUE #1927: Decouple OTP generation/verification and delivery using a new SMS executor
### Current Limitation

The current `SMSOTPAuthExecutor` combines OTP generation, delivery, and verification in a single executor, making it difficult to customize or extend the delivery mechanism (e.g., to use a generic SMS executor or alternative channels). This tight coupling limits flexibility for new flows and delivery methods.

### Suggested Improvement

- If a new generic SMSExecutor (see #1926) is implemented, refactor the OTP authentication and registration flows to decouple OTP generation/verification from delivery.
- Introduce a otp service to the authn package (need to think on the package structuring as this is not a direct authn service, rather a supporting service)
- Introduce a generic OTPExecutor that supports generate and verify modes.
- Update flows to use the following pattern:
    - OTPExecutor in generate mode to create the OTP.
    - SMSExecutor (or EmailExecutor) in send mode to deliver the OTP.
    - OTPExecutor in verify mode to validate the OTP.

### Version

_No response_

### LINKED ISSUE #2016: Add email OTP authentication
## What problem are we solving?
Thunder does not support authentication using email-based one-time passwords (OTP).

## Who are we solving this for?
Users and organizations needing email OTP authentication.

## Why should we solve this now?
Expands authentication options and flexibility.

## Proposed solution
Add support for email OTP authentication in flows and APIs.

## Discussion
Depends on: generic OTP executor ([#1927](https://github.com/asgardeo/thunder/issues/1927))

Dilusha-Madushan:
### Purpose

`SMSOTPAuthExecutor` tightly coupled OTP generation, SMS delivery, and OTP verification into a single executor. This made it impossible to reuse OTP logic across other delivery channels (email, push, etc.) and prevented the flow from being composed modularly.

This PR replaces `SMSOTPAuthExecutor` with a channel-agnostic `OTPExecutor` and a new three-node flow pattern:

```text
OTPExecutor(generate) → SMSNotificationSender(send) → OTPExecutor(verify)
```

#### New OTPExecutor

- Dispatches on `executorMode`:
  - `generate` identifies the user via `IdentifyEntity`, calls `GenerateOTP`, and stores the OTP session token in `RuntimeData`.
  - `verify` reads the session token, calls `VerifyOTP`, and completes authentication.

- Stores `userID` (not `mobileNumber`) as the Recipient in the OTP JWT, with `RecipientAttr: "userID"`. The verify path resolves the user directly by ID via `GetEntity`, making it channel-agnostic — the same executor can verify an OTP regardless of how it was delivered.

- `maxAttempts` is read from `NodeProperties["maxAttempts"]` with a default of `3`, making it configurable per flow node rather than hardcoded.

#### AMR / Authenticator Identifier

- Introduced `AuthenticatorOTP = "OTPAuthenticator"` (with `FactorPossession`) to replace the channel-specific `AuthenticatorSMSOTP` for AMR claims.
- The authenticator name reflects the factor type, not the delivery channel.

#### Cleanup

- Deleted `SMSOTPAuthExecutor` and its tests.
- Removed `insertPhoneInputPromptIfNeeded` from `inference.go` — this was a workaround to inject a phone prompt node for `SMSOTPAuthExecutor`. With `OTPExecutor`, missing inputs are handled inline via the standard `ExecUserInputRequired` mechanism (same as all other executors), so the inference-layer injection is no longer needed.
- Removed two unused error helper functions (`errAttributeNotFoundFor`, `errFailedToRetrieveAttribute`) and the now-dead `createInputPromptNode` inference helper that were left unused after the deletions.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/1927
- https://github.com/thunder-id/thunderid/issues/2016

### Related PRs
- https://github.com/thunder-id/thunderid/issues/1926

## Summary by CodeRabbit

* **New Features**
  * Added an updated OTP flow that separates OTP generation, SMS delivery, and verification into dedicated steps/executors.
  * Added email OTP template support and OTP runtime values to enable downstream steps.
  * Authentication now recognizes OTP as a dedicated authenticator type.
* **Bug Fixes**
  * Improved OTP session/verification behavior, including recipient-attribute handling and clearer error mapping.
  * Updated registration, recovery, and MFA flow graphs to use the new OTP path consistently.
  * Refreshed SMS/OTP-related UI flow configuration to match the new executor setup.
* **Tests**
  * Refactored and expanded OTP/unit and flow integration tests to reflect the new lifecycle.


backend/internal/authn/common/constants.go
```diff
@@ -29,6 +29,7 @@ const DefaultHTTPTimeout = 5 * time.Second
 const (
 	AuthenticatorCredentials = "CredentialsAuthenticator"
 	AuthenticatorSMSOTP      = "SMSOTPAuthenticator"
```
ThaminduDilshan, line 31: Do we need to keep this still?

backend/internal/flow/common/constants.go
```diff
@@ -231,11 +231,17 @@ const (
...
 	RuntimeKeyPresentedOptionalInputs = "presentedOptionalInputs"
 	// RuntimeKeySMSOTPMobileNumber holds the resolved mobile number for SMS OTP verification.
-	// TODO: Revisit when the generic OTP executor is implemented.
 	RuntimeKeySMSOTPMobileNumber = "smsOTPMobileNumber"
 	// RuntimeKeySMSOTPPhoneAttr holds the schema attribute name used to look up the mobile number.
-	// TODO: Revisit when the generic OTP executor is implemented.
 	RuntimeKeySMSOTPPhoneAttr = "smsOTPPhoneAttr"
+	// RuntimeKeyOTPValue holds the plaintext OTP produced by OTPExecutor in generate mode.
+	// SMSExecutor reads this key via its generic RuntimeData loop and passes it to the SMS template as "otp".
+	RuntimeKeyOTPValue = "otp"
```
ThaminduDilshan: Rather than adding generated otp to the runtime data, we can utilize forwarded data for this similar to magic link executor right? The otp executor can set the `common.ForwardedDataKeyTemplateData` with the required data

backend/internal/flow/common/constants.go
```diff
@@ -231,11 +231,17 @@ const (
 	// that have already been prompted to the user, even if the user left them empty.
 	RuntimeKeyPresentedOptionalInputs = "presentedOptionalInputs"
 	// RuntimeKeySMSOTPMobileNumber holds the resolved mobile number for SMS OTP verification.
-	// TODO: Revisit when the generic OTP executor is implemented.
```
ThaminduDilshan, line 176: I think we have to generalize these two keys to be used with any delivery method (sms, whatsapp, email, etc) right?
If we use otp executor with email delivery (for email otp), runtime keys should store email address and the attribute key.

backend/internal/authn/otp/service.go
```diff
@@ -47,21 +47,20 @@ var supportedChannels = []notifcommon.ChannelType{notifcommon.ChannelTypeSMS}
 type OTPAuthnServiceInterface interface {
```
ThaminduDilshan, line 42: IMO we can simplify this interface to just have GenerateOTP() and Authenticate() methods. SendOTP is no longer a responsibility of this.
Internal authn service is just responsible for generating and verifying (or authenticating) the otp. There are two layers which uses this. Flow does it's own delivery via a pluggable executor. Only atomic APIs need a way to do explicit delivery. But that should also go into the caller similar to flow. Let's move that logic to the internal/authn/service.go

backend/internal/authn/otp/service.go
```diff
@@ -47,21 +47,20 @@ var supportedChannels = []notifcommon.ChannelType{notifcommon.ChannelTypeSMS}
...
+	GenerateOTP(ctx context.Context, recipient, recipientAttr string) (
+		sessionToken string, otpValue string, svcErr *serviceerror.ServiceError)
 	Authenticate(ctx context.Context, sessionToken, otp string) (*common.AuthnResult, *serviceerror.ServiceError)
 }
 
 // otpAuthnService is the default implementation of OTPAuthnServiceInterface.
 type otpAuthnService struct {
-	otpService     notification.OTPServiceInterface
-	entityProvider entityprovider.EntityProviderInterface
+	otpService notification.OTPServiceInterface
```
ThaminduDilshan: We don't need to have two otp services; one in authn and another in notification package. Ideally we should take out otp service from notification package and move into the authn/otp package.
Let's align the implementation similar to the magic link authn service

backend/internal/authn/otp/service.go
```diff
@@ -47,21 +47,20 @@ var supportedChannels = []notifcommon.ChannelType{notifcommon.ChannelTypeSMS}
 type OTPAuthnServiceInterface interface {
 	SendOTP(ctx context.Context, senderID string, channel notifcommon.ChannelType,
 		recipient string) (string, *serviceerror.ServiceError)
+	GenerateOTP(ctx context.Context, recipient, recipientAttr string) (
+		sessionToken string, otpValue string, svcErr *serviceerror.ServiceError)
 	Authenticate(ctx context.Context, sessionToken, otp string) (*common.AuthnResult, *serviceerror.ServiceError)
 }
 
 // otpAuthnService is the default implementation of OTPAuthnServiceInterface.
 type otpAuthnService struct {
```
ThaminduDilshan, line 49: This might still require `entityProvider` for user verification during `Authenticate()` right?
cc: @ThumulaPerera

backend/internal/authn/otp/service.go
```diff
@@ -90,6 +89,34 @@ func (s *otpAuthnService) SendOTP(ctx context.Context, senderID string, channel
...
 }
 
+// GenerateOTP generates an OTP and session token for the recipient without delivering it.
+// recipientAttr identifies which attribute the recipient value represents (e.g. "userID",
+// "mobileNumber"). Use authnprovidercm.UserAttributeUserID for authentication flows and
+// the actual attribute name (e.g. common.AttributeMobileNumber) for registration flows.
+func (s *otpAuthnService) GenerateOTP(ctx context.Context,
+	recipient, recipientAttr string) (string, string, *serviceerror.ServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
+	logger.Debug(ctx, "Generating OTP without delivery", log.MaskedString("recipient", recipient))
```
ThaminduDilshan:
```suggestion
	logger.Debug(ctx, "Generating OTP", log.MaskedString("recipient", recipient))
```

backend/internal/authn/otp/service.go
```diff
@@ -90,6 +89,34 @@ func (s *otpAuthnService) SendOTP(ctx context.Context, senderID string, channel
...
+	if strings.TrimSpace(recipient) == "" {
+		return "", "", &ErrorInvalidRecipient
+	}
+	if recipientAttr == "" {
+		recipientAttr = authnprovidercm.UserAttributeUserID
+	}
+
+	result, svcErr := s.otpService.GenerateOTP(ctx, notifcommon.GenerateOTPDTO{
+		Recipient:     recipient,
+		RecipientAttr: recipientAttr,
```
ThaminduDilshan: Ideally otp generation shouldn't require recipient or the attribute. But authn notification service will require it to perform user verification. Anyway this will get refactored with my previous comment

backend/internal/authn/init.go
```diff
@@ -63,6 +63,10 @@ func Initialize(
 		Name:    common.AuthenticatorSMSOTP,
 		Factors: []common.AuthenticationFactor{common.FactorPossession},
 	})
+	common.RegisterAuthenticator(common.AuthenticatorMeta{
```
ThaminduDilshan, line 67: Rather than introducing a new one, let's replace the existing sms otp authn service with the refactored otp service

backend/internal/flow/executor/error_constants.go
```diff
@@ -1175,18 +1175,3 @@ func errMaxOTPAttemptsReachedFor(count int) *serviceerror.ServiceError {
 		"The maximum number of OTP verification attempts (%d) has been reached", count)
 	return &e
 }
-
-// errAttributeNotFoundFor returns a ServiceError for a specific attribute not found for the user.
-func errAttributeNotFoundFor(attrName string) *serviceerror.ServiceError {
```
ThaminduDilshan: Hope usages of these methods get removed with your changes

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,372 @@
...
+		return execResp, nil
+	}
+
+	if !e.HasRequiredInputs(ctx, execResp) {
+		logger.Debug(ctx.Context, "Required inputs for OTP verification are not provided")
+		execResp.Status = common.ExecUserInputRequired
+		return execResp, nil
+	}
+
+	if err := e.getAuthenticatedUser(ctx, execResp); err != nil {
```
ThaminduDilshan, line 185: Shouldn't we call otp authn service's authenticate method during otp verify?

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,372 @@
...
+		execResp.Error = errMaxOTPAttemptsReachedFor(attemptCount)
+		return 0, nil
+	}
+
+	return attemptCount, nil
+}
+
+// getMaxOTPAttempts returns the maximum OTP generation attempts from NodeProperties,
+// falling back to 3 if not set or invalid.
+func getMaxOTPAttempts(ctx *core.NodeContext) int {
```
ThaminduDilshan:
```suggestion
func (e *otpExecutor) getMaxOTPAttempts(ctx *core.NodeContext) int {
```
Let's convert this to a receiver function

backend/internal/flow/executor/sms_executor.go
```diff
@@ -125,11 +125,11 @@ func (e *smsExecutor) Execute(ctx *core.NodeContext) (*common.ExecutorResponse,
 	}
 	scenario := template.ScenarioType(tmplStr)
 
-	templateData := template.TemplateData{
-		"appName":        ctx.Application.Name,
-		"inviteLink":     ctx.RuntimeData[common.RuntimeKeyInviteLink],
-		"bindingMessage": ctx.RuntimeData[common.RuntimeKeyBindingMessage],
+	templateData := template.TemplateData{}
+	for k, v := range ctx.RuntimeData {
```
ThaminduDilshan: This adds everything in runtime data to the template data. Is this required? Better to only add a selective set of runtime data to the template

backend/internal/notification/common/model.go
```diff
@@ -95,16 +95,29 @@ type VerifyOTPResponse struct {
 
 // SendOTPDTO represents the service layer data structure for sending an OTP.
 type SendOTPDTO struct {
-	Recipient string
-	SenderID  string
-	Channel   string
+	Recipient     string
+	SenderID      string
+	Channel       string
+	RecipientAttr string
```
ThaminduDilshan: Why is this required to be in here?

backend/internal/authn/common/constants.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/common/constants.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/common/constants.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/init.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/executor/error_constants.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/notification/common/model.go
Dilusha-Madushan: @coderabbitai check this with new changed

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai is it resolved now?

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai is it resolved now?

backend/cmd/server/config/resources/templates/sms_otp.yaml
```diff
@@ -3,4 +3,4 @@ displayName: "SMS OTP Verification"
 scenario: "OTP"
 type: "sms"
 contentType: "text/plain"
-body: "Your verification code is: {{ctx(otp)}}. This code will expire in {{ctx(expiryMinutes)}} minutes."
+body: "Your verification code is: {{ctx(generatedOtpCode)}}. Expires in {{ctx(expiryMinutes)}} minutes."
```
ThaminduDilshan: 1. Is this due to this being conflicted with otp user input? If not I prefer to keep "otp". Otherwise shall we simply use "otpCode"?

2. "Expires in {{ctx(expiryMinutes)}} minutes." sentence feels incomplete. WDYT?

backend/internal/notification/handler.go
```diff
@@ -223,7 +223,11 @@ func (h *messageNotificationSenderHandler) HandleOTPSendRequest(w http.ResponseW
 		return
 	}
 
-	otpDTO := common.SendOTPDTO(*request)
```
ThaminduDilshan, line 226: Is this a required change? 

@coderabbitai verify. Provide a short reply

backend/internal/notification/init.go
```diff
@@ -22,6 +22,7 @@ import (
 	"context"
 	"net/http"
 
+	authnOTP "github.com/thunder-id/thunderid/internal/authn/otp"
```
ThaminduDilshan: IMO this pattern is wrong. We still have otp service in two different packages; authn and notification. My initial idea is to merge these two services into a single service similar to the magic link service. There's nothing much logic to split this into two services. Responsibility of the otp service is to provide methods to generate and verify otp. If we're to keep two services, we should make the notification package one the core service. It shouldn't have a dependency to the authn/otp package. 

We have two options here;
1. Merge notification/otp_service into authn/otp/service keeping all the logic there
2. Keep notification/otp_service as a generic otp service to handle otp generation and verification. Any service that has a otp requirement can use this (future use cases). Make authn/otp/service an authentication specific service that uses notification/otp_service underline.

If we're to keep two services, shall we follow the second approach?

backend/cmd/server/config/resources/templates/sms_otp.yaml
Dilusha-Madushan: Is this due to this being conflicted with otp user input? If not I prefer to keep "otp". Otherwise shall we simply use "otpCode"? Yes, I usded that name to avoid the conflicts and I go with your suggestion for this variable name.

"Expires in {{ctx(expiryMinutes)}} minutes." sentence feels incomplete. WDYT? I will improve this sentence.

backend/internal/notification/init.go
Dilusha-Madushan: I will go with option 2.

backend/internal/authn/otp/service.go
```diff
@@ -38,141 +37,82 @@ const (
...
-	result, svcErr := s.otpService.SendOTP(ctx, otpData)
-	if svcErr != nil {
-		return "", s.handleOTPServiceError(ctx, svcErr, false, logger)
+	if recipientAttr == "" {
+		recipientAttr = authnprovidercm.UserAttributeUserID
 	}
 
-	logger.Debug(ctx, "OTP sent successfully, session token generated")
-	return result.SessionToken, nil
+	return s.notifOTPService.GenerateOTP(ctx, recipient, recipientAttr)
```
ThaminduDilshan: We should wrap and return a otp authn service error. Cannot return a notification otp service error from the caller service

backend/internal/authn/otp/service.go
```diff
@@ -38,141 +37,82 @@ const (
...
-			return tidcommon.CustomServiceError(ErrorClientErrorFromOTPService, tidcommon.I18nMessage{
-				Key:          "error.otpauthnservice.error_verifying_otp_description",
-				DefaultValue: fmt.Sprintf("Error verifying OTP: %s", svcErr.ErrorDescription.DefaultValue),
-			})
-		} else {
-			return tidcommon.CustomServiceError(ErrorClientErrorFromOTPService, tidcommon.I18nMessage{
-				Key:          "error.otpauthnservice.error_sending_otp_description",
-				DefaultValue: fmt.Sprintf("Error sending OTP: %s", svcErr.ErrorDescription.DefaultValue),
-			})
+		if svcErr.Code == notification.ErrorInvalidSessionToken.Code {
```
ThaminduDilshan: We should handle other client errors like invalid otp here too right? Otherwise a internal server error will be return for all other cases.
We should handle client errors

backend/internal/authn/otp/service.go
```diff
@@ -38,141 +37,82 @@ const (
...
 	}
-
 	if result.Recipient == "" {
 		logger.Error(ctx, "Recipient not found in OTP verification result")
 		return nil, &tidcommon.InternalServerError
 	}
 
+	recipientAttr := result.RecipientAttr
+	if recipientAttr == "" {
+		recipientAttr = userAttributeMobileNumber
```
ThaminduDilshan, line 125: IMO we shouldn't default to mobile anymore

backend/internal/authn/otp/service.go
ThaminduDilshan: Seems Invalid otp is handled with the verification status. But we should handle client errors

backend/internal/authn/service.go
```diff
@@ -195,10 +204,31 @@ func (as *authenticationService) AuthenticateWithCredentials(ctx context.Context
...
 }
 
-// SendOTP sends an OTP to the specified recipient for authentication.
+// SendOTP generates an OTP, renders the SMS template, and delivers it to the recipient.
 func (as *authenticationService) SendOTP(ctx context.Context, senderID string, channel notifcommon.ChannelType,
 	recipient string) (string, *tidcommon.ServiceError) {
-	return as.otpService.SendOTP(ctx, senderID, channel, recipient)
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, svcLoggerComponentName))
+
+	sessionToken, otpValue, _, svcErr := as.otpService.GenerateOTP(ctx, recipient, "mobile_number")
```
ThaminduDilshan, line 212: Seems atomic API is hard coded to use `mobile_number` attribute. We should generalize it to use any attribute. Even with the attributes being flexible with the user type concept, we cannot hard code attributes this level.
Currently we only take "senderId" and "recipient" in the request. Maybe we can allow sending this in the authentication request as well. Otherwise need to think on a configuration model for the atomic APIs.

@coderabbitai can you create a follow up issue?

backend/internal/authn/service.go
```diff
@@ -195,10 +204,31 @@ func (as *authenticationService) AuthenticateWithCredentials(ctx context.Context
...
+	if svcErr != nil {
+		return "", svcErr
+	}
+
+	otpCfg := config.GetServerRuntime().Config.Notification.OTP
+	expiryMinutes := strconv.FormatInt(int64(otpCfg.ValidityPeriodSeconds)/60, 10)
+	templateData := template.TemplateData{"otpCode": otpValue, "expiryMinutes": expiryMinutes}
+	rendered, renderErr := as.templateService.Render(ctx, template.ScenarioOTP, template.TemplateTypeSMS, templateData)
+	if renderErr != nil {
+		logger.Error(ctx, "Failed to render OTP template", log.String("error", renderErr.Code))
```
ThaminduDilshan: We should handle client errors. Should only log the server errors. Also we should return a wrapped service error in all the places where we call other services.

Same applies to the places where we invoke notifSenderSvc

backend/internal/flow/common/constants.go
```diff
@@ -168,12 +168,16 @@ const (
...
-	RuntimeKeySMSOTPMobileNumber = "smsOTPMobileNumber"
-	// RuntimeKeySMSOTPPhoneAttr holds the schema attribute name used to look up the mobile number.
-	// TODO: Revisit when the generic OTP executor is implemented.
-	RuntimeKeySMSOTPPhoneAttr = "smsOTPPhoneAttr"
+	// RuntimeKeyOTPValue is the key used for the plaintext OTP within the ForwardedData template
+	// data map produced by OTPExecutor in generate mode. Sender executors (SMSExecutor,
+	// EmailExecutor) read this key from ForwardedData[ForwardedDataKeyTemplateData] when rendering
+	// the OTP notification template.
+	RuntimeKeyOTPValue = "otpCode"
```
ThaminduDilshan: We don't need to add otp value to the runtime data right? We can use forwarded data to forward it to the next executor

backend/internal/flow/core/utils.go
```diff
@@ -159,10 +159,14 @@ func collectMissingInputs(ctx *providers.NodeContext, presentedOptionalInputs ma
 		if _, ok := ctx.UserInputs[input.Identifier]; ok {
 			continue
 		}
-		if _, ok := ctx.RuntimeData[input.Identifier]; ok {
-			logger.Debug(ctx.Context, "Input available in runtime data, skipping",
-				log.String("identifier", input.Identifier), log.Bool("isRequired", input.Required))
-			continue
+		// OTP inputs require explicit user submission; skip the runtime data check entirely
+		// to prevent a generated OTP value in RuntimeData from bypassing the user prompt.
+		if input.Type != providers.InputTypeOTP {
```
ThaminduDilshan: Do we need this change? If we're checking for a specific otp key in runtime data, it will be differed from otp user input right?

Also if we do https://github.com/thunder-id/thunderid/pull/3477/changes#r3499309823, this change is no longer required

backend/internal/flow/executor/email_executor.go
```diff
@@ -123,7 +123,10 @@ func (e *emailExecutor) executeSend(ctx *providers.NodeContext) (*providers.Exec
 			return nil, fmt.Errorf("email template property is empty in node configuration")
 		}
 		scenario = template.ScenarioType(tmplStr)
+		logger.Info(ctx.Context, "EmailExecutor: resolved email template", log.String("scenario", tmplStr))
```
ThaminduDilshan:
```suggestion
		logger.Debug(ctx.Context, "EmailExecutor: resolved email template", log.String("scenario", tmplStr))
```

backend/internal/flow/executor/email_executor.go
```diff
@@ -123,7 +123,10 @@ func (e *emailExecutor) executeSend(ctx *providers.NodeContext) (*providers.Exec
...
 	} else {
+		logger.Info(ctx.Context, "EmailExecutor: emailTemplate property missing from node configuration",
```
ThaminduDilshan: This log is not required right?

backend/internal/flow/executor/email_executor.go
ThaminduDilshan: Do we need this log at all?

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	systemutils "github.com/thunder-id/thunderid/internal/system/utils"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// otpExecutor implements a channel-agnostic OTP generate/verify executor.
+// Generate mode: identifies the user from declared node inputs, generates an OTP session
+// token keyed to the userID, and forwards the plaintext OTP via ForwardedData so that a
+// downstream sender executor (SMSExecutor, EmailExecutor, etc.) can render it into a template
+// without persisting it for the life of the flow session.
+// Verify mode: validates the OTP code against the session token and authenticates the user.
```
ThaminduDilshan, line 39: Shall we optimize the comment? Lot of unnecessary details are mentioned here

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+		return e.executeGenerate(ctx, execResp)
+	case ExecutorModeVerify:
+		return e.executeVerify(ctx, execResp)
+	default:
+		return execResp, fmt.Errorf("invalid executor mode: %s", ctx.ExecutorMode)
+	}
+}
+
+// executeGenerate identifies the user from declared node inputs, generates an OTP, and
+// stores the session token and plaintext OTP value in RuntimeData for downstream executors.
```
ThaminduDilshan: Here it says plaintext otp value is added to runtime data. But otpExecutor struct comments says otherwise.
Anyway this should be refactored with my previous comment

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+//
+// For authentication flows the recipient is the resolved userID. For registration flows
+// the user does not exist yet, so when identification fails the first PHONE_INPUT nodeInput
+// attribute value is used directly as the recipient (matching the MagicLink pattern).
+func (e *otpExecutor) executeGenerate(ctx *providers.NodeContext,
```
ThaminduDilshan, line 116: Is it correct to use `PHONE_INPUT` attribute for user identification? This is no longer a sms otp executor

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	execResp *providers.ExecutorResponse) (*providers.ExecutorResponse, error) {
+	logger := e.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+
+	logger.Info(ctx.Context, "OTP generate started",
```
ThaminduDilshan:
```suggestion
	logger.Debug(ctx.Context, "OTP generate started",
```

Let's not add unnecessary info logs

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	if execResp.Status == providers.ExecFailure {
+		return execResp, nil
+	}
+
+	userID, err := e.resolveUserID(ctx, execResp)
+	if err != nil {
+		return execResp, err
+	}
+
+	logger.Info(ctx.Context, "OTP generate: resolveUserID result",
```
ThaminduDilshan:
```suggestion
	logger.Debug(ctx.Context, "OTP generate: resolveUserID result",
```

ThaminduDilshan: - Shall we reword the logs to follow project's existing pattern?
- Also let's see what data to be logged along with logs. Seems too much data is added to the logs

ThaminduDilshan: Check all other places as well. Seems you have added lot of info logs unnecessarily.
Also check for the log text and data as well

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+		log.String("status", string(execResp.Status)),
+	)
+
+	recipient := userID
+	recipientAttr := authnprovidercm.UserAttributeUserID
+
+	if recipient == "" && execResp.Status == providers.ExecFailure {
+		if ctx.FlowType != providers.FlowTypeRegistration {
+			logger.Info(ctx.Context, "OTP generate: user not found, non-registration flow — aborting",
+				log.String("errorCode", func() string {
```
ThaminduDilshan: Need to embed this inside a isDebug check

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+					if execResp.Error != nil {
+						return execResp.Error.Code
+					}
+					return ""
+				}()),
+			)
+			return execResp, nil
+		}
+		// Registration: user does not exist yet — use the declared nodeInput attribute value
```
ThaminduDilshan, line 141:
```suggestion

		// Registration: user does not exist yet — use the declared nodeInput attribute value
```

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+			log.Bool("hasDestValue", destValue != ""),
+		)
+		if destValue == "" {
+			logger.Info(ctx.Context, "OTP generate: no destination value in registration flow, prompting for input")
+			execResp.Status = providers.ExecUserInputRequired
+			execResp.Error = nil
+			execResp.Inputs = e.getGenerateInputs(ctx)
+			return execResp, nil
+		}
+		execResp.Status = ""
```
ThaminduDilshan, line 151:
```suggestion

		execResp.Status = ""
```

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	sessionToken, otpValue, expirySeconds, svcErr := e.otpService.GenerateOTP(ctx.Context, recipient, recipientAttr)
+	if svcErr != nil {
+		return execResp, fmt.Errorf("failed to generate OTP: %s", svcErr.ErrorDescription.DefaultValue)
+	}
+
+	execResp.RuntimeData[common.RuntimeKeyOTPSessionToken] = sessionToken
+	execResp.RuntimeData[common.RuntimeKeyOTPAttemptCount] = strconv.Itoa(attemptCount + 1)
+	execResp.ForwardedData[common.ForwardedDataKeyTemplateData] = map[string]interface{}{
+		common.RuntimeKeyOTPValue: otpValue,
+		"expiryMinutes":           strconv.FormatInt(expirySeconds/60, 10),
```
ThaminduDilshan: Shall we define a constant for `expiryMinutes`?

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	logger := e.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+
+	logger.Info(ctx.Context, "OTP generate started",
+		log.String("flowType", string(ctx.FlowType)),
+		log.Int("nodeInputCount", len(ctx.NodeInputs)),
+		log.Int("userInputCount", len(ctx.UserInputs)),
+		log.Bool("hasUserIDInRuntime", ctx.RuntimeData[userAttributeUserID] != ""),
+	)
+
+	attemptCount, err := e.validateAttempts(ctx, execResp, logger)
```
ThaminduDilshan, line 120: Since we return a otp session token, we should be able to embed attempt validation logic into the core otp service. This way same attempt validations can be made available for the atomic APIs as well.

@coderabbitai can you create a followup issue to address this?

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	if err != nil {
+		return execResp, err
+	}
+	if execResp.Status == providers.ExecFailure {
+		return execResp, nil
+	}
+
+	userID, err := e.resolveUserID(ctx, execResp)
```
ThaminduDilshan, line 128: There are two different flavors for OTP generally; send OTP blindly to any valid number and send OTP only upon valid user verification. We follow the second pattern. But we may need to allow configuring the user verification behaviour maybe with a node property.

This can be generalized for any executor as well. @coderabbitai can you create a followup issue?

backend/internal/flow/executor/otp_executor.go
```diff
@@ -0,0 +1,480 @@
...
+	return *identifiedUserID, nil
+}
+
+// buildSearchAttributes collects non-OTP input values from UserInputs, RuntimeData, and ForwardedData
+// using the identifiers declared in NodeInputs.
+func (e *otpExecutor) buildSearchAttributes(ctx *providers.NodeContext) map[string]interface{} {
+	logger := e.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
+	attrs := make(map[string]interface{})
+
+	inputs := ctx.NodeInputs
```
ThaminduDilshan, line 241: We can expect here to define a single input for the node right? Do we need to have this much complex logic defined here?

Also we already have methods in base executor to get input values. Can we reuse them rather than duplicating the logic here?

backend/internal/flow/executor/sms_executor.go
```diff
@@ -128,10 +128,20 @@ func (e *smsExecutor) Execute(ctx *providers.NodeContext) (*providers.ExecutorRe
 	}
 	scenario := template.ScenarioType(tmplStr)
 
-	templateData := template.TemplateData{
-		"appName":        ctx.Application.Name,
-		"inviteLink":     ctx.RuntimeData[common.RuntimeKeyInviteLink],
-		"bindingMessage": ctx.RuntimeData[common.RuntimeKeyBindingMessage],
+	templateData := template.TemplateData{"appName": ctx.Application.Name}
+	if ctx.ForwardedData != nil {
+		if fwdTemplateData, ok := ctx.ForwardedData[common.ForwardedDataKeyTemplateData]; ok {
```
ThaminduDilshan, line 133: Hope this doesn't break existing invite and magic link executor behaviour.

backend/internal/notification/otp_service.go
```diff
@@ -38,8 +39,27 @@ import (
 	"github.com/thunder-id/thunderid/internal/system/template"
 )
 
+const otpSessionAudience = "otp-svc"
+
+// otpSessionData holds the data encoded in the OTP session JWT.
+// JSON field names match those used by authn/otp for session token compatibility.
+type otpSessionData struct {
```
ThaminduDilshan, line 46: Does authn/otp service needs to decode and validate fields inside session token? I feel like that's not required.
If required, we should export this model and let authn service reuse the same model. But let's have a chat first and decide

backend/internal/notification/otp_service.go
```diff
@@ -65,6 +85,40 @@ func newOTPService(notifSenderSvc NotificationSenderMgtSvcInterface,
 	}
 }
 
+// GenerateOTP generates an OTP and session token for the recipient without delivering it.
+func (s *otpService) GenerateOTP(ctx context.Context, recipient, recipientAttr string) (
+	string, string, int64, *tidcommon.ServiceError) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "OTPService"))
```
ThaminduDilshan: Shall we move logger to otpService struct rather than initializing in each method?

backend/internal/notification/otp_service.go
```diff
@@ -198,68 +232,12 @@ func (s *otpService) validateOTPVerifyRequest(request common.VerifyOTPDTO) *tidc
...
-// getOTPCharset returns the character set for OTP generation.
-func (s *otpService) getOTPCharset() string {
-	if s.useOnlyNumericChars() {
-		return "9245378016"
-	}
-	return "KIGXHOYSPRWCEFMVUQLZDNABJT9245378016"
-}
-
-// resolveOTPConfig returns the effective OTP configuration for this otpService.
-func (s *otpService) resolveOTPConfig() config.OTPConfig {
```
ThaminduDilshan, line 239: We have an idea to allow overriding these configs with the flow configs. Ideally configs will be passed to the otp service in the `GenerateOTP` method. Better to keep it

backend/internal/notification/otp_service.go
```diff
@@ -38,8 +39,27 @@ import (
...
+type generatedOTP struct {
+	Value              string
+	ExpiryTimeInMillis int64
+}
+
 // OTPServiceInterface defines the interface for OTP operations.
 type OTPServiceInterface interface {
+	GenerateOTP(ctx context.Context, recipient, recipientAttr string) (
+		sessionToken string, otpValue string, expirySeconds int64, svcErr *tidcommon.ServiceError)
 	SendOTP(ctx context.Context, request common.SendOTPDTO) (*common.SendOTPResultDTO, *tidcommon.ServiceError)
```
ThaminduDilshan, line 63: Feels incorrect to have `SendOTP()` method here. But let's revisit it later.
@coderabbitai can you create an issue?

frontend/apps/console/src/features/flows/components/resources/steps/execution/execution-factory/ExecutionFactory.tsx
```diff
@@ -59,10 +58,6 @@ function ExecutionFactory({resource}: ExecutionFactoryPropsInterface): ReactElem
     return <GithubExecution resource={resource} />;
   }
 
-  if (executorName === ExecutionTypes.SMSOTPAuth) {
```
ThaminduDilshan, line 62: Don't we need to add OTP executor?

frontend/apps/console/src/features/flows/validation/validation-rules.ts
```diff
@@ -178,12 +178,6 @@ export const VALIDATION_RULES: ValidationRuleDefinition[] = [
     fields: [{name: 'data.properties.idpId', errorMessageKey: 'flows:core.validation.fields.input.idpId'}],
     generalMessageKey: 'flows:core.validation.fields.executor.general',
   },
-  // SMS OTP executor
```
ThaminduDilshan, line 181: I think we need to add required validations for the otp executor

frontend/apps/console/src/features/login-flow/components/resource-property-panel/extended-properties/execution-properties/SmsOtpProperties.tsx
```diff
@@ -1,142 +0,0 @@
...
-import {Alert, FormHelperText, FormLabel, MenuItem, Select, Stack, Typography} from '@wso2/oxygen-ui';
-import {useMemo, type ReactNode} from 'react';
-import {useTranslation} from 'react-i18next';
-import {SMS_OTP_MODES} from './constants';
-import type {CommonResourcePropertiesPropsInterface} from './types';
-import useValidationStatus from '@/features/flows/hooks/useValidationStatus';
-import type {StepData} from '@/features/flows/models/steps';
-import useNotificationSenders from '@/features/notification-senders/api/useNotificationSenders';
-
-function SmsOtpProperties({resource, onChange}: CommonResourcePropertiesPropsInterface): ReactNode {
```
ThaminduDilshan, line 28: Similar otp properties should be there for otp executor right?

tests/e2e/utils/server-setup/mfa-flow-nodes.json
```diff
@@ -340,9 +356,9 @@
...
                 "action": {
                     "ref": "resend_6o42",
-                    "nextNode": "send_otp"
+                    "nextNode": "generate_otp"
                 }
             }
         ]
     }
-]
+]
```
ThaminduDilshan: Add an empty line at end of each file

backend/internal/authn/otp/service.go
ThaminduDilshan: Check other places as well

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: I added some logs for debugging the code for testing. I will remove the unnecessary logs from changed files.

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: If we plan to use multiple send executors then those cases we can have multiple node inputs right? That's why I considered multiple node inputs possible for the otp executor in generate mode time.

backend/internal/authn/otp/service.go
Dilusha-Madushan: what should we use as fallback option here? or if recipientAttr is missing then do hard failure.

backend/internal/notification/otp_service.go
Dilusha-Madushan: @coderabbitai check status with new changes

frontend/apps/console/src/features/flows/components/resources/steps/execution/execution-factory/ExecutionFactory.tsx
Dilusha-Madushan: @coderabbitai check status with new changes

frontend/apps/console/src/features/flows/validation/validation-rules.ts
Dilusha-Madushan: @coderabbitai check status with new changes

frontend/apps/console/src/features/login-flow/components/resource-property-panel/extended-properties/execution-properties/SmsOtpProperties.tsx
Dilusha-Madushan: @coderabbitai check status with new changes

tests/e2e/utils/server-setup/mfa-flow-nodes.json
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/notification/otp_service.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/notification/otp_service.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/sms_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/email_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/executor/email_executor.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/core/utils.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/authn/otp/service.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/authn/service.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/flow/common/constants.go
Dilusha-Madushan: @coderabbitai check status with new changes

backend/internal/authn/service.go
Dilusha-Madushan: @coderabbitai check this now

backend/internal/notification/otp_service.go
Dilusha-Madushan: @coderabbitai 
resolveOTPConfig() is already the active config accessor — it's called by getOTPCharset(), getOTPLength(), getOTPValidityPeriodInMillis(), and sendSMSOTP(). The reviewer's comment ("Better to keep it") was addressing a previous suggestion to inline the config access, not requesting that we wire flow-level overrides now. Introducing a config parameter on GenerateOTP is out of scope for this PR, which focuses on decoupling the OTP executor. The method exists as the single choke point for future extensibility when that override is eventually needed.

backend/internal/flow/executor/otp_executor.go
Dilusha-Madushan: @coderabbitai comment on this

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #3545, opened 2026-06-25
Update default agent type schema with structured attributes

Dilusha-Madushan:
## Purpose

Replace the default agent type schema fields (`model`, `department`, `purpose`) with a structured, semantically meaningful set of attributes that better describe an agent's identity and operational context.

The previous schema used free-form strings with no enforcement. The new schema introduces enum-validated fields for **provider** and **function**, and a free-text field for the **model identifier**, keeping the contract minimal and decoupled from model release cycles.

---

## Approach

Replaced the `POST /agent-types` payload with the following focused fields:

| Field | Type | Rationale |
|-------|------|-----------|
| `modelProvider` | `string` (enum, required) | Identifies the inference runtime (e.g., `anthropic`, `gemini`). The enum is currently limited to well-known providers (`openai`, `anthropic`, `gemini`, `mistral`, `custom`) and can define custom enums from schema. `custom` covers any non-standard runtime. |
| `model` | `string` (required) | A free-text field rather than an enum to avoid coupling the schema to model release cycles. New model versions can be introduced without requiring schema updates. |
| `function` | `string` (enum, optional) | Describes the functional role of the agent (e.g., `task-automation`, `rag-retrieval`, `code-gen`). This field is optional because not all deployments require function classification during registration. The `custom` value supports non-standard agent functions. |

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3491

## Summary by CodeRabbit

* **Documentation**
  * Updated agent type create/view/edit examples and API schema to use the latest required fields: model provider, model, agent function, and action mode.
  * Added a `displayName` label to schema property definitions for clearer UI naming.

* **Bug Fixes**
  * Updated the default agent type setup payload to align with the new required and validated schema fields.

* **Tests**
  * Refreshed frontend agent-type schema fixtures to reflect the updated contract.

### DIFF
--- a/api/agent.yaml
+++ b/api/agent.yaml
@@ -1421,18 +1421,17 @@ paths:
               systemAttributes:
                 display: "name"
               schema:
-                name:
+                modelProvider:
                   type: "string"
                   required: true
-                  unique: true
-                description:
-                  type: "string"
+                  enum: ["openai", "anthropic", "gemini", "mistral", "custom"]
                 model:
                   type: "string"
                   required: true
-                  enum: ["gpt-4", "gpt-4o", "claude-sonnet", "claude-opus"]
-                isActive:
-                  type: "boolean"
+                function:
+                  type: "string"
+                  required: false
+                  enum: ["task-automation", "rag-retrieval", "code-gen", "data-analysis", "orchestrator", "sub-agent", "assistant", "custom"]
       responses:
         "201":
           description: Agent type created
@@ -1448,18 +1447,17 @@ paths:
                 systemAttributes:
                   display: "name"
                 schema:
-                  name:
+                  modelProvider:
                     type: "string"
                     required: true
-                    unique: true
-                  description:
-                    type: "string"
+                    enum: ["openai", "anthropic", "gemini", "mistral", "custom"]
                   model:
                     type: "string"
                     required: true
-                    enum: ["gpt-4", "gpt-4o", "claude-sonnet", "claude-opus"]
-                  isActive:
-                    type: "boolean"
+                  function:
+                    type: "string"
+                    required: false
+                    enum: ["task-automation", "rag-retrieval", "code-gen", "data-analysis", "orchestrator", "sub-agent", "assistant", "custom"]
         "400":
           description: Bad request
           content:
@@ -1549,23 +1547,17 @@ paths:
                 systemAttributes:
                   display: "name"
                 schema:
-                  name:
+                  modelProvider:
                     type: "string"
                     required: true
-                    unique: true
-                  description:
-                    type: "string"
+                    enum: ["openai", "anthropic", "gemini", "mistral", "custom"]
                   model:
                     type: "string"
                     required: true
-                    enum: ["gpt-4", "gpt-4o", "claude-sonnet", "claude-opus"]
-                  capabilities:
-                    type: "array"
-                    items:
-                      type: "string"
-                      enum: ["chat", "tools", "vision", "code"]
-                  isActive:
-                    type: "boolean"
+                  function:
+                    type: "string"
+                    required: false
+                    enum: ["task-automation", "rag-retrieval", "code-gen", "data-analysis", "orchestrator", "sub-agent", "assistant", "custom"]
         "404":
           description: Agent type not found
           content:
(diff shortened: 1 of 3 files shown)


api/agent.yaml
```diff
@@ -2697,25 +2735,28 @@ components:
...
-              enum: ["gpt-4", "gpt-4o", "claude-sonnet", "claude-opus"]
-            temperature:
-              type: "number"
-            isActive:
-              type: "boolean"
-            capabilities:
+            agentFunction:
+              type: "string"
+              required: true
+              enum: ["task-automation", "rag-retrieval", "code-gen", "data-analysis", "orchestrator", "sub-agent", "assistant"]
```
sahandilshan: Can user add custom value on the runtime here?

api/agent.yaml
```diff
@@ -2697,25 +2735,28 @@ components:
...
+            autonomyLevel:
+              type: "string"
+              required: true
+              enum: ["suggest", "propose", "execute"]
```
sahandilshan: What is the purpose of `autonomyLevel`

api/agent.yaml
```diff
@@ -2697,25 +2735,28 @@ components:
...
+            tags:
               type: "array"
               items:
                 type: "string"
-                enum: ["chat", "tools", "vision", "code"]
+            deploymentEnvironment:
+              type: "string"
+              enum: ["development", "staging", "production"]
```
sahandilshan: Seems this is not an agent specific attribute

api/agent.yaml
Dilusha-Madushan: This says about the agent decision depth. Added this because now a days developers build the agents with different decision depth and this will be very helpful in agent traceability. This indirectly says the agent required scopes level as well.

api/agent.yaml
Dilusha-Madushan: Added again this as special tag for the agent. But since I added general tag attribute, we can remove this from default schema.

api/agent.yaml
Dilusha-Madushan: Yes, user can define enum values from agent schema.

api/agent.yaml
```diff
@@ -1421,18 +1421,28 @@ paths:
               systemAttributes:
                 display: "name"
               schema:
-                name:
+                provider:
```
ayshsandu: modelProvider instead of just provider?

api/agent.yaml
```diff
@@ -1421,18 +1421,28 @@ paths:
...
-                  type: "boolean"
+                agentFunction:
+                  type: "string"
+                  required: true
+                  enum: ["task-automation", "rag-retrieval", "code-gen", "data-analysis", "orchestrator", "sub-agent", "assistant"]
+                autonomyLevel:
+                  type: "string"
+                  required: true
+                  enum: ["suggest", "propose", "execute"]
+                tags:
```
ayshsandu: Can remove tags and keep the default schema minimal.

api/agent.yaml
```diff
@@ -1421,18 +1421,28 @@ paths:
...
+                  type: "array"
+                  items:
+                    type: "string"
+                deploymentEnvironment:
```
ayshsandu: Environments ideally should come from the deployments. No need to have it in the default schema.

api/agent.yaml
```diff
@@ -1448,18 +1458,28 @@ paths:
...
                   display: "name"
                 schema:
-                  name:
+                  provider:
                     type: "string"
                     required: true
-                    unique: true
-                  description:
-                    type: "string"
+                    enum: ["openai", "anthropic", "google", "meta", "mistral", "aws-bedrock", "ibm", "custom"]
```
sahandilshan: What is the purpose of this provider?
I can see that openai, antropic are model providers. But what is aws-bedrock doing here?

sahandilshan: Also it would be better if we have to select the provider and a model

Dilusha-Madushan: modelProvider here represents the inference runtime that serves the model, not just the model's original creator. AWS Bedrock, for example, hosts models from Anthropic, Meta, and others an agent running anthropic/claude-3 via Bedrock is a different deployment target from one calling the Anthropic API directly.

api/agent.yaml
Dilusha-Madushan: Addressed

api/agent.yaml
Dilusha-Madushan: Addressed

api/agent.yaml
Dilusha-Madushan: Addressed

api/agent.yaml
Dilusha-Madushan: Yes, the approach is to add an optional **dependsOn** block to the string property definition in the schema, which the frontend reads to filter model options based on the selected provider. The backend stores it as passthrough metadata. This feature touching the schema engine and the renderSchemaField utility, so tracking it as a follow-up task.

api/agent.yaml
Dilusha-Madushan: Addressed

sahandilshan (dismissed):

sahandilshan (approved):

PR #3555, opened 2026-06-26
Allow hyphens in database filter keys

KD23243:
## Purpose
User schema property names can be defined with hyphens (for example `silver-mail`), but the database filter key validator only permitted alphanumeric characters, underscores, and dots. When such a property was marked `unique: true`, uniqueness validation and identification built a filter map keyed by the property name and the key was rejected, causing user creation and identification to fail with a generic internal server error.

## Related Issue
- https://github.com/thunder-id/thunderid/issues/1696

## Behaviour
ScreenRecording2026-06-26at09 25 36-ezgif com-video-to-gif-converter

## Implementation
- Added the hyphen to the allowed character set in `ValidateKey` (`backend/internal/system/database/utils/querybuilder.go`), the single chokepoint that all filter key validation routes through. Characters that could break out of the JSON path (quotes, semicolons, whitespace) remain rejected, so the existing protection is preserved. Hyphenated keys work in both dialects (`attributes->>'silver-mail'` on PostgreSQL and `json_extract(attributes, '$.silver-mail')` on SQLite), so no path builder changes were needed.
- Updated `querybuilder_test.go`: added hyphenated keys to the valid set, removed the hyphen case from the invalid set, and adjusted the invalid filter key fixture to use a still-rejected character.

## Summary by CodeRabbit

* **Bug Fixes**
  * Keys used in queries can now include hyphens, improving compatibility with more column and filter names.
  * Validation behavior has been aligned so hyphenated keys are no longer incorrectly rejected.
  * Related validation checks were updated to continue flagging truly invalid key formats.

### DIFF
--- a/backend/internal/system/database/utils/querybuilder.go
+++ b/backend/internal/system/database/utils/querybuilder.go
@@ -91,12 +91,13 @@ func BuildSQLiteJSONCondition(columnName, key string) string {
 	return fmt.Sprintf(" AND json_extract(%s, '$.%s') = ?", columnName, key)
 }
 
-// ValidateKey ensures that the provided key contains only safe characters (alphanumeric, underscores, and dots).
+// ValidateKey ensures that the provided key contains only safe characters
+// (alphanumeric, underscores, dots, and hyphens).
 // This validation prevents SQL injection by ensuring keys can be safely used in queries.
 func ValidateKey(key string) error {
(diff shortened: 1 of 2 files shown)


KD23243:
Superseded by #3556, which rejects invalid property names at schema creation time instead of broadening the allowed filter key characters.

PR #3556, opened 2026-06-26
Reject invalid characters in schema property names

### LINKED ISSUE #1696: User schema properties with hyphens fail uniqueness validation
## Problem
When a user schema property contains a hyphen (e.g., `silver-mail`) and is marked as `unique: true`, the uniqueness validation fails with an error about invalid filter keys.

## Error Messages
```
time=2026-03-06T06:04:24.093Z level=ERROR msg="Failed to identify user" component=UserService error="failed to build identify query: invalid filter key: key 'silver-mail' contains invalid characters"
time=2026-03-06T06:04:24.093Z level=ERROR msg="Failed during uniqueness validation" component=UserSchemaService error="Internal server error"
time=2026-03-06T06:04:24.093Z level=ERROR msg="Failed to validate user schema" component=UserService
```

## Root Cause
1. User schema properties can be defined with hyphens in their names (e.g., `silver-mail`)
2. When validating uniqueness for such properties, the system calls `IdentifyUser()` with a filter map containing the property name as a key: `{"silver-mail": value}`
3. The `BuildFilterQuery()` function in `internal/system/database/utils/querybuilder.go` calls `ValidateKey()` to prevent SQL injection
4. `ValidateKey()` only allows alphanumeric characters, underscores, and dots - hyphens are rejected as invalid characters
5. This validation failure propagates up and causes the entire user creation/identification flow to fail

## Affected Code
- [`backend/internal/system/database/utils/querybuilder.go`](backend/internal/system/database/utils/querybuilder.go#L98-L104) - `ValidateKey()` function
- [`backend/internal/userschema/model/string.go`](backend/internal/userschema/model/string.go#L75-L85) - `validateUniqueness()` method
- [`backend/internal/user/store.go`](backend/internal/user/store.go#L416) - `IdentifyUser()` method

## Steps to Reproduce
1. Create a user schema with a property containing hyphens, marked as unique:
   ```json
   {
     "type": "object",
     "properties": {
       "silver-mail": {
         "type": "string",
         "unique": true
       }
     }
   }
2. Create a user with a value for this property
3. The user creation will fail with the error above

### Version

v0.24.0

_No response_

KD23243:
## Purpose
User schema property names could be defined with characters such as hyphens (for example `silver-mail`), but property names are used directly as JSON filter keys during value and uniqueness lookups. The query builder only accepts letters, digits, underscores, and dots as filter keys, so such names passed schema compilation but later failed at lookup time. When the property was marked `unique: true`, user creation and identification failed with a generic internal server error.

## Related Issue
- https://github.com/thunder-id/thunderid/issues/1696

## Implementation
- Added `validatePropertyName` in `backend/internal/entitytype/model/schema.go` and call it at the start of `compileProperty`, which is the single point every property name flows through. Property names are now restricted to letters, digits, and underscores. Invalid names are rejected during schema compilation, which surfaces as a clear client-facing validation error on entity type create/update rather than a runtime failure during user creation. This covers both top-level and nested object properties.
- Added tests in `schema_test.go` for hyphenated top-level and nested property names (rejected) and an underscore name (accepted).

## Summary by CodeRabbit

* **Bug Fixes**
  * User-type schema property names are now validated during compilation, rejecting invalid characters (for example hyphens) early with a clear error.
* **New Features**
  * The user-type configuration UI validates property names inline (letters, digits, underscores only), shows helpful guidance, and prevents saving when any property name is invalid.
* **Tests**
  * Added coverage for valid/invalid property names, including nested cases, on both backend compilation and the frontend validator.
* **Documentation**
  * Added the new English validation message for invalid property names.

### DIFF
--- a/backend/internal/entitytype/model/schema.go
+++ b/backend/internal/entitytype/model/schema.go
@@ -283,7 +283,29 @@ func CompileSchema(schema json.RawMessage) (*Schema, error) {
 	return compiled, nil
 }
 
+// validatePropertyName ensures a schema property name only contains characters
+// that are safe to use as a database filter key (letters, digits, and underscores).
+// Names are used directly as JSON filter keys during value and uniqueness lookups,
+// so characters such as hyphens would later be rejected by the query builder.
+func validatePropertyName(name string) error {
+	if name == "" {
+		return fmt.Errorf("property name must not be empty")
+	}
+	for _, char := range name {
+		if !(char >= 'a' && char <= 'z' || char >= 'A' && char <= 'Z' ||
+			char >= '0' && char <= '9' || char == '_') {
+			return fmt.Errorf("property name '%s' contains invalid characters; "+
+				"only letters, digits, and underscores are allowed", name)
+		}
+	}
+	return nil
+}
+
 func compileProperty(propName string, propRaw json.RawMessage) (property, error) {
+	if err := validatePropertyName(propName); err != nil {
+		return nil, err
+	}
+
 	var propMap map[string]json.RawMessage
 	if err := json.Unmarshal(propRaw, &propMap); err != nil {
 		return nil, fmt.Errorf("property definition must be an object")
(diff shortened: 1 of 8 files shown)


backend/internal/entitytype/model/schema.go
Malith-19, line 2: Let's update the license headers.

Malith-19: Check other places as well.

PR #3560, opened 2026-06-26
Added captcha interceptor and service impl interface

### LINKED ISSUE #3513: Add captcha interceptor and the captcha service interface
## Summary

Add a built-in flow interceptor, `CaptchaInterceptor`, that runs in `PRE_NODE` mode and
validates a captcha token submitted with a node's user inputs before the node executor
runs. The interceptor does not verify the token itself — it delegates to a
`CaptchaService`. This PR introduces **only the interceptor and `CaptchaServiceInterface`**;
the concrete provider-backed service implementation should be taken up in a follow-up PR.

It fits the existing interceptor framework with no engine or runner changes:

- Implements `core.InterceptorInterface`, mirroring `challengeTokenInterceptor`.
- Configurable, not default — flows opt in per node via an `InterceptorDefinition`
  (`mode: PRE_NODE`, `scope`, `applyTo`). Not added to `DefaultInterceptors`.
- Reads the token from the user-input field named by its `fieldKey` property, declared per
  node in the flow graph (e.g. `properties: { fieldKey: "captcha_token" }`).
- On an invalid token (negative verdict from the service), returns
  `InterceptorStatusFailure` with a client-type `ServiceError`, which the runner converts
  into the engine failure response. A missing/empty token passes — the interceptor does not
  fail on absence; the prompt node enforces presence by declaring the captcha field as a
  required input.

Scope of changes: new `captcha` package (interface + result model), new interceptor file,
registration wiring (`InterceptorDependencies` + `builtinRegistrars`), threading interceptor
`Properties` through the runner into `InterceptorContext` (not done today), new error
constants, and unit tests with a mocked service.

## Goals

- A `PRE_NODE` `CaptchaInterceptor` that gates node execution on a valid captcha token,
  integrated through the existing framework with no engine/runner changes.
- Define `CaptchaServiceInterface` as the seam to any future captcha provider; the
  interceptor depends only on this interface.
- Read the token from the user-input field named by the interceptor's `fieldKey` property,
  declared per node in the flow graph.
- Thread interceptor `Properties` into `InterceptorContext` so configurable interceptors
  can read their per-declaration config.
- Reject only invalid tokens (negative verdict) with a clear client error. A missing/empty
  token passes — presence is enforced by the prompt node's required-input declaration.
- Surface operational verification failures (provider down/misconfigured) as server errors.
- Per-node opt-in via `InterceptorDefinition` (`scope` / `applyTo`).
- Unit tests covering the interceptor branches with a mocked service (~80% coverage).

## Non-goals

- No concrete captcha provider implementation (reCAPTCHA, hCaptcha, Turnstile, etc.).
- No provider configuration, secrets, HTTP client, or captcha store.
- No engine, graph-builder, or default-interceptor-set changes. (The runner gains only the
  minimal `Properties` threading described above.)
- No `POST_NODE` / `PRE_REQUEST` / `POST_REQUEST` behavior.
- No frontend/SDK changes for rendering or submitting the captcha.
- No rate limiting, replay detection, or scoring policy.

## Acceptance criteria

- `CaptchaServiceInterface` and its result model exist in a new `captcha` package
  following existing service conventions; a generated mock is available.
- `CaptchaInterceptor` implements `core.InterceptorInterface`, is registered as a built-in
  configurable interceptor, and is not part of `DefaultInterceptors`.
- A flow can declare it (`PRE_NODE`, `scope`, `applyTo`, `properties.fieldKey`) and
  graph-builder validation accepts it.
- The runner threads each interceptor unit's `Properties` into `InterceptorContext`.
- The interceptor reads the token from `UserInputs[fieldKey]`, where `fieldKey` comes from
  its `Properties`; when `fieldKey` is absent it falls back to `captcha_token`.
- Valid token (service returns success) → `Complete`, node executes.
- Missing/empty token → `Complete` (no validation, no service call); presence is enforced
  by the prompt node's required input, not the interceptor.
- Service returns a negative verdict → `Failure` with an "invalid captcha" client error.
- Service returns an operational error → engine maps it to an internal server error
  (request is not allowed through).
- Non-`PRE_NODE` mode → `Failure` without calling the service.
- Registration fails fast if the flow factory or captcha service dependency is absent.
- `go vet` / `gofmt` clean, tests pass, no `thunder`/`mosip` literals, doc comments on new
  declarations.

## Known edge cases / risks

- **Service not yet implemented.** The interface is wired through dependencies, but no
  concrete value exists this PR. The fail-fast registrar means the interceptor can't be
  enabled until the impl lands. Decision: guard the registrar now, or defer adding it to
  `builtinRegistrars` until the impl PR.
- **Presence is delegated to the node.** The interceptor never fails on an absent token; it
  relies on the prompt node declaring the captcha field as a required input. If a flow
  attaches the interceptor but does not mark that field required, the captcha can be
  bypassed by simply omitting the token. Flow authoring must keep the two in sync. (A
  possible follow-up: graph-builder validation that the `fieldKey` field is declared
  required on the targeted node.)
- **Mode misconfiguration.** A flow could declare the interceptor in the wrong mode;
  `Execute` guards on mode, but graph-builder rejection of non-`PRE_NODE` declarations is a
  possible follow-up.

sacrana0:
### Purpose
Added interceptor to capture captcha token and validate it. The actual captcha validation is not provided in this pr.

### Related Issues
- #3513 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added captcha validation as a pre-node interceptor, consuming a one-time captcha token input and blocking execution on failed verification.
  * Introduced a dedicated “Invalid captcha” service error with i18n messaging.
* **Bug Fixes**
  * Improved missing/invalid captcha handling to consistently surface the invalid-captcha error.
  * Updated interceptor registration to safely skip captcha checks when required components aren’t available.
* **Tests**
  * Added comprehensive coverage for captcha interceptor behavior, including success, failure, and error scenarios.


backend/internal/captcha/service.go
```diff
@@ -0,0 +1,34 @@
...
+package captcha
+
+import (
+	"context"
+
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// CaptchaServiceInterface defines the contract for verifying captcha tokens.
+type CaptchaServiceInterface interface {
```
ThaminduDilshan: Captcha is a separate service right? This shouldn't be added to flow/interceptor package

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,96 @@
...
+// Execute validates the captcha token carried in the node's user inputs. A missing token passes
+// through, leaving presence enforcement to the prompt node's required-input declaration; an invalid
+// token fails with a client error; an operational verification failure is surfaced as an execution
+// error so the engine rejects the request with a server error.
+func (c *captchaInterceptor) Execute(ctx *core.InterceptorContext) (*common.InterceptorResponse, error) {
+	if ctx.Mode != providers.InterceptorModePreNode {
+		return &common.InterceptorResponse{
+			Status: common.InterceptorStatusFailure,
+		}, nil
+	}
```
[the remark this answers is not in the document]
sacrana0: @coderabbitai  This is the behaviour we have agreed on for now. There is not flow configuration validation currently and will be added later. The same thing you can see in challenge_token interceptor. When flow validation be added then it will be taken care of.

backend/internal/captcha/service.go
sacrana0: Done

backend/internal/captcha/service.go
```diff
@@ -0,0 +1,34 @@
...
+// Package captcha defines the contract for verifying captcha tokens against a provider.
+package captcha
+
+import (
+	"context"
+
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// CaptchaServiceInterface defines the contract for verifying captcha tokens.
```
ThaminduDilshan: With the recent work, we may have to define a provider and move this to the pkg right?

backend/internal/flow/interceptor/constants.go
```diff
@@ -30,6 +30,18 @@ const (
 const (
 	// ChallengeTokenInterceptor is the registered name of the challenge token interceptor.
 	ChallengeTokenInterceptor = "ChallengeTokenInterceptor"
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/interceptor/constants.go
```diff
@@ -30,6 +30,18 @@ const (
...
+	// CaptchaInterceptor is the registered name of the captcha interceptor.
+	CaptchaInterceptor = "CaptchaInterceptor"
+)
+
+// Captcha interceptor configuration constants.
+const (
+	// propertyKeyFieldKey is the interceptor property naming the user-input field that carries the captcha token.
+	propertyKeyFieldKey = "fieldKey"
+
```
ThaminduDilshan:
```suggestion
```

backend/internal/flow/interceptor/constants.go
```diff
@@ -30,6 +30,18 @@ const (
...
+	// defaultCaptchaFieldKey is the user-input field read when the fieldKey property is unset.
+	defaultCaptchaFieldKey = "captcha_token"
```
ThaminduDilshan: Shall we stick to a single pattern; either camel case or snake case? I think we use camel case in most of the places

backend/internal/flow/interceptor/constants.go
sacrana0: Why to remove this new line? The formatting is same as the rest of the file.

backend/internal/flow/interceptor/constants.go
sacrana0: Same here.

backend/internal/flow/interceptor/constants.go
```diff
@@ -30,6 +30,18 @@ const (
 const (
 	// ChallengeTokenInterceptor is the registered name of the challenge token interceptor.
 	ChallengeTokenInterceptor = "ChallengeTokenInterceptor"
+
+	// CaptchaInterceptor is the registered name of the captcha interceptor.
+	CaptchaInterceptor = "CaptchaInterceptor"
+)
+
+// Captcha interceptor configuration constants.
```
NutharaNR: Shall we change this so that interceptor property keys are grouped consistently? Within the const, we could organize them into separate commented sections, with each section containing the properties for a specific interceptor.

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,96 @@
...
```
NutharaNR: Shall we refactor to use the same coding format similar to challenge token interceptor? 
```
switch ctx.mode{
 case providers.InterceptorModePreNode:
   return c.validateCaptcha(ctx)
default:
    return &common.InterceptorResponse{
			Status: common.InterceptorStatusFailure,
		}, nil

Comment (sacrana0) [reply backend/internal/flow/interceptor/captcha_interceptor.go]: Is there any specific reason for that? The current code does the same thing with less line of code. In case of challenge_token_interceptor there are multiple modes but not here.

File: backend/internal/flow/core/model.go [hunk 9]
```diff
@@ -95,6 +95,9 @@ type InterceptorContext struct {
 	ForwardedData       map[string]interface{}
 	AdditionalData      map[string]string
 
+	// Properties holds the per-declaration configuration of the interceptor unit being executed.
+	Properties map[string]any
```
NutharaNR:
```suggestion
	Properties map[string]interface{}
```

backend/internal/flow/flowexec/interceptor_runner.go
```diff
@@ -168,6 +168,7 @@ func (s *interceptorRunner) executeInterceptor(
 	ic core.InterceptorInterface,
 	name string,
 	mode providers.InterceptorMode,
+	properties map[string]any,
```
NutharaNR:
```suggestion
	properties map[string]interface{},
```

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// resolveFieldKey returns the user-input field name carrying the captcha token, read from the
+// interceptor's fieldKey property and falling back to the default when it is unset.
+func (c *captchaInterceptor) resolveFieldKey(ctx *core.InterceptorContext) string {
+	if fieldKey, ok := ctx.Properties[propertyKeyFieldKey].(string); ok && fieldKey != "" {
+		return fieldKey
+	}
+	return defaultCaptchaFieldKey
+}
```
NutharaNR: Shall we move this property key resolution to a utility(utils.go), as property key resolution will be used across all interceptors?

```
func resolvePropertyKeys(properties map[string]interface{}) string

File: backend/internal/flow/interceptor/captcha_interceptor.go (hunk 8 above)
Comment (NutharaNR) [reply backend/internal/flow/interceptor/captcha_interceptor.go]: It is suggested to maintain the coding format and improve readability.

File: backend/internal/flow/core/model.go (hunk 9 above)
Comment (sacrana0) [reply backend/internal/flow/core/model.go]: any and interface{} are equivalent. And it is suggested to use any.

File: backend/internal/flow/interceptor/captcha_interceptor.go (hunk 11 above)
Comment (sacrana0) [reply backend/internal/flow/interceptor/captcha_interceptor.go]: Currently it is not used anywhere. When it is needed then it can be taken out. It also only useful when you have a default value for the property. Let wait for any actual usage of it in other places then we can refactor it.

File: backend/internal/flow/flowexec/interceptor_runner.go (hunk 10 above)
Comment (sacrana0) [reply backend/internal/flow/flowexec/interceptor_runner.go]: Same here.

File: backend/internal/flow/interceptor/constants.go (hunk 7 above)
Comment (sacrana0) [reply backend/internal/flow/interceptor/constants.go]: The existing grouping is not interceptor based. Interceptor names are in different group and shared constants are in another. I will rearrange it as per your comment.

File: backend/internal/flow/interceptor/constants.go (hunk 4 above)
Comment (ThaminduDilshan) [reply backend/internal/flow/interceptor/constants.go]: Not a must change. But we don't need to add new lines unnecessarily. It just increases the file length.
AFAIR this is followed in executor package.

File: backend/internal/flow/interceptor/captcha_interceptor.go [hunk 12]
```diff
@@ -0,0 +1,96 @@
...
+	token := ctx.UserInputs[c.resolveFieldKey(ctx)]
+	if token == "" {
+		return &common.InterceptorResponse{
+			Status: common.InterceptorStatusComplete,
+		}, nil
+	}
+
+	result, svcErr := c.captchaService.Verify(ctx.Context, token)
+	if svcErr != nil {
+		return nil, fmt.Errorf("captcha verification failed: %s", svcErr.Code)
```
ThaminduDilshan, line 93: We should handle client errors here. Should only return a go error for the server errors

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return &common.InterceptorResponse{
+		Status: common.InterceptorStatusComplete,
+	}, nil
+}
+
+// resolveFieldKey returns the user-input field name carrying the captcha token, read from the
+// interceptor's fieldKey property and falling back to the default when it is unset.
+func (c *captchaInterceptor) resolveFieldKey(ctx *core.InterceptorContext) string {
```
ThaminduDilshan: Need to revisit this. 
 @ThaminduDilshan Tagging to track

backend/internal/flow/core/model.go
ThaminduDilshan: I think go 1.26+ recommends using any

backend/internal/flow/interceptor/constants.go
NutharaNR: Sorry for the confusion. What I meant was having a single generic group for all interceptor property keys instead of separate groups per interceptor. The current comment makes the new group sound Captcha-specific, but it could be generalized to cover property keys for any interceptor. Interceptor names and shared data keys could be grouped separately from interceptor property keys, as that is cleaner.
For example:
```
// Interceptor property key constants
const (
	// Captcha interceptor property keys
	// propertyKeyFieldKey is the interceptor property naming the user-input field that carries the captcha token.
	propertyKeyFieldKey = "fieldKey"
	// defaultCaptchaFieldKey is the user-input field read when the fieldKey property is unset.
	defaultCaptchaFieldKey = "captcha_token"

	// ABC interceptor property keys
    . . .
)
```

backend/internal/flow/interceptor/captcha_interceptor.go
sacrana0: This is a server error only. If the validation fails svcErr will be nil and result will contain the failure status. @ThaminduDilshan

backend/internal/flow/interceptor/constants.go
sacrana0: The second value is identifier value. Should it also be camel case? There are no other values for reference in this case. @ThaminduDilshan

backend/internal/flow/interceptor/captcha_interceptor.go
ThaminduDilshan: @sacrana0 created this discussion for the concern. What's your thoughts on this?

https://github.com/thunder-id/thunderid/discussions/3667

backend/internal/flow/interceptor/constants.go
ThaminduDilshan: But both are interceptor properties right? We should stick to a consistent pattern with property keys.
Anyway this will likely get removed with https://github.com/thunder-id/thunderid/discussions/3667

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,101 @@
...
+		InterceptorInterface: base,
+		captchaService:       captchaService,
+		logger:               log.GetLogger().With(log.String(log.LoggerKeyComponentName, CaptchaInterceptor)),
+	}
+}
+
+// Execute delegates to the appropriate handler based on the interceptor mode.
+func (c *captchaInterceptor) Execute(ctx *core.InterceptorContext) (*common.InterceptorResponse, error) {
+	switch ctx.Mode {
+	case providers.InterceptorModePreNode:
```
ThaminduDilshan, line 71: @sacrana0 please check https://github.com/thunder-id/thunderid/discussions/2375#discussioncomment-17519120 discussion point. This aligns to some extent with the approach you proposed (let node collect and be responsible for the interceptor inputs). To resolve other issue, we can make the captcha interceptor a post node interceptor right? Then prompt node will make sure captcha input is present in ctx and post node interceptor can validate for the token.

Also we're introducing one time use inputs with https://github.com/thunder-id/thunderid/pull/3724. Also this allows defining inputs in interceptors. With this if we mark captcha input as onetime, it will be cleared once consumed. This way, flow will collect the same input again during next time

sacrana0: > This aligns to some extent with the approach you proposed (let node collect and be responsible for the interceptor inputs). To resolve other issue, we can make the captcha interceptor a post node interceptor right?

What other issues? Does it make any difference if it is prenode or postnode? Anyway both are going to run twice right? and on the first run of postnode the captcha token will not be there and it has to return success. @ThaminduDilshan

sacrana0: Is there any functional difference between PreNode and PostNode interceptor for a PromptNode?

ThaminduDilshan: If it's a pre node interceptor, with your approach, you're blindly marking status as success even though it didn't run. That is conceptually wrong right?
If we make it a post node interceptor, node will make sure captcha token input is present. Hence captcha interceptor can only be run once with the correct behaviour

backend/internal/flow/interceptor/captcha_interceptor.go
ThaminduDilshan: @sacrana0 https://github.com/thunder-id/thunderid/pull/3724 PR is merged today. If you use `ConsumeInput()`, it will clear one time use inputs

backend/internal/flow/interceptor/register.go
```diff
@@ -186,6 +184,18 @@ func registerChallengeTokenInterceptor(deps InterceptorDependencies, registry In
 	return nil
 }
 
+// registerCaptchaInterceptor registers the captcha interceptor in the registry.
+//nolint:unparam // error return kept for signature consistency with other register* functions
+func registerCaptchaInterceptor(deps InterceptorDependencies, registry InterceptorRegistryInterface) error {
+	if deps.FlowFactory == nil || deps.CaptchaService == nil {
+		log.GetLogger().Warn(context.Background(), "Skipping captcha interceptor registration: missing dependencies",
```
ThaminduDilshan: Shall we have a debug log? Yes a warn log would be ideal here. But since this is not available for thunder default distribution, this will be printed in every startup

backend/internal/flow/interceptor/captcha_interceptor.go
```diff
@@ -0,0 +1,107 @@
...
+
+var _ core.InterceptorInterface = (*captchaInterceptor)(nil)
+
+// captchaInterceptorInputs declares the inputs consumed by the captcha interceptor. The captcha
+// token is one-time-use: the captcha provider invalidates it on first verification, so the engine
+// clears it from the flow context once it has been consumed.
+var captchaInterceptorInputs = []providers.Input{
+	{
+		Identifier: captchaTokenFieldKey,
+		Type:       providers.InputTypeText,
```
ThaminduDilshan: shall we use hidden type for this?

backend/internal/flow/interceptor/register_test.go
```diff
@@ -208,18 +207,6 @@ func (s *InterceptorRegistryTestSuite) TestregisterInterceptors_NameWithWhitespa
 	assert.True(s.T(), registry.IsRegistered(ChallengeTokenInterceptor))
 }
 
-func (s *InterceptorRegistryTestSuite) TestregisterInterceptors_RegistryNotUpdated_ReturnsError() {
```
ThaminduDilshan, line 211: Any reason to remove this test? Seems not related to this PR

sacrana0: This test is removed because for captcha interceptor we are not throwing error and because of that we have removed these lines as well - 

`if !registry.IsRegistered(name) {
	 return fmt.Errorf("interceptor %s was not registered successfully", name)
}`

That testcase is failing if added now.

backend/internal/flow/interceptor/register.go
```diff
@@ -144,10 +146,6 @@ func registerInterceptors(deps InterceptorDependencies, registry InterceptorRegi
 		if err := registrar(deps, registry); err != nil {
 			return fmt.Errorf("failed to register interceptor %s: %w", name, err)
 		}
-
-		if !registry.IsRegistered(name) {
-			return fmt.Errorf("interceptor %s was not registered successfully", name)
-		}
```
sacrana0, line 150: This check we have removed because of captcha interceptor @ThaminduDilshan

ThaminduDilshan (approved):

ThaminduDilshan:
Skipping from changelog as not supported in the default distribution

PR #3634, opened 2026-07-01
Add search and pagination to the application listing API

madurangasiriwardena:
### Purpose
The Applications listing page's search box was wired to nothing, and `GET /applications` supported neither search nor pagination — it ignored all query parameters and returned every application at once. This PR adds search and pagination, following the `?filter=` convention from #369 and the `limit`/`offset` + `links` shape used by the other list endpoints.

### Approach
`GET /applications` now accepts three optional, composable query parameters:

| Param | Type | Behavior | Notes |
|-------|------|----------|-------|
| `filter` | string | omitted → no filtering; malformed, unsupported attribute/operator, or multi-expression → **HTTP 400** | A **single expression**: `attribute op "value"`. Attributes: `name`, `clientId`, `description`. Operators: `co` (contains, case-insensitive) and `eq` (exact, case-insensitive). `and`/`or` connectors are **not** supported. |
| `limit` | int | omitted → defaults to **30**; if present must be **> 0**, else **HTTP 400** | Page size. |
| `offset` | int | omitted → defaults to **0**; if present must be **≥ 0**, else **HTTP 400** | Records to skip. |

**Examples**

```
GET /applications
GET /applications?filter=name co "portal"
GET /applications?filter=clientId eq "my-spa-client"
GET /applications?limit=10&offset=20
GET /applications?limit=10&offset=0&filter=name co "portal"
```

### Response

Two additive fields — `startIndex` and `links` — bring the response in line with the other list endpoints; existing fields are unchanged.

| Field | Type | Description |
|-------|------|-------------|
| `totalResults` | int | Total applications matching the filter. |
| `startIndex` | int | 1-based index of the first item in this page (`offset + 1`). **New.** |
| `count` | int | Number of applications in this page. |
| `applications` | array | The page of application objects. |
| `links` | array | Pagination links (`first` / `prev` / `next` / `last`, as applicable). **New.** |

### What's new (Console UI)

- **The search box now works** — plain free text, matching the **application name** (case-insensitive substring). Typing settles ~300 ms before searching; clearing restores the full list.
- **Server-side pagination** — the table pages through results via the API (page-size selector 5 / 10 / 25 / 50, starting at 10) instead of loading everything at once. Changing the search term returns to the first page.

### Design notes

- Filtering reuses the SCIM `?filter=` convention from #369; the `co` operator is added to the shared filter parser.
- `and`/`or` connectors are intentionally **not** implemented — kept to a single expression until resource filtering is standardized across all resources.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks


darshanasbg:
+1 for the approach. Lets validate the filter query is consistent with the one we are using (ref: https://github.com/thunder-id/thunderid/pull/369)

Also, lets not implement the or/and operators right now, if we don't have a pressing requirement. I would like to have the resource filtering and limiting support consistent across all the resources, before supporting more operators on specific resources.

PR #3553, opened 2026-06-26
Add dynamic cors support.

ImalshaD:
### Purpose

Resolves : https://github.com/thunder-id/thunderid/issues/3260

Make CORS allowed origins **runtime-mutable without a server restart**. Until now origins were fixed in `deployment.yaml` (`cors.allowed_origins`) and could only change with a config reload.

This is **PR2** of the server-config effort tracked in #3260. PR1 (#3512) added the generic server-config store; this PR makes the CORS middleware enforce against that store's `cors` section, so origins can be set **declaratively** (a read-only baseline loaded at boot) and/or **at runtime** via `PUT /server-config/cors` — taking effect immediately, no restart.

**Scope note:** the original #3260 proposal stored origins *per-application*. Following the design discussion on that issue, the approach was changed to a **server-wide, runtime-mutable server-config `cors` section** (more general, reuses PR1's store, no per-app/per-route coupling). Origins are now sourced **entirely** from that section; the static `deployment.yaml cors.allowed_origins` baseline is removed.

---
#### 🔧 Summary of Breaking Changes
- CORS origins are **no longer read from `deployment.yaml`** — the `cors.allowed_origins` block is removed. Origins come only from the server-config `cors` section.
- The `cors` section **value shape changed** from a bare list `["https://..."]` to an object `{ "allowedOrigins": ["https://..."] }`. The old bare-list form is rejected.

#### 💥 Impact
- Deployments relying on `deployment.yaml`'s `cors.allowed_origins` lose their allow-list until origins are migrated to the server-config `cors` section.
- Declarative resources or API clients using the old bare-list `cors` value are rejected on load / `PUT`.

#### 🔄 Migration Guide
Move origins out of `deployment.yaml`:
```yaml
# deployment.yaml — REMOVE
cors:
  allowed_origins:
    - "https://app.example.com"
```
into a declarative server-config resource (loaded at boot when `server_config.store` is `declarative` or `composite`):
```yaml
# config/resources/server_configs/cors.yaml
name: cors
value:
  allowedOrigins:
    - "https://app.example.com"
```
or apply at runtime, no restart:
```bash
curl -k -X PUT https://<host>/server-config/cors \
  -H "Authorization: Bearer <system-scope token>" \
  -H "Content-Type: application/json" \
  -d '{"allowedOrigins":["https://app.example.com"]}'
```

---
### Approach

[...]

PR #3657, opened 2026-07-02
Provide support for versioning to server configurations and optimize the CORs runtime.

### LINKED ISSUE #3643: Optimize CORs runtime.
### Current Limitation

Hot-reloadable server configurations such as CORS are re-resolved on every request so runtime
changes take effect without a restart. For CORS specifically, each cross-origin request currently:

- reads the merged `cors` server-config,
- decodes **both** configuration layers (declarative `readOnly` + editable `writable`),
- merges them, and
- computes a hash over the **full merged** origin list to detect whether the configuration
  changed since the previous request.

However, the declarative (`readOnly`) layer is fixed for the lifetime of the process — it is loaded
once at boot and can only change with a restart. Only the small, operator-editable (`writable`)
layer changes at runtime (via `PUT /server-config`). Decoding, merging, and hashing the static
layer on every request is therefore wasted work on a performance-critical path, and the cost scales
with the size of the full declarative allowlist rather than the (usually tiny) set of runtime-
editable origins.

### Suggested Improvement

Split the runtime resolution along the layer boundary the server-config already tracks
(`readOnly` / `writable`) and treat the two layers differently, entirely **in-process** (no schema
change):

- Expose the layers separately from the server-config service (decoded `readOnly` / `writable` /
  merged getters) so a consumer reads only the layer it needs.
- Compile the static `readOnly` layer **once** per process (lazily, on first request) and cache it
  for the process lifetime.
- On each request, check only the small `writable` layer for changes — a hash over just the
  editable entries — and recompile only that layer when its hash changes.
- Evaluate both layers through a **single combined matcher**, rebuilt only when the writable layer
  changes, so the hot path stays one match and runtime coherence / fail-closed behavior is
  unchanged.

This approach would:

- Eliminate the per-request decode + merge + full-list hash of the static declarative layer.
- Scale per-request overhead with the size of the small editable layer, not the full allowlist.
- Preserve today's runtime coherence (the editable layer is still read on every request) and
  fail-closed semantics.
- Require **no database schema change or migration**.

ImalshaD:
### Purpose

Server configurations had no version tracking, so hot-reloadable consumers like CORS had to perform a
read-and-hash of the effective config on **every request** to detect changes (#3643) — unnecessary
per-request database/decode/merge overhead.

This PR adds a `VERSION` to each `SERVER_CONFIG` row and switches the CORS runtime from the per-request
**signature hash** to a cheap **version compare**, removing the per-request decode + merge + hash from the
hot path. The version is a generic, reusable change-detection signal for any server-level configuration; it
is **internal only** — no change to the `/server-config` API wire shape or behaviour.

### Approach

- **Schema (`dbscripts/configdb/{postgres,sqlite}.sql`):** add `VERSION INTEGER NOT NULL DEFAULT 1` to
  `SERVER_CONFIG`. No migration is required — this schema has not shipped in a released version yet.
- **DB-side rotating increment (`SCF-02` upsert):** the version is incremented atomically in the
  `ON CONFLICT … DO UPDATE` clause as `VERSION = (VERSION % 2000000000) + 1`. It's a **rotating counter**
  bounded below the 32-bit `INTEGER` max so it can never overflow, and it is compared **only for equality**
  (never ordering), so wrapping is safe.
- **Store/service:** the version is threaded through the store layers (db → stored value, file → `0`,
  composite → the db layer's). A generic `GetConfigVersion(ctx, name)` is added to `ServerConfigService`
  that reads the version **without decoding or merging** (the cheap path); `GetMergedConfig` is unchanged.
- **CORS runtime (`internal/system/cors/dynamic.go`):** the dynamic matcher now caches `{version, *Matcher}`
  and gates recompilation on a version compare — reading the version first (safe under concurrent writes),
  and only reading/decoding/compiling the merged config on a version mismatch. `signature()` (and its
  `hash/fnv` usage) is deleted. The consumer interface was renamed `MergedConfigReader → ServerConfigReader`
  to reflect that it now exposes the version too.
- Mocks regenerated via `make mockery`; no hand-edits.

### Related Issues
- Fixes #3643

### Related PRs
- #3553 (dynamic CORS) — this implements the "version-based memo" deferred from that PR.

### Checklist
    - [x] Unit Tests 
    - [ ] Integration Tests 

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Server configuration now tracks a version number, enabling version-aware reads and updates.
  * CORS settings now refresh based on configuration version changes, improving consistency after updates.

* **Bug Fixes**
  * Configuration caching now preserves version data correctly across read paths.
  * Database updates now advance configuration versions automatically, helping avoid stale settings.


senthalan:
@ImalshaD Are we proceeding with this approach?

ImalshaD:
> @ImalshaD Are we proceeding with this approach?

Hi @senthalan,

No, as agreed offline this approach will not be continued and this PR was raised prior to having the offline disccusion, hence closing the PR.

PR #3727, opened 2026-07-03
Add RFC 7009 token revocation with deny-list enforcement

indeewari:
### Purpose
Implements RFC 7009 OAuth 2.0 Token Revocation and enforces it on the authorization server's hot paths.

ThunderID issues stateless JWTs, so a token stays cryptographically valid until it expires — revoking it has no effect unless the AS actively checks a deny list wherever the token is presented for use. This PR adds both halves:

1. A POST /oauth2/revoke endpoint (RFC 7009) so clients can revoke a token.
2. Enforcement of that revocation on the three paths where a token round-trips to the AS — introspection, the refresh-token grant, and token exchange — so a revoked token is rejected in real time rather than remaining usable until expiry.

### Approach
Single-token revocation is modeled as a JTI deny list in the operation database, with a clear read/write separation kept in one revocation package.

Write path (RFC 7009 endpoint)
- POST /oauth2/revoke behind the clientauth middleware; signature-only validation, so expired tokens remain revocable per §2.2 and an invalid/unknown/jti-less token is a no-op success.
- Ownership is enforced via the token's client_id claim — a mismatch returns invalid_grant.
- The deny-list write is idempotent (ON CONFLICT DO NOTHING), and the store constructor is unexported so the write path can't be reached from the hot paths, only through the service's validation + ownership checks.

Read/enforcement path
- A CheckerInterface.EnsureNotRevoked consults the deny list under a fail-closed policy: if the list can't be consulted, the token is rejected, never allowed.
- To avoid a self-inflicted DoS + alert storm when the operation DB is down, reads sit behind a consecutive-failure circuit breaker (5 failures / 30s) that short-circuits a failing DB and publishes OPERATION_DB_UNAVAILABLE once per trip.
- A shared enforceRevocation helper wires the checker into each hot path, mapping outcomes to each endpoint's native error vocabulary:

| Path | Revoked | Enforcement unavailable |
|---|---|---|
| Introspection | `active: false` | 5xx |
| Refresh grant  | invalid_grant | server_error |
| Token exchange (subject + actor) | invalid_request | server_error |

- Only self-issued tokens carry a jti into enforcement (an external issuer's jti has no meaning in this server's deny list); the validator extracts jti for refresh tokens and self-issued subject tokens only.
- Discovery advertises revocation_endpoint; a TOKEN_REVOKED audit event is emitted on revocation.

Testing — Unit tests cover the service, handler, writer, checker (fail-closed, circuit breaker, once-per-trip alerting), store, and each hot path. Integration tests cover revoke/idempotency/invalid-client/cross-client, and assert that introspection and token exchange reflect revocation.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3603

### Related PRs
- N/A

### Checklist
- [X] Followed the contribution guidelines.
- [X] Manual test round performed and verified.
- [X] Tests provided. (Add links if there are any)
    - [X] Unit Tests
    - [X] Integration Tests

### Security checks
- [X] Followed secure coding standards in [WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)
- [X] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.

## Summary by CodeRabbit

* **New Features**
  * Added OAuth 2.0 RFC 7009 token revocation endpoint and storage-backed deny-list enforcement.
  * Discovery metadata now advertises a populated revocation endpoint.
  * Revocation checks are enforced during refresh token grants, token exchange, and token introspection.
  * Refresh/subject tokens now expose JTI needed for revocation lookups.
* **Bug Fixes**
  * Updated behavior to fail closed when revocation enforcement can’t be verified.
  * Revoked tokens are now rejected consistently with appropriate OAuth error responses.
* **Tests**
  * Added integration and unit coverage for revocation endpoint and fail-closed enforcement (including circuit breaker behavior).


backend/internal/oauth/oauth2/granthandlers/refresh_token.go
```diff
@@ -370,6 +381,30 @@ func (h *refreshTokenGrantHandler) extendCacheTTL(
 	return nil
 }
 
+// checkRevocation enforces the deny list for the refresh token's JTI. A revoked token is rejected
+// with invalid_grant; an operation-DB outage fails closed with server_error (transient).
+func (h *refreshTokenGrantHandler) checkRevocation(
```
ThaminduDilshan: Is this function going to be repeated in each grant handler? If possible let's extract it to a util.

backend/internal/oauth/oauth2/introspect/service.go
```diff
@@ -71,8 +77,19 @@ func (s *tokenIntrospectionService) IntrospectToken(
...
+	// Enforce the RFC 7009 deny list. A revoked token is inactive; if the deny list cannot be
+	// consulted we fail closed (surface a server error) rather than asserting the token is active.
+	jti, _ := payload[constants.ClaimJTI].(string)
+	if revErr := s.revocationChecker.EnsureNotRevoked(ctx, jti); revErr != nil {
+		if errors.Is(revErr, revocation.ErrTokenRevoked) {
+			return &IntrospectResponse{
+				Active: false,
+			}, nil
+		}
+		logger.Error(ctx, "Token revocation status could not be verified; failing closed",
```
ThaminduDilshan, line 89: Hope we're not logging client errors here. And for server errors, we're returning a system wide server error

backend/internal/oauth/oauth2/revocation/checker.go
```diff
@@ -0,0 +1,112 @@
...
+// and alerts (via an observability event) when the breaker trips.
+type checker struct {
+	store            RevokedTokenStoreInterface
+	breaker          *circuitBreaker
+	observabilitySvc providers.ObservabilityProvider
+}
+
+// NewChecker creates a deny-list checker backed by the operation DB, guarded by a circuit breaker
+// and the fail-closed policy.
+func NewChecker(observabilitySvc providers.ObservabilityProvider) CheckerInterface {
```
ThaminduDilshan: Shall we make this constructor method private and plug it through the dependency injection pattern?

backend/internal/oauth/oauth2/revocation/checker.go
```diff
@@ -0,0 +1,112 @@
...
+		store:            newRevokedTokenStore(),
+		breaker:          newCircuitBreaker(enforcementFailureThreshold, enforcementOpenDuration),
+		observabilitySvc: observabilitySvc,
+	}
+}
+
+// EnsureNotRevoked checks the deny list for the given jti, applying the circuit breaker and the
+// fail-closed policy.
+func (c *checker) EnsureNotRevoked(ctx context.Context, jti string) error {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "RevocationChecker"))
```
ThaminduDilshan: Let's initialize logger inside struct. Then we can reuse it in methods. We don't have to keep initializing it in each method

backend/internal/oauth/oauth2/revocation/enforcement_service.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	revoked, err := c.store.IsTokenRevoked(ctx, jti)
+	if err != nil {
+		logger.Error(ctx, "Failed to consult token revocation deny list; failing closed",
+			log.Error(err))
+		if c.breaker.recordFailure() {
+			c.publishOperationDBUnavailableEvent(ctx, err)
+		}
+		return ErrEnforcementUnavailable
```
ThaminduDilshan, line 89: Shouldn't we return the system wide server error?

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
```diff
@@ -0,0 +1,107 @@
+/*
```
ThaminduDilshan, line 1: Shall we rename this file to `circuit_breaker.go`?

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
```diff
@@ -0,0 +1,107 @@
...
+
+const (
+	circuitClosed circuitState = iota
+	circuitOpen
+	circuitHalfOpen
+)
+
+// circuitBreaker is a minimal consecutive-failure circuit breaker guarding deny-list reads.
+// While open it short-circuits calls so a failing operation DB is not hammered on every request.
+type circuitBreaker struct {
```
ThaminduDilshan, line 46: Shall we define a interface for this and make constructor method return the interface?

backend/internal/oauth/oauth2/revocation/circuitbreaker_test.go
```diff
@@ -0,0 +1,85 @@
+/*
```
ThaminduDilshan: Rename file to have underscore

backend/internal/oauth/oauth2/revocation/circuitbreaker_test.go
```diff
@@ -0,0 +1,85 @@
...
+package revocation
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCircuitBreaker_StartsClosedAndAllows(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/oauth/oauth2/revocation/handler.go
```diff
@@ -0,0 +1,85 @@
...
+	tokenTypeHint := r.FormValue(constants.RequestParamTokenTypeHint)
+
+	clientID := ""
+	if client := clientauth.GetOAuthClient(ctx); client != nil {
+		clientID = client.ClientID
+	}
+
+	revokeOutcome, err := h.service.RevokeToken(ctx, token, tokenTypeHint, clientID)
+	if err != nil {
+		h.logger.Error(ctx, "Failed to revoke token", log.Error(err))
```
ThaminduDilshan, line 70: As per the current practice we log all of these errors at service layer. So this log will be double logged right?

@rajithacharith @UdeshAthukorala if we have already started to log it in handler, we can follow the same practice here removing logging at service. WDYT?

backend/internal/oauth/oauth2/revocation/model.go
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Package revocation provides the OAuth2 token revocation endpoint (RFC 7009) and the write path
```
ThaminduDilshan: Shall we define package comment only in a single file? Current practice is to define it in service or init file

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
indeewari: The error returns depends on the grant handler, however we can check the possibility of optimizing.

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
+ * Copyright (c) 2025-2026, WSO2 LLC. (https://www.wso2.com).
```
senthalan, line 2: Do we usually perform this change?

@ThaminduDilshan @rajithacharith

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -177,6 +177,7 @@ const (
 	ErrorInvalidScope             string = "invalid_scope"
 	ErrorInvalidTarget            string = "invalid_target"
 	ErrorServerError              string = "server_error"
+	ErrorUnsupportedTokenType     string = "unsupported_token_type" // #nosec G101
```
senthalan: Why we need to skip this ` #nosec G101`

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
```diff
@@ -0,0 +1,107 @@
...
+type circuitState int
+
+const (
+	circuitClosed circuitState = iota
+	circuitOpen
+	circuitHalfOpen
+)
+
+// circuitBreaker is a minimal consecutive-failure circuit breaker guarding deny-list reads.
+// While open it short-circuits calls so a failing operation DB is not hammered on every request.
```
senthalan, line 45: Why this circuitBreaker is implemented only for this service? 

 Shouldn't it be needed for other services as well ? Other services also have DB operations.

backend/internal/oauth/oauth2/revocation/checker.go
```diff
@@ -0,0 +1,112 @@
...
+var ErrTokenRevoked = errors.New("token has been revoked")
+
+// ErrEnforcementUnavailable indicates the deny list could not be consulted (operation DB
+// unavailable or the circuit is open). Under the fail-closed policy callers MUST reject the token.
+var ErrEnforcementUnavailable = errors.New("token revocation enforcement is unavailable")
+
+// CheckerInterface enforces the single-token deny list on the AS hot path (introspection, refresh
+// grant, token exchange) under a fail-closed policy: when the deny list cannot be consulted, tokens
+// are rejected rather than allowed.
+type CheckerInterface interface {
```
senthalan: Why we need a separate interface for this? what is the issue on adding this to the service interface ?

backend/internal/system/observability/event/category.go
```diff
@@ -72,9 +72,11 @@ func (e *UnmappedEventTypeError) Error() string {
...
 var eventTypeToCategory = map[providers.EventType]EventCategory{
 	// Authentication events
-	EventTypeTokenIssuanceStarted: CategoryAuthentication,
-	EventTypeTokenIssued:          CategoryAuthentication,
-	EventTypeTokenIssuanceFailed:  CategoryAuthentication,
+	EventTypeTokenIssuanceStarted:   CategoryAuthentication,
+	EventTypeTokenIssued:            CategoryAuthentication,
+	EventTypeTokenIssuanceFailed:    CategoryAuthentication,
+	EventTypeTokenRevoked:           CategoryAuthentication,
+	EventTypeOperationDBUnavailable: CategoryAuthentication,
```
senthalan, line 79: Why we have this? shouldn't the revoke operation fails when we have DB unavailable ?

backend/internal/oauth/oauth2/revocation/checker.go
senthalan: If this is for separate of concerns, it's okay

backend/internal/oauth/oauth2/constants/constants.go
indeewari: cc @Malith-19

backend/internal/oauth/oauth2/revocation/checker.go
indeewari: Yes — it's exactly separation of concerns / least privilege: CheckerInterface is the read-only enforcement path imported by the hot paths (introspection, grant handlers), while writes go through RevocationService. Keeping them as distinct interfaces means the hot-path consumers can't reach the deny-list writer.

backend/internal/system/observability/event/category.go
indeewari: This event comes from the enforcement read path (the checker), not the revoke write op. And the operation does fail — fail-closed rejects the token (introspection → inactive/5xx, grant → server_error). OPERATION_DB_UNAVAILABLE is not instead of failing; it's the throttled operator alert (emitted once per circuit trip, not per request) that tells why tokens are being fail-closed.

backend/internal/oauth/oauth2/constants/constants.go
indeewari: This follows the standard header practice of bumping the copyright year range whenever a file is modified.

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
indeewari: It's here specifically because enforcement is on the fail-closed hot path. When the operation DB is down, fail-closed means every introspection/refresh/exchange both rejects and, without a breaker, re-hammers the dead DB and floods the logs with one error per request.

Other DB operations aren't fail-closed on the auth hot path.

A cross-cutting DB circuit breaker in the provider layer is a reasonable future direction, but I'd keep it scoped to enforcement here rather than generalize preemptively.

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/checker.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/circuitbreaker_test.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/circuitbreaker_test.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/model.go
indeewari: Fixed

backend/internal/oauth/oauth2/constants/constants.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/handler.go
indeewari: Checked — not today. service.go only logs Debug-level no-op decisions (bad signature, undecodable, no-jti, wrong client); the propagated server error isn't logged there, it bubbles up and is logged once here at the handler. So there's no double log. Happy to standardize on handler-vs-service logging as a separate convention call.

backend/internal/oauth/oauth2/revocation/enforcement_service.go
indeewari: ErrEnforcementUnavailable - rather than a system-wide server error, so each hot path maps it to its own protocol vocabulary: introspection → HTTP 500, refresh → server_error, token exchange → server_error. If the checker returned a generic HTTP error, the OAuth grant handlers couldn't emit the correct {"error":"server_error"} body. The mapping-to-standard-error happens at each boundary; the shared checker only classifies.

backend/internal/oauth/oauth2/introspect/service.go
indeewari: We don't log the revoked outcome (it returns active:false as a normal result); only the fail-closed/unavailable case logs at Error. The handler maps to a standard server_error — confirmed

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
indeewari: Is it a convention ?

 It's an unexported, package-internal helper with a single implementation, used only by checker and exercised directly in tests — no second impl and nothing cross-package mocks it.

ThaminduDilshan: Not a convention, but a common pattern followed in most of the places. Unless we define a interface, mockery won't be able to generate mocks. If there's a requirement (current or future) to mock the behaviour of circuit breaker, we should define a interface and let mockery generate the mocks. We shouldn't write any custom mock implementations.

If this is not a requirement, you can ignore this comment

ThaminduDilshan: Answering "It's an unexported, package-internal helper" generally, if it's a package internal, you can define the interface private.

backend/internal/oauth/oauth2/revocation/checker.go
ThaminduDilshan: Maybe we can rename this to `RevocationEnforcementService`.

backend/internal/oauth/oauth2/revocation/checker.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/checker.go
indeewari: Renamed

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if cb.state == circuitOpen {
+		if time.Since(cb.openedAt) >= cb.openDuration {
+			cb.state = circuitHalfOpen
+			return true
+		}
+		return false
+	}
+	return true
+}
```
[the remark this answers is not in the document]
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
indeewari: it's tested directly through its real state transitions rather than mocked — the enforcement-service tests inject a real breaker. So there's no mocking requirement to satisfy here. Happy to introduce a (private) interface if a future consumer needs to mock it.

backend/internal/oauth/oauth2/granthandlers/grant_handler.go
```diff
@@ -37,6 +41,32 @@ type GrantHandlerInterface interface {
 		*model.TokenResponseDTO, *model.ErrorResponse)
 }
 
+// enforceRevocation runs the single-token deny-list check for jti and maps the outcome to an OAuth
+// error response shared by the grant handlers. It returns nil when the token may proceed, revokedErr
+// (the caller's grant-specific error) when the token is on the deny list, and a fail-closed
+// server_error when the deny list cannot be consulted.
+func enforceRevocation(
```
ThaminduDilshan, line 48: Shall we move this to a utils.go file?

backend/internal/oauth/oauth2/granthandlers/provider.go
```diff
@@ -56,16 +57,17 @@ func newGrantHandlerProvider(
 	resourceService providers.ResourceServerProvider,
 	cibaService ciba.CIBAServiceInterface,
 	cfg oauthconfig.Config,
+	enforcementService revocation.EnforcementServiceInterface,
```
ThaminduDilshan, line 60:
```suggestion
	revEnforcementSvc revocation.EnforcementServiceInterface,
```

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
```diff
@@ -41,12 +42,13 @@ import (
...
-	tokenValidator   tokenservice.TokenValidatorInterface
-	attrCacheService attributecache.AttributeCacheServiceInterface
-	resourceService  providers.ResourceServerProvider
+	cfg                oauthconfig.Config
+	jwtService         jwt.JWTServiceInterface
+	tokenBuilder       tokenservice.TokenBuilderInterface
+	tokenValidator     tokenservice.TokenValidatorInterface
+	attrCacheService   attributecache.AttributeCacheServiceInterface
+	resourceService    providers.ResourceServerProvider
+	enforcementService revocation.EnforcementServiceInterface
```
ThaminduDilshan, line 51:
```suggestion
	revEnforcementSvc revocation.EnforcementServiceInterface
```

ThaminduDilshan: Check other grant handlers too

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
```diff
@@ -118,6 +122,18 @@ func (h *refreshTokenGrantHandler) HandleGrant(ctx context.Context, tokenRequest
 		return nil, errResp
 	}
 
+	// Reject refresh tokens that have been revoked (RFC 7009 deny list). Fail closed when the
+	// operation DB cannot be consulted.
+	revokedErr := &model.ErrorResponse{
```
rajithacharith, line 127: Shall we move this error definition inside this if condition

backend/internal/oauth/oauth2/revocation/enforcement_service.go
```diff
@@ -0,0 +1,114 @@
...
+	"errors"
+
+	syscontext "github.com/thunder-id/thunderid/internal/system/context"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/system/observability/event"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// ErrTokenRevoked indicates the presented token's JTI is on the deny list.
+var ErrTokenRevoked = errors.New("token has been revoked")
```
ThaminduDilshan, line 32: Shall we define these error constants in a `error_constants.go` file?

ThaminduDilshan: Also if these are not returned outside revocation package, let's make them private

backend/internal/oauth/oauth2/granthandlers/token_exchange.go
```diff
@@ -151,6 +155,17 @@ func (h *tokenExchangeGrantHandler) HandleGrant(ctx context.Context, tokenReques
 		return nil, errResp
 	}
 
+	// Reject a revoked subject token (RFC 7009 deny list). JTI is set only for self-issued tokens.
+	subjectRevoked := &model.ErrorResponse{
+		Error:            constants.ErrorInvalidRequest,
+		ErrorDescription: "Invalid subject_token",
+	}
```
rajithacharith, line 162: Same comment as https://github.com/thunder-id/thunderid/pull/3727/changes#diff-19ce85dd7746918fa4de5ae1d38f3b703437b0310170c484c773917ab57a67adR127

backend/internal/oauth/oauth2/introspect/service.go
```diff
@@ -71,8 +77,19 @@ func (s *tokenIntrospectionService) IntrospectToken(
...
+	// Enforce the RFC 7009 deny list. A revoked token is inactive; if the deny list cannot be
+	// consulted we fail closed (surface a server error) rather than asserting the token is active.
+	jti, _ := payload[constants.ClaimJTI].(string)
+	if revErr := s.enforcementService.EnsureNotRevoked(ctx, jti); revErr != nil {
+		if errors.Is(revErr, revocation.ErrTokenRevoked) {
+			return &IntrospectResponse{
+				Active: false,
+			}, nil
+		}
+		logger.Error(ctx, "Token revocation status could not be verified; failing closed",
```
rajithacharith, line 89:
```suggestion
		logger.Error(ctx, "Token revocation status could not be verified;",
```
Do we need this failing closed part in the log?

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
+	syscontext "github.com/thunder-id/thunderid/internal/system/context"
+	"github.com/thunder-id/thunderid/internal/system/jose/jwt"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/system/observability/event"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// RevokeOutcome is the protocol-level result of a revocation request, mapped to HTTP by the handler.
+type RevokeOutcome int
```
ThaminduDilshan, line 35: Shall we move these to the model.go file?

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -0,0 +1,159 @@
...
+// RevokeToken validates and revokes the presented token. The token_type_hint parameter is ignored
+// (blank identifier) for the reasons documented on RevocationServiceInterface.RevokeToken.
+//
+// Per RFC 7009: signature is verified but expiry is intentionally not checked (expired tokens remain
+// revocable). An invalid, unparseable, or unknown token is a successful no-op. A token issued to a
+// different client is rejected with invalid_grant.
+func (s *revocationService) RevokeToken(
+	ctx context.Context, token, _, authenticatedClientID string,
+) (RevokeOutcome, error) {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "RevocationService"))
```
ThaminduDilshan, line 88: Let's initialize logger inside constructor method and use here rather than initializing in each method

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -0,0 +1,159 @@
...
+	jti, _ := payload[constants.ClaimJTI].(string)
+	if jti == "" {
+		logger.Debug(ctx, "Revocation request for a token without a jti claim; nothing to revoke")
+		return RevokeOutcomeRevoked, nil
+	}
+
+	// Ownership enforcement: a client may only revoke tokens issued to it. ThunderID tokens carry the
+	// owning client in the client_id claim (no azp), so ownership is checked against client_id; a
+	// mismatch is rejected with invalid_grant per RFC 7009 §2.1.
+	tokenClientID, _ := payload["client_id"].(string)
```
ThaminduDilshan, line 113: We should already have defined a constant for this. Let's reuse it

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	s.publishTokenRevokedEvent(ctx, authenticatedClientID, jti)
+	return RevokeOutcomeRevoked, nil
+}
+
+// extractExpiryTime returns the token's exp claim as a time, falling back to now when absent
+// (an absent/expired exp simply makes the deny-list row immediately cleanup-eligible).
+func extractExpiryTime(payload map[string]interface{}) time.Time {
+	if exp, ok := payload["exp"].(float64); ok {
```
ThaminduDilshan, line 136: Let's reuse a constant. Should have been defined already

backend/internal/oauth/oauth2/revocation/store.go
```diff
@@ -0,0 +1,95 @@
...
+
+	"github.com/thunder-id/thunderid/internal/system/config"
+	"github.com/thunder-id/thunderid/internal/system/database/provider"
+	"github.com/thunder-id/thunderid/internal/system/utils"
+)
+
+// RevokedTokenStoreInterface defines the deny-list persistence for single-token revocation: the
+// write path (InsertRevokedToken) used by the RFC 7009 revocation service and the read path
+// (IsTokenRevoked) used by the enforcement service on the AS hot path.
+type RevokedTokenStoreInterface interface {
```
ThaminduDilshan, line 34: We can make this private

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -332,6 +334,13 @@ func (tv *tokenValidator) extractSubjectTokenClaims(
 		return nil, err
 	}
 
+	// Only self-issued tokens participate in deny-list (revocation) enforcement; an external
+	// issuer's jti has no meaning in this server's deny list.
+	var jti string
+	if tv.isSelfIssuer(iss) {
+		jti, _ = extractStringClaim(claims, "jti")
```
ThaminduDilshan, line 341: Let's reuse a constant. Should have been defined already

backend/internal/oauth/init.go
```diff
@@ -82,12 +83,19 @@ func Initialize(
 	if err != nil {
 		return err
 	}
+	// Wire the revocation feature: registers the RFC 7009 endpoint (write path) and returns the
```
ThaminduDilshan, line 86: Do we need this comment?

backend/internal/oauth/init.go
```diff
@@ -82,12 +83,19 @@ func Initialize(
...
+	// shared deny-list enforcement service (read path) — a single enforcement instance with a circuit breaker
+	// and fail-closed policy, injected across the hot path: refresh grant, token exchange, and
+	// introspection.
+	_, enforcementService := revocation.Initialize(
```
ThaminduDilshan, line 90: If revocation service is not used by outside package, we don't need to return the instance

rajithacharith (approved):

ThaminduDilshan (approved):

rajithacharith:
We need to update the api/*.yaml

indeewari:
> We need to update the api/*.yaml

Fixed https://github.com/thunder-id/thunderid/pull/3765

backend/internal/oauth/oauth2/constants/constants.go
senthalan: Then we should follow this across other PRs as well.

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
senthalan: In the auth code flow, CIBA flow we are performing DB operations. So having this circuit breaker as a overall system improvement will be helpful for the product. 

For the time being shall we move this code based out from the revocation package and move to system package. It doesn't make any sense to have it here. Then we can have other DB operations to use it

backend/internal/system/observability/event/category.go
senthalan: What is the use case of these events?

backend/internal/oauth/oauth2/granthandlers/grant_handler.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/granthandlers/provider.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/enforcement_service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/granthandlers/token_exchange.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/introspect/service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/store.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/tokenservice/validator.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/init.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/init.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Fixed with https://github.com/thunder-id/thunderid/pull/3772

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -0,0 +1,159 @@
...
+	if jti == "" {
+		logger.Debug(ctx, "Revocation request for a token without a jti claim; nothing to revoke")
+		return RevokeOutcomeRevoked, nil
+	}
+
+	// Ownership enforcement: a client may only revoke tokens issued to it. ThunderID tokens carry the
+	// owning client in the client_id claim (no azp), so ownership is checked against client_id; a
+	// mismatch is rejected with invalid_grant per RFC 7009 §2.1.
+	tokenClientID, _ := payload["client_id"].(string)
+	if tokenClientID != "" && authenticatedClientID != "" && tokenClientID != authenticatedClientID {
```
thiva-k, line 114: Refresh tokens do not have client_id right? If not so this check will get always skipped.

backend/internal/oauth/oauth2/revocation/enforcement_service.go
```diff
@@ -0,0 +1,114 @@
...
+	breaker          *circuitBreaker
+	observabilitySvc providers.ObservabilityProvider
+	logger           *log.Logger
+}
+
+// newEnforcementService creates a deny-list enforcement service backed by the operation DB, guarded
+// by a circuit breaker and the fail-closed policy. It is unexported and constructed once via
+// Initialize so the shared enforcement instance — and its circuit breaker — cannot be duplicated by
+// external callers.
+func newEnforcementService(observabilitySvc providers.ObservabilityProvider) EnforcementServiceInterface {
```
thiva-k, line 61: Why do we need a separate service for this? Can't we move this to the revocation service? 

Currently the store is also being initialized  two times. Lets have two separate interfaces on the service if needed

backend/internal/oauth/oauth2/revocation/circuit_breaker.go
```diff
@@ -0,0 +1,113 @@
...
+	mu               sync.Mutex
+	state            circuitState
+	consecutiveFails int
+	failureThreshold int
+	openDuration     time.Duration
+	openedAt         time.Time
+}
+
+// newCircuitBreaker creates a circuit breaker in the closed state.
+func newCircuitBreaker(failureThreshold int, openDuration time.Duration) *circuitBreaker {
```
thiva-k, line 56: Hope we are planning to remove this from service layer

PR #3737, opened 2026-07-05
Split access token config by subject (userConfig/clientConfig) and embed client attributes as claims

Dilusha-Madushan:
### Purpose
Agents currently authenticate as themselves via the `client_credentials` grant (an M2M access token), but that token could only carry OU claims (`ouId`/`ouName`/`ouHandle`) — there was no way to embed any of the agent's own declared attributes (e.g. `modelProvider`, `model`) into it. Resource servers consuming the token had no way to learn anything about the calling agent beyond its OU, forcing a separate out-of-band lookup on every M2M call.

This PR lets an agent's own attributes be selectively embedded as claims in its `client_credentials` access token, via a new `clientAttributes` allow-list on `AccessTokenConfig`.

### Approach
- **Config**: added `AccessTokenConfig.ClientAttributes []string` (`clientAttributes` in JSON/YAML) — an allow-list of attribute names, mirroring the existing `UserAttributes` allow-list pattern already used for `authorization_code`/token-exchange.
- **Token building**: `BuildClientAttributes()` now also resolves the agent's own attributes (in addition to the existing OU-claim resolution), sourced via the grant handler's existing `actorProvider` dependency — no new service dependency was needed. This is gated on `EntityCategory == agent`, so it's a no-op for Applications.
- **Validation**: a new `validateClientAttributes()` in the agent service checks configured attribute names against the agent's own type schema (`entitytype.TypeCategoryAgent`), reusing the same schema infrastructure already used to validate the agent's attribute values. Wired into `ValidateAgent()`, so it runs on both create and update.
- **Reserved-claim safety**: since configured attribute names are owner-controlled, a name colliding with a reserved claim (`scope`, `aud`, `client_id`, `sub`, `iss`, `act`, `ouId`, etc.) could otherwise silently overwrite that claim. Added a `ReservedAccessTokenClaimNames()` guard, enforced both at config-validation time (rejects the config outright) and defensively at claim-merge time (drops the key rather than overwrite).
- **Scope**: deliberately agent-only for now, not "any OAuth client." The gate lives in two places: `resolveClientOwnAttributes()` checks `EntityCategory == agent` before doing anything, and `validateClientAttributes()` is only invoked from the agent service's `ValidateAgent()` — Applications never call it. This is because only agents currently have a declared, schema-validated attribute store (`entitytype.TypeCategoryAgent`); Applications have no attribute schema to source values from or validate against. The config field itself lives on the shared `AccessTokenConfig` struct so it can be extended to Applications later without another schema change, if an Application-side attribute schema is ever introduced.
- Backend-only change — no console/UI or OpenAPI spec updates included in this PR (tracked separately).

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3505

## Summary by CodeRabbit

* **New Features**
  * Access-token configuration is now split by token subject (end-user vs client-credentials) via `userConfig` and `clientConfig`.
  * Client-credentials access tokens can embed allowed client/agent-owned attributes using configured allowlists.

* **Bug Fixes**
  * Reserved claim names are protected from being overwritten.
  * Access-token validity and attribute allowlisting/embedding are applied consistently across authorization, refresh, CIBA, and token exchange.

* **Documentation**
  * Updated API schemas, console UI/config examples, samples, and request/response fixtures to use the new `userConfig`/`clientConfig` structure.


api/application.yaml
```diff
@@ -1188,17 +1190,33 @@ components:
     AccessTokenConfig:
       type: object
       description: |
-        Access token configuration for OAuth applications.
+        Access token configuration for OAuth applications, split by token subject.
+      properties:
```
ThaminduDilshan, line 1194: What if we go ahead with SubjectConfig and ActorConfig instead of this? This applies to both application and agent

api/application.yaml
```diff
@@ -1188,17 +1190,33 @@ components:
...
+        userConfigs:
```
ThaminduDilshan: Let's make sure these changes appear properly in swagger editor/ docs. Some of these endpoints uses individual examples and hence require adding these to multiple places

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -418,12 +422,46 @@ func buildClaimsFromRequest(
 	return result
 }
 
+// ReservedAccessTokenClaimNames returns the claim names set by the token builder itself.
+// Configured attribute allow-lists (userAttributes, clientAttributes) must not be allowed to use these.
+func ReservedAccessTokenClaimNames() map[string]bool {
```
ThaminduDilshan, line 427: @coderabbitai hope this list is not duplicated elsewhere, even as a array

ThaminduDilshan:
@Dilusha-Madushan this is not only a fix for agents right? Shall we update PR title accordingly?

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -582,10 +581,27 @@ type OAuthTokenConfig struct {
 	RefreshToken *RefreshTokenConfig `json:"refreshToken,omitempty" yaml:"refreshToken,omitempty" jsonschema:"Refresh token configuration."`
 }
 
-// AccessTokenConfig is the access token configuration.
+// AccessTokenConfig is the access token configuration, split by token subject: an end user
+// (UserConfig) or the OAuth client itself, issued only via the client_credentials grant
+// (ClientConfig).
 type AccessTokenConfig struct {
```
ThaminduDilshan, line 587: Hope this is addressed for declarative resources too. @coderabbitai check

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -581,10 +581,19 @@ type OAuthTokenConfig struct {
...
+	ClientConfigs *AccessTokenSubConfig `json:"clientConfigs,omitempty" yaml:"clientConfigs,omitempty" jsonschema:"Access token configuration applied when the token subject is the OAuth client itself, issued only via the client_credentials grant."`
+}
+
+// AccessTokenSubConfig holds the validity period and attribute selection for one access
+// token subject type (user or client).
+type AccessTokenSubConfig struct {
 	ValidityPeriod int64    `json:"validityPeriod,omitempty" yaml:"validityPeriod,omitempty" jsonschema:"Access token validity period in seconds."`
-	UserAttributes []string `json:"userAttributes,omitempty" yaml:"userAttributes,omitempty" jsonschema:"User attributes to embed in the access token."`
+	Attributes     []string `json:"attributes,omitempty"     yaml:"attributes,omitempty"     jsonschema:"Attributes to embed in the access token."`
 }
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai
Thanks — the concern is valid: this is an intentional breaking change to the persisted AccessTokenConfig shape, and we're accepting it rather than adding a legacy decoder or a data migration.

Rationale:

Access-token config is authored as declarative resources (application/agent YAML) that are re-imported, and via the Console UI which writes the new shape — so the fix for any existing definition is a re-apply, not a silent data loss in normal operation.
We're pre-release here and explicitly opted for a clean break over a back-compat shim, per the team decision on this feature. No production records need preserving.
A dual-read/legacy decoder would carry the old flat validityPeriod/userAttributes shape forward indefinitely, which is exactly the ambiguity the subject-split is removing.
Also note this comment is on an earlier revision — the fields have since been renamed to singular userConfig/clientConfig and AccessTokenSubConfig.UserAttributes was removed (now just validityPeriod + attributes). Could you re-check against the latest commit?

backend/pkg/thunderidengine/providers/model.go
ThaminduDilshan: retry

ThaminduDilshan: @Dilusha-Madushan can you make sure changes doesn't break wayfinder try out?

Dilusha-Madushan: @coderabbitai

declarative resources are covered. The declarative parser reuses the same providers.AccessTokenConfig type (there's only one definition), so token.accessToken.userConfig/clientConfig in imported YAML deserializes through the identical path as the REST API — there's no separate declarative DTO to update. The shipped 01-default-resources.yaml bootstrap and all sample resource YAMLs already use the new shape, and no declarative docs referenced the old one.

ThaminduDilshan (approved):

api/application.yaml
Dilusha-Madushan: "Actor" already has a specific meaning in our codebase. We already use buildActorClaim() for the RFC 8693 act claim, which represents an entity acting on behalf of the subject (delegation/token exchange). That's a different concept from ClientConfigs, which is used when the OAuth client itself is the subject (for example, in the client_credentials flow). Calling it ActorConfigs could be confusing because it overlaps with the existing actor/delegation terminology.
"Client" matches the OAuth terminology. The distinction here is essentially User vs. Client—the subject is either an end user (delegated flows) or the OAuth client itself (client_credentials). "Client" is the standard OAuth term and pairs naturally with "User", whereas "Actor" refers to a different concept.
Also, even if the caller is an application, service, or agent, it's still an OAuth client, so the name is broad enough to cover all of those cases.

Dilusha-Madushan: @coderabbitai explain about your idea about these naming sugestions

Issue #3601, opened 2026-06-30
[SCIM] implement SCIM v2.0 Users endpoint — Phase 1 custom schema provisioning


ravindu439:
### What problem are we solving?

ThunderID has its own internal users system with existing user management endpoints, but it does not yet expose a SCIM 2.0 compliant `/scim/v2/Users` surface. Enterprise SCIM clients require standard SCIM provisioning endpoints `(GET, POST, PUT, DELETE /scim/v2/Users)` to manage user lifecycle externally. Without these, the SCIM discovery handshake (completed on `SCIM-serviceproviderconfig`) succeeds but every downstream provisioning operation fails — SCIM clients can discover what ThunderID supports, but cannot act on it.

Phase 1 uses ThunderID's own custom entity type schemas — surfaced through the Schemas discovery endpoint — as the provisioning payload format. The client first calls `GET /scim/v2/Schemas` to discover the correct ThunderID extension URN for the target user type, then sends that URN in the provisioning request.

### Who are we solving this for?

Primary: Enterprise Identity Administrators / SCIM Client Integrators

- Goal: Successfully provision, read, update, and deprovision ThunderID users from an external IdP or HR system (e.g. BambooHR, custom SCIM client) via SCIM v2.0 — building on the discovery layer already in place.
- Pain: SCIM clients can complete the discovery handshake but have no provisioning surface to call. RFC 7644 §3 mandates these operations for any compliant SCIM provider.

### Why should we solve this now?

Why now:

- Value to Persona: The Users endpoints are the core deliverable every SCIM integrator expects after discovery. Discovery without provisioning is a dead-end in any SCIM integration flow.
- Value to Org: Delivers Phase 1 user lifecycle management (create, read, update, delete) — the primary use case driving the entire SCIM implementation. Also introduces the reusable internal SCIM validator (`validator.go`) that is prerequisite for all future SCIM resources (Groups, Bulk).

### Proposed Solution

Implement five SCIM 2.0 Users endpoints, scoped to `backend/internal/scim`. The SCIM layer acts as a **pure translation proxy** — it calls existing `user.UserService` and `entitytype.EntityTypeService` functions and adds no new DB code, authz logic, or credential handling.

#### Endpoints to Implement

| Endpoint | Method | Description |
|---|---|---|
| `/scim/v2/Users` | `GET` | List users (OU-scoped, paginated); reject `?filter=` with SCIM-1014 |
[...]

PR #3743, opened 2026-07-06
feat(scim): implement Phase 1 SCIM Users CRUD endpoints

ravindu439:
### Purpose
This PR implements **Phase 1 of the SCIM 2.0 Users endpoint** (`/scim/v2/Users`) for the ThunderID SCIM package (`backend/internal/scim`), as specified in the ThunderID SCIM Master Proposal.

It introduces full CRUD provisioning for users via a custom Thunder schema payload (Phase 1). External SCIM clients (e.g. Okta, BambooHR) discover the custom URN via the `/scim/v2/Schemas` endpoint, then send provisioning requests that include the Thunder extension URN (e.g. `urn:thunderid:params:scim:schemas:employee:2.0:User`). The SCIM layer acts as a pure translation proxy — it performs RFC 7643/7644 protocol validation and delegates all business logic to the existing `user.UserService` and `entitytype.EntityTypeService`. No new database or authorization logic is added.

---

### Approach

**Key design decisions:**

1. **Internal SCIM Validator (`scim_validator.go`):** A reusable, zero-Thunder-import validator that checks all RFC 7643/7644 protocol rules (JSON validity, `schemas` key, no duplicate URNs, exactly one ThunderID extension URN, URN well-formedness, extension object presence). Returns a structured `SCIMUserPayload` on success with the extension URN, resolved user type name, core attributes, and extension attributes.

2. **Users Handler (`users_handler.go`):** Five HTTP handlers wired in `init.go`:
   - `GET /scim/v2/Users` — OU-scoped, paginated list (rejects `?filter` with SCIM-1014)
   - `POST /scim/v2/Users` — Validate → resolve entity type → create user
   - `GET /scim/v2/Users/{id}` — Fetch by ID; returns 404 (SCIM-1010) if not found
   - `PUT /scim/v2/Users/{id}` — Full replace; blocked for declarative (file-loaded) users
   - `DELETE /scim/v2/Users/{id}` — Delete; blocked for declarative users

3. **Users Service (`users_service.go`):** Orchestration layer implementing `SCIMUserServiceInterface`. Calls `entitytype.GetEntityTypeByName` to resolve OU from the extension URN, then delegates to `user.UserService`. Uses `u.OUID = entityType.OUID` directly (not `ResolveUserOUHandle`, which is reserved for declarative file loading).

4. **Users Resource Builder (`users_resource.go`):** Builds the SCIM response (`SCIMUser`, `SCIMUserListResponse`), strips credential-only attributes fetched via `entitytype.GetAttributes(credentialOnly: true)`.

5. **Users Model (`users_model.go`):** `SCIMUser` with a custom `MarshalJSON` to produce the correct extension namespace structure in the response.

6. **Error constant added:** `ErrorFilterNotSupported` (SCIM-1014) added to `error_constants.go`.

**Authorization flow:** The auth middleware injects caller identity into `ctx`. The SCIM service passes `ctx` unchanged to `user.UserService`, which calls `sysauthz.IsActionAllowed`. No authz logic lives in the SCIM package.

**Declarative guard:** Both PUT and DELETE check `entity.IsEntityDeclarative()` and return `400 invalidValue` if true, preventing SCIM from overwriting file-loaded users.

---

### Related Issues
- #3601 

### Related PRs
- #3298 

---

### Checklist
- [x] Tests provided.
    - [x] Unit Tests — `service_test.go`, `handler_test.go`, `SCIMServiceInterface_mock_test.go`, `init_test.go`
    - [ ] Integration Tests

---

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added SCIM user support for creating, viewing, updating, deleting, and listing users.
  * Added support for SCIM-compliant request/response formatting, including user resource metadata and paginated list responses.
  * Added SCIM-specific error responses and validation for content type, schemas, filters, and user attributes.

* **Bug Fixes**
  * Improved handling of authentication and authorization failures on SCIM routes.
  * Ensured SCIM endpoints return the correct status codes and error details for common failure cases.


backend/internal/scim/scim_validator.go
```diff
@@ -0,0 +1,106 @@
...
+
+	// Step 3: Check for duplicate URNs in the "schemas" array.
+	seen := make(map[string]struct{}, len(schemas))
+	for _, urn := range schemas {
+		lower := strings.ToLower(strings.TrimSpace(urn))
+		if _, exists := seen[lower]; exists {
+			return nil, &ErrorDuplicateSchemas
+		}
+		seen[lower] = struct{}{}
+	}
```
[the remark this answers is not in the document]
ravindu439: since we only use core users id attributes and it does not sent through the payload , we don't need to pass the core schemas in the schemas array.

backend/tests/mocks/scim/SCIMUsersServiceInterface_mock.go
```diff
@@ -0,0 +1,413 @@
+// Code generated by mockery; DO NOT EDIT.
```
sadilchamishka, line 1: This generated mock looks dead and inconsistent with the project convention:

- **No consumers** — nothing imports `tests/mocks/scim`.
- The `_test.go` suffix means it **can't be imported** by any other package, which defeats the purpose of a `tests/mocks/*` shared mock. Every other mock there is `{Interface}_mock.go` in a dedicated `*mock` package (`usermock`, `entitytypemock`).
- It's not referenced in `.mockery.private.yml`/`.public.yml`, so `make verify_mocks` won't manage it.

The mock the tests actually use is the in-package `backend/internal/scim/SCIMUsersServiceInterface_mock_test.go`. Suggest **deleting this copy**; or if a shared mock is intended, rename to `_mock.go`, move to a `scimmock` package, and add a mockery config entry.

backend/internal/scim/users_service.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	attrsJSON, err := json.Marshal(payload.ExtensionAttrs)
+	if err != nil {
+		logger.Error(ctx, "SCIM ReplaceUser: failed to marshal extension attrs", log.Error(err))
+		return nil, &ErrorInvalidRequestBody
+	}
+	updatedUser := &user.User{
+		ID:         userID,
+		OUID:       et.OUID,
+		Type:       canonicalName,
```
sadilchamishka, line 180: `ReplaceUser` resolves `Type`/`OUID` solely from the extension URN in the request body and passes them to `UpdateUser`, without checking they match the user identified by `{id}`. So `PUT /Users/{employee-id}` carrying a `...:contractor:...` URN will attempt to reclassify the user's entity type and move its OU.

It's gated by `UpdateUser`'s destination-OU authz + `validateOrganizationUnitForUserType`, so not a security hole — but SCIM PUT treats the resource type as immutable, and silent reclassification is surprising. Consider verifying the resolved type matches the stored user (else reject), or ignoring type/OU on replace.

backend/internal/scim/scim_validator.go
```diff
@@ -0,0 +1,106 @@
...
+	// Collect core attributes (those that are not "schemas" or the extension URN)
+	coreAttrs := make(map[string]json.RawMessage)
+	for k, v := range raw {
+		if strings.EqualFold(k, "schemas") {
+			continue
+		}
+		if strings.EqualFold(k, extensionURN) {
+			continue
+		}
+		coreAttrs[k] = v
```
sadilchamishka, line 98: `CoreAttrs` (top-level fields like `userName`, `externalId`, `active`) is collected here but never persisted or echoed back — Create/Replace only marshal `ExtensionAttrs`. For the custom-schema Phase 1 design that may be intentional, but **`externalId`** is what IdPs (Okta/BambooHR) use to correlate resources, so silently dropping it tends to bite integrators. Worth confirming this is intended; if so, rejecting unexpected core fields would be friendlier than dropping them quietly.

backend/internal/system/security/middleware.go
```diff
@@ -60,3 +74,13 @@ func writeSecurityError(ctx context.Context, w http.ResponseWriter, err error) {
 
 	utils.WriteErrorResponse(ctx, w, http.StatusUnauthorized, apierror.ErrUnauthorized)
 }
+
+func writeSCIMSecurityError(w http.ResponseWriter, status int, detail string) {
+	w.Header().Set("Content-Type", "application/scim+json")
```
sadilchamishka: Minor: this hardcodes `"application/scim+json"` and the error-schema URN literal. The content type is available as `constants.SCIMContentType` (importable here) — worth reusing for consistency. (The `scim` package's `SCIMErrorResponse` struct can't be imported here due to the package cycle, so the inline map is fine.)

backend/internal/scim/users_handler.go
```diff
@@ -0,0 +1,187 @@
...
+	"io"
+	"net/http"
+	"strconv"
+
+	"github.com/thunder-id/thunderid/internal/system/log"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+// scimUsersHandler handles all /scim/v2/Users HTTP requests.
+const loggerComponentName = "scim"
```
sadilchamishka, line 13: Nit: this `const` sits between the `scimUsersHandler` doc comment (line above) and the struct, detaching the doc comment from its type. Move the const above the comment (or below the struct).

backend/internal/scim/users_handler.go
```diff
@@ -0,0 +1,187 @@
...
+	if svcErr != nil {
+		h.handleSCIMError(w, r, svcErr)
+		return
+	}
+
+	writeSCIMSuccessResponse(ctx, w, http.StatusOK, replaced)
+	logger.Debug(ctx, "SCIM User replaced", log.String("userID", userID))
+}
+
+// handleDeleteUserRequest handles DELETE /scim/v2/Users/{id}
```
sadilchamishka: Nit: doc comment says `handleDeleteUserRequest` but the function is `HandleUsersDeleteRequest`.

sadilchamishka (changes requested):

backend/internal/scim/users_handler.go

backend/internal/system/security/middleware.go
ravindu439: fixed

backend/tests/mocks/scim/SCIMUsersServiceInterface_mock.go
ravindu439: fixed

backend/internal/scim/users_service.go
ravindu439: Fixed. ReplaceUser now fetches the existing user right after resolving the schema-derived type, and rejects with ErrorImmutableUserType (SCIM-1020) if it doesn't match — before any entity type lookup or update is attempted.

backend/internal/scim/scim_validator.go
ravindu439: Only id (from the URL path, not the body) is used from the core SCIM schema — userName, externalId, active, etc. aren't mapped to stored fields yet; identity/type come entirely from the ThunderID extension schema. CoreAttrs is parsed but deliberately not persisted at this stage.

Agreed externalId is the one that'll bite integrators (Okta/BambooHR correlation) — tracking as follow-up for a later phase rather than blocking this PR. Will consider your suggestion then: either map externalId to a stored field, or explicitly reject unexpected core attributes instead of silently dropping them.

sadilchamishka (approved):

PR #3751, opened 2026-07-06
Block connection deletion when in use

### LINKED ISSUE #3750: Block deletion of connections (identity providers) and notification senders that are in use
### Describe the issue

Identity providers (connections) and notification senders can currently be deleted even when other resources reference them. Deleting a resource that is still in use leaves those references dangling — for example, an active flow that references an identity provider or a notification sender will break when the referenced resource is removed.

### Expected behavior

Deletion of an identity provider or a notification sender should be blocked when a resource with a `restrict` delete behavior depends on it (e.g. an active flow that references it). The API should return a client error explaining which resources depend on it, so the user can remove or reassign them first.

Because deletion is destructive, the check must fail closed: if dependency information cannot be determined, the deletion is refused rather than allowed.

### Suggested approach

Extend the existing resource dependency registry (already used to block theme, flow, and user deletion) to cover:
- `idp` and `notificationSender` resource types.
- A flow dependency provider that reports which active flows reference an identity provider or notification sender via their node properties.
- Wire the dependency registry into the IDP and notification sender management services and enforce the check on delete.

### Version

main

rajithacharith:
### Purpose

Blocks deletion of identity providers (connections) and notification senders while other resources still depend on them. Previously either resource could be deleted even when an active flow referenced it, leaving dangling references that break the flow.

### Approach

Extends the existing resource dependency registry (already used to block theme, flow, and user deletion):

- Adds `idp` and `notificationSender` resource types to the registry, plus `SummarizeBlockingUsages` to render a deterministic, human-readable summary of blocking dependencies grouped by resource type.
- The flow management service now implements the dependency `Provider` interface (`GetResourceDependencies`): an active flow references an identity provider or notification sender when one of its nodes carries the matching ID in its node properties. A new store query `ListActiveFlowsWithNodes` loads all active flows with node definitions in a single query for this lookup.
- The IDP service and notification sender service each gain a `SetDependencyRegistry` injector and an `ensureNoBlockingDependencies` check on delete. Both fail closed: if the registry is unset or dependency data cannot be determined, deletion is refused with an internal server error rather than allowed.
- `servicemanager` wiring is refactored to a `dependencyConsumers` struct and now injects the registry into the IDP and notification sender services; `flowMgtService` is registered as both a consumer and a provider (two-phase init to avoid cyclic imports).
- Adds client error constants (`IDP-1013`, notification sender equivalent) and i18n default messages.

### Related Issues
- Fixes thunder-id/thunderid#3750

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Deleting identity providers and notification senders is now dependency-aware: removal is blocked when other resources still reference them.
  * Dependency evaluation now considers active flows along with their node details for more accurate impact detection.
* **Bug Fixes**
  * Dependency-blocking deletion failures now consistently return **409 Conflict** instead of an unexpected error status.
  * Deletion now “fails closed” when dependency information can’t be determined, preventing unsafe removals.
* **Tests**
  * Added/expanded unit tests for dependency-aware dependency checks and deletion refusal scenarios.


backend/internal/idp/handler.go
```diff
@@ -251,7 +251,8 @@ func getClientErrorStatusCode(errorCode string) int {
 	switch errorCode {
 	case ErrorIDPNotFound.Code:
 		return http.StatusNotFound
-	case ErrorIDPAlreadyExists.Code:
+	case ErrorIDPAlreadyExists.Code,
+		ErrorIDPHasBlockingDependencies.Code:
```
UdeshAthukorala, line 255: Is it ok to use conflict error code for `ErrorIDPHasBlockingDependencies` scenario as well?
If we are using `conflict` error code for this scenario it needs to be update in `internal/connection/mapping.go — writeServiceError (~lines 119–129)` as well right?

rajithacharith: Fixed this

UdeshAthukorala (approved):

PR #3765, opened 2026-07-06
Document the OAuth 2.0 token revocation endpoint in the API spec

indeewari:
### Purpose
The RFC 7009 token revocation endpoint (`POST /oauth2/revoke`) shipped in #3727 but was never documented in the OpenAPI spec (`api/oauth2.yaml`), which only described `/oauth2/introspect`. This adds the missing endpoint definition so the spec matches the implementation.

### Approach
Mirrored the existing conventions in `api/oauth2.yaml` (modeled on `/oauth2/introspect`):
- Added a `Revocation` tag (`Token revocation per RFC 7009`).
- Added the `POST /oauth2/revoke` path with responses that match the handler exactly:
  - `200` (empty body) — success, or a no-op success for a token that cannot be validated (bad signature / malformed / unknown). Expired-but-validly-signed tokens are genuinely revocable, per the signature-only verification in the service.
  - `400` — missing `token` (`invalid_request`) or token issued to a different client (`invalid_grant`).
  - `401` — client authentication failed (`invalid_client`).
  - `500` — revocation could not be recorded (e.g. operation DB unavailable).
- Added a `RevokeRequest` schema documenting `token` and `token_type_hint` (accepted per RFC 7009 §2.1 but not acted on, since ThunderID tokens are self-describing).

No behavior change — documentation only.

### Related Issues
- N/A

### Related PRs
- #3727 (introduced the revocation endpoint)

### Checklist
- [x] Manual test round performed and verified. (YAML validated; responses cross-checked against the handler.)
- [x] Documentation provided. (This PR is the API spec documentation.)
- [ ] Tests provided. (N/A — OpenAPI spec change only.)
- [ ] Breaking changes. (None.)

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added support for token revocation in the OAuth2 API.
  * Introduced a new revocation endpoint for access and refresh tokens.
  * Added request fields for the token and an optional token type hint.

* **Bug Fixes**
  * Invalid or unverifiable tokens are now handled as successful no-ops, matching expected OAuth2 behavior.
  * Documented clear response codes for missing, invalid, unauthorized, and server error cases.

### DIFF
--- a/api/oauth2.yaml
+++ b/api/oauth2.yaml
@@ -23,6 +23,8 @@ tags:
     description: OAuth 2.0 token endpoint for issuing access tokens, refresh tokens, and ID tokens.
   - name: Introspection
     description: Token introspection per RFC 7662.
+  - name: Revocation
+    description: Token revocation per RFC 7009.
   - name: PAR
     description: Pushed Authorization Requests per RFC 9126.
   - name: JWKS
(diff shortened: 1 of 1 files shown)


api/oauth2.yaml
```diff
@@ -657,6 +704,21 @@ components:
...
+      properties:
+        token:
+          type: string
+          description: The access or refresh token to revoke.
+        token_type_hint:
+          type: string
+          description: >-
+            Hint about the type of the submitted token (`access_token` or `refresh_token`). Accepted
+            per RFC 7009 §2.1 but optional; ThunderID tokens are self-describing, so the hint does not
+            affect processing.
```
rajithacharith, line 720: Don't we need client_id?

rajithacharith (approved):

PR #3779, opened 2026-07-06
Add flow-centric browser SSO

madurangasiriwardena:
### Purpose
Introduces **flow-centric browser SSO** — the ability to reuse an existing browser login session across applications that share the same authentication flow, so a returning user can skip re-authenticating.

SSO is expressed as a property of the **authentication flow graph** rather than a separate subsystem: the flow author places SSO nodes to decide what an existing session lets a user skip. A session belongs to exactly one flow (keyed by `flow_id`) and is referenced by an opaque per-flow handle cookie; two apps SSO with each other **iff** they are configured with the same flow.

This is a first-phase (POC-level) implementation. A few data-handling decisions are deliberately interim, pending a follow-up flow-context data-classification effort (called out under Approach → Deferred).

Design discussion (architecture, storage model, DB operations): https://github.com/thunder-id/thunderid/discussions/3673

### Approach
**Flow graph — two node types, used as one or more checkpoint pairs**
- **SSO-Check** (`SSOCheckExecutor`, utility node) — resolves whether a live, compatible session already holds a given checkpoint and routes **skip** vs. **authenticate** (via `onSuccess`/`onFailure`).
- **Session** (`SessionExecutor`, authentication node) — the join where the SSO and fresh-auth branches converge. On the fresh path it saves the checkpoint's context and mints a handle; on the SSO path it loads the saved context so downstream nodes continue authenticated.
- A flow may contain **multiple SSO-Check/Session checkpoint pairs** (e.g. password as one checkpoint, step-up/MFA as another). Each SSO-Check binds to its Session node via a `checkpointRef` node property; the checkpoint id is that join node's id. All checkpoints of one login share a **single session per flow execution**, established idempotently on `flow_execution_id`, so each stage can be skipped independently on reuse. Create-vs-append is decided from the database (by handle or flow execution id), not in-memory ordering, so it holds under multi-request / divergent-branch flows.

image

**Storage (runtime DB)**
- `SSO_SESSION` — lean row read on every resolve/touch (one row per establishing flow execution; unique on `flow_execution_id`).
- `SSO_SESSION_CONTEXT` — one encrypted row **per checkpoint** (`(session_id, checkpoint_id)`), loaded only on the SSO restore path.
- `SSO_SESSION_PARTICIPANT` — apps that used the session (basis for future session-wide logout/revocation).

**Session lifecycle** — governed by configurable idle (sliding) and absolute deadlines; the resolver rejects a session past either.

**Engine-agnostic integration** — the reusable flow engine (`pkg/thunderidengine`) stays SSO-unaware. SSO inputs ride on the request `context.Context`; the minted handle is returned on the engine's generic `AdditionalData` channel and the product transport layer turns it into the per-flow cookie. No SSO-specific field is added to the engine contract.

**Deferred to a follow-up (flow-context data classification)**
- The resolved `AuthUser` is currently snapshotted **as-is** (materialized attributes), and `RuntimeData` is persisted in full apart from a deny-list of transient SSO control keys and request-scoped keys (`requested_permissions`, `required_*_attributes`, `required_locales`, `clientId`, `authorizationRequestId`, `applicationId`). A later phase will introduce a proper durability classification and attribute minimization/re-resolution. Tracked in `SSO_TODO.md`.
- Session end/logout fan-out and an SSO cleanup job are defined as seams but not wired yet.

### Related Issues
- https://github.com/thunder-id/thunderid/discussions/3673
- https://github.com/thunder-id/thunderid/issues/3672

### Related PRs
- https://github.com/thunder-id/javascript-sdks/pull/4

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added SSO session support, allowing authentication flows to reuse active sessions across steps and flows.
  * Added session cookies with configurable idle and absolute expiration times.
  * Added checkpoint-based session save and restore behavior.
  * Added support for OAuth `max_age` and authentication assurance checks, including step-up authentication when requirements are unmet.
  * Added a “Basic with SSO” authentication flow template.

* **UI Improvements**
  * Flow execution nodes now support custom descriptions and outcome labels in the console.


madurangasiriwardena:
@coderabbitai resume

backend/internal/flow/graphbuilder/testdata/sso_flow.json
```diff
@@ -0,0 +1,127 @@
+{
```
ThaminduDilshan, line 1: Shall we add this as a flow builder template rather than introducing a new bootstrap flow?

backend/dbscripts/runtimedb/postgres.sql
```diff
@@ -67,6 +67,61 @@ CREATE INDEX idx_flow_context_deployment_id ON "FLOW_CONTEXT" (DEPLOYMENT_ID);
 -- Index for expiry time on FLOW_CONTEXT
 CREATE INDEX idx_flow_context_expiry_time ON "FLOW_CONTEXT" (EXPIRY_TIME);
 
+-- Table to store SSO sessions, grouped by flow (FLOW_ID) and resolved by an opaque handle.
+CREATE TABLE "SSO_SESSION" (
```
ThaminduDilshan: This is persistent runtime data right? Shouldn't this go into operations DB?

ThaminduDilshan: If going with runtime DB, there's some other refactoring happening to move this into a key value pair based table. Other existing tables are refactored progressively, but since this is a new table, we can directly follow this pattern adding data to `RUNTIME_STORE`.

https://github.com/thunder-id/thunderid/pull/3818

backend/internal/flow/graphbuilder/testdata/sso_flow.json
```diff
@@ -0,0 +1,127 @@
...
+            "onSuccess": "sso_check"
+        },
+        {
+            "id": "sso_check",
+            "type": "TASK_EXECUTION",
+            "executor": {
+                "name": "SSOCheckExecutor"
+            },
+            "properties": {
+                "checkpointRef": "session"
```
ThaminduDilshan, line 18: This `checkpointRef` is sort of a checkpoint key within the session entry right?. SSOCheckExecutor lookup for a checkpoint with a key constructed by appending this ref in the session record.

But is this something intuitive for a user to configure? Does this always need to have id of the SessionExecutor? If so same info is already available via the onSuccess edge. I'm thinking whether we can set this explicitly by looking at the onSuccess edge at the graph construction time.

Anyway this is fine for the initial cut.

backend/cmd/server/deployment.yaml
```diff
@@ -60,6 +60,11 @@ crypto:
 jwt:
     preferred_key_id: "default-key"
 
+session:
```
ThaminduDilshan: Shall we move these to default.json and remove from deployment.yaml?

backend/cmd/server/servicemanager.go
```diff
@@ -349,6 +352,16 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 			GithubSvc:             githubAuthnService,
 			GoogleSvc:             googleAuthnService,
 			OpenID4VPVerifierSvc:  openid4vpSvc,
+			SessionStore: flowsession.NewStore(
```
ThaminduDilshan: Plugging store implementations is not something we follow currently. But something we thought of having in future.
But is this a requirement currently? If this is not used outside flow package, we can keep it flow internal. 

There are use cases which uses flow execution and core, but nothing else. Need to see how they can unplug session implementation via the thunderidengine/ service manager.

cc: @senthalan

backend/internal/flow/common/constants.go
```diff
@@ -199,8 +207,27 @@ const (
...
+	// It is scoped per checkpoint via SSOCheckpointKey; the paired Session node reads it to choose
+	// load vs save. "true" commits the join to loading (fail closed if the snapshot is gone).
+	RuntimeKeySSOSessionPresent = "ssoSessionPresent"
+	// RuntimeKeySSOSessionSaved is the prefix of the per-checkpoint guard holding the handle under
+	// which a checkpoint's context was saved, making the save idempotent if the join re-executes. It
+	// is scoped per checkpoint via SSOCheckpointKey.
+	RuntimeKeySSOSessionSaved = "ssoSessionSaved"
+	// RuntimeKeyAuthTime holds the Unix timestamp (seconds) at which the subject authenticated
+	// for the current session, carried across the SSO path for downstream assurance checks.
+	RuntimeKeyAuthTime = "authTime"
```
ThaminduDilshan:
```suggestion
	RuntimeKeyAuthTime = "ssoAuthTime"
```

What if we append sso prefix for this too?

backend/internal/flow/session/store.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// store implements StoreInterface against the runtime relational store.
+type store struct {
+	dbProvider   provider.DBProviderInterface
+	deploymentID string
+}
+
+// NewStore creates a session store backed by the given runtime DB provider.
+func NewStore(dbProvider provider.DBProviderInterface, deploymentID string) StoreInterface {
```
ThaminduDilshan: Having public constructor method doesn't align with the dependency injection pattern we follow. In order to prevent unintentional object creation, we keep all the constructor methods private and exposes them via a package Initialize() method placed in a init.go file.

Shall we follow the same pattern in session stores too?

backend/internal/flow/executor/error_constants.go
```diff
@@ -1158,6 +1158,60 @@ var (
...
+				"acr_values or max_age",
+		},
+	}
+
+	// ErrSSOSessionLoadFailed is returned by the Session node when a session resolved by SSO-Check
+	// cannot be loaded (the session or its context is gone, or the snapshot is unreadable). On the
+	// SSO path the credential steps were already skipped, so a load failure leaves no authenticated
+	// subject and the flow cannot proceed: it is a hard failure rather than a graceful degrade.
+	ErrSSOSessionLoadFailed = tidcommon.ServiceError{
+		Type: tidcommon.ServerErrorType,
```
ThaminduDilshan: We don't define server errors in package level. We log them in service layer and return `InternalServerError` in backend/pkg/thunderidengine/common/error.go.

But in executors we should return a go error if it needed to be treated as a server error

backend/internal/flow/flowexec/engine.go
```diff
@@ -644,6 +652,16 @@ func (fe *flowEngine) processNodeResponse(ctx *EngineContext, nodeResp *common.N
 		return nil, false, &tidcommon.InternalServerError
 	}
 
+	// Carry any SSO handle minted by this node onto the flow step so the transport layer can
+	// emit it. The Session node emits the handle on the generic AdditionalData channel (keeping SSO
+	// off the engine contract). Stamped here (not only at completion) so it survives an immediately
+	// following prompt step that returns the flow as incomplete.
+	if handle := nodeResp.AdditionalData[session.SessionHandleKey]; handle != "" {
```
ThaminduDilshan: Data added to additional data will be returned in the response. Seems a special case handling is done by adding this to additional data and removing it upon usage.
Better to follow a different approach for this.

backend/internal/flow/flowexec/engine.go
```diff
@@ -1462,3 +1480,32 @@ func processNodeResponseErrorForEventPublish(nodeResp *common.NodeResponse) map[
...
+	if ctx == nil || ctx.Graph == nil {
+		return ""
+	}
+	return ctx.Graph.GetID()
+}
+
+// flowUsesSSOSession reports whether the flow graph contains a node that establishes or
+// consults an SSO session (the Session or SSO-Check executors). It gates the active-flow-version
+// lookup so only SSO-capable flows pay for it.
+func flowUsesSSOSession(graph core.GraphInterface) bool {
```
ThaminduDilshan: Seems this method is used in service.go file. Shall we move it there or move to a utils.go file?

backend/internal/system/config/config.go
```diff
@@ -133,6 +133,30 @@ func (c *OTPConfig) Validate() error {
 	return nil
 }
 
+// SessionConfig holds the SSO session lifetime configuration. Durations are in seconds; a zero or
+// omitted value falls back to the built-in default. SSO sessions are a product feature, so this is
+// product config (top-level `session`), kept out of the reusable engine's flow config.
+type SessionConfig struct {
```
ThaminduDilshan: Alternatively this can be moved to the new server config API

backend/internal/flow/flowexec/flowVersionLookup_mock_test.go
```diff
@@ -0,0 +1,110 @@
+// Code generated by mockery; DO NOT EDIT.
```
ThaminduDilshan: Do we have a interface for this mockery generated file? Couldn't find a matching interface.
Is this something generated manually?

backend/internal/flow/executor/session_executor.go
```diff
@@ -0,0 +1,475 @@
...
+
+// newSessionExecutor creates a new Session executor. The transactioner is used to write the
+// session and its session context atomically on the fresh path. The authn provider resolves the
+// subject's entity reference and attributes when saving, and is the contract downstream nodes
+// use to read the subject reconstructed on the SSO load path.
+func newSessionExecutor(flowFactory core.FlowFactoryInterface, store session.StoreInterface,
+	sessionContextStore session.SessionContextStoreInterface, participantStore session.ParticipantStoreInterface,
+	transactioner transaction.Transactioner, authnProvider providers.AuthnProviderManager,
+	timeouts session.Timeouts) *sessionExecutor {
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, sessionLoggerComponentName),
```
ThaminduDilshan:
```suggestion
	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, "SessionExecutor"),
```
No need for a variable as it'll be available package wide

backend/internal/flow/executor/session_executor.go
```diff
@@ -0,0 +1,475 @@
...
+
+	execResp := &providers.ExecutorResponse{
+		Status:      providers.ExecComplete,
+		RuntimeData: make(map[string]string),
+	}
+
+	checkpoint := ctx.CurrentNodeID
+	if ctx.RuntimeData[common.SSOCheckpointKey(common.RuntimeKeySSOSessionPresent, checkpoint)] == dataValueTrue {
+		if err := e.loadCheckpoint(ctx, execResp, checkpoint, logger); err != nil {
+			logger.Error(ctx.Context, "Failed to load SSO checkpoint", log.Error(err))
```
ThaminduDilshan: If an error needs to be handled as client error, we should set it to the exec response. Client errors are never logged in the server.
If an error needs to be handled as server error, we should return a go error from Execute method and it'll be logged in the task execution node. Otherwise error will be double logged

backend/internal/flow/executor/session_executor.go
```diff
@@ -0,0 +1,475 @@
...
+		CheckpointID:   checkpoint,
+		RuntimeData:    sanitizeSnapshotRuntimeData(ctx.RuntimeData),
+		AuthUser:       authUserJSON,
+		CompletedSteps: buildCompletedSteps(ctx.ExecutionHistory),
+		ContextVersion: 1,
+	}
+
+	// Write this checkpoint's context (upsert) and the joining participant in one transaction.
+	now := time.Now().UTC()
+	saveErr := e.transactioner.Transact(ctx.Context, func(txCtx context.Context) error {
```
ThaminduDilshan: Rather than handling transactions in executors, what if we introduce a session service (or sso service or session enforcer service) to a sub package in `internal/authn` and call that service from the executors?

backend/internal/flow/executor/session_executor_test.go
```diff
@@ -0,0 +1,663 @@
...
+	"github.com/thunder-id/thunderid/internal/system/cache"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	"github.com/thunder-id/thunderid/internal/system/transaction"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+	"github.com/thunder-id/thunderid/tests/mocks/authnprovider/managermock"
+)
+
+// txStub is a test Transactioner that runs the function and models commit/rollback: on error
+// it invokes onRollback (which test code uses to discard the fakes' staged writes).
+type txStub struct {
```
ThaminduDilshan: We should be able to use mockery to generate stubs rather than writing custom mocks

backend/internal/flow/executor/session_executor_test.go
```diff
@@ -0,0 +1,663 @@
...
+			// Completed authentication step: recorded with its completion time (ms → s).
+			"basic_auth": {NodeID: "basic_auth", ExecutorName: "CredentialsAuthExecutor",
+				ExecutorType: providers.ExecutorTypeAuthentication, Status: providers.FlowStatusComplete,
+				EndTime: 1700000005000},
+		},
+		Application: providers.Application{ID: "app-123"},
+	}
+}
+
+func TestSession_FreshSave(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/flow/executor/session_executor.go
ThaminduDilshan: If our plan is to keep sessions flow specific, we can add this service to flow/session package. internal/authn package provides shared authn services that can be used in both flows and atomic APIs

backend/internal/flow/executor/sso_check_executor.go
```diff
@@ -0,0 +1,170 @@
...
+// using the decrypt-free checkpoint listing. On lookup error it fails closed (treats it as absent),
+// forcing full authentication rather than skipping on incomplete information.
+func (e *ssoCheckExecutor) checkpointPresent(ctx *providers.NodeContext, sessionID, checkpoint string,
+	logger *log.Logger) bool {
+	if e.checkpoints == nil {
+		return false
+	}
+	ids, err := e.checkpoints.ListCheckpointIDs(ctx.Context, sessionID)
+	if err != nil {
+		logger.Error(ctx.Context, "Failed to list SSO session checkpoints", log.Error(err))
```
ThaminduDilshan: Shouldn't log server errors at this level. We should return a go error and it'll be logged in task exec node

backend/internal/flow/executor/sso_check_executor.go
```diff
@@ -0,0 +1,170 @@
...
+// at an incompatible flow version (which forces full authentication).
+func (e *ssoCheckExecutor) resolveSession(ctx *providers.NodeContext, logger *log.Logger) *session.Session {
+	in := session.SSOInputsFrom(ctx.Context)
+	if e.resolver == nil || in.Handle == "" {
+		return nil
+	}
+
+	s, err := e.resolver.Resolve(ctx.Context, in.Handle, time.Now().UTC())
+	if err != nil {
+		logger.Error(ctx.Context, "Failed to resolve SSO session", log.Error(err))
```
ThaminduDilshan: Same here

backend/internal/flow/executor/session_executor_test.go
ThaminduDilshan: Same applies to backend/internal/flow/executor/sso_check_executor_test.go and backend/internal/flow/flowexec/service_sso_test.go

backend/internal/flow/flowexec/init.go
```diff
@@ -66,7 +68,14 @@ func Initialize(
 	flowExecService := newFlowExecService(flowProvider, flowStore, flowEngine,
 		actorProvider, observabilitySvc, transactioner, cryptoSvc, graphBuilder, cfg)
 
-	handler := newFlowExecutionHandler(flowExecService)
+	// Mark the SSO cookie Secure unless the deployment is configured to serve over plain HTTP, and
+	// bound its lifetime to the session's configured absolute timeout (same fallback as the session
+	// executor's timeouts).
+	secureCookies := !config.GetServerRuntime().Config.Server.HTTPOnly
```
ThaminduDilshan: Need to inject these too `flowconfig.Config` struct. Accessing server runtime directly is demoted

backend/internal/flow/flowexec/handler.go
```diff
@@ -32,11 +34,18 @@ import (
...
+	ssoHandleTTL time.Duration
 }
 
-func newFlowExecutionHandler(flowExecService FlowExecServiceInterface) *flowExecutionHandler {
+func newFlowExecutionHandler(flowExecService FlowExecServiceInterface, secure bool,
+	ssoHandleTTL time.Duration) *flowExecutionHandler {
 	return &flowExecutionHandler{
 		flowExecService: flowExecService,
+		// The SSO cookie is marked Secure when the deployment serves over TLS (i.e. not HTTP-only).
+		ssoTransport: session.NewCookieTransport(secure),
```
ThaminduDilshan: Should be injected as a dependency rather than invoking public constrcutor method

backend/internal/flow/mgt/graph_builder_sso_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/thunder-id/thunderid/internal/system/cache"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	engineconfig "github.com/thunder-id/thunderid/pkg/thunderidengine/config"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// TestSSOFlowDefinitionBuilds validates that the bootstrap SSO sample flow parses, references
+// only registered executors, and builds into a graph with all expected nodes. It exercises the
+// real executor registry so a typo'd executor name or dangling node reference fails here.
+func TestSSOFlowDefinitionBuilds(t *testing.T) {
```
ThaminduDilshan: Shall we move this to the existing graph builder test suite and the same file?

backend/internal/flow/session/errors.go
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package session
+
+import "errors"
+
+var (
+	// ErrVersionConflict is returned by Update when the optimistic-lock version no longer
+	// matches (the row was updated concurrently or no longer exists).
+	ErrVersionConflict = errors.New("session version conflict")
```
ThaminduDilshan: Shall we keep these error constants unexported?

Current practice is to make services return `ServiceError` struct. Go errors can be returned from internal layers like stores. But no need to expose them

backend/internal/flow/session/inputs.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return SSOInputs{}
+}
+
+// SessionHandleKey is the key for the SSO session handle. It is used both as the RuntimeData key that
+// carries the run's session handle across nodes (resolved on reuse, minted on a fresh login) and as the
+// ExecutorResponse.AdditionalData key the Session node uses to hand a freshly minted handle to the
+// transport layer for the per-flow cookie. Using the generic AdditionalData channel keeps SSO concepts
+// out of the reusable engine contract.
+const SessionHandleKey = "ssoSessionHandle"
```
ThaminduDilshan: If this is used as a runtime data key, shall we keep an alias for this in runtime data constants?

backend/internal/flow/session/model.go
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+// Package session provides the persistent SSO session model and relational store.
+//
+// A session is the unit that carries authenticated state across separate flow
+// executions. It is grouped by flow: the flow ID is the group key, so only
+// applications configured with the same flow can share a session (SSO). The
+// session is referenced by an opaque handle, decoupled from the transport that
+// carries it (a cookie is one such transport; see HandleTransport).
+package session
```
ThaminduDilshan, line 19: Shall we move package comment to main service file or init.go file? Also we shouldn't define this in multiple files

backend/internal/flow/session/participant.go
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package session
+
+import "time"
+
+// Participant records an application that has used (joined) an SSO session. A session is shared
+// across the applications that authenticate through its flow; each such application is tracked so
+// the session's audience is known — the basis for logout and subject-scoped revocation.
+type Participant struct {
```
ThaminduDilshan: Since all of these inputs.go, participant.go defines session related models, we can move all of them to model.go file. Fine to split into multiple files if the content is high and need separation for clarity

backend/internal/flow/session/session_context.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// StepFact is a per-node completed authentication-step fact.
+type StepFact struct {
+	Executor string `json:"executor,omitempty"`
+	Status   string `json:"status,omitempty"`
+	// CompletedAt is the Unix time (seconds) at which the authentication step completed.
+	CompletedAt int64 `json:"completedAt,omitempty"`
+	// TODO(sso): enrich with the authentication method reference (amr). amr is a protocol-specific
+	// fact; see SSO_TODO.md.
```
ThaminduDilshan: Is this intentional comment?

backend/internal/flow/session/session_context_store.go
```diff
@@ -0,0 +1,175 @@
...
+	GetByCheckpoint(ctx context.Context, sessionID, checkpointID string) (*SessionContext, error)
+	// Delete removes all of a session's checkpoint contexts.
+	Delete(ctx context.Context, sessionID string) error
+	CheckpointLister
+}
+
+// CheckpointLister exposes the lightweight existence check the SSO-Check node uses to decide
+// checkpoint availability. It is a narrow view of the context store so the hot check path never
+// loads a context payload.
+type CheckpointLister interface {
```
ThaminduDilshan: This is like a main service right? Shall we move this out from store into a separate file?

backend/internal/flow/session/participant_store.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/system/database/provider"
+)
+
+// ParticipantStoreInterface defines the persistence operations for SSO session participants.
+type ParticipantStoreInterface interface {
```
ThaminduDilshan: We can keep a single store implementation merging all of these different stores right? 

Is there a requirement to split invocations into multiple levels like config time, runtime, etc?

backend/internal/flow/session/store_constants.go
```diff
@@ -0,0 +1,125 @@
+/*
```
ThaminduDilshan, line 1: Current practice is to keep errors in a error_constants.go file and queries in a store_constants.go file. But I'd like to adopt this pattern in throughout the code base (I.e. Keep errors in errors.go and queries in queries.go).

backend/internal/flow/session/queries.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"github.com/thunder-id/thunderid/internal/system/database/model"
+)
+
+var (
+	// QueryCreateSession inserts a new SSO session. It is idempotent per establishing flow execution:
+	// on a FLOW_EXECUTION_ID conflict it does nothing, so concurrent joins in one execution converge
+	// on the single session that won the race (the caller re-reads it via QueryGetSessionByExecutionID).
+	// The ON CONFLICT ... DO NOTHING form is valid in both PostgreSQL and SQLite.
+	QueryCreateSession = model.DBQuery{
```
ThaminduDilshan: Shall we make queries package private?

backend/internal/flow/session/store.go
```diff
@@ -0,0 +1,303 @@
...
+		return time.Time{}
+	}
+	t, err := parseTime(value, "")
+	if err != nil {
+		return time.Time{}
+	}
+	return t
+}
+
+// parseTime parses a required time column across the string and time.Time forms drivers return.
```
ThaminduDilshan: If possible let's reuse `ParseDBTimeField()` in backend/internal/system/utils/time_util.go

backend/internal/flow/session/state_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewTimeouts_Defaults(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/flow/session/transient_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// TestSessionRowCarriesNoTransientExecutionState guards the lean, hot-path SESSION row: transient
+// per-execution state (current node, partial inputs, runtime data, execution history, challenge
+// token) must never land on it, so liveness/activity checks never pull execution state.
+//
+// The session-context sibling intentionally snapshots runtime state (RuntimeData, the resolved
+// AuthUser, completed steps) so the SSO join can replay the effective attribute set — so this
+// guard deliberately does NOT cover SessionContext.
+func TestSessionRowCarriesNoTransientExecutionState(t *testing.T) {
```
ThaminduDilshan: We should be able to move this to a existing test suite right?

backend/internal/flow/session/transport_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCookieName(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/flow/session/state.go
```diff
@@ -0,0 +1,73 @@
...
+	if t.Idle > t.Absolute {
+		t.Idle = t.Absolute
+	}
+	return t
+}
+
+// Revoker is the seam for revoking a subject's sessions when their identity attributes or
+// standing change. It is intentionally unimplemented in this POC.
+// TODO(sso): implement subject-scoped revocation on identity-attribute/standing change.
+type Revoker interface {
```
ThaminduDilshan: Better to move this to a dedicated service file? Or else we can move all service interfaces to a interface.go file

backend/internal/flow/flowexec/service.go
```diff
@@ -168,6 +181,57 @@ func (s *flowExecService) Execute(ctx context.Context,
...
+
+// resolveActiveFlowVersion returns the current active version of the flow, or 0 when it
+// cannot be determined. A 0 (unknown) version makes the SSO-Check node treat any saved
+// session as version-incompatible, i.e. it falls back to full authentication.
+func (s *flowExecService) resolveActiveFlowVersion(ctx context.Context, engineCtx *EngineContext,
+	logger *log.Logger) int {
+	if engineCtx.Graph == nil {
+		return 0
+	}
+	def, svcErr := s.flowProvider.GetFlow(ctx, engineCtx.Graph.GetID())
```
ThaminduDilshan, line 226: Here we'll be fetching the flow again right? Can't we pass this version from previous fetch/ add to ctx and reuse?

backend/internal/flow/graphbuilder/testdata/sso_flow.json
ThaminduDilshan, line 1: Do we need to keep this here?

backend/internal/flow/session/participant_store.go
```diff
@@ -0,0 +1,100 @@
+/*
```
ThaminduDilshan, line 1: Shall we merge this to the same store.go file?

backend/internal/flow/session/store_constants.go
```diff
@@ -0,0 +1,125 @@
...
+			`LAST_ACTIVE_AT = $3, IDLE_EXPIRES_AT = $4, ABSOLUTE_EXPIRES_AT = $5, STATE = $6, ` +
+			`VERSION = VERSION + 1, UPDATED_AT = CURRENT_TIMESTAMP ` +
+			`WHERE SESSION_ID = $7 AND DEPLOYMENT_ID = $8 AND VERSION = $9`,
+	}
+
+	// queryCreateSessionContext upserts a checkpoint's session context. Re-saving the same checkpoint
+	// (re-execution or a concurrent request) overwrites it rather than erroring on the primary key.
+	// The ON CONFLICT ... DO UPDATE form is valid in both PostgreSQL and SQLite.
+	queryCreateSessionContext = model.DBQuery{
+		ID: "SSO-SESS-AC-01",
```
ThaminduDilshan, line 73: Not a must change. But we should be able to update all queries to use the same prefix since we have a single store interface now

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -872,6 +872,10 @@ type ExecutorResponse struct {
 	Assertion      string                 `json:"assertion,omitempty"`
 	Error          *common.ServiceError   `json:"error,omitempty"`
 	AuthUser       AuthUser               `json:"-"`
+	// EngineData carries executor output the flow engine consumes internally (for example, a
+	// transport signal such as a minted session handle). Unlike AdditionalData, it is never
+	// serialized to the client.
+	EngineData map[string]string `json:"-"`
```
ThaminduDilshan, line 878: We already have a forwarded data concept, but that's used to communicate data from current node to the next. If possible let's reuse that for this requirement

backend/internal/flow/session/session_context_store.go
```diff
@@ -0,0 +1,141 @@
+/*
```
ThaminduDilshan, line 1: Should be able to merge to the same store.go file

ThaminduDilshan (dismissed):

rajithacharith (approved):

PR #3793, opened 2026-07-07
Block connection deletion when dependencies exist

rajithacharith:
### Purpose

Extend the "block deletion when dependencies exist" experience — already implemented for users — to connections. The backend already refuses to delete a connection that has blocking dependencies (e.g. flows that reference it), but the console surfaced no advance warning: the delete action just failed. This PR adds a connection usages endpoint and updates the connection delete dialog to check for, list, and block on dependent resources before deletion.

### Approach

**Backend — connection usages endpoint**
- Added `GetIDPUsages` to the identity-provider service (mirrors `GetUserUsages`): validates the ID, confirms the IDP exists, and returns dependency data from the resource-dependency registry. Being informational, it returns "unknown" when the registry is unset rather than failing closed (deletion remains gated separately by `ensureNoBlockingDependencies`).
- Added `usagesByType` to the connection service (verifies the instance is the expected vendor type, then delegates) and exposed `GET /connections/{type}/{id}/usages`.

**Frontend — connection delete dialog**
- New `useGetConnectionUsages` hook + `ConnectionUsage`/`ConnectionUsagesResponse` models and query key.
- `ConnectionDeleteDialog` now fetches usages when open, shows a loading alert while checking, lists resources with `behaviorOnDelete: restrict` in an error alert (capped at 5 with a "+N more" row), and disables the Delete button while loading or when blocked. `ConnectionDetailPage` passes the connection type and id through.
- Added the `connections` i18n strings (`delete.usages.loading`, `delete.usages.more`, `delete.blocking.title`).

This mirrors the existing user delete dialog for consistency.

### Related Issues
- Fixes #3792

### Related PRs
- N/A

### Checklist
- [x] Followed the contribution guidelines.
- [x] Manual test round performed and verified.
- [ ] Documentation provided. (Add links if there are any)
    - [ ] Ran Vale and fixed all errors and warnings
- [x] Tests provided. (Add links if there are any)
    - [x] Unit Tests
    - [ ] Integration Tests
- [ ] Breaking changes. (Fill if applicable)
    - [ ] Breaking changes section filled.
    - [ ] `breaking change` label added.

### Security checks
- [x] Followed secure coding standards in [WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)
- [x] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.

[...]

PR #3712, opened 2026-07-03
Add usage support for users

rajithacharith:
### Purpose

Adds usage support for users. Introduces a new `GET /users/{id}/usages` endpoint that returns the resources referencing a given user (currently agents that list the user as their owner), aggregated across resource types. This is informational only — it drives the pre-delete confirmation dialog in the Console and does not gate deletion on the server.

Each usage entry includes a `behaviorOnDelete` field (`fallback` keeps the reference as-is, `cascade` deletes the referencing resource with the user). When usage data is unavailable, `totalResults` and `summary` are returned as `null` rather than `0`/empty, so consumers can distinguish "unknown" from "confirmed empty".

### Approach

- **Backend**: Added `HandleUserUsagesGetRequest` handler and `GetUserUsages` service method. The user service resolves usages through the shared `resourcedependency.Registry`, which is injected via `SetDependencyRegistry` from the service manager after provider services initialise (avoiding a cyclic import). Registered the user resource in the dependency registry.
- **API**: Documented the new endpoint and the `ResourceUsage`, `ResourceUsagesSummary`, and `ResourceUsagesResponse` schemas in `api/user.yaml`.
- **Frontend**: Added `useGetUserUsages` query hook and models, and wired the results into `UserDeleteDialog` so the delete confirmation surfaces referencing resources. Added i18n keys.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
- [x] Followed the contribution guidelines.
- [ ] Manual test round performed and verified.
- [ ] Documentation provided. (Add links if there are any)
    - [ ] Ran Vale and fixed all errors and warnings
- [x] Tests provided. (Add links if there are any)
    - [x] Unit Tests
    - [ ] Integration Tests
- [ ] Breaking changes. (Fill if applicable)

### Security checks
- [x] Followed secure coding standards in [WSO2 Secure Coding Guidelines](https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)
- [x] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.

## Summary by CodeRabbit

* **New Features**
  * Added `GET /users/{id}/usages` to fetch aggregated user usage details with per-resource summaries.
  * Updated user deletion flow to display affected owning agents/resources (including overflow via “+N more”) using the new usages API.
* **Bug Fixes**
  * Clarified and surfaced “unknown” vs “confirmed empty” usage states in the UI.
[...]

PR #3362, opened 2026-06-17
Add usage service registry and flow usages support

rajithacharith:
### Purpose

Generalises the resource-usage lookup introduced for themes into a reusable **usage service registry**, and adds a flow usages endpoint + pre-delete confirmation UI on top of it.

Previously, "what references this theme?" was wired ad-hoc: `thememgt` defined an `ApplicationUsageReader` interface, `servicemanager` built a bespoke adapter, and only applications were considered (agents — which share the inbound-client store — were miscounted as applications). Every new "X references Y" relationship would have needed another reader interface, adapter, and `SetXxx` call.

This PR replaces that with a registry any resource-owning service can query, and uses it to answer flow usages (`GET /flows/{flowId}/usages`).

Key changes:
- New `internal/system/usage` package: `Registry` + `Provider` (`GetResourceUsages(ctx, resourceType, id)`) + aggregated `UsagesResponse` (per-type `summary` map; `totalResults`/`summary` are `null` when unavailable = "unknown").
- `application` and `agent` are registered as providers and now filter by entity `Category`, fixing the agents-counted-as-applications bug.
- Generalised inbound-client lookup `GetEntityIDsByReference(refType, refID, …)` (replaces `GetEntityIDsByThemeID`), keyed by reference type — adding a new referenced type no longer requires editing existing providers.
- Theme migrated from `ApplicationUsageReader` to the registry; new flow usages endpoint, handler, route, and service method.
- Console: `useGetFlowUsages` hook + `FlowDeleteDialog` now lists affected applications/agents before deletion (mirrors the theme delete dialog).
- Registry is hardened against nil/uninitialised providers; consumers degrade to "unknown" if the registry was never injected.

### Approach

- **Registry over point-to-point wiring.** A service that owns a resource asks the registry which other resources reference `(resourceType, id)`; the registry fans out to all registered providers and concatenates results. Providers are blind to each other.
- **Reference mapping lives in the inbound-client store**, not in each provider. `GetEntityIDsByReference` maps `theme → THEME_ID` and `flow → (AUTH_FLOW_ID | REGISTRATION_FLOW_ID | RECOVERY_FLOW_ID)`. Unknown reference types resolve to empty, so providers stay generic (no per-type `switch`).
- **Aggregated, non-paginated response.** The endpoint is informational (drives the pre-delete dialog), so the registry returns one aggregated array with a per-`resourceType` count map; each provider is bounded by `MaxCompositeStoreRecords`.
[...]

PR #3802, opened 2026-07-07
Cascade delete role and group assignments on user deletion

### LINKED ISSUE #3801: Cascade-delete a user's role assignments and group memberships on user deletion
### Description

When a user is deleted, resources that reference the user as a principal are left behind pointing at an identity that no longer exists:

- The user's **role assignments** remain in `ROLE_ASSIGNMENT`.
- The user's **group memberships** remain in `GROUP_MEMBER_REFERENCE`.

These should be removed as part of deleting the user.

### Scope

This issue covers the **cascade cleanup** of a deleted user's dependents. The complementary **blocking** behavior (e.g. refusing to delete a user that still owns an agent, with `409 Conflict`) is already handled by the dependency registry and is out of scope here.

### Proposal

Use the existing `resourcedependency` registry's `behaviorOnDelete = cascade` semantics and add a write path:

- Add an optional `CascadeDeleter` provider interface and a `Registry.CascadeDelete` fan-out.
- The role-assignment and group services implement it to delete a principal's assignments/memberships.
- `DeleteUser` invokes the cascade after the blocking guard and before the entity delete, so a cleanup failure aborts the operation and leaves the user retriable.

### Notes / follow-ups

- Role assignments and group memberships live in different databases from the user entity, with polymorphic assignee/member references (no FK), so a DB-level `ON DELETE CASCADE` is not possible — cleanup is application-level, idempotent, and fail-closed.
- Dependents residing in the immutable (file-based/declarative) store cannot be cascade-deleted; handling those as blocking is a known follow-up.

rajithacharith:
### Purpose

When a user is deleted, remove the resources that reference the user as a principal, which would otherwise be left dangling:

- the user's **role assignments** (`ROLE_ASSIGNMENT`), and
- the user's **group memberships** (`GROUP_MEMBER_REFERENCE`).

This builds on the existing dependency registry (the same mechanism behind the usages endpoint and the delete-blocking guard) — no parallel machinery. Delete-blocking (e.g. owned agents → `409`) is already merged and is not part of this PR.

### Approach

- **Cascade behavior:** added `BehaviorCascade` and an optional `CascadeDeleter` provider interface, plus a `Registry.CascadeDelete` fan-out that invokes every provider implementing it and stops on the first error.
- **Providers declare, target enforces.** The role-assignment and group services implement `CascadeDeleter` to remove a principal's assignments/memberships (`DeleteAssignmentsByAssignee` / `DeleteMembershipsByMember`, added across the db/composite/file stores; users/apps/agents are stored as `entity` assignees/members).
- **`DeleteUser` sequence:** existing checks → blocking guard → **cascade cleanup** → entity delete. Cascade runs before the entity delete so a cleanup failure aborts the operation and leaves the user retriable.
- **Cross-DB note:** role assignments and group memberships live in different databases from the user entity with polymorphic references (no FK), so `ON DELETE CASCADE` is not possible — cleanup is application-level, idempotent, and fail-closed. The immutable (declarative) dependent case is a documented follow-up.
- Also fixes pre-existing duplicate `DBQuery` IDs in the group store (`GRQ-GROUP_MGT-03`/`-04` were reused across a query builder's branches → `-22`/`-23`).

### Related Issues

- Fixes #3801

### Related PRs

- Builds on the dependency-registry and delete-blocking work (#3362, #3712, #3793).

### Checklist
- [x] Tests provided.
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * User deletion now triggers automatic cascade cleanup for related group memberships and role assignments for supported principal types.
  * Resource-dependency handling now includes cascade-deletion behavior to remove dependent data when appropriate.
* **Bug Fixes**
  * Deletion now aborts when cascade cleanup fails, preventing partial removals.
* **Tests**
  * Expanded unit test coverage and updated test doubles to support cascade-deletion, membership deletion, and role-assignment removal behavior.

### DIFF
--- a/backend/internal/role/store_constants.go
+++ b/backend/internal/role/store_constants.go
@@ -116,6 +116,14 @@ var (
 		Query: `DELETE FROM "ROLE_ASSIGNMENT" WHERE ROLE_ID = $1 AND DEPLOYMENT_ID = $2`,
 	}
 
+	// queryDeleteRoleAssignmentsByAssignee deletes all assignments for a given assignee across roles
+	// (used to cascade-delete assignments when the assignee principal is deleted).
+	queryDeleteRoleAssignmentsByAssignee = dbmodel.DBQuery{
+		ID: "RLQ-ROLE_MGT-25",
+		Query: `DELETE FROM "ROLE_ASSIGNMENT" ` +
+			`WHERE ASSIGNEE_TYPE = $1 AND ASSIGNEE_ID = $2 AND DEPLOYMENT_ID = $3`,
+	}
+
 	// queryCheckRoleNameExists checks if a role name already exists for a given organization unit.
 	queryCheckRoleNameExists = dbmodel.DBQuery{
 		ID:    "RLQ-ROLE_MGT-14",
--- a/backend/internal/system/resourcedependency/registry.go
+++ b/backend/internal/system/resourcedependency/registry.go
@@ -116,10 +116,18 @@ type Provider interface {
 	GetResourceDependencies(ctx context.Context, resourceType, id string) ([]ResourceDependency, error)
 }
 
+// CascadeDeleter is an optional interface a Provider may implement to delete its cascade-behavior
+// dependents of a target resource when the target is deleted. Implementations must be idempotent
+// and return the number of dependents removed.
+type CascadeDeleter interface {
+	CascadeDeleteDependencies(ctx context.Context, resourceType, id string) (int, error)
+}
+
 // Registry fans out dependency lookups to all registered providers.
 type Registry interface {
 	RegisterProvider(p Provider)
 	GetDependencies(ctx context.Context, resourceType, id string) (*DependenciesResponse, error)
+	CascadeDelete(ctx context.Context, resourceType, id string) (int, error)
 }
 
 // registry is the default implementation of Registry.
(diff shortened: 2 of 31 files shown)


backend/internal/role/store_constants.go
```diff
@@ -116,6 +116,14 @@ var (
 		Query: `DELETE FROM "ROLE_ASSIGNMENT" WHERE ROLE_ID = $1 AND DEPLOYMENT_ID = $2`,
 	}
 
+	// queryDeleteRoleAssignmentsByAssignee deletes all assignments for a given assignee across roles
+	// (used to cascade-delete assignments when the assignee principal is deleted).
+	queryDeleteRoleAssignmentsByAssignee = dbmodel.DBQuery{
+		ID: "RLQ-ROLE_MGT-23",
```
UdeshAthukorala: RLQ-ROLE_MGT-23 is already used by queryGetRoleListByOUID (and -24 by queryGetRoleListCountByOUID). The next free ID is RLQ-ROLE_MGT-25.

backend/internal/system/resourcedependency/registry.go
```diff
@@ -40,6 +40,10 @@ const BehaviorFallback = "fallback"
 // reference exists; the caller must remove or reassign the reference before the target can be deleted.
 const BehaviorRestrict = "restrict"
 
+// BehaviorCascade indicates the referencing resource is deleted along with the target when the
+// target is deleted (e.g. a user's role assignments are removed when the user is deleted).
+const BehaviorCascade = "cascade"
```
UdeshAthukorala: Are we using this constant in anywhere in the source code?

backend/internal/role/store_constants.go
rajithacharith: Fixed

backend/internal/system/resourcedependency/registry.go
rajithacharith: Fixed

UdeshAthukorala (approved):

PR #3837, opened 2026-07-08
Simplify public URL configuration

Osara-B:
### Purpose
Changing ThunderID's public URL (hostname, port, or protocol) previously required editing the same information in many places: the backend server config, the gate_client block, both frontend config.js files, and the Helm deployment. There was no single source of truth, so the host/port/scheme was duplicated everywhere and easy to get partially wrong, leading to redirect and login failures.

This PR makes the served/accessed URL the effective source of truth for the common case:

The frontend apps (Console and Gate) derive the backend API URL from the origin they are served from when nothing is configured, so the server block in config.js becomes optional.
The backend gate_client settings default to the server's own public URL when unset, so gate_client only needs configuring when Gate is hosted separately from the server.
The result: for the common deployment, changing the hostname requires no edits to the default configuration. public_url remains the explicit override for cases where the externally reachable URL differs from what the browser accesses.

Excludes the CORS and passkey allowed-origin auto-include work discussed in the issue. 

### Approach

  Frontend (runtime derivation from the served URL):
- Made the `server` block and its `hostname`, `port`, and `http_only` fields optional in the config types.
- `getServerUrl()` and the trusted-issuer fallback that depends on it now return `window.location.origin` when no `public_url` or `hostname`/`port` is configured, matching the pattern `getClientUrl()` already used.
- Removed the hardcoded `server` blocks from the Console and Gate `config.js` files, replacing them with guidance comments. The default configs now ship with no hardcoded host, so the apps use whatever URL they are served from.
- Added unit tests for the served-origin fallback in `getServerUrl()` and `getTrustedIssuerUrl()`.

Backend (`gate_client` defaults to the server's URL):

- `LoadConfig` now derives `gate_client.hostname`, `port`, and `scheme` from the server's public URL (via `GetServerURL`) when they are not explicitly set, falling back to the scheme's default port (80 for http, 443 for https) when the URL carries no explicit port. Removed the hardcoded `gate_client` defaults from `default.json`.
- Added a startup guard: if the resolved `gate_client` hostname is a bind-all address (`0.0.0.0` or `::`), config loading fails with an actionable message telling the operator to set `server.public_url` (or `gate_client.hostname`) to a browser-reachable host. This prevents a silent misconfiguration that would otherwise produce broken login and error redirects.
- Added unit tests covering derivation from `hostname`/`port`, derivation from `public_url` (with and without an explicit port, over http and https), partial overrides, and the bind-all failure case.

Local development:

- The dev apps are served on their own Vite origin rather than by the backend, so each app's `vite.config` injects a build-time constant `__DEV_SERVER_URL__` (from `THUNDERID_DEV_SERVER_URL`, defaulting to `https://localhost:8090`) for the dev server only. Production builds receive an empty string.
- `main.tsx` seeds the runtime config with this value only under an `import.meta.env.DEV` guard, so the block is removed from production bundles by dead-code elimination and never affects a real deployment. The `build.sh` and `build.ps1` scripts pass the backend's public URL through this variable.

Documentation:

- Updated the configuration guide, quick-start README, and Helm `values.yaml` to document the new optional defaults for `gate_client` and the frontend `server` block, and to describe `public_url` as the single override point.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3406

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * The app now derives backend connection defaults from the server’s public-facing URL when separate Gate client settings aren’t provided.
  * Dev mode automatically wires the frontend to the current backend origin via the dev server URL.

* **Bug Fixes**
  * Added stricter validation to fail fast when backend hostname is unreachable (e.g., bind-all addresses without a browser-reachable public URL).
  * Improved client-side URL fallback behavior when server configuration is missing or partial.

* **Documentation**
  * Updated configuration guidance to describe `server.public_url` and the new defaulting behavior.

### DIFF
--- a/docs/content/guides/getting-started/configuration.mdx
+++ b/docs/content/guides/getting-started/configuration.mdx
@@ -31,17 +31,20 @@ Controls the <ProductName /> server's network settings and identity.
 | `server.hostname` | `localhost` | The hostname or IP address the server binds to |
 | `server.port` | `8090` | Port the server listens on |
 | `server.http_only` | `false` | If `true`, disables HTTPS and uses HTTP only (not recommended for production) |
+| `server.public_url` | _(derived from server hostname, port and protocol)_ | The public URL clients use to reach the server, if it differs from the bind address. Derived when unset. |
 | `server.identifier` | `default-deployment` | Unique identifier for this deployment instance |
+
+When `server.public_url` is set, the `gate_client` settings below default to the corresponding hostname, port, and protocol from that URL. Most deployments only need to configure the server section.
 
 ## Gate Client Configuration
 
-Configures the connection to <ProductName /> Gate (the login UI).
+Configures the connection to <ProductName /> Gate (the login UI). Every setting is optional. By default, the `gate_client` settings are derived from `server.public_url`, so you only need to configure this section when Gate is hosted separately from the server.
 
 | Setting | Default | Description |
 |---------|---------|-------------|
-| `gate_client.hostname` | `localhost` | The hostname where <ProductName /> Gate is hosted |
-| `gate_client.port` | `8090` | Port where <ProductName /> Gate is accessible |
-| `gate_client.scheme` | `https` | Protocol scheme (`http` or `https`) |
+| `gate_client.hostname` | _(server's hostname)_ | The hostname where <ProductName /> Gate is hosted |
+| `gate_client.port` | _(server's port)_ | Port where <ProductName /> Gate is accessible |
+| `gate_client.scheme` | _(server's scheme)_ | Protocol scheme (`http` or `https`) |
 | `gate_client.path` | `/gate` | Base path for <ProductName /> Gate |
 
 ## TLS Configuration
--- a/backend/cmd/server/config/default.json
+++ b/backend/cmd/server/config/default.json
@@ -15,9 +15,6 @@
     }
   },
   "gate_client": {
-    "hostname": "localhost",
-    "port": 8090,
-    "scheme": "https",
     "path": "/gate"
   },
   "tls": {
--- a/backend/internal/system/config/config.go
+++ b/backend/internal/system/config/config.go
@@ -22,10 +22,13 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"net"
+	"net/url"
 	"os"
 	urlpath "path"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/thunder-id/thunderid/internal/system/log"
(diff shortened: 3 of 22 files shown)


brionmario (dismissed):

rajithacharith (dismissed):

docs/content/guides/getting-started/configuration.mdx
```diff
@@ -31,17 +31,20 @@ Controls the <ProductName /> server's network settings and identity.
 | `server.hostname` | `localhost` | The hostname or IP address the server binds to |
 | `server.port` | `8090` | Port the server listens on |
 | `server.http_only` | `false` | If `true`, disables HTTPS and uses HTTP only (not recommended for production) |
+| `server.public_url` | _(derived from server hostname, port and scheme)_ | The public URL clients use to reach the server, if it differs from the bind address. Derived when unset. |
```
himeshsiriwardana: protocol is appropriate here instead of "scheme"

docs/content/guides/getting-started/configuration.mdx
```diff
@@ -31,17 +31,20 @@ Controls the <ProductName /> server's network settings and identity.
...
 | `server.identifier` | `default-deployment` | Unique identifier for this deployment instance |
 
+When `public_url` is set, the `gate_client` settings below default to it, so most deployments only need to configure the server section above.
```
himeshsiriwardana:
```suggestion
When `server.public_url` is set, the `gate_client` settings below default to the corresponding hostname, port, and protocol from that URL. Most deployments only need to configure the server section.
```

docs/content/guides/getting-started/configuration.mdx
```diff
@@ -31,17 +31,20 @@ Controls the <ProductName /> server's network settings and identity.
...
+
 ## Gate Client Configuration
 
-Configures the connection to <ProductName /> Gate (the login UI).
+Configures the connection to <ProductName /> Gate (the login UI). Every setting is optional. When unset, the `gate_client` settings default to the server's public URL, so this section is only needed when Gate runs separately from the server.
```
himeshsiriwardana:
```suggestion
Configures the connection to <ProductName /> Gate (the login UI). Every setting is optional. By default, the `gate_client` settings are derived from `server.public_url`, so you only need to configure this section when Gate is hosted separately from the server.
```

docs/content/guides/getting-started/configuration.mdx
Osara-B: Resolved in [Simplify URL configuration](https://github.com/thunder-id/thunderid/pull/3837/commits/252c2af58341aa8066d54d33d0c652b028523275)

rajithacharith (approved):

PR #3863, opened 2026-07-09
Attribute Cache is updated to use RuntimeStoreProvider

anushasunkada:
### Purpose

Attribute Cache is updated to use RuntimeStoreProvider

---
`ATTRIBUTE_CACHE` table is removed from the runtime database, instead uses `RUNTIME_STORE` table.

#### 💥 Impact
The data stored in the ATTRIBUTE_CACHE will be lost. This is not a critical impact as it's have only the information related to the flow. The users might need go through the login, recovery or registration flows

#### 🔄 Migration Guide
Create new table in the runtimestore

---

### Related Issues
- #3037 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Attribute cache is now stored in the shared runtime store under the `attribute:cache` namespace.
  * Added runtime-store TTL extension support used by attribute-cache refresh flows.
* **Bug Fixes**
  * Refresh-token flows now extend attribute-cache TTL whenever a cache entry exists.
  * JWT assertion generation now relies on the app’s validity period and default issuer behavior.
* **Chores**
  * Updated runtime database schema and cleanup scripts to match the new attribute-cache storage layout.
* **Tests**
  * Refreshed attribute-cache and runtime-store TTL test coverage and updated not-found/error expectations.


backend/internal/attributecache/service.go
```diff
@@ -146,10 +146,6 @@ func (s *attributeCacheService) ExtendAttributeCacheTTL(
 
 	err := s.store.ExtendAttributeCacheTTL(ctx, id, ttlSeconds)
 	if err != nil {
-		if errors.Is(err, errAttributeCacheNotFound) {
```
ThaminduDilshan, line 149: With this, if someone try to extend ttl for a non existing cache, it will be treated as a server error right? IMO that's wrong behaviour for the service

backend/internal/attributecache/service.go
```diff
@@ -171,10 +167,6 @@ func (s *attributeCacheService) DeleteAttributeCache(
 
 	err := s.store.DeleteAttributeCache(ctx, id)
 	if err != nil {
-		if errors.Is(err, errAttributeCacheNotFound) {
```
ThaminduDilshan, line 174: Same as previous comment. Instead here we should silently return if the error is errAttributeCacheNotFound

ThaminduDilshan: Since DB store not returning an error for this case and it's only server errors this implementation is fine

ThaminduDilshan (approved):

backend/internal/attributecache/service.go
ThaminduDilshan: Fixed

PR #3872, opened 2026-07-09
Revoke the consumed refresh token on rotation to enforce single-use

indeewari:
### Purpose

Enforce **single-use refresh tokens**. When refresh-token rotation is enabled, the authorization server now records the *consumed* refresh token on the revocation deny list so it cannot be replayed, per [RFC 9700 (BCP 240) §4.14.2](https://www.rfc-editor.org/rfc/rfc9700#section-4.14.2). This is the grant-independent slice of implicit revocation (M3-a).

Gated by a new setting `oauth.refresh_token.revoke_previous_on_renew` (default `true`); it only takes effect when `oauth.refresh_token.renew_on_grant` is enabled.

### Approach

- **New `RefreshTokenRevokerInterface`** in the `revocation` package — a narrow *write* seam that exposes only rotation revocation (`RevokeRotatedRefreshToken`), not the client-facing RFC 7009 path. This keeps the refresh grant's reach least-privilege (read/write split).
- The refresh grant records the consumed RT's `jti` with reason `refresh_rotation`. **Fail-closed:** if the revocation cannot be recorded, the rotation is rejected so the old token is never left usable; the client retries with the still-valid token.
- `RefreshTokenClaims.Exp` is now carried so the deny-list entry is bounded to the consumed token's original expiry.
- Replay *rejection* then flows through the existing AS deny-list enforcement (`ValidateRefreshToken` → `EnsureNotRevoked`, #3727) — no new read path.

**Deferred (needs a grant/family identifier):** grant-scoped behaviours — cascade to related tokens (RFC 7009 §2.1) and reuse-triggered grant revocation (RFC 9700 §4.14.2) — cannot be done on the current jti-only model and are out of scope here.

### Related Issue
- https://github.com/thunder-id/thunderid/issues/3605

### Related PRs
- Builds on #3839 (RS RevocationCache), #3803 (validator-centralized enforcement), and #3727 (deny-list write + AS enforcement) — all merged to `main`.

### Checklist
- [ ] Documentation provided. (separate follow-up PR)
- [x] Tests provided.
    - [x] Unit Tests (revocation 95.6%, granthandlers 92.4%)
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added support for revoking the consumed refresh token during refresh-token renewal (rotation).
  * Added the `revokePreviousOnRenew` / `revoke_previous_on_renew` configuration toggle.

* **Bug Fixes**
  * Renewal now fails if the consumed refresh token revocation cannot be recorded.
  * Revocation uses the consumed token’s expiry (`exp`) to correctly bound invalidation.

* **Documentation**
  * Updated Helm chart configuration docs and values with the new option.

* **Tests**
  * Added/expanded tests covering success, empty-JTI no-op behavior, and revocation failure.


backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -53,12 +63,14 @@ type revocationService struct {
...
-// newRevocationService creates a new revocationService (internal use).
+// newRevocationService creates a new revocationService (internal use). It returns the concrete type so
+// the caller can expose it as both RevocationServiceInterface (RFC 7009 write path) and
+// RefreshTokenRevokerInterface (rotation single-use), which it both satisfies.
 func newRevocationService(
 	jwtService jwt.JWTServiceInterface,
 	store RevokedTokenStoreInterface,
 	observabilitySvc providers.ObservabilityProvider,
-) RevocationServiceInterface {
+) *revocationService {
```
ThaminduDilshan: Shall we use the interface in the constructor method signature? Any reason to do this change?

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -119,6 +131,26 @@ func (s *revocationService) RevokeToken(
 	return RevokeOutcomeRevoked, nil
 }
 
+// RevokeRotatedRefreshToken records a consumed (rotated) refresh token on the deny list with the
+// refresh_rotation reason, enforcing single-use. The token was already validated by the refresh grant,
+// so no signature or ownership check is repeated here. An empty jti is a no-op.
+func (s *revocationService) RevokeRotatedRefreshToken(ctx context.Context, jti string, expiryTime time.Time) error {
```
ThaminduDilshan:
```suggestion
func (s *revocationService) RevokeRefreshToken(ctx context.Context, jti string, expiryTime time.Time) error {
```

Can't we keep this as a generic refresh token revoke method? Rotate is unnecessary info here right?
Need to update logs and errors too

backend/internal/oauth/oauth2/revocation/service.go
ThaminduDilshan: `RefreshTokenRevokerInterface` is a provider interface that the caller consumes right? We can define all methods in the `RevocationServiceInterface` and let the constructor return it. Modify the caller to only accept `RefreshTokenRevokerInterface`.
This way we can plug the same `revocationService` instance to the caller, but since the interface has defined only a limited set of methods, caller won't be able to invoke other methods

backend/pkg/thunderidengine/config/config.go
```diff
@@ -162,8 +162,9 @@ type AuthClassConfig struct {
 
 // RefreshTokenConfig holds the refresh token configuration details.
 type RefreshTokenConfig struct {
-	RenewOnGrant   bool  `yaml:"renew_on_grant"  json:"renew_on_grant"`
-	ValidityPeriod int64 `yaml:"validity_period" json:"validity_period"`
+	RenewOnGrant          bool  `yaml:"renew_on_grant"           json:"renew_on_grant"`
```
ThaminduDilshan, line 165: We need to add these configs to deployment resources in install directory too. Otherwise we won't be able to template/customize values

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Good call — reverted. `newRevocationService` now returns `RevocationServiceInterface` again (no concrete return). That interface **embeds** `RefreshTokenRevokerInterface`, so the single `revocationService` instance satisfies both, and `Initialize` hands it to the refresh grant narrowed to `RefreshTokenRevokerInterface` — the grant sees only `RevokeRefreshToken` and can't reach the full API. Done in `b131bda`.

backend/internal/oauth/oauth2/revocation/service.go
indeewari: Done — renamed to `RevokeRefreshToken` and updated the log/error strings accordingly (`b131bda`). Kept the recorded reason as `refresh_rotation` since rotation is the only caller today; if a future caller needs a different reason we can add a reason parameter then.

backend/pkg/thunderidengine/config/config.go
indeewari: Done — added `revoke_previous_on_renew` to the Helm resources in `b131bda`: `install/helm/conf/deployment.yaml` (templated), `install/helm/values.yaml` (`configuration.oauth.refreshToken.revokePreviousOnRenew: true`), and the README config table.

backend/internal/oauth/oauth2/revocation/init.go
```diff
@@ -34,21 +34,22 @@ import (
...
 	discoveryService discovery.DiscoveryServiceInterface,
 	observabilitySvc providers.ObservabilityProvider,
-) EnforcementServiceInterface {
+) (EnforcementServiceInterface, RefreshTokenRevokerInterface) {
 	enforcementService := newEnforcementService(observabilitySvc)
 	revocationService := newRevocationService(jwtService, newRevokedTokenStore(), observabilitySvc)
 	revocationHandler := newRevocationHandler(revocationService)
 	registerRoutes(mux, revocationHandler, actorProvider, authnProvider, jwtService, discoveryService)
-	return enforcementService
+	return enforcementService, revocationService
```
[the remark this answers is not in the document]
indeewari: Documentation for the new `oauth.refresh_token.revoke_previous_on_renew` setting and the single-use rotation behavior will be handled in a **separate docs follow-up PR** (as noted in the PR checklist), to keep this change focused.

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -43,6 +43,21 @@ type RevocationServiceInterface interface {
 	//
 	// It returns an error only on server errors; all token-state outcomes are conveyed via RevokeOutcome.
 	RevokeToken(ctx context.Context, token, tokenTypeHint, authenticatedClientID string) (RevokeOutcome, error)
+
+	// RefreshTokenRevokerInterface is embedded so a single revocationService instance can be handed to
+	// the refresh grant as the narrow RefreshTokenRevokerInterface (a subset of this interface),
+	// preventing the grant from reaching the full revocation API.
+	RefreshTokenRevokerInterface
```
ThaminduDilshan: Can we move this to top and remove the comment?

indeewari: Done in `7fdce56` — moved the embedded `RefreshTokenRevokerInterface` to the top of `RevocationServiceInterface` and removed the comment.

ThaminduDilshan (approved):

PR #3893, opened 2026-07-10
Add Identity Assertion Authorization Grant (ID-JAG) support

sahandilshan:
### Purpose
Implement both legs of draft-ietf-oauth-identity-assertion-authz-grant-04 at the /oauth2/token endpoint.

Leg 1 (issuance): extend the RFC 8693 token-exchange handler so a confidential client can exchange a self-issued ID token for an ID-JAG (requested_token_type=urn:ietf:params:oauth:token-type:id-jag). The issued JWT uses typ=oauth-id-jag+jwt with iss/sub/aud/client_id/jti/ iat/exp/scope, and the response carries issued_token_type=id-jag, token_type=N_A, and no refresh token. Issuance is gated per app by a new token.idJag config (allowedAudiences exact-match, validityPeriod).

Leg 2 (consumption): add the urn:ietf:params:oauth:grant-type:jwt-bearer grant, which validates an ID-JAG issued by a trusted external IdP (new id_jag_enabled IdP property) — enforcing the typ header, JWKS signature, exp/iat/nbf/jti, aud bound to this server's issuer, and the client_id claim bound to the authenticated client — then issues an access token whose subject is the federated principal, tagged with an idp provenance claim.

Both grants are restricted to confidential clients at request and create time. jti/iat presence is required; jti replay caching and DPoP binding are deferred.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2440

## Summary by CodeRabbit

* **New Features**
  * Added **jwt-bearer** grant support and **ID-JAG** issuance via **token exchange**, including token type handling.
  * Introduced **ID-JAG app configuration** (enabled, allowed audiences, validity with defaulting).
  * Extended **OIDC connection** requests/responses with an optional **idJagEnabled** flag.
  * Updated OAuth/OpenAPI docs and discovery examples to advertise **jwt-bearer** support.

* **Bug Fixes**
  * Strengthened validation to reject the **`none`** token endpoint auth method when confidentiality is required.
  * Improved ID-JAG configuration handling and defaulting.

* **Tests**
  * Expanded test coverage for jwt-bearer and ID-JAG grant flows, token building, and validation/defaulting.

### DIFF
--- a/backend/internal/idp/constants.go
+++ b/backend/internal/idp/constants.go
@@ -36,6 +36,7 @@ const (
 	PropIssuer                = "issuer"
 	PropTokenExchangeEnabled  = "token_exchange_enabled"
 	PropTrustedTokenAudience  = "trusted_token_audience"
+	PropIDJagEnabled          = "id_jag_enabled"
 )
 
 // Known endpoints for Google OAuth2/OIDC.
@@ -97,6 +98,7 @@ var idpPropertyConfigs = map[providers.IDPType]idpPropertyConfig{
 			PropIssuer,
 			PropTokenExchangeEnabled,
 			PropTrustedTokenAudience,
+			PropIDJagEnabled,
 		},
 		Defaults: map[string]string{},
 	},
--- a/backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
+++ b/backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
@@ -1,0 +1,199 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package granthandlers
+
+import (
+	"context"
+	"slices"
+
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/dpop"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/model"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/resourceindicators"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/tokenservice"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// jwtBearerGrantHandler handles the jwt-bearer grant type used to present an ID-JAG assertion
+// (draft-ietf-oauth-identity-assertion-authz-grant) issued by a trusted external IdP.
+type jwtBearerGrantHandler struct {
+	tokenBuilder    tokenservice.TokenBuilderInterface
+	tokenValidator  tokenservice.TokenValidatorInterface
+	resourceService providers.ResourceServerProvider
+}
+
+// newJWTBearerGrantHandler creates a new instance of jwtBearerGrantHandler.
+func newJWTBearerGrantHandler(
+	tokenBuilder tokenservice.TokenBuilderInterface,
+	tokenValidator tokenservice.TokenValidatorInterface,
+	resourceService providers.ResourceServerProvider,
+) GrantHandlerInterface {
+	return &jwtBearerGrantHandler{
+		tokenBuilder:    tokenBuilder,
+		tokenValidator:  tokenValidator,
+		resourceService: resourceService,
+	}
+}
+
+// ValidateGrant validates the jwt-bearer grant type request.
+func (h *jwtBearerGrantHandler) ValidateGrant(ctx context.Context, tokenRequest *model.TokenRequest,
+	oauthApp *providers.OAuthClient) *model.ErrorResponse {
+	if providers.GrantType(tokenRequest.GrantType) != providers.GrantTypeJWTBearer {
+		return &model.ErrorResponse{
+			Error:            constants.ErrorUnsupportedGrantType,
+			ErrorDescription: "Unsupported grant type",
+		}
+	}
+
+	if tokenRequest.Assertion == "" {
+		return &model.ErrorResponse{
+			Error:            constants.ErrorInvalidRequest,
+			ErrorDescription: "Missing required parameter: assertion",
+		}
+	}
+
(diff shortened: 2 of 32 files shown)


backend/internal/idp/constants.go
```diff
@@ -36,6 +36,7 @@ const (
 	PropIssuer                = "issuer"
 	PropTokenExchangeEnabled  = "token_exchange_enabled"
 	PropTrustedTokenAudience  = "trusted_token_audience"
+	PropIDJagEnabled          = "id_jag_enabled"
```
rajithacharith, line 39: Sync this properly with connections service as the UI is only available for these via Connections

backend/internal/oauth/oauth2/granthandlers/token_exchange.go
```diff
@@ -237,6 +258,109 @@ func (h *tokenExchangeGrantHandler) HandleGrant(ctx context.Context, tokenReques
...
+				Error:            constants.ErrorServerError,
+				ErrorDescription: "Token revocation status could not be verified",
+			}
+		}
+		return nil, &model.ErrorResponse{
+			Error:            constants.ErrorInvalidRequest,
+			ErrorDescription: "subject_token must be an ID token issued to this client",
+		}
+	}
+	if subjectClaims.Iss != config.GetServerRuntime().Config.JWT.Issuer {
```
thiva-k: Can't we move this to ValidateIDJAGSubjectToken? And cfg should be injected right?

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -580,6 +580,20 @@ type OAuthTokenConfig struct {
...
+
+// DefaultIDJAGValidityPeriod is the default validity period, in seconds, of an issued ID-JAG when the
+// application does not configure one.
+const DefaultIDJAGValidityPeriod int64 = 300
+
+// IDJAGConfig is the Identity Assertion Authorization Grant (ID-JAG) configuration. Enabled must be
+// true and AllowedAudiences must be non-empty for the application to request ID-JAGs via token
+// exchange.
+type IDJAGConfig struct {
+	Enabled          bool     `json:"enabled"                    yaml:"enabled"                    jsonschema:"Enable ID-JAG issuance for this application."`
```
thiva-k, line 590: Why do we need this separate flag? Based on the allowed audiences cant we make this decision?

sahandilshan: So pure purpose of this is to have a UI level option to show whether the ID jag is enabled or not. But we can remove this and backed with whether the AUD is empty or not

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -580,6 +580,20 @@ type OAuthTokenConfig struct {
 	AccessToken  *AccessTokenConfig  `json:"accessToken,omitempty"  yaml:"accessToken,omitempty"  jsonschema:"Access token configuration."`
 	IDToken      *IDTokenConfig      `json:"idToken,omitempty"      yaml:"idToken,omitempty"      jsonschema:"ID token configuration."`
 	RefreshToken *RefreshTokenConfig `json:"refreshToken,omitempty" yaml:"refreshToken,omitempty" jsonschema:"Refresh token configuration."`
+	IDJAG        *IDJAGConfig        `json:"idJag,omitempty"        yaml:"idJag,omitempty"        jsonschema:"Identity Assertion Authorization Grant (ID-JAG) configuration."`
+}
+
+// DefaultIDJAGValidityPeriod is the default validity period, in seconds, of an issued ID-JAG when the
+// application does not configure one.
+const DefaultIDJAGValidityPeriod int64 = 300
```
thiva-k: Shall we move this constant out from model.go?

backend/internal/idp/constants.go
sahandilshan: This is synced with the `/conncetion`

thiva-k:
Let's update the openAPI spec too

backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
```diff
@@ -0,0 +1,157 @@
...
+	// claim so that this external `sub` is not mistaken for a local user id by downstream consumers.
+	accessToken, err := h.tokenBuilder.BuildAccessToken(ctx, &tokenservice.AccessTokenBuildContext{
+		Subject:           assertionClaims.Sub,
+		Audiences:         audiences,
+		ClientID:          tokenRequest.ClientID,
+		Scopes:            grantedScopes,
+		SubjectAttributes: make(map[string]interface{}),
+		GrantType:         string(providers.GrantTypeJWTBearer),
+		OAuthApp:          oauthApp,
+		SourceIDP:         assertionClaims.Iss,
```
thiva-k, line 167: These access tokens can be dpop bound right? Shouldn't we sent dpopjkt here like other grants?

backend/internal/oauth/oauth2/granthandlers/token_exchange_test.go
```diff
@@ -1235,7 +1235,7 @@ func (suite *TokenExchangeGrantHandlerTestSuite) TestValidateGrant_UnsupportedID
 	assert.NotNil(suite.T(), errResp)
 	assert.Equal(suite.T(), constants.ErrorInvalidRequest, errResp.Error)
 	assert.Contains(suite.T(), errResp.ErrorDescription, "Unsupported requested_token_type")
-	assert.Contains(suite.T(), errResp.ErrorDescription, "Only access tokens and JWT tokens are supported")
+	assert.Contains(suite.T(), errResp.ErrorDescription, "Only access tokens, JWT tokens, and ID-JAGs are supported")
```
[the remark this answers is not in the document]
sahandilshan: Will update the documentation when revamping agent documentations

backend/internal/oauth/oauth2/granthandlers/token_exchange.go
sahandilshan: This is fixed now

backend/pkg/thunderidengine/providers/model.go
sahandilshan: @thiva-k in second thought, we should keep this as a seperate flag. In a scenario where user temporary wants disbale the id jag without removing the audiences, then we won't be able to achieve that use case if we bound the the enable/disable with the audiences

backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
sahandilshan: THis is fixed now. Added the dpop support as well

thiva-k (approved):

backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
```diff
@@ -0,0 +1,199 @@
...
+		logger.Debug(ctx, "Failed to validate ID-JAG assertion", log.Error(err))
+		return nil, &model.ErrorResponse{
+			Error:            constants.ErrorInvalidGrant,
+			ErrorDescription: "Invalid assertion",
+		}
+	}
+
+	// Granted scopes = (assertion scope ∩ the app's registered scopes), further intersected with the
+	// request scope parameter when present.
+	grantedScopes := intersectScopes(assertionClaims.Scopes, oauthApp.Scopes)
```
thiva-k, line 106: Is it intentional to intersect with oauth app scopes here? Oauth app scopes is not being used in any validations in other grants AFAIK. 

IMO better to remove this intersection logic and use app-resource subscription here once implemented.

Issue #3939, opened 2026-07-13
[SCIM] Implement SCIM Group management endpoints


ravindu439:
### What problem are we solving?

When an admin wants to manage groups of users via SCIM (RFC 7644), they currently have no way to create, read, update, or delete Groups through the SCIM API, which causes them to fall back to manual per-user provisioning in the Console or rely on external IdP tooling that ThunderID's SCIM endpoint doesn't yet support.

### Who are we solving this for?

Primary: Identity/IT Admin integrating ThunderID with an external IdP (Okta, Azure AD, etc.) via SCIM

- Goal: Provision and sync groups (not just users) automatically from the IdP
- Pain: SCIM Users CRUD exists (Phase 1), but Group resource is missing, so group membership/roles can't be synced, forcing manual reconciliation

### Why should we solve this now?

- Value to Persona: Full SCIM 2.0 compliance (RFC 7643/7644) lets admins provision groups end-to-end without manual steps
- Value to Org: SCIM Users (Phase 1) and ServiceProviderConfig/discovery endpoints already implemented  — Group support completes the core SCIM resource set expected by IdP integrations, unblocking enterprise SCIM adoption

### Proposed Solution

Implement SCIM Group endpoints (/Groups) per RFC 7644:

- POST /Groups — create group
- GET /Groups/{id} and GET /Groups — retrieve/list groups (with filtering/pagination consistent with Users endpoints)
- PUT/PATCH /Groups/{id} — update group and membership
- DELETE /Groups/{id} — delete group
- Group resource schema per RFC 7643 (displayName, members, etc.), built on top of the existing SCIM Users implementation (shared handler/service/constants patterns from backend/internal/scim).

PR #3971, opened 2026-07-14
feat(scim): Implement SCIM Group management

ravindu439:
### Purpose
Implements SCIM Group management endpoints (`/scim/v2/Groups`), completing CRUD and PATCH support for Groups per RFC 7643/7644, following on from the earlier SCIM Users and ServiceProviderConfig work. Also adds RFC 7644 §3.14 optimistic concurrency control (`ETag`/`If-Match`) for both Groups and Users, since the service already advertised `ETagSupported: true` in ServiceProviderConfig but never enforced it.

### Approach
- Added `scimGroupsHandler` (`group_handler.go`) implementing List, Create, Get, Replace (PUT), Patch, and Delete against `/scim/v2/Groups` and `/scim/v2/Groups/{id}`, plus `OPTIONS` for CORS preflight.
- Added `scimGroupsService` (`group_service.go`) as the adapter between the SCIM wire format and the existing `group.GroupServiceInterface`, translating Thunder `group.Group`/`group.Member` to SCIM `SCIMGroup`/`SCIMGroupMember` resources (`group_resource.go`, `group_model.go`) and mapping group-service errors to SCIM errors.
- Added `group_patch_validator.go` to parse and validate PATCH bodies per RFC 7644 §3.5.2 — supports `add`/`remove`/`replace` on `displayName` and `members`, including the `members[value eq "<id>"]` filtered-remove path.
- Added weak-ETag versioning (`version.go`) — a SHA-256 hash over the mutable fields (`displayName`, `members`) so `meta.version` only changes when resource content does.
- Wired new handler into `Initialize`/`registerRoutes` (`init.go`), removed `/scim/v2/Groups*` from the unimplemented stub list, added `PATCH` to allowed CORS methods.
- Added `GET/POST/PUT/PATCH/DELETE /scim/v2/Groups[/**]` entries to permission table (`permissions.go`), reusing existing `p.Group`/`p.GroupView` permissions.
- Added `If-Match` optimistic concurrency enforcement (RFC 7644 §3.14): `ReplaceGroup`, `PatchGroup`, `DeleteGroup`, `ReplaceUser`, and `DeleteUser` now compare an incoming `If-Match` header against the resource's current weak ETag (`checkIfMatch`/`normalizeETag` in `version.go`) before applying the mutation, and reject stale writes with `412 Precondition Failed` (new `ErrorPreconditionFailed`, `SCIM-1027`). The header is optional — omitting it preserves the previous unconditional-write behavior.
- Added the HTTP `ETag` response header (mirroring `meta.version`) on all Group/User responses that return a resource representation (GET, POST-create, PUT, PATCH), which was previously only present in the JSON body.
- Regenerated `SCIMGroupsServiceInterface` and `SCIMUsersServiceInterface` mocks via `make mockery`.

### Behavior Corrections
This PR also corrects two previously-shipped `ServiceProviderConfig`/schema values that were inaccurate; both are corrections, not intentional feature changes, and were caught in review:
- **`FilterSupported: true → false`** (`config/config.go`) — this was overclaiming. The Users filter implementation only supports the `eq` operator, not the full RFC 7644 §3.4.2.2 filter grammar, so advertising full filter support was inaccurate. `false` now reflects actual capability.
- **`CaseExact: false → true`** for user extension attributes (`schema_builder.go`) — corrects a latent matching-semantics bug; these attribute values should compare case-sensitively per their intended semantics.

### Related Issues
- #3939

### Related PRs
- #3743

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,361 @@
...
+			continue
+		}
+		// GetGroup does not populate Members for database-backed groups, so fetch separately.
+		members, err := s.fetchAllGroupMembers(ctx, g.ID)
+		if err != nil {
+			continue
+		}
+		full.Members = members
+		scimGroups = append(scimGroups, buildSCIMGroupResource(*full, baseURL))
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

sadilchamishka (commented):
Reviewed the SCIM Group implementation + ETag/If-Match work. Overall this is clean, idiomatic, and unusually well-tested (handler/service/validator/version paths all covered). Left inline comments on the points worth addressing; #1 (non-atomic mutations) is the one I'd want resolved or explicitly deferred before merge.

One cross-file note that didn't fit a single line: two changes touch **already-shipped** behavior and aren't called out in the PR description — `FilterSupported` flipped `true → false` in `config/config.go` (changes what ServiceProviderConfig advertises for the Users endpoint) and `CaseExact` flipped `false → true` for user extension attributes in `schema_builder.go:70`. Both may be intentional corrections, but since they affect previously-released behavior they deserve a line in the PR body (or a separate commit) so consumers aren't surprised.

backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,365 @@
...
+	if svcErr != nil {
+		return nil, mapGroupServiceErrorToSCIM(svcErr)
+	}
+	if len(existingMembers) > 0 {
+		if _, svcErr = s.groupService.RemoveGroupMembers(ctx, groupID, existingMembers); svcErr != nil {
+			return nil, mapGroupServiceErrorToSCIM(svcErr)
+		}
+	}
+	if len(thunderMembers) > 0 {
+		if _, svcErr = s.groupService.AddGroupMembers(ctx, groupID, thunderMembers); svcErr != nil {
```
sadilchamishka, line 159: **Non-atomic mutation — partial failure can wipe members.** `ReplaceGroup` runs `UpdateGroup` → `RemoveGroupMembers` → `AddGroupMembers` as three separate service calls with no surrounding transaction. If `AddGroupMembers` fails here after `RemoveGroupMembers` already succeeded, the group is left with **all members removed and none added**, and the client gets an error implying nothing changed.

`PatchGroup` (the sequential `for _, action := range actions` loop) has the same issue — a multi-op PATCH that fails on op 2 leaves op 1 committed, which violates RFC 7644 §3.5.2's all-or-nothing requirement.

If the group service exposes a transactional API, wrap these; otherwise please add a comment documenting the non-atomic behavior as a known limitation.

backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+	scimGroups := make([]SCIMGroup, 0, len(listResp.Groups))
+	for _, g := range listResp.Groups {
+		// GroupBasic has no Members; fetch full group for members
+		full, svcErr := s.groupService.GetGroup(ctx, g.ID, true)
+		if svcErr != nil {
+			return SCIMGroupListResponse{}, mapGroupServiceErrorToSCIM(svcErr)
+		}
+		// GetGroup does not populate Members for database-backed groups, so fetch separately.
+		members, svcErr := s.fetchAllGroupMembers(ctx, g.ID, true)
```
sadilchamishka, line 56: **N+1 (really N×2+1) queries.** For each group in the page this calls both `GetGroup` *and* `fetchAllGroupMembers` (which itself pages), so listing 20 groups = 1 list + 20 `GetGroup` + 20+ `GetGroupMembers` round-trips.

The per-group `GetGroup` also looks redundant: its result feeds `buildSCIMGroupResource` but `Members` is immediately overwritten, and the `GroupBasic` from `GetGroupList` already carries `ID` and `Name`. Constructing the `group.Group` from the basic entry would drop the extra 20 round-trips.

backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,365 @@
...
+	case "GRP-1003":
+		return &ErrorResourceNotFound
+	case "GRP-1015":
+		return &ErrorMutabilityViolation
+	case "GRP-1007", "GRP-1008":
+		return &ErrorInvalidGroupMember
+	case "GRP-1004":
+		return &ErrorUniquenessConflict
+	case "GRP-1014":
+		return &ErrorUnsupportedMemberType
```
sadilchamishka: **Prefer the exported error vars over literal codes.** These switch on `"GRP-1003"`, `"GRP-1015"`, etc., while the accompanying test references `group.ErrorGroupNotFound`, `group.ErrorImmutableGroup`, `group.ErrorInvalidMemberID`, … Using `group.ErrorGroupNotFound.Code` (etc.) here keeps the mapping in sync if the group package ever renumbers, and self-documents which error each case maps to.

backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,365 @@
...
+
+	g, svcErr := s.groupService.GetGroup(ctx, groupID, false)
+	if svcErr != nil {
+		return nil, mapGroupServiceErrorToSCIM(svcErr)
+	}
+	if g.IsReadOnly {
+		return nil, &ErrorMutabilityViolation
+	}
+
+	if trimmed := strings.TrimSpace(ifMatch); trimmed != "" {
```
sadilchamishka, line 131: **TOCTOU window in the If-Match check.** This reads the current version, compares, then mutates without a lock or transaction spanning the two steps. Two concurrent `If-Match` writers can both pass the check and both apply, so §3.14's guarantee isn't fully delivered. Still better than nothing — worth a short comment noting the window (or closing it if the service can do a conditional/transactional update).

backend/internal/scim/group_handler.go
```diff
@@ -0,0 +1,202 @@
...
+		DisplayName string            `json:"displayName"`
+		Members     []SCIMGroupMember `json:"members"`
+	}
+	if err := json.Unmarshal(body, &payload); err != nil || payload.DisplayName == "" {
+		h.handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
+	hasGroupSchema := false
+	for _, s := range payload.Schemas {
+		if s == SCIMCoreGroupSchemaURN {
```
sadilchamishka: Minor consistency nit: schema URN matched here with exact `==`, but the PATCH validator uses `strings.EqualFold` for the PatchOp URN. Aligning the two (EqualFold) avoids rejecting a technically-valid request that differs only in case.

backend/internal/scim/config/config.go
```diff
@@ -47,7 +47,7 @@ const (
 
 	// FilterSupported indicates whether SCIM filtering is supported
 	// per RFC 7644 §3.4.2.2.
-	FilterSupported = true
+	FilterSupported = false
```
sadilchamishka, line 50: `FilterSupported` flipped `true → false` here. This changes what `ServiceProviderConfig` advertises for the already-shipped Users endpoint too, and it isn't mentioned in the PR description — worth calling out explicitly (see the general review comment).

backend/internal/scim/group_service.go
ravindu439: Confirmed, real gap. `group.GroupServiceInterface` has no atomic replace-members operation, each of `AddGroupMembers/RemoveGroupMembers` wraps its own transaction independently, nothing spans both. Adding true atomicity means a new transactional method at the group service layer, out of scope for this SCIM-only PR. Documented as known limitation on `ReplaceGroup, PatchGroup`, and the PATCH replace-op path (see inline comments) per your suggested fallback.

backend/internal/scim/group_service.go
ravindu439: fixed

backend/internal/scim/group_service.go
ravindu439: fixed

backend/internal/scim/group_service.go
ravindu439: The version isn't a persisted column, `generateVersion` derives it on the fly from `displayName+members` each call `(version.go)`, so there's no `WHERE version=?` to compare-and-swap against, and `group.GroupServiceInterface` doesn't expose a locking primitive spanning the check and the mutation (each `AddGroupMembers/RemoveGroupMembers/UpdateGroup` call opens its own independent transaction). Closing it for real needs either a persisted version column with conditional update, or a transaction with a row lock spanning `check+mutate`, both group-service-layer changes out of scope here.

Documented the window on `checkIfMatch` in `version.go` (single shared function, covers Groups and Users).

backend/internal/scim/group_handler.go
ravindu439: fixed

Issue #3672, opened 2026-07-02
Basic SSO: reuse an existing login session across applications


madurangasiriwardena:
### What problem are we solving?

Thunder authenticates every application independently. There is no notion of a user being logged in *at Thunder* — only of a user having completed authentication for one authorization request. Consequences:

- A user who has already authenticated in the browser is prompted again for every other application in the same deployment, and again for each new authorization request to the same application.
- Any deployment with more than one application exposes this immediately, so Thunder cannot be adopted for multi-application estates without users noticing repeated logins.

### Who are we solving this for?

- **End users of applications secured by Thunder** - want to prove who they are once per browser session, not once per application.
- **Application developers** - want to deploy several applications against one Thunder instance and have the login experience behave the way users expect from any other IdP.
- **Flow authors** - want control over what an existing session is allowed to satisfy and what must always run, so that reuse doesn't silently weaken a flow designed for a sensitive operation.

### Why should we solve this now?

SSO across applications is a baseline IdP expectation.

- SSO across applications is a baseline expectation of an IdP. Its absence is not a missing convenience; it disqualifies Thunder from evaluation in any multi-application deployment.
- Introducing a session commits us to companion work - session termination, forced re-authentication, and lifetime policy — none of which exist as problems today. Doing SSO now means choosing a session model deliberately, with that follow-on work in view, at the only point where changing that choice is still cheap: once sessions are released, the model is persisted state and observable protocol surface, and every feature built on it raises the cost of revising it.

### Use Cases

1. As an end user, I want a second application to sign me in without prompting me again, so that I authenticate once per browser session.
2. As an end user returning to an application I already signed into, I want a new authorization request to be satisfied without re-entering credentials.
3. As a flow author, I want to group applications into separate session boundaries, so that session reuse is confined to applications with comparable security expectations.
4. As a flow author, I want to declare which steps an existing session satisfies and which always execute, so that a flow protecting a sensitive operation is not weakened by session reuse.
### Proposed Solution

Detailed design will be added later.

madurangasiriwardena:
Related discussion https://github.com/thunder-id/thunderid/discussions/3673

PR #3973, opened 2026-07-14
Implement session sign-out

madurangasiriwardena:
### Purpose
Implements session sign-out on the flow-graph model, closing the loop on SSO: #3672 / #3779 established and reused flow-centric sessions — this ends them.

Direction is set by discussion #3916: sign-out is authored and configured per flow, on the same flow-graph model login/SSO use, rather than as a hardcoded path in the OAuth2 layer.

Two entry points converge on one termination machinery:
- RP-initiated logout — OIDC RP-Initiated Logout 1.0 §2: a browser redirect to the OP end_session_endpoint (GET/POST /oauth2/logout).
- Native logout — an API-driven run of the client's sign-out flow through the existing /flow/execute machinery.

Both resolve the target session, revoke it and its stored context, clear the per-flow SSO cookie, and — on the browser path — confirm with the user and land them on a validated post_logout_redirect_uri.

### Approach
Two separate resolutions — the crux. A session is keyed by the login flow's ID (cookie tid_sso_<hash(loginFlowID)>), but the flow we run on sign-out is a different flow (the client's sign-out flow). These are kept distinct:

- Which flow to run — sign-out is a first-class flow type (FlowTypeSignOut = "SIGNOUT"), resolved exactly like recovery via getFlowGraph(appID, flowType) → client.SignOutFlowID.
- Which session to end — the login flow's session. The login flow id (client.AuthFlowID) is threaded explicitly onto the engine context as SessionFlowID (set only for sign-out). applyInboundSSO uses it in place of the running flow's Graph.GetID(), so the correct login-flow cookie is read.

Termination machinery. SessionSignOutExecutor reads the inbound handle + target flow id, calls session.Terminate (resolve by handle → State=ENDED, bump version, delete contexts + participants; idempotent on absent/ended sessions), and emits a cookie-clear signal the flow handler applies via the existing ssoTransport.Clear seam with CookieName(session.FlowID).

RP-initiated endpoint. New oauth2/logout package mirrors authz/: validates id_token_hint → aud → client → AuthFlowID, reads the per-flow cookie, and redirects the browser to the gate sign-out page with the flow executionId. post_logout_redirect_uri is validated against the client's allow-list (PostLogoutRedirectURIs) or falls back to a default. end_session_endpoint is advertised in OIDC discovery.

Gate. A flow-type-agnostic SignOutBox resumes the execution via a plain /flow/execute fetch (credentials: 'include'), rendering any confirmation step with FlowComponentRenderer and echoing the per-step challengeToken on submit. It runs on the published @thunderid/react 0.4.0 — no bespoke SDK sign-out surface, so this is a single-repo change (no SDK PR).

Terminology. User-facing text and internal identifiers use "Sign Out"; OIDC spec terms are preserved as-is (/oauth2/logout, post_logout_redirect_uri, end_session_endpoint, logout_hint).

Screenshot 2026-07-14 at 9 48 22 AM
Screenshot 2026-07-14 at 9 49 03 AM

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2036

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added RP-initiated OAuth/OIDC sign-out (`id_token_hint` + optional `state`), including validation and redirecting to the configured post-logout target.
  * Introduced configurable per-application **SIGNOUT** flows with new console controls, templates, and a sign-out executor step; gate now includes a Sign Out page that resumes execution and redirects on completion.
  * Added **Post-Logout Redirect URIs** end-to-end (console UI, validation, persistence, and export/import support).
* **Bug Fixes**
  * Sign-out now reliably terminates sessions and clears the correct per-flow cookie; post-logout redirect routing is preserved, and sign-out can be disabled per application.


backend/internal/oauth/oauth2/logout/handler.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+// logoutHandler serves the RP-initiated logout endpoint. It validates the request, initiates the
+// application's sign-out flow (seeding the validated post-logout landing into the flow's runtime data),
+// and redirects the browser to the gate sign-out page to run the flow (confirmation + session
+// termination). The gate executes the flow via /flow/execute, which clears the per-flow cookie and
+// lands the browser on the validated post-logout redirect URI once the flow completes.
+type logoutHandler struct {
+	service         LogoutServiceInterface
+	flowExecService flowexec.FlowExecServiceInterface
```
rajithacharith: Do we need a direct dependency to flowExec service? Can we proxy this through the logout service?

backend/internal/flow/session/service.go
```diff
@@ -210,6 +218,48 @@ func (s *service) LoadCheckpoint(ctx context.Context, handle, checkpoint, appID
...
+	sess, err := s.store.GetByHandle(ctx, handle)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load session for termination: %w", err)
+	}
+	if sess == nil {
+		return nil, nil
+	}
+	// The handle must belong to the expected flow: never end a session grouped under a different flow.
+	if flowID != "" && sess.FlowID != flowID {
+		s.logger.Debug(ctx, "Session handle belongs to a different flow; not terminating")
```
rajithacharith: Shouldn't this be a error scenario?

backend/internal/flow/common/constants.go
```diff
@@ -225,6 +225,19 @@ const (
...
+	// concepts off the reusable engine contract.
+	RuntimeKeySSOSessionCleared = "ssoSessionCleared"
+	// RuntimeKeyPostLogoutRedirectURI carries the validated post-logout landing URI seeded into a
+	// sign-out flow's runtime data by the logout endpoint. The session sign-out node echoes it into the
+	// completion response's additional data so the browser lands there after the session is
+	// terminated. It is server-seeded (never taken from the browser mid-flow), so the landing cannot
+	// be forged by tampering with the flow request.
+	RuntimeKeyPostLogoutRedirectURI = "postLogoutRedirectUri"
+	// RuntimeKeyPostLogoutState carries the OAuth state value to append to the post-logout landing.
+	RuntimeKeyPostLogoutState = "postLogoutState"
```
ThaminduDilshan: Seems this is returned in additional data. Is this something consumed by the sdk/ client? Couldn't find any other usage in backend

backend/internal/flow/common/constants.go
```diff
@@ -225,6 +225,19 @@ const (
...
+	// raises once it has terminated the session, telling the transport layer to clear the per-flow
+	// cookie. Like RuntimeKeySSOSessionHandle it rides the engine-only EngineData channel, keeping SSO
+	// concepts off the reusable engine contract.
+	RuntimeKeySSOSessionCleared = "ssoSessionCleared"
+	// RuntimeKeyPostLogoutRedirectURI carries the validated post-logout landing URI seeded into a
+	// sign-out flow's runtime data by the logout endpoint. The session sign-out node echoes it into the
+	// completion response's additional data so the browser lands there after the session is
+	// terminated. It is server-seeded (never taken from the browser mid-flow), so the landing cannot
+	// be forged by tampering with the flow request.
+	RuntimeKeyPostLogoutRedirectURI = "postLogoutRedirectUri"
```
ThaminduDilshan: If I understood correctly, this is returned in additional data to be consumed by sdk/ client right? Client reads this value and redirect to the post logout endpoint.

But can there be a requirement to perform some post protocol level actions after the signout flow completion? Ex: for oauth, trigger token revocation, update revoke list.

If there's such requirement:
Since we're keeping flow execution/ engine protocol agnostic, triggering it within the flow may not be the ideal approach. Need to think whether we have to perform a client post oauth redirection here too similar to authorization code flow.

In that case, I'm thinking whether we need to add post redirect uri and state to the flow context. If oauth is maintaining a temporary session until flow completion, we can keep this there.

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -150,6 +150,20 @@ var ErrorRecoveryFlowDisabled = tidcommon.ServiceError{
 	},
 }
 
+// ErrorSignOutFlowDisabled defines the error response for sign-out flow disabled errors.
+var ErrorSignOutFlowDisabled = tidcommon.ServiceError{
```
ThaminduDilshan: Shall we move this to the end of constant list to maintain sequential error code order?

backend/internal/inboundclient/service.go
```diff
@@ -522,6 +530,10 @@ func (s *inboundClientService) resolveFlowDefaults(ctx context.Context, c *inbou
 		// If a recovery flow is not defined, disable recovery flow for the application.
 		c.IsRecoveryFlowEnabled = false
 	}
+	if c.SignOutFlowID == "" {
+		// If a sign-out flow is not defined, disable sign-out for the application.
+		c.IsSignOutFlowEnabled = false
```
ThaminduDilshan, line 603: Is it a valid requirement to have an application without signout flow? I'm thinking whether we should have a default signout flow (configured in deployment.yaml) and attach it by default to the apps.

Also if that's the case, we don't need a flag to enable signout flows as well right?
Open for discussion

backend/internal/flow/session/service.go
```diff
@@ -210,6 +218,45 @@ func (s *service) LoadCheckpoint(ctx context.Context, handle, checkpoint, appID
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to load session for termination: %w", err)
+	}
+	if sess == nil {
+		return nil, nil
+	}
+	// The handle must belong to the expected flow. A per-flow handle resolving to a session grouped
+	// under a different flow should never happen; surface it as an error rather than silently skipping.
+	if flowID != "" && sess.FlowID != flowID {
+		return nil, fmt.Errorf("session handle belongs to flow %q, expected %q", sess.FlowID, flowID)
```
ThaminduDilshan, line 236: Seems this is handled as a server error in the executor. Hence should be fine.

But if there's a case where we may have to return this to the client, are we revealing correct flow id with the error description when someone performs logout with a invalid flow id?

backend/internal/oauth/oauth2/logout/handler.go
```diff
@@ -0,0 +1,109 @@
...
+		return
+	}
+
+	executionID, svcErr := h.service.InitiateSignOutFlow(r.Context(), resolution)
+	if svcErr != nil {
+		status := http.StatusInternalServerError
+		if svcErr.Type == tidcommon.ClientErrorType {
+			status = http.StatusBadRequest
+		}
+		h.logger.Error(r.Context(), "Failed to initiate sign-out flow",
```
ThaminduDilshan: Seems we're logging client errors too here. Is it expected?

backend/internal/oauth/oauth2/logout/service.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// LogoutServiceInterface validates an RP-initiated logout request, resolves its target, and initiates
+// the application's sign-out flow.
+type LogoutServiceInterface interface {
+	Resolve(ctx context.Context, req LogoutRequest) (*LogoutResolution, error)
+	InitiateSignOutFlow(ctx context.Context, resolution *LogoutResolution) (string, *tidcommon.ServiceError)
+}
+
+type logoutService struct {
```
ThaminduDilshan, line 81: Shall we add a comment to the struct?

backend/internal/oauth/oauth2/logout/service.go
```diff
@@ -0,0 +1,190 @@
...
+			clientID = hintClientID
+		}
+	}
+	if clientID == "" {
+		return nil, errClientRequired
+	}
+
+	client, svcErr := s.actorProvider.GetOAuthClientByClientID(ctx, clientID)
+	if svcErr != nil {
+		s.logger.Error(ctx, "Failed to resolve client for logout", log.String("clientId", clientID))
```
ThaminduDilshan: Are we logging a client error here? Maybe a existing problem in the oauth packages

backend/internal/oauth/oauth2/logout/service.go
```diff
@@ -0,0 +1,190 @@
...
+		return "", errInvalidIDTokenHint
+	}
+	return audienceClientID(payload), nil
+}
+
+// audienceClientID extracts the client id from an ID token. When the token has multiple audiences the
+// authorized party (azp) claim identifies the client, so it is preferred; otherwise the aud claim,
+// which may be a single string or an array of strings, is used.
+func audienceClientID(payload map[string]interface{}) string {
+	if azp, ok := payload["azp"].(string); ok && azp != "" {
```
ThaminduDilshan: Shall we define constants for these?

frontend/apps/console/src/features/flows/components/resources/steps/call/Call.tsx
```diff
@@ -56,6 +56,7 @@ const FLOW_TYPE_TO_ROUTE_SEGMENT: Record<string, string> = {
   [FlowType.AUTHENTICATION]: 'signin',
   [FlowType.REGISTRATION]: 'registration',
   [FlowType.RECOVERY]: 'recovery',
+  [FlowType.SIGNOUT]: 'signout',
```
ThaminduDilshan: It's possible in the backend, but do we need a signout flow to be invoked via a call node from another flow?

frontend/apps/console/src/features/flows/data/templates.json
```diff
@@ -12806,5 +12806,222 @@
...
+            },
+            "position": {
+              "x": 62,
+              "y": 278
+            }
+          },
+          "onSuccess": "prompt_confirm"
+        },
+        {
+          "id": "prompt_confirm",
```
ThaminduDilshan, line 12940: Technically this is a consent we collect right? Currently it's handled as just user prompt. We may need to think whether we have to engage consent executor here to collect and record this decision.

IMO not required for initial cut.

backend/pkg/thunderidengine/providers/constants.go
```diff
@@ -55,6 +55,8 @@ const (
 	FlowTypeUserOnboarding FlowType = "USER_ONBOARDING"
 	// FlowTypeRecovery represents a flow execution for account recovery (e.g., password reset).
 	FlowTypeRecovery FlowType = "RECOVERY"
+	// FlowTypeSignOut represents a flow execution for terminating an SSO session.
+	FlowTypeSignOut FlowType = "SIGNOUT"
```
ThaminduDilshan, line 59: Shall we also add some integration tests for this new flow type? Fine to add in a separate PR

madurangasiriwardena: Will send as a separate PR

frontend/apps/console/src/features/flows/data/templates.json
madurangasiriwardena: Will move this to the discussion and finalize.

frontend/apps/console/src/features/flows/components/resources/steps/call/Call.tsx
madurangasiriwardena: As of now I don't see any strong requirement. So I'll remove this now and add it back if we see a requirement.

backend/internal/inboundclient/service.go
madurangasiriwardena: Will move this to the discussion.

backend/internal/flow/session/service.go
madurangasiriwardena: As of now, we are logging this error and throwing a server error. So didn't change the error message.

backend/internal/oauth/oauth2/logout/handler.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	redirectURI, err := h.service.CompleteSignOut(r.Context(), body.LogoutID)
+	if err != nil {
+		h.logger.Error(r.Context(), "Failed to complete sign-out", log.Error(err))
+		http.Error(w, "logout failed", http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
```
ThaminduDilshan: Shall we use constants from system package?

backend/internal/oauth/oauth2/logout/service_test.go
```diff
@@ -0,0 +1,303 @@
...
+func (suite *LogoutServiceTestSuite) SetupTest() {
+	suite.Require().NoError(config.InitializeServerRuntime(suite.T().TempDir(), &config.Config{}))
+}
+
+func (suite *LogoutServiceTestSuite) TearDownTest() {
+	config.ResetServerRuntime()
+}
+
+// fakeLogoutRequestStore is an in-memory logoutRequestStoreInterface for tests.
+type fakeLogoutRequestStore struct {
```
ThaminduDilshan: Should be able to use mockery to generate mocks

ThaminduDilshan (dismissed):

ThaminduDilshan (dismissed):

ThaminduDilshan (dismissed):

ThaminduDilshan (dismissed):

rajithacharith (approved):

Issue #2861, opened 2026-05-20
OAuth Token Exchange Allows Arbitrary audience Claim Injection


shashimalcse:
### Description

token exchange grant accepts the audience request parameter and copies it into the issued access token without validating it against a per-client allow-list or registered resource server. A client with the token-exchange grant can request an access token for an arbitrary audience causing AS to mint a valid JWT with requested aud.

### Steps to Reproduce

1. Get a valid subject token by completing a login flow.
2. Exchange it with an arbitrary audience:

```bash
curl -k "$BASE/oauth2/token" \
  -u "$CLIENT_ID:$CLIENT_SECRET" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data-urlencode "grant_type=urn:ietf:params:oauth:grant-type:token-exchange" \
  --data-urlencode "subject_token=$SUBJECT_TOKEN" \
  --data-urlencode "subject_token_type=urn:ietf:params:oauth:token-type:jwt" \
  --data-urlencode "audience=https://api.internal/admin"
```
3. Decode the returned `access_token`.

### Version

0.39.0

Sadeesha-Sath:
Decided to drop support for explicit audience claims

thiva-k:
Removing audience parameter support: https://github.com/thunder-id/thunderid/issues/4145

PR #3979, opened 2026-07-14
Add validation logic for token exchange grant audiences

Sadeesha-Sath:
### Purpose

Currently the Token Exchange grant merges all the explicit audience values from the requests with the resolved audiences from the scopes and resource parameter. This makes it possible to include arbitrary audience values and get valid tokens from them.

All the explicit audience values are validated against the resolved Audience values from the resource servers through scope lookup and the audience values are downscoped with the explicit audience parameter if multiple valid audiences exist.

#### 💥 Impact
_What will break? Who is affected?_
- Any integration that was passing an opaque/logical audience not backed by a registered Resource Server will start getting a 400 error where it previously got a 200
- A client relying on resource implicitly widening the aud list alongside an explicit audience will see a smaller aud claim than before, even though the request itself still succeeds.

#### 🔄 Migration Guide
_How should users update their code/configuration to adapt to the breaking changes? Include examples if helpful_
- Convert all the opaque/logical explicit audience values to be valid resource server identifiers. 
- Use the audience parameter to downscope.

### Approach

Removed the audience merging which was not validating and created a valid audience list from the valid resource servers. Then explicit audience claim is used to filter and downscope the available valid resource servers.

### Related Issues
- #2861 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened OAuth token-exchange `audience` handling to fail-closed: unregistered/invalid audiences are rejected with `invalid_target`, and no token is issued.
  * Updated issued token `aud` behavior when combining `audience` with `resource` to use validated/downscoped audiences only (no union), with deduplication and order preservation.

* **Tests**
  * Added and updated regression coverage for rejected unregistered audiences and the correct `aud` downscoping/self-audience behavior (including agent and integration flows).

* **Documentation**
  * Updated the Token Exchange guide with precise `audience` validation rules, `aud` computation, and `invalid_target` examples.

### DIFF
--- a/backend/internal/oauth/oauth2/granthandlers/token_exchange.go
+++ b/backend/internal/oauth/oauth2/granthandlers/token_exchange.go
@@ -207,9 +207,10 @@ func (h *tokenExchangeGrantHandler) HandleGrant(ctx context.Context, tokenReques
 	}
 
 	// Determine final audiences per RFC 8693 §2.1: audience and resource parameters may be
-	// combined. audience values are opaque logical names passed verbatim; resource values are
-	// RFC 8707 URIs that resolve to registered Resource Servers and participate in scope
-	// downscoping.
+	// combined. resource values are RFC 8707 URIs that resolve to registered Resource Servers and
+	// participate in scope downscoping. audience values must themselves correspond to a resolved
+	// Resource Server (or the client's own clientID) — an audience that matches neither is
+	// rejected rather than minted verbatim (see resolveAudiences).
 	resolvedRSes, resErr := resourceindicators.ResolveResourceServers(ctx, h.resourceService, tokenRequest.Resources)
 	if resErr != nil {
 		return nil, resErr
@@ -228,7 +229,10 @@ func (h *tokenExchangeGrantHandler) HandleGrant(ctx context.Context, tokenReques
 	if audErr != nil {
 		return nil, audErr
 	}
-	finalAudiences := mergeAudiences(tokenRequest.Audiences, rsAudiences, tokenRequest.ClientID)
+	finalAudiences, audFilterErr := resolveAudiences(tokenRequest.Audiences, rsAudiences, tokenRequest.ClientID)
+	if audFilterErr != nil {
+		return nil, audFilterErr
+	}
 
 	// Build access token using token builder
 	userSubConfig := oauthApp.UserAccessTokenConfig()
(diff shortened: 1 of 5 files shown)


backend/internal/oauth/oauth2/granthandlers/token_exchange.go
```diff
@@ -278,45 +282,47 @@ func (h *tokenExchangeGrantHandler) getScopes(
...
 	}
 
-	if len(merged) == 0 {
-		if clientID != "" {
-			return []string{clientID}
-		}
-		return []string{}
-	}
-	return merged
+	return filtered, nil
```
Yathusiga27, line 450: I think this return drops valid resource-server audiences whenever an explicit `audience` parameter is present.

At this point `filtered` contains only the validated values from `tokenRequest.Audiences`; it does not include `rsAudiences`, even when those came from a valid `resource` parameter. 

Could we either preserve the resolved `rsAudiences` in the final audience set after validating explicit audiences, or reject unsupported `audience` + `resource` combinations instead of silently dropping the resource target?

Sadeesha-Sath: In this implementation both resource and audience parameters are used to downscope. There are a few possible scenarios.

1. Both resource and audience parameters are not present - use `rsAudiences`
2. Only resource parameter is present - use `rsAudiences`
3. Only audience parameter is present - filter `rsAudiences` and return `filtered`
4. Either resource or audience value is invalid - return `Error`
5. Both resource and audience values are valid and more audiences are present than resources - return `Error`
6. Both resource and audience values are valid and more resources are present than audiences - (This scenario) - Send the intersection value `filtered`

I agree that behaviours of 5 and 6 are inconsistent. 

According to the spec, [RFC 8693](https://datatracker.ietf.org/doc/html/rfc8693#section-2.1),

> The audience and resource parameters may be used together to indicate multiple target services with a mix of logical names and resource URIs.

Both audience and resource parameters may be used together, and it discourages to send an error. Since it is important to have least privilege in the token, the intersection of both parameters need to be sent. 

So I would propose to change the behaviour of scenario 5 to send the intersection. 

Kindly give your inputs on this

thiva-k:
Let's remove the audience parameter support from token exchange request altogether for now. Since we have scope downscoping and aud parameter override support via the resource parameter now, having an audience parameter is redundant. We also cannot validate and downscope permissions based on the audience parameter based on our current RS model.

 Let's revisit this with the app-resource subscription story later if needed.

Sadeesha-Sath:
Closing due to #4009 Removing support for --audience field

PR #3997, opened 2026-07-14
Remove the delimiter from the hint text

### LINKED ISSUE #3328: Add resource UI doesn't allow colon (`:`) character to be added to the name
### Description

In the UI, it says the colon is allowed. However, the colon can't be typed. 

Image

### Steps to Reproduce

1. Create a resource server
2. Try to create a resource with a colon

### Version

0.44.0

_No response_

Sadeesha-Sath:
### Purpose

Previously the hint for adding a new resource or an action for a resource server included the delimiter character of the server as a valid character. Each resource server can have different delimiter characters.

### Approach

Replaced the relevant delimiter character of the resource server in the Dialogbox hint.

### Related Issues
- Fixes #3328 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved helper text shown for resource handles missing the required delimiter by removing redundant delimiter text from the translated hint for clearer guidance.


frontend/packages/configure-resource-servers/src/components/resource-tree/AddNodeDialog.tsx
```diff
@@ -277,7 +277,10 @@ export default function AddNodeDialog({
                     )
                   : t(
                       'resourceServers:tree.fields.handleHint',
-                      'Lowercase, alphanumeric, and . _ - : / characters. Cannot be changed after creation.',
```
KaveeshaPiumini, line 279: The fix is applied to the i18n fallback, not the translation, we might still observe this issue with a second language.

Sadeesha-Sath: Applied the fix to the whole i18n block

KaveeshaPiumini: This is definitely an improvement, but I'm still a bit concerned that we're modifying the translated string after it's has been resolved. This relies on the translated text having the expected format, which may not hold true across all locales. Would it be better to pass the character list as an interpolation parameter instead of modifying the translated text?

Sadeesha-Sath: Added as an interpolation parameter. Also fixed the failure message to use interpolation parameter as well

KaveeshaPiumini (approved):

ThaminduDilshan:
@Sadeesha-Sath shall we squash changes into a single commit? We generally keep a single commit per PR

Sadeesha-Sath:
Squashed to a single commit

PR #4006, opened 2026-07-15
Add support for importBehaviour Flag for serverconfig import yaml files

### LINKED ISSUE #4005: Importing a CORS config through /import endpoint replaces the current CORS allowed origins
### Description

Applying any declarative `server_config` bundle whose cors section sets `allowedOrigins` overwrites the entire writable CORS allow-list instead of adding to it. 

##### Developer Setup:

This breaks the Console and Gate app immediately in the dev build after using the "Try sample" welcome screen flow with the Wayfinder sample, since the samplt's own cors doc contains only its own origin, and running the config replaces the cors values for console and gate apps.

This also results in cors list resetting to console and gate only and losing the other cors origin values each time the bootstrap portion of the `make run` command executes.

##### Production:

When running production, importing any cors configs from samples using the /import endpoint will result on overwriting previously configured cors endpoints.

### Steps to Reproduce

1. Start via `make run` (this sets cors to `allowedOrigins: ["https://localhost:5190", "https://localhost:5191"]`).
2. From the Console welcome screen, click "Try sample" and import the Wayfinder sample bundle.
3. All subsequent Console to backend requests fail due to CORS errors.

### Version

0.48.0

_No response_

Sadeesha-Sath:
### Purpose

Importing a `server_config` bundle always replaced the writable layer wholesale via SetConfig. When two `YAML` bundles each declared allowed origins for the same section, the second import silently dropped every origin the first one added. There was also no way to add origins from an import without stomping on the writable layer. Since the Console's own dev-mode CORS origins live in that same DB row, any later import that includes a `cors` section, such as using the Wayfinder sample via the welcome screen's "Try sample" flow, replaces them, making the console to crash with cors errors, and need to run the bootstrap process again through `make run`. Also there is another issue of running `make run` resetting the CORS list to just the console and gate ports.

### Approach

- Added an optional `importBehavior` field to server-config declarative documents (`replace`, the default when omitted, or `merge`). Unknown values fail the import instead of silently falling back to replace.
- `serverconfig.SetConfig` takes a variadic `merge ...bool`, defaulting to `false` when omitted, so existing 3-arg call sites (the `PUT /server-config/{name}` handler, the passkey origin sync) keep compiling unchanged.
- A handler can now optionally implement `ComposeWritable(readOnly, existing, incoming any) any` to control what gets stored in the writable layer. `SetConfig` calls it whenever present, passing the current writable value only when `merge` is true, so a replace composes from the incoming value alone and a merge adds to what's already stored.
- Implemented `ComposeWritable` on the CORS `OriginHandler`: merges incoming origins with the existing writable set (when merging) and always excludes origins the declarative (read-only) layer already allows, so a stored origin can't outlive its declarative entry being removed. This dedup now runs on the replace path too, not just merge.
- The importer routes `importBehavior: merge` through `SetConfig(..., true)` and everything else through the default (`false`).
- Added `importBehavior: merge` to every shipped bundle that imports the CORS section (react-sdk-sample, react-api-based-sample, wayfinder app-native/redirect, the E2E config, and the generated dev bootstrap doc in `build.sh`/`build.ps1`), since those are exactly the imports that previously clobbered each other.
- Updated the deployment and import-resources docs to describe `importBehavior` and the declarative-origin dedup on `PUT /server-config/cors`.

### Related Issues
- Fixes #4005

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * CORS origins imported via server configuration now merge with existing writable origins rather than replacing them.
  * Importing CORS settings continues even when the existing CORS configuration can’t be retrieved or interpreted.
* **Tests**
  * Added test coverage for successful CORS merging and the fallback behavior when existing CORS data is unavailable.

### DIFF
--- a/backend/internal/serverconfig/service_test.go
+++ b/backend/internal/serverconfig/service_test.go
@@ -11,8 +11,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 
+	"github.com/thunder-id/thunderid/internal/system/cors"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/common"
 )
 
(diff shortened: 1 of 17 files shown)


backend/internal/system/importer/server_config_import.go
```diff
@@ -81,12 +83,32 @@ func (s *importService) importServerConfig(ctx context.Context, doc parsedDocume
...
+
 	if svcErr := s.serverConfigService.SetConfig(ctx, serverconfig.ConfigName(req.Name), value); svcErr != nil {
 		return serviceErrorOutcome(resourceTypeServerConfig, req.Name, req.Name, operationUpdate, svcErr)
 	}
 	return successOutcome(resourceTypeServerConfig, req.Name, req.Name, operationUpdate)
 }
 
+// Makes sure that the CORS origins are merged with the existing ones instead of overwriting them.
+// This is to avoid breaking existing clients that rely on the existing CORS origins.
+func (s *importService) mergeCORSOrigins(ctx context.Context, incoming json.RawMessage) json.RawMessage {
```
ThaminduDilshan: Method comments should start with method name: mergeCORSOrigins... as per the go conventions

backend/internal/system/importer/server_config_import.go
```diff
@@ -81,12 +83,32 @@ func (s *importService) importServerConfig(ctx context.Context, doc parsedDocume
...
+	existing, svcErr := s.serverConfigService.GetWritableConfig(ctx, string(serverconfig.ConfigNameCORS))
+	if svcErr != nil {
+		return incoming
```
ThaminduDilshan: If there's an error, this still overrides the existing cors configs right? Is this intentional?

backend/internal/system/importer/server_config_import.go
```diff
@@ -81,12 +83,32 @@ func (s *importService) importServerConfig(ctx context.Context, doc parsedDocume
...
+	}
+	decodedIncoming, err := cors.OriginHandler{}.Decode(incoming)
+	if err != nil {
+		return incoming
+	}
+	// Merge always returns an OriginConfig
+	merged, _ := json.Marshal(cors.OriginHandler{}.Merge(existing, decodedIncoming))
```
ThaminduDilshan: Shouldn't we handle json marshall errors here?

docs/content/guides/declarative-configurations/import-resources.mdx
```diff
@@ -180,6 +175,17 @@ Example response:
 
 When import runs with `dryRun=false`, <ProductName /> persists resource changes in runtime stores.
 
+A `server_config` document can set `merge` alongside `name` and `value`. For a section that holds a collection, such as the `cors` section's allowed origins, `merge: true` (the default) adds the incoming entries to what the writable layer already has instead of replacing it; `merge: false` replaces it wholesale, the way `PUT /server-config/{name}` does, which is how to remove an entry through import. `merge` has no effect on a section whose value isn't a collection: it is always replaced regardless of the field.
```
ThaminduDilshan: Term "document" feels off here. Shall we try to improve it?
Also writing style seems off compared to the existing content

cc: @himeshsiriwardana

backend/internal/serverconfig/service_test.go
```diff
@@ -288,6 +290,92 @@ func (suite *ServiceTestSuite) TestSetConfig_OK() {
...
+	assert.Nil(suite.T(), svcErr)
+	assert.Equal(suite.T(), readOnlyVal, handler.readOnly)
+	assert.Equal(suite.T(), writableVal, handler.existing)
+	assert.Equal(suite.T(), incomingVal, handler.incoming)
+}
+
+// TestImportConfig_RealCORSHandlerIsAdditive wires the real CORS handler through the service. The
+// composition is duck-typed, so a change to either side of the contract would otherwise revert imports
+// to replacing the writable layer without a compile error; only this test fails.
+func TestImportConfig_RealCORSHandlerIsAdditive(t *testing.T) {
```
ThaminduDilshan: Shall we move all tests to test suite?

PR #4088, opened 2026-07-17
Move Direct Auth Secret gating into the authn service

### LINKED ISSUE #4042: Move Direct Auth Secret gating from server middleware into the authn service
## Description

The Direct API endpoints (`/auth/**`, `/register/passkey/**`, `/access/**`) are currently gated by a secret key (`Direct-Auth-Secret` header) that is validated in the **system-wide security middleware**, not in the service that owns those APIs.

Today this lives in `internal/system/security`:

- Secret configured via `server.security.direct_auth_secret` and threaded through `main.go` → `security.Initialize()` ([backend/cmd/server/main.go:230-281](backend/cmd/server/main.go#L230-L281), [backend/internal/system/security/init.go:30-40](backend/internal/system/security/init.go#L30-L40)).
- Protected paths hard-coded in the security package ([backend/internal/system/security/permissions.go:26-30](backend/internal/system/security/permissions.go#L26-L30)).
- Validation (constant-time compare, secure-by-default) in the security service ([backend/internal/system/security/service.go:112-120](backend/internal/system/security/service.go#L112-L120)).

This couples authentication-specific access control to the generic server middleware. The path list and secret handling belong to the authn service that owns these endpoints ([backend/internal/authn/](backend/internal/authn/)).

## Proposed Change

Move the Direct Auth Secret gating out of `internal/system/security` and into the authn service (`internal/authn`) so the service that owns the endpoints owns their protection:

- Enforce the `Direct-Auth-Secret` check at the authn service/handler level (e.g. as an authn-scoped middleware or handler guard).
- Own the protected-path list within authn rather than in the generic security permissions.
- Preserve current behavior: secure-by-default (empty secret blocks the endpoints), constant-time comparison, and the `401` + RFC 6750 `WWW-Authenticate` response.
- Keep configuration source-compatible (or document any config key change).

## Rationale

- Keeps authentication access control colocated with the authn service instead of the shared server middleware.
- Removes authn-specific knowledge (path list, secret) from the generic `system/security` package.
- Makes the boundary and ownership clearer for future direct-API changes.

## Acceptance Criteria

- [ ] Direct Auth Secret validation is enforced within the authn service, not the system security middleware.
- [ ] `internal/system/security` no longer references the direct-auth secret or direct-auth path list.
- [ ] Behavior is unchanged: secure-by-default, constant-time compare, `401` with `WWW-Authenticate` header.
- [ ] Tests cover the moved enforcement.

Malith-19:
### Purpose

Move the Direct Auth Secret gating for the Direct API endpoints (`/auth/**`, `/register/passkey/**`, `/access/**`) out of the generic system security middleware and into the authn service that owns those endpoints, as described in #4042.

Previously the `Direct-Auth-Secret` check lived in `internal/system/security`, coupling authentication-specific access control to the shared server middleware. This PR relocates the path list and secret handling to the service that owns the endpoints, making ownership and the boundary clearer for future Direct API changes.

Behavior is unchanged: secure by default (an empty secret blocks the endpoints), constant-time comparison, and a `401` with the RFC 6750 `WWW-Authenticate: Bearer` challenge. The `direct_auth_secret` configuration key is unchanged (source-compatible).

### Approach

- Added a new `internal/authn/directauth` package with a `Guard` that owns the `Direct-Auth-Secret` header and its enforcement. `Guard.Wrap` wraps a handler and admits the request only when a secret is configured and the caller presents a matching value (constant-time compare); otherwise it writes the `401` + Bearer challenge.
- Wrapped the Direct API handlers at route registration:
  - `internal/authn/init.go` gates the `/auth/**` and `/register/passkey/**` `POST` handlers.
  - `internal/authzen/init.go` gates the `/access/v1/**` `POST` handlers (the public AuthZEN discovery endpoint stays open).
  - The guard is applied outside CORS so a rejected request is answered with a bare `401` and no CORS headers, matching the previous middleware short-circuit. CORS preflight (OPTIONS) handlers stay ungated.
- Threaded the configured secret from `runtime.Config.Server.SecurityConfig.DirectAuthSecret` into `authn.Initialize` and `authzen.Initialize`.
- Removed the direct-auth secret, header constant, path list, errors, and gate logic from `internal/system/security`. The three Direct API path prefixes remain in `publicPaths` so those endpoints stay unauthenticated at the middleware layer; their secret gating is now owned by the authn service.
- Corrected the `DirectAuthSecret` config doc comment to include `/access/**` and note the new ownership.

### Related Issues
- Fixes #4042

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Security Improvements**
  * Expanded Direct Auth protection to additional credential, OTP, OAuth, and passkey endpoints, plus AuthZEN access evaluation/search endpoints.
  * Unauthorized Direct Auth requests are denied with `401 Unauthorized` and no CORS headers.
  * CORS preflight (`OPTIONS`) remains available without Direct Auth.
* **Documentation**
  * Updated Direct Auth configuration notes to include `/access/**` endpoints.


backend/internal/authn/directauth/service.go
```diff
@@ -0,0 +1,62 @@
...
+	"crypto/subtle"
+	"net/http"
+
+	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
+	"github.com/thunder-id/thunderid/internal/system/error/apierror"
+	sysutils "github.com/thunder-id/thunderid/internal/system/utils"
+)
+
+// GuardServiceInterface enforces the Direct Auth Secret on Direct API endpoints.
+type GuardServiceInterface interface {
```
ThaminduDilshan: What if we keep this in authn root package rather than introducing a new sub package?

Malith-19: Good call, done. Moved the guard into the authn root package as internal/authn/direct_auth.go (NewDirectAuthGuard + DirectAuthGuardInterface) and dropped the sub-package.

One nuance: authzen also uses this guard for /access/**. To avoid authzen importing the whole authn package (nothing under internal/ depends on the authn root today), authzen accepts the guard through a small local interface that the authn guard satisfies structurally, and servicemanager wires one guard instance into both. Happy to switch authzen to import authn directly instead if you'd prefer that.

backend/internal/authn/direct_auth.go
```diff
@@ -0,0 +1,65 @@
...
+	Wrap(next http.HandlerFunc) http.HandlerFunc
+}
+
+// directAuthGuard is secure by default: an empty secret blocks every wrapped endpoint.
+type directAuthGuard struct {
+	secret string
+}
+
+// NewDirectAuthGuard creates the Direct Auth Secret guard.
+func NewDirectAuthGuard(secret string) DirectAuthGuardInterface {
```
ThaminduDilshan: This violates dependency injection pattern. We can return `DirectAuthGuardInterface` from Initialize method in addition to what it returns rt now

backend/internal/authzen/init.go
```diff
@@ -27,21 +27,29 @@ import (
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
+// DirectAuthGuard gates the Direct API access handlers with the Direct Auth Secret. It is satisfied
+// by the guard owned by the authn service.
+type DirectAuthGuard interface {
```
ThaminduDilshan:
```suggestion
type DirectAuthGuardProvider interface {
```

Just to align with pattern in pkg. Also if possible let's make this private

backend/internal/authn/direct_auth.go
Malith-19: Updated.

backend/internal/authn/directauth/service.go
Malith-19: Removed the new package and included inside the authn package.

backend/internal/authzen/init.go
Malith-19: Updated.

ThaminduDilshan (approved):

PR #4136, opened 2026-07-20
Fix invite page hanging on Add Another User

### LINKED ISSUE #4104: Invite user page hangs on loader when clicking "Add Another User"
### Description

On the Console invite flow, after inviting a user and generating the invite link, clicking Add Another User navigates to /console/users/add/invite and shows a loading spinner that never resolves. The first invite works correctly but the subsequent "Add Another User" hangs.

https://github.com/user-attachments/assets/7829e2d6-b804-4234-b177-57a6dec86045

### Steps to Reproduce

1. Start ThunderID (server + console on https://localhost:8090).
2. Log in to the Console as an admin.
3. Go to Users → Add User → Invite User.
4. Complete the invite and generate the email/invite link.
5. On the completion screen, click Add Another User.

The page navigates to /console/users/add/invite and stays on the loading animation indefinitely. It never renders the form.

### Version

1.0.0-alpha

OS: macOS
Database: default SQLite

Osara-B:
### Purpose
Fixes #4104 where the Invite User page hangs on an infinite loading spinner when the user clicks Add Another User after completing an invite.

The invite route auto-submits the onboarding-mode prompt (the "create vs invite" fork) so the user skips straight to inviting. That auto-submit is guarded by a one-shot `useRef`(`autoInviteTriggeredRef`) so it fires only once. Clicking "Add Another User" calls the SDK's `resetFlow()`, which restarts the flow without remounting `UserInvitePage`. Because the component is not remounted, the guard stays `true` from the first invite, so the restarted onboarding prompt is never auto-submitted and the page returns \<PageLoadingAnimation /> indefinitely.

This is purely a frontend state bug on the onboarding-flow path. The /flow/execute call on reset succeeds and returns the correct next step; the UI simply never acts on it. It does not involve the manual /users fallback.

### Approach

The fix ties the guard reset to the user action. The button already calls resetFlow() and onResetLocalState() together, so we wrap onResetLocalState to also clear the guard:
```
const handleReset = useCallback(() => {
  autoInviteTriggeredRef.current = false;
  onResetLocalState();
}, [onResetLocalState]);
```
and pass handleReset down to InviteUserStepContent. On restart the guard is false again, so the prompt is auto-submitted and the flow proceeds instead of stalling.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4104

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the “Add Another User” invitation flow so auto-submit can trigger again after the onboarding flow is reset.
  - Prevented the invitation process from stalling when flow components are cleared and restored.

- **Tests**
  - Added coverage verifying that invitations are submitted again after resetting the flow.

### DIFF
--- a/frontend/packages/configure-users/src/pages/UserInvitePage.tsx
+++ b/frontend/packages/configure-users/src/pages/UserInvitePage.tsx
@@ -810,6 +810,12 @@ function InviteUserFlowBridge({
     }
   }, [inviteAction, renderProps]);
 
+  // Clear the auto-invite guard on reset so the restarted prompt is auto-submitted again.
+  const handleReset = useCallback(() => {
+    autoInviteTriggeredRef.current = false;
+    onResetLocalState();
+  }, [onResetLocalState]);
+
   // Derive current step label from the HEADING_1 component
   const currentStepLabel = components?.length ? deriveStepLabel(components, resolve, t) : '';
 
@@ -847,7 +853,7 @@ function InviteUserFlowBridge({
       renderProps={renderProps}
       flowError={flowError}
       handleClose={handleClose}
-      onResetLocalState={onResetLocalState}
+      onResetLocalState={handleReset}
     />
   );
 }
--- a/frontend/packages/configure-users/src/pages/__tests__/UserInvitePage.test.tsx
+++ b/frontend/packages/configure-users/src/pages/__tests__/UserInvitePage.test.tsx
@@ -549,6 +549,35 @@ describe('UserInvitePage', () => {
       expect(mockResetFlow).toHaveBeenCalled();
     });
 
+    it('should re-trigger the auto-invite after "Add Another User" resets the flow', async () => {
+      const invitePrompt: EmbeddedFlowComponent[] = [
+        heading('How would you like to add a user?'),
+        block([submitAction('Invite user', {id: 'action_invite_user'})]),
+      ];
+
+      // Auto-invite fires once on arrival.
+      mockInviteUserRenderProps.components = invitePrompt;
+      const {rerender} = render(<UserInvitePage />);
+
+      await waitFor(() => {
+        expect(mockHandleSubmit).toHaveBeenCalledTimes(1);
+      });
+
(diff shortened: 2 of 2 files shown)


frontend/packages/configure-users/src/pages/UserInvitePage.tsx
```diff
@@ -803,6 +803,14 @@ function InviteUserFlowBridge({
   const autoInviteTriggeredRef = useRef(false);
   const inviteAction = findActionComponentById(components, ONBOARDING_MODE_INVITE_ACTION_ID);
 
+  // Clear the one-shot guard when the flow resets (components emptied) so the
+  // restarted onboarding prompt is auto-submitted again instead of stalling.
+  useEffect(() => {
+    if (!components?.length) {
+      autoInviteTriggeredRef.current = false;
+    }
+  }, [components]);
```
DonOmalVindula: Wdyt about resetting the ref when the user actually resets, instead of inferring it from components going empty? The Add Another User button already calls `resetFlow()` + `onResetLocalState()` together, and `onResetLocalState` passes through this component anyway, so we could just wrap it here:

```tsx
const handleResetLocalState = useCallback(() => {
  autoInviteTriggeredRef.current = false;
  onResetLocalState();
}, [onResetLocalState]);
```

and pass that down to `InviteUserStepContent`.

The components-empty state during reset is more of an SDK implementation detail, so I'd rather not depend on it. It also avoids the edge case where components empties transiently for some other reason (e.g. while a failing auto-submit is in flight) and the auto-submit re-fires in a loop.

Osara-B: Updated in [Reset onboarding auto invite guard on flow reset](https://github.com/thunder-id/thunderid/pull/4136/commits/a6ff33a226294f718a55336bacaedc3e783b2880)

DonOmalVindula (approved):

Issue #3871, opened 2026-07-09
Improve feature discoverability and overall UX of the flow builder


DonOmalVindula:
### Current Limitation

The flow builder in the ThunderID Console has a discoverability problem: most of its capabilities are hidden or buried deep inside the UI, making the builder hard to learn and slow to use.

The recurring patterns behind this:

- **Actions gated behind node selection** — configuring, renaming, and deleting elements are only reachable through the property panel, with no visible affordance on the canvas suggesting these actions exist.
- **Hover-reveal-only controls** — key interactions (e.g., connection handles, connection/edge actions, container menus) are invisible until the user happens to hover in the right place.
- **Missing common editor affordances** — several capabilities users expect from a visual editor have no UI surface at all, and there is no in-product guidance introducing the builder's interaction model.
- **Deep interaction paths** — frequent configuration tasks require several clicks through nested panels.

As a result, new users struggle to discover the builder's core mechanics, and experienced users are slowed down on routine tasks.

### Suggested Improvement

Audit the flow builder's UX with a focus on discoverability, and improve it so that core capabilities are visible, reachable with minimal interaction depth, and learnable without external documentation. A detailed pain-point inventory is being compiled with the team and will inform the concrete improvements, which can be tracked as follow-up tasks under this issue.

PR #4150, opened 2026-07-20
Lazy-load the ELK layout engine in the flow builder

DonOmalVindula:
### Purpose

Speed up opening the flow builder. Clicking a flow in the flows list showed a noticeable delay before the canvas appeared.

The builder page is code-split, so opening it downloads and parses a single large chunk before anything but the loading spinner renders. `applyAutoLayout` statically imported `elkjs` (the ELK layout engine, ~1.5MB) at module top level, so the entire engine was bundled into the builder's entry chunk and parsed on every open, even though auto-layout only runs on demand (the toolbar action, or opening a flow whose nodes have no stored positions). Saved flows carry node positions, so the common editing path never uses ELK at all.

### Approach

Dynamically import ELK on first layout and cache the instance, instead of importing it eagerly at module load. The static import becomes type-only. No behavior change: `applyAutoLayout` is already async, so the layout call simply awaits the engine before running.

Result (production build of the console app):

| | Before | After |
|---|---|---|
| Builder entry chunk (raw) | 2,399 KB | 985 KB |
| Builder entry chunk (gzipped) | 727 KB | 308 KB |
| ELK | in entry chunk | separate on-demand chunk |

ELK now downloads only when a layout is actually requested.

### Related Issues
- Refs #3871

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests (existing `applyAutoLayout` suite, 25 tests, still passing)
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Performance**
  * Improved flow editor loading by loading auto-layout functionality only when needed.
  * Preserved existing layout behavior and fallback handling when layout operations fail.

### DIFF
--- a/frontend/apps/console/src/features/flows/utils/applyAutoLayout.ts
+++ b/frontend/apps/console/src/features/flows/utils/applyAutoLayout.ts
@@ -17,7 +17,7 @@
  */
 
 import type {Edge, Node} from '@xyflow/react';
-import ELK from 'elkjs/lib/elk.bundled.js';
+import type {ELK} from 'elkjs/lib/elk-api';
 
 /**
  * Configuration options for auto-layout.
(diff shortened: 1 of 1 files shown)


DonOmalVindula:
Folded into #4153 (same file, combined as one change) together with the CodeRabbit retry fix for the cached dynamic import.

PR #4162, opened 2026-07-20
Fix Reset Bahaviour on All Edit Pages

### LINKED ISSUE #3895: Reset button on edit pages (Agent/Application) does not revert all unsaved changes
### Description

On the Application and Agent edit pages, clicking the Reset button does not discard the every unsaved changes as expected. The button is meant to roll back all unsaved edits and restore the form to its last-saved state. Currently this behavior work in partial manner: some edits remain in place, yet the Reset and Save buttons disappear. This leaves the UI in an inconsistent state where some unsaved changes are still shown but no longer look unsaved.

### Steps to Reproduce

1. Navigate to Applications (or Agents).
2. Select an application (or agent) to open its edit page.
3. Make change in redirect uri like remove some added uri, the Reset and Save buttons appear at the bottom.
4. Click the Reset button.

### Version

0.47.0

_No response_

### LINKED ISSUE #3953: [Bug] User attributes form does not update UI after saving without page refresh
### Description

When a user edits their attributes in the console and clicks save, the changes are saved successfully but the interface does not update automatically. The user is forced to refresh the page to see their newly updated information.

https://github.com/user-attachments/assets/ca277a33-864e-4992-934e-5998a0c50e37

### Steps to Reproduce

1. Log in to the console.
2. Try to edit and save user attribute.

### Version

v0.48.0

_No response_

Sadeesha-Sath:
### Purpose

Fixes inconsistent "unsaved changes" and Reset behavior across all Edit pages in the console. The unsaved changes bar and Save button stayed active even when a field was edited back to its original value, and Reset did not properly clear the populated form fields, because `hasChanges` was computed as `Object.keys(editedX).length > 0`, treating any touched key as a real change regardless of whether the value actually differed from the saved original. Empty-vs-null/undefined mismatches (e.g. clearing a field that was already unset) were also incorrectly flagged as changes. See #3895

Also fixes saved values not updating correctly on Save. This was due to a race condition on form remounting before data is refetched, due to `await refetch()` happening after the remounting. See #3953

Affected pages: Application, Agent, Group, Role, Organization Unit, Resource Server, and User edit pages, plus the resource-tree detail panel and the Agent Type / User Type view pages.

### Approach

* Added `isEqual`, `isPlainObject`, and `isEqualIgnoringEmpty` to `@thunderid/utils`, and `isEqualIgnoringEmpty` recursively normalizes `undefined`, `null`, `''`, `[]`, and `{}` to the same "empty" value at every level of nesting before deep-comparing, so a field cleared back to blank compares equal to one that was never set.
* Replaced each page's `hasChanges` derivation with a comparison of each edited field against its saved original via `isEqualIgnoringEmpty` (Application, Agent, Group, Role, Organization Unit, Resource Server, User edit pages, and the Agent Type / User Type view pages), so only real differences count as unsaved changes.
* `ResourceServerEditPage` also had its `handleFieldChange` simplified to an unconditional setter and its description `onBlur`/`onKeyDown` handlers simplified to call it directly, now that `hasChanges` handles the equality check centrally instead of each field guarding itself.
* `OrganizationUnitEditPage` had its duplicated description `onBlur`/`onKeyDown` equality checks consolidated into one shared `commitDescription` helper, so the inline-edit-on-blur field stays deduplicated while the centralized `hasChanges` reconciles the rest.
* The resource-tree `ResourceDetailPanel` replaced its one-way `dirty` touched-flag with a `baseline` snapshot (advanced on mount, Save, and Discard) that each field is compared against on every render, so reverting a field by hand clears the bar.
* Moved each edit page's `await refetch()` call to happen before bumping its remount key, fixing the race where the child section remounted with stale data before the refetch resolved.
* Standardized the remount-key naming to `sectionResetKey` across pages (`AgentEditPage` previously called it `attributesResetKey`), bumped on both Save (after `refetch()` resolves) and Reset, and passed down to child sections so they remount and drop local component state that `setEditedX({})` alone wouldn't clear.
* Added `@thunderid/utils` as a dependency of `configure-agent-types`, the only affected package manifest that didn't already declare it.
* Added `.coderabbit.yaml` review rules that flag Edit pages with stateful child sections missing the `sectionResetKey` prop, and child sections with local state that don't accept/use it, so future PRs follow the same pattern.

### Related Issues
- Fixes #3895 
- Fixes #3953 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Unsaved-changes indicators now hide only when edits truly match saved data, using empty-insensitive comparisons across edit flows (agents, applications, users, groups, roles, organization units, resource servers, and type/configure views).
  * Save/Reset now more reliably clears stale section state by coordinating section resets/remounts (including token validity, access redirects, URLs, attributes, and MCP sections) at the right time.
  * Failed saves preserve in-progress edits and keep the unsaved-changes UI visible.
* **Tests**
  * Expanded revert/reset and “section reset” coverage to verify correct bar visibility plus remount vs in-place reset behavior.


Sadeesha-Sath:
@coderabbitai review

frontend/apps/console/src/features/applications/components/edit-application/token-settings/__tests__/EditTokenSettings.test.tsx
```diff
@@ -110,18 +112,64 @@ vi.mock('../ScopeSection', () => ({
 // TokenValidationSection is called with tokenType="oauth" in OAuth mode and
 // tokenType="shared" in native mode. The mock splits "oauth" into separate
 // access, id, and refresh testids to match test expectations.
+//
+// The mock also carries its own local "which sub-tab is selected" state (mirroring the real
+// component's independent activeValidationTab) plus a real react-hook-form Controller bound to
+// the refreshTokenValidity/validityPeriod field it's given — this lets reset-behavior tests prove
+// both halves of the bug fix: the field value reverts on a sectionResetKey bump, AND this local
```
Dilusha-Madushan: Please avoid using phrases like "bug fixes..." in code comments. Comments should describe the code itself rather than the intent of the PR. Once the PR is merged, those comments become verbose. Check other comments as well.

Dilusha-Madushan (approved):

frontend/packages/utils/src/object/isEqualIgnoringEmpty.ts
```diff
@@ -0,0 +1,57 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import isEqual from 'lodash-es/isEqual';
+import isPlainObject from 'lodash-es/isPlainObject';
```
brionmario: We are trying to get rid of `lodash` usages and maintain the utils locally. https://github.com/thunder-id/thunderid/issues/1951

Try to keep the `isEqual` & `isPlainObject` in `thunderid/utils -> object`

frontend/apps/console/src/features/applications/components/edit-application/customization-settings/EditCustomizationSettings.tsx
```diff
@@ -40,6 +40,11 @@ interface EditCustomizationSettingsProps {
    * @param value - The new value for the field
    */
   onFieldChange: (field: keyof Application, value: unknown) => void;
+  /**
+   * Bumped by the parent on Save/Reset to force UrlsSection to remount and drop its stale
+   * react-hook-form defaults.
+   */
+  sectionResetKey?: number;
```
brionmario, line 47: Any way we can mantate this?

If not, lets add a CodeRabbit instruction to validate this for any new implementations.

frontend/packages/configure-resource-servers/src/pages/ResourceServerEditPage.tsx
Dilusha-Madushan: Other edit pages we label the changes rollback button as 'Reset' but in this page we show it as 'Discard'. Better if we keep same label for same action in all edit pages. Check this in other edit pages as well.

cc: @brionmario

Dilusha-Madushan (changes requested):

frontend/packages/configure-resource-servers/src/pages/ResourceServerEditPage.tsx
Sadeesha-Sath: Since this PR changes mostly the behaviour and logic, and this 'Discard' label is there in the connection module as well, shall we open a new issue to track this effort and address that with another PR?

Dilusha-Madushan: sure, handle that issue in separate PR

frontend/packages/utils/src/object/isEqualIgnoringEmpty.ts
Sadeesha-Sath: Removed `lodash` and implemented utils locally

frontend/apps/console/src/features/applications/components/edit-application/customization-settings/EditCustomizationSettings.tsx
Sadeesha-Sath: Added coderabbit instructions.

frontend/packages/configure-resource-servers/src/pages/ResourceServerEditPage.tsx
Sadeesha-Sath: Created a new issue to track this #4243

Sadeesha-Sath: Created a new PR for this #4249

JayaShakthi97:
The description says this adds `lodash-es` as a dependency. Hope it is not. Please verify.

Sadeesha-Sath:
> The description says this adds `lodash-es` as a dependency. Hope it is not. Please verify.

It was removed. Forgot to update the description. I'll update

JayaShakthi97 (dismissed):

Issue #4128, opened 2026-07-20
[SCIM] Core User Schema Attribute Mapping (Phase 2)


ravindu439:
### What problem are we solving?

When SCIM clients (IdP, provisioning tools) want to manage ThunderID users via full core user schema (RFC 7643), they hit limited attribute support from Phase 1 CRUD implementation, which causes incomplete interoperability — clients can't map/sync all standard SCIM core attributes to ThunderID user resource fields.

### Who are we solving this for?

Primary: IdP admin / provisioning integrator
- Goal: provision and sync users via SCIM with full core attribute fidelity
- Pain: Phase 1 only covers basic CRUD, no proper core schema attribute mapping between SCIM resource and ThunderID user model

Secondary: ThunderID backend maintainers
- Goal: consistent, spec-compliant SCIM resource/schema layer
- Pain: without mapping layer, schema and resource handling diverge from RFC 7643

### Why should we solve this now?

Why now:
- Value to Persona: unblocks full SCIM client compatibility, not just partial CRUD
- Value to Org: RFC 7643 compliance required for enterprise IdP integrations; builds directly on Phase 1 ( #3743 ) — natural next increment while SCIM context fresh

### Proposed Solution

Implement Phase 2: core user attribute mapping.
- Add core attribute mapper (`core_attr_mapper.go`) translating SCIM core schema attributes to/from ThunderID user resource fields
- Add schema builder (`schema_builder.go`) for SCIM schema representation
- Update `users_model.go`, `users_resource.go`, `users_service.go` to use mapping layer
- Cover with unit tests (`core_attr_mapper_test.go`)
- Depends on / extends Phase 1 SCIM Users CRUD

PR #4184, opened 2026-07-21
feat(scim): Add core attribute mapping for SCIM Users

ravindu439:
### Purpose
Adds core attribute mapping between the ThunderID user schema and SCIM core attributes. This enables translating ThunderID's flat user attributes (e.g. `username`, `email`, `given_name`) into SCIM core schema fields (`userName`, `emails`, `name.givenName`, etc.) and back, so SCIM user resources correctly expose and accept core attributes regardless of how the underlying user type schema names them.

### Approach
- Added `core_attr_mapper.go` with a pre-configured `coreAttrRules` table — one entry per ThunderID library attribute candidate, matched case-insensitively.
- Each rule declares an `attrKind`:
  - `kindSimpleString` — direct 1:1 fields (`userName`, `displayName`, `title`, `locale`, etc.)
  - `kindMultiComplex` — array-valued fields with a default type/value (`emails`, `phoneNumbers`, `photos`)
  - `kindSubAttr` — nested sub-attributes of a complex parent object (e.g. `name.givenName`, `name.familyName`)
  - `kindAddrPart` — sub-attributes rolled into the single `addresses[0]` element
- `mapToCoreAttrs` builds the outbound SCIM JSON representation from ThunderID user attributes.
- `reverseMapCoreAttrsForSchema` translates SCIM attribute paths back to ThunderID attribute names against a given user type's schema, so the mapping stays correct across differently-configured user types.
- Updated `users_resource.go` to use the new mapper when constructing SCIM user resources.

### Related Issues
- #4128 

### Related PRs
- #3971 

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/scim/group_handler.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	if err := json.Unmarshal(body, &payload); err != nil || payload.DisplayName == "" {
+		h.handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
+	hasGroupSchema := false
+	for _, s := range payload.Schemas {
+		if s == SCIMCoreGroupSchemaURN {
+			hasGroupSchema = true
+			break
```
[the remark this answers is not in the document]
ravindu439: fixed it

backend/internal/scim/schema_builder.go
```diff
@@ -72,7 +70,7 @@ func mapRawPropertyToSCIMAttribute(name string, def rawPropertyDef) SCIMSchemaAt
 		Name:        name,
 		Description: def.DisplayName,
 		Required:    def.Required,
-		CaseExact:   false,
+		CaseExact:   true,
```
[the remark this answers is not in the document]
ravindu439: In thunder natively it consider as case exact for all attributes ,so keeping it true is right way in thunder.

backend/internal/scim/init.go
```diff
@@ -45,21 +47,23 @@ func Initialize(
 	uSvc := newSCIMUsersService(userService, entityTypeService)
 	uh := newSCIMUsersHandler(uSvc, cfg.PublicURL)
 
-	registerRoutes(mux, h, uh)
+	gSvc := newSCIMGroupsService(groupService)
+	gh := newSCIMGroupsHandler(gSvc, cfg.PublicURL)
+	registerRoutes(mux, h, uh, gh)
```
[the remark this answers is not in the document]
ravindu439: Groups in separate PR , this PR is on top of that commit.

backend/internal/scim/core_attr_mapper.go
```diff
@@ -0,0 +1,546 @@
...
+		scimField: scimFieldWebsite,
+		kind:      kindSimpleString,
+	},
+	{
+		candidate: "title",
+		scimField: scimFieldTitle,
+		kind:      kindSimpleString,
+	},
+	{
+		candidate: "employee_number",
```
sadilchamishka: `employee_number`, `department`, `division`, `organization`, `cost_center`, and `manager` are **Enterprise User Extension** attributes (`urn:ietf:params:scim:schemas:extension:enterprise:2.0:User`), not core `User` attributes (RFC 7643 §4.3). Emitting them as top-level fields of the User resource is not SCIM-compliant, so strict clients will ignore or reject them. They should live under the enterprise extension URN object, or be dropped from this PR. Also note `manager` is a **complex** type (`{value, $ref, displayName}`) in the enterprise schema, not a simple string as modeled here (`kindSimpleString`).

backend/internal/scim/core_attr_mapper.go
```diff
@@ -0,0 +1,546 @@
...
+		subAttr:     "formatted",
+	},
+	{
+		candidate:   "middle_name",
+		kind:        kindSubAttr,
+		parentField: scimFieldName,
+		subAttr:     "middleName",
+	},
+	{
+		candidate: "nickname",
```
sadilchamishka, line 98: The rule table emits top-level `nickName`, `photos`, `locale`, `profileUrl`, and `website` (plus the enterprise fields), but `coreUserAttributes()` in `schema_builder.go` doesn't declare any of them. As a result these appear on resources but aren't described by `/Schemas`. The emit set here and the advertised schema should match exactly.

backend/internal/scim/schema_builder.go
```diff
@@ -178,6 +176,368 @@ func coreUserAttributes() []SCIMSchemaAttribute {
...
+			Type:        scimAttrTypeString,
+			Description: "The user's title, such as \"Vice President.\"",
+			Required:    false,
+			CaseExact:   false,
+			Mutability:  scimMutabilityReadWrite,
+			Returned:    scimReturnedDefault,
+			Uniqueness:  scimUniquenessNone,
+		},
+		{
+			Name: "userType",
```
sadilchamishka: `userType` is advertised in the User schema here, but there is no corresponding entry in `coreAttrRules`, so it's never populated by `mapToCoreAttrs`. Either add a mapping rule or drop it from the advertised schema (see the related mismatch in `core_attr_mapper.go`).

backend/internal/scim/schema_builder.go
sadilchamishka, line 73: This flips `CaseExact` from `false` to `true` for every extension string attribute. Per RFC 7643 the default is `false`, and `caseExact` governs value comparison/filtering semantics (not attribute-name matching). This looks unrelated to core-attribute mapping and possibly unintended. If it's deliberate, a comment explaining why would help; otherwise it should be reverted.

backend/internal/scim/users_service.go
```diff
@@ -114,6 +115,15 @@ func (s *scimUsersService) CreateUser(
 			log.String("userTypeName", canonicalName), log.Any("error", svcErr))
 		return nil, &ErrorUnknownUserType
 	}
+
+	if len(payload.CoreAttrs) > 0 {
+		reverseMapped := reverseMapCoreAttrsForSchema(payload.CoreAttrs, et.Schema)
```
sadilchamishka, line 120: Minor: reverse-mapped core values only fill keys not already in `ExtensionAttrs` (`if _, exists := ...; !exists`). If a client sends both a core field (e.g. `userName`) and its extension counterpart (`username`) with different values, the core value is silently dropped. Reasonable precedence, but worth a short comment documenting it. (Safe from a nil-map panic only because `ValidateSCIMUserRequest` guarantees a non-nil `ExtensionAttrs`.)

sadilchamishka (commented):
Core attribute mapping looks clean and is well tested. A few correctness/spec points worth addressing before merge (inline). Overall the mapper design (declarative rule table + outbound/reverse) is nice. Note: the diff also bundles unrelated SCIM Group / permission / i18n changes, which suggests the branch needs a rebase on the current `feature/scim-support` tip so the diff reflects only this PR.

backend/internal/scim/core_attr_mapper.go
ravindu439: Agreed, dropped them. `employee_number/department/division/organization/cost_center/manager` removed from `coreAttrRules` entirely rather than relocated under the enterprise extension URN, keeping this PR scoped to core User attributes only. manager's complex-type shape `({value, $ref, displayName})` is exactly the kind of thing that'd need its own extension-schema modeling work, better done as a separate follow-up if there's a need for it.

backend/internal/scim/core_attr_mapper.go
ravindu439: fixed

backend/internal/scim/schema_builder.go
ravindu439: Since we handle `userType` in a separate way we don't need to populate the `userType` in the SCIM payload for now.

backend/internal/scim/schema_builder.go
ravindu439: In thunder natively it consider as case exact for all attributes ,so keeping it true is right way in thunder.

backend/internal/scim/users_service.go
ravindu439: Kept extension-wins precedence,
Reasoning: reverseMapCoreAttrsForSchema (core → native) is lossy — it round-trips through multiComplex normalization, subAttr merging, addrPart array collapsing — so the extension value is the raw, native-shaped value the client sent directly, while the core value survived a reverse-engineering step to reach the same shape. Also, on read buildSCIMUserResource always derives CoreAttrs from the stored native attributes (mapToCoreAttrs), so native storage is already the system's source of truth architecturally; this just keeps the write path consistent with that. Confirmed ExtensionAttrs non-nil guarantee via ValidateSCIMUserRequest too, noted in the comment.

sadilchamishka (approved):

Issue #4129, opened 2026-07-20
[SCIM] Implement filter query support for Users endpoint (RFC 7644)


ravindu439:
### What problem are we solving?

When SCIM clients (IdP, provisioning tools) want to query ThunderID users with the `filter` query parameter, they get rejected entirely — ThunderID's SCIM Users list endpoint has no tracked filter support: filtering was never scoped or implemented as a feature, which causes clients to fall back to fetching the full user list and filtering client-side, or to fail integration where `filter` is mandatory.

### Who are we solving this for?

Primary: IdP admin / provisioning integrator
- Goal: query/filter ThunderID users using standard SCIM filter syntax (e.g. `userName eq "john"`, `userName sw "j"`, `active eq true and userType eq "Employee"`)
- Pain: no filter capability tracked or designed for ThunderID's SCIM Users endpoint

### Why should we solve this now?

Why now:
- Value to Persona: `filter` is a standard, frequently-required SCIM query capability; without it many IdP provisioning flows can't integrate
- Value to Org: RFC 7644 compliance gap on a core SCIM capability; scoping it properly now avoids re-architecture later once partial/ad hoc support exists in more places

### Proposed Solution

Design and implement SCIM filter query support for the Users endpoint from the ground up:
- Define filter expression grammar per RFC 7644 §3.4.2.2 ABNF
- Comparison operators: eq, ne, co, sw, ew, pr, gt, lt, ge, le
- Logical operators: and, or, not
- Grouping via `()`
- Complex attribute filters via `[]` (e.g. `emails[type eq "work" and value co "@example.com"]`)
- Parser (tokenizer + recursive-descent or equivalent) producing a filter expression tree
- Attribute path resolution against SCIM core/extension schema (including URN-prefixed sub-attributes)
- Integration with `ListUsers` service and underlying store query layer to evaluate the filter tree
- Error handling per RFC 7644 (`invalidFilter` SCIM error response) for malformed/unsupported expressions
- Test coverage across operators, logical combinations, grouping, and complex attribute filters

### Alternatives

- Client-side filtering only, no server `filter` support (rejected: standard SCIM clients expect server-side filtering; poor scalability for large user sets)

PR #4185, opened 2026-07-21
feat(scim): Add filter query (eq) support for SCIM Users

ravindu439:
### Purpose
Adds SCIM filter query support (`eq` operator) for the SCIM Users list/search endpoint. Enables clients to filter users by core attributes, e.g. `userName eq "john"`, `emails.value eq "john@example.com"`, using ThunderID's underlying flat user attributes for the actual comparison.

### Approach
- `users_handler.go`: parses the SCIM `filter` query parameter on list requests, restricted to a single `eq` expression against a supported attribute path.
- `core_attr_mapper.go`: added `scimToThunderAttrIndex`, a lowercase-keyed reverse lookup built once at package init from `coreAttrRules`, mapping SCIM filter attribute paths (e.g. `username`, `emails.value`, `name.givenname`) to their ThunderID attribute names in O(1). `translateSCIMFilterAttr` resolves a filter attribute through this index; attributes with no core mapping (e.g. `id`, `active`, custom schema attrs) pass through unchanged.
- Multi-valued sub-attributes ThunderID doesn't store as a flat field (`type`, `primary`) are explicitly unsupported for filtering — `scimUnsupportedFilterAttrs` returns an error for paths like `emails.type` instead of silently matching nothing, since no flat ThunderID attribute exists to compare against.
- `users_service.go` wires the translated attribute/value into the existing user query path.

### Related Issues
- #4129 

### Related PRs
- #4184 

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/internal/scim/config/config.go
+++ b/backend/internal/scim/config/config.go
@@ -31,7 +31,7 @@ import (
 const (
 	// PatchSupported indicates that the SCIM PATCH operation is supported
 	// per RFC 7644 §3.5.2.
-	PatchSupported = true
+	PatchSupported = false
 
 	// BulkSupported indicates that SCIM Bulk operations are not yet
 	// implemented per RFC 7644 §3.7.
@@ -47,7 +47,7 @@ const (
 
 	// FilterSupported indicates whether SCIM filtering is supported
 	// per RFC 7644 §3.4.2.2.
-	FilterSupported = false
+	FilterSupported = true
 
 	// FilterMaxResults caps the number of resources returned in a single
 	// filtered query, guarding against excessively large result sets.
--- a/backend/internal/scim/users_handler.go
+++ b/backend/internal/scim/users_handler.go
@@ -4,13 +4,22 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strconv"
+	"strings"
 
+	"github.com/thunder-id/thunderid/internal/system/database/utils"
 	"github.com/thunder-id/thunderid/internal/system/log"
 	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
 )
 
 const loggerComponentName = "scim"
+
+var (
+	scimFilterQuotedStringRe = regexp.MustCompile(`"([^"\\]|\\.)*"`)
+	scimFilterEqRe           = regexp.MustCompile(`(?i)^((?:[A-Za-z0-9][A-Za-z0-9.\-_]*:)*)` +
+		`([A-Za-z][A-Za-z0-9.\-_]*)\s+eq\s+(.+)$`)
+)
 
 // scimUsersHandler handles all /scim/v2/Users HTTP requests.
 type scimUsersHandler struct {
@@ -28,10 +37,20 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
 	ctx := r.Context()
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
 
-	// ?filter is not supported in this implementation.
-	if r.URL.Query().Get("filter") != "" {
-		h.handleSCIMError(w, r, &ErrorFilterNotSupported)
-		return
+	// Parse optional SCIM filter — only single "eq" expressions are supported.
+	var parsedFilters map[string]interface{}
+	if filterStr := r.URL.Query().Get("filter"); filterStr != "" {
+		var err error
+		parsedFilters, err = parseSCIMFilterForEq(filterStr)
+		if err != nil {
+			writeSCIMErrorResponse(ctx, w, http.StatusBadRequest, SCIMErrorResponse{
+				Schemas:  []string{SCIMErrorSchemaURN},
+				Status:   "400",
+				ScimType: "invalidFilter",
+				Detail:   err.Error(),
+			})
+			return
+		}
 	}
 	startIndex := 1
 	count := 20
@@ -45,7 +64,7 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
 			count = n
 		}
 	}
-	listResp, svcErr := h.svc.ListUsers(ctx, startIndex, count, h.baseURL)
+	listResp, svcErr := h.svc.ListUsers(ctx, startIndex, count, parsedFilters, h.baseURL)
 	if svcErr != nil {
 		h.handleSCIMError(w, r, svcErr)
 		return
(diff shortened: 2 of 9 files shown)


backend/internal/scim/users_handler.go
```diff
@@ -185,3 +196,87 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+	return map[string]interface{}{attribute: value}, nil
+}
+
+// parseSCIMCompValue converts a raw SCIM compValue token into a typed Go value.
+// compValue = false / null / true / number / string  (RFC 7159 JSON rules)
+func parseSCIMCompValue(raw string) (interface{}, error) {
+	// Quoted string — strip the surrounding double-quotes.
+	if len(raw) >= 2 && raw[0] == '"' && raw[len(raw)-1] == '"' {
+		return raw[1 : len(raw)-1], nil
+	}
```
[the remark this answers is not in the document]

backend/internal/scim/schema_builder.go
```diff
@@ -72,7 +70,7 @@ func mapRawPropertyToSCIMAttribute(name string, def rawPropertyDef) SCIMSchemaAt
 		Name:        name,
 		Description: def.DisplayName,
 		Required:    def.Required,
-		CaseExact:   false,
+		CaseExact:   true,
```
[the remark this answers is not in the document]
ravindu439: In thunder natively it consider as case exact for all attributes ,so keeping it true is right way in thunder.

backend/internal/scim/init.go
```diff
@@ -157,13 +161,48 @@ func registerRoutes(mux *http.ServeMux, h *scimHandler, uh *scimUsersHandler) {
 		optsCRUD,
 	))
 
+	mux.HandleFunc(middleware.WithCORS(
+		"GET "+SCIMBasePath+"/Groups",
+		gh.HandleGroupsListRequest,
+		optsCRUD,
+	))
```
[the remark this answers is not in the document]
ravindu439: this branch only contain the filter implementations , those implementations are on top of previous Groups implementations.

backend/internal/scim/group_service.go
```diff
@@ -0,0 +1,365 @@
...
+	listResp, svcErr := s.groupService.GetGroupList(ctx, count, offset, true)
+	if svcErr != nil {
+		return SCIMGroupListResponse{}, mapGroupServiceErrorToSCIM(svcErr)
+	}
+	scimGroups := make([]SCIMGroup, 0, len(listResp.Groups))
+	for _, g := range listResp.Groups {
+		// GroupBasic has no Members; fetch full group for members
+		full, svcErr := s.groupService.GetGroup(ctx, g.ID, true)
+		if svcErr != nil {
+			return SCIMGroupListResponse{}, mapGroupServiceErrorToSCIM(svcErr)
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/users_handler.go
```diff
@@ -185,3 +196,87 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+			// Extract the actual operator token for the error message.
+			return nil, fmt.Errorf(
+				"the specified filter operator is not supported; only 'eq' is supported",
+			)
+		}
+	}
+	// Match: [optional-URN-prefix:]attrPath eq compValue
+	// attrPath allows alphanumeric, underscore, hyphen, and dot (for sub-attributes).
+	// compValue is a quoted string, a boolean literal, or a number.
+	pattern := `^((?:[A-Za-z][A-Za-z0-9\-]*:)*)?([A-Za-z][A-Za-z0-9_\-\.]*)\s+eq\s+(.+)$`
```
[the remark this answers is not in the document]

backend/internal/scim/users_handler.go
```diff
@@ -185,3 +196,87 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+	lower := strings.ToLower(filterStr)
+	// Reject all compound/complex expressions up front.
+	if strings.Contains(lower, " and ") ||
+		strings.Contains(lower, " or ") ||
+		strings.HasPrefix(lower, "not ") ||
+		strings.ContainsAny(filterStr, "()[]") {
+		return nil, fmt.Errorf(
+			"compound filter expressions are not supported; only a single 'eq' expression is supported",
+		)
+	}
```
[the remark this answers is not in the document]

backend/internal/scim/config/config.go
```diff
@@ -31,7 +31,7 @@ import (
 const (
 	// PatchSupported indicates that the SCIM PATCH operation is supported
 	// per RFC 7644 §3.5.2.
-	PatchSupported = true
+	PatchSupported = false
```
sadilchamishka, line 34: This flips `PatchSupported` from `true` to `false`, but PATCH is implemented (routes registered at `init.go:185`/`208`, `group_service.PatchGroup`, PATCH handling in `handler.go`). This advertises PATCH as unsupported in ServiceProviderConfig while the endpoints still work, and it's unrelated to filter support. Looks like an accidental merge artifact — please revert or split it out.

backend/internal/scim/users_handler.go
```diff
@@ -184,3 +196,92 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+			// Extract the actual operator token for the error message.
+			return nil, fmt.Errorf(
+				"the specified filter operator is not supported; only 'eq' is supported",
+			)
+		}
+	}
+	// Match: [optional-URN-prefix:]attrPath eq compValue
+	// attrPath allows alphanumeric, underscore, hyphen, and dot (for sub-attributes).
+	// compValue is a quoted string, a boolean literal, or a number.
+	pattern := `^((?:[A-Za-z][A-Za-z0-9.\-_]*:)*)?([A-Za-z][A-Za-z0-9_.\-_]*)\s+eq\s+(.+)$`
```
sadilchamishka: The attr-path regex allows `-` (and RFC 7643 permits `-` in SCIM attribute names), but `utils.ValidateKey` at the store layer rejects `-`. A hyphenated unmapped/custom attribute passes here, then fails deep in the store and surfaces as a 500 rather than a 400 `invalidFilter`. Consider validating the translated key against the same charset here to return a clean 400.

backend/internal/scim/users_handler.go
```diff
@@ -184,3 +196,92 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+	re := regexp.MustCompile(`(?i)` + pattern)
+	matches := re.FindStringSubmatch(filterStr)
+	if len(matches) == 0 {
+		return nil, fmt.Errorf(
+			"invalid filter expression; expected format: 'attrPath eq value'",
+		)
+	}
+	// matches[1] = optional URN prefix (e.g. "urn:thunderid:params:scim:schemas:employee_hier:2.0:User:")
```
sadilchamishka: This comment advertises support for URN-prefixed paths like `...:employee_hier:2.0:User:profile.manager.id`, but the URN-prefix group requires each segment to start with `[A-Za-z]`, so the numeric `2.0` segment breaks the match and the whole filter is rejected as invalid. Fix the regex or correct the comment.

backend/internal/scim/users_handler.go
```diff
@@ -184,3 +196,92 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
+// parseSCIMFilterForEq parses a SCIM filter string that contains exactly one
+// "eq" comparison and no logical operators, grouping, or square brackets.
+// Returns a native filter map suitable for userService.GetUserList, or an
+// error if the expression uses any unsupported syntax.
+func parseSCIMFilterForEq(filterStr string) (map[string]interface{}, error) {
+	filterStr = strings.TrimSpace(filterStr)
+	if filterStr == "" {
+		return nil, nil
+	}
+	sanitized := regexp.MustCompile(`"([^"\\]|\\.)*"`).ReplaceAllString(filterStr, `""`)
```
sadilchamishka: `regexp.MustCompile` runs here (and again below for the eq pattern) on every request. Hoist both to package-level vars compiled once.

backend/internal/scim/schema_builder.go
```diff
@@ -625,6 +609,22 @@ func coreGroupAttributes() []SCIMSchemaAttribute {
 	}
 }
 
+// buildCoreGroupSchema returns the static SCIM Core Group schema (RFC 7643 §4.2).
+func buildCoreGroupSchema(baseURL string) SCIMSchema {
```
sadilchamishka: `buildCoreGroupSchema` is only moved below `coreGroupAttributes` with no functional change. Unrelated churn — please drop it to keep the diff focused on the filter feature.

sadilchamishka (commented):
Reviewed the SCIM `eq` filter support. The core approach is solid: the O(1) reverse index, explicit rejection of unsupported multi-valued sub-attrs (`emails.type`/`emails.primary`), and good unit coverage on translation. SQL injection is mitigated downstream via `utils.ValidateKey` + parameterized args. A few things to address before merge:

**1. Unrelated `PatchSupported` flip (blocking).** `config.go` changes `PatchSupported = true → false`. PATCH is implemented (routes at `init.go:185`/`208`, `group_service.PatchGroup`, handler PATCH handling). This flips an advertised capability in ServiceProviderConfig while the routes keep working, and it's out of scope for a filter PR. Looks like an accidental merge/rebase artifact. Please drop it (or split into its own PR with justification).

**2. Hyphenated attribute paths → 500 instead of 400.** The handler regex allows `-` in the attr path, and RFC 7643 permits `-` in SCIM attribute names, but `utils.ValidateKey` at the store layer rejects `-`. A syntactically valid hyphenated (unmapped/custom) attribute passes handler parsing, then fails deep in the store and surfaces as a 500 InternalServerError rather than a 400 `invalidFilter`. Consider validating the translated key against the same charset in the handler so it returns a clean 400.

**3. Misleading URN-prefix comment / non-working case.** The comment in `parseSCIMFilterForEq` documents support for `urn:thunderid:params:scim:schemas:employee_hier:2.0:User:profile.manager.id`, but the URN-prefix group requires each segment to start with `[A-Za-z]`, so the numeric version segment (`2.0`) breaks the match and the whole filter is rejected as invalid. Either fix the regex or correct the comment to reflect what's actually supported.

**Minor:**
- The two `regexp.MustCompile` calls run on every request inside `parseSCIMFilterForEq`; hoist them to package-level `var`s compiled once.
- `schema_builder.go` only reorders `buildCoreGroupSchema` below `coreGroupAttributes` with no functional change — unrelated churn, please drop to keep the diff focused.
- Boolean/number comp values (`active eq true`) are translated but the real store comparison (JSON `->>` returns text vs. a Go bool/int param) isn't exercised by any test and may not match at the DB. Worth an integration test or a documented limitation, since only string core attrs are truly covered.
- `parseSCIMCompValue` uses `strconv.Unquote` (Go string rules) rather than JSON unquoting; fine for typical double-quoted values but slightly off-spec vs RFC 7159.

backend/internal/scim/config/config.go
ravindu439: Since we didn't implemented PATCH support for Users , lets keep this as false .

backend/internal/scim/users_handler.go
ravindu439: Fixed. `parseSCIMFilterForEq` now runs the translated attribute through `utils.ValidateKey` (same charset the store layer enforces) right after `translateSCIMFilterAttr`, before building the filter map. Invalid chars (hyphen included) now fail fast with 400 `invalidFilter` instead of surfacing as 500 downstream.

backend/internal/scim/users_handler.go

backend/internal/scim/users_handler.go

backend/internal/scim/schema_builder.go
ravindu439: fixed

sadilchamishka (approved):

PR #4198, opened 2026-07-21
Add client-level default token audience config

### LINKED ISSUE #3725: The default `aud` value of an access token is set to the client_id, which is ambiguous
Image

It violates the recommendations in the spec - https://datatracker.ietf.org/doc/html/rfc9068#figure-2

In the standard OAuth world, the client is not the audience unless specified under certain conditions.

Image

thiva-k:
### Purpose

Fixes #3725. Access tokens that aren't bound to a resource server (OIDC-only or scopeless requests) defaulted their `aud` claim to the `client_id`, which is ambiguous and deviates from RFC 9068 (the client isn't the audience). This adds a configurable **default audience** per application/agent for exactly those unbound tokens.

### Approach

- New optional `token.accessToken.defaultAudience` on OAuth applications and agents.
- When a request resolves to no resource server (no `resource` parameter and no permission scopes), the access token's `aud` uses this value, falling back to `client_id` when unset. Resource-bound tokens are unaffected; they still get the resource server identifier. Precedence: explicit `resource` → `defaultResourceServer` → per-app `defaultAudience` → `client_id`.
- Applied uniformly across all grants that mint access tokens (`authorization_code`, `client_credentials`, `jwt_bearer`, `token_exchange`, `ciba`); refresh tokens inherit the original binding.
- Console: free-text field under **Advanced Settings** for both applications and agents.

### Related Issues
- #3725

## Summary by CodeRabbit

* **New Features**
  * Added `defaultAudience` configuration for access tokens when no resource server is targeted (aud falls back to client ID when unset).
  * Updated token flows so OIDC-only/scopeless requests use the configured default audience, while resource-bound requests still use the resolved resource server audience.
  * Added console UI controls for setting default token audience (application and agent advanced settings).

* **Documentation**
  * Updated OAuth/OIDC, token, resource indicators, CIBA, token exchange, and authentication guides to reflect the new `aud` claim behavior.

* **Tests**
  * Added unit and integration tests covering configured default audience behavior and resource override.


backend/pkg/thunderidengine/providers/oauth_client.go
```diff
@@ -107,6 +107,17 @@ func (o *OAuthClient) ClientAccessTokenConfig() *AccessTokenSubConfig {
...
+// ResolveDefaultAudiences returns the aud claim for an access token that is not bound to a
+// resource server (an OIDC-only or scopeless request). It returns the application's configured
+// default audience when set; otherwise it falls back to the given client_id.
+func (o *OAuthClient) ResolveDefaultAudiences(clientID string) []string {
+	if o != nil && o.Token != nil && o.Token.AccessToken != nil &&
+		o.Token.AccessToken.DefaultAudience != "" {
+		return []string{o.Token.AccessToken.DefaultAudience}
+	}
+	return []string{clientID}
+}
```
sahandilshan: We are only support one default audiences, right?

Dilusha-Madushan:
@thiva-k Should defaultAudience live under the Token tab rather than Advanced Settings?

Dilusha-Madushan (approved):

backend/pkg/thunderidengine/providers/oauth_client.go
thiva-k: Updated to return as string

thiva-k:
> @thiva-k Should defaultAudience live under the Token tab rather than Advanced Settings?

Keeping it under advanced settings for now since the token section is planned to be fully revamped

sahandilshan (approved):

PR #4209, opened 2026-07-21
Return default OIDC scope-to-claims mappings in API responses

thiva-k:
### Purpose

Return default OIDC scope-to-claims mappings in the application and agent management API responses so the Console UI shows them.

Previously the backend stored only explicit overrides for `scopeClaims` (nil when uncustomized) and omitted the field from responses, so the Console rendered every scope with zero mapped attributes even though the runtime applied the standard OIDC defaults. Responses now include the effective mapping (standard OIDC defaults from `StandardOIDCScopes` overlaid with any stored per-scope overrides), matching exactly what tokens and userinfo emit at runtime.

---
Application and Agent management API responses (GET, create, update, and Export, DCR, and MCP responses) now always include the effective OIDC `scopeClaims` mapping: the standard OIDC scope-to-claim defaults merged with any stored per-scope overrides. Previously the field was omitted from the response when a client had not customized it.

#### 💥 Impact
- Response shape: `scopeClaims` is now always present and populated with all standard OIDC scopes, instead of being omitted when not customized. Consumers can no longer infer "using defaults" from the field being absent.
- Round-trip writes: a client that reads a configuration and re-submits it (GET then PUT) now persists the full default mapping, converting an implicit "inherit defaults" into an explicit stored map (a frozen snapshot). The stored record changes, though token and UserInfo output do not.
- Declarative/GitOps and export: exported application and agent resources now include the full `scopeClaims` map, so a re-apply shows a nil to explicit diff on the first apply and pins the defaults into the resource.
- Not affected: runtime token and UserInfo claim issuance is unchanged. The emitted claims are identical.

#### 🔄 Migration Guide
- No action is required for token or UserInfo behavior; it is unchanged.
- Consumers that treated an absent `scopeClaims` as "using defaults" should treat a populated map as the effective mapping. To detect explicit customization, compare each scope's claims against the standard OIDC defaults.
- Declarative/GitOps users should re-export resources after upgrading so stored files reflect the effective mapping, or ignore the additive `scopeClaims` entries in diffs.
---

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3710

## Summary by CodeRabbit

* **New Features**
  * Scope claims now automatically include standard OIDC defaults.
  * Application-specific scope-claim overrides and custom scopes are preserved and merged with the defaults.
  * Providing an explicit empty claim list can suppress default claims for a scope.

* **Bug Fixes**
  * When no stored scope-claims are provided, the effective scope-claims now correctly resolve to the full default set (rather than an empty result).
  * Resolved scope-claims no longer risk altering the shared default claims.

### DIFF
--- a/backend/internal/inboundclient/service.go
+++ b/backend/internal/inboundclient/service.go
@@ -38,6 +38,7 @@ import (
 	flowmgt "github.com/thunder-id/thunderid/internal/flow/mgt"
 	inboundmodel "github.com/thunder-id/thunderid/internal/inboundclient/model"
 	oauth2const "github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
+	oauthutils "github.com/thunder-id/thunderid/internal/oauth/oauth2/utils"
 	"github.com/thunder-id/thunderid/internal/system/config"
 	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	syshttp "github.com/thunder-id/thunderid/internal/system/http"
@@ -1346,7 +1347,8 @@ func applyInboundDefaults(c *inboundmodel.InboundClient, oauthProfile *providers
 		IDJAG:        resolveIDJAG(oauthProfile.Token),
 	}
 	oauthProfile.UserInfo = resolveUserInfo(oauthProfile.UserInfo, idToken)
-	oauthProfile.ScopeClaims = resolveScopeClaims(oauthProfile.ScopeClaims)
+	// Persist the effective scope-to-claims mapping: standard OIDC defaults merged with any overrides.
+	oauthProfile.ScopeClaims = oauthutils.ResolveEffectiveScopeClaims(oauthProfile.ScopeClaims)
 }
 
 // getDefaultAssertionFromDeployment returns the assertion config from the deployment-level JWT settings.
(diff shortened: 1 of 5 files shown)


backend/internal/application/service.go
```diff
@@ -1820,7 +1821,7 @@ func buildApplicationResponse(dto *model.ApplicationProcessedDTO) *providers.App
 					Token:                              oauthAppConfig.Token,
 					Scopes:                             oauthAppConfig.Scopes,
 					UserInfo:                           oauthAppConfig.UserInfo,
-					ScopeClaims:                        oauthAppConfig.ScopeClaims,
+					ScopeClaims:                        effectiveScopeClaims,
```
ThaminduDilshan: Rather than duplicating this in both application and agent, can't we handle this in inbound client?

backend/internal/inboundclient/service.go
```diff
@@ -1513,12 +1495,11 @@ func resolveUserInfo(in *providers.UserInfoConfig,
 	return out
 }
 
-// resolveScopeClaims returns the input scope claims map, defaulting to an empty map if nil.
+// resolveScopeClaims returns the effective scope-to-claims mapping to persist: the standard OIDC
+// defaults merged with any per-scope overrides in the input (nil yields the full defaults). Persisting
+// the effective map keeps stored configuration aligned with what tokens and API responses contain.
 func resolveScopeClaims(in map[string][]string) map[string][]string {
```
ThaminduDilshan: We can get rid of this method and directly invoke `oauthutils.ResolveEffectiveScopeClaims(in)` right?

ThaminduDilshan (approved):

backend/internal/application/service.go
thiva-k: Updated

backend/internal/inboundclient/service.go
thiva-k: Updated

PR #4219, opened 2026-07-21
Add an Enable SSO toggle to the flow builder for login flows

### LINKED ISSUE #4066: Add an Enable SSO toggle to the flow builder for login flows

When a builder user wants an existing sign-in flow to reuse an active session (SSO), the only path today is starting over from the "Basic with SSO" starter template. There is no way to add SSO to a flow that already exists, and hand-building it requires knowing the exact executor pairing and wiring conventions: an `SSOCheckExecutor` whose `checkpointRef` property references a `SessionExecutor` node, correct success/failure branch routing, and a session join placed at the authentication-complete boundary. Getting any of it wrong is only caught by backend validation.

Primary: identity admins and developers composing sign-in flows in the flow builder who want SSO behavior without rebuilding their flow from a template or learning the executor-level wiring contract.

### Proposed Solution

An **Enable SSO toggle** for AUTHENTICATION (login) flows that inserts or removes the SSO wiring as a graph transformation. Scope is auth flows only; registration and recovery flows are out of scope.

**Canonical pattern** (as produced by the `BASIC_SSO` template):

```
start -> sso_check --(success: "Skip to")------------------> session -> authorization -> assert -> end
              \--(failure: "Authenticate")-> prompt -> credentials_auth --(success)--^
```

- `SSOCheckExecutor` sits after Start; success means a session exists and jumps to the checkpoint, failure routes to fresh authentication.
- `SessionExecutor` is the join node: the fresh path saves the session there, the SSO path loads it; the node id is the checkpoint id.
- The pairing is an explicit backend contract: `SSOCheckExecutor.properties.checkpointRef` must reference a `SessionExecutor` node, and every `SessionExecutor` must be referenced by at least one SSO check (`backend/internal/flow/mgt/validator.go`, `validateSSOCheckExecutor` / `validateSessionExecutor`).

**Toggle UX**

- Flow-level toggle in the builder header (near Preview/Save), shown only for AUTHENTICATION flows.
- Toggle state is **derived from the graph** (ON when the flow contains an `SSOCheckExecutor`), not stored as a flag, so template-created SSO flows read as enabled with no migration.
- On enable: run the transformation, auto-layout only the affected region, select the inserted SSO check so its properties open, and show a snackbar summarizing the change.
- On disable: confirmation dialog ("Remove N SSO checkpoints from this flow?"), then remove the pairs and splice the edges back.

**Enable transformation (single checkpoint, default)**

1. Entry splice point: Start's `onSuccess` target `T_entry`.
2. Join point: the authentication-complete boundary, found heuristically as the node feeding `AuthAssertExecutor` (or the `AuthorizationExecutor` immediately before it).
3. Insert `session_<suffix>` with `onSuccess -> T_join`; retarget all fresh-path edges that pointed at `T_join` to the session node.
4. Insert `sso_check_<suffix>` with `checkpointRef = session_<suffix>`, `onSuccess -> session_<suffix>`, `onFailure -> T_entry`; retarget Start to it.
5. The pair shares an id suffix (purposeful-prefix convention) so the two nodes read as related on the canvas.

**Ambiguity handling:** when the join heuristic finds multiple candidates or no `AuthAssertExecutor`, do not guess. Enter a lightweight placement mode where candidate edges highlight on the canvas (same visual language as drop zones) and the user clicks where the checkpoint goes. If the flow has no Start wiring at all, the toggle is disabled with an explanatory tooltip.

**Disable transformation**

For each `SSOCheckExecutor`: splice its predecessor back to the check's `onFailure` target; retarget fresh-path edges entering its referenced session node to the session's `onSuccess` target; delete both nodes.

**Multiple checkpoints (phase 2)**

The backend supports several checkpoints per flow, each SSO check naming its checkpoint via `checkpointRef` (`backend/internal/flow/executor/session_executor.go`). Rather than more global toggles, phase 2 adds a per-location "Add SSO checkpoint here" action on edges / the node context menu (dovetails with #4063), plus a `checkpointRef` picker (dropdown of the flow's session nodes) in the SSO check's properties panel. Before building phase 2, confirm with the backend team what a mid-flow checkpoint skips when multiple checks chain (e.g. whether an MFA flow can skip factor 1 but still require factor 2).

**Validation and safety**

- The client transformation must produce graphs that pass the backend pairing validation by construction.
- Mirror the `checkpointRef` pairing checks in the console's live validation rules so hand-deleting half a pair surfaces immediately in the notifications panel; warn on deleting a session node that SSO checks still reference.
- Removal touches configured nodes, so this feature benefits from undo/redo (#4062) landing first.

### Phasing

- Phase 1 (MVP): header toggle, derived state, single-checkpoint auto-insert with join heuristic, interactive placement fallback, confirmed removal, mirrored validation rules.
- Phase 2: per-location checkpoint insertion, `checkpointRef` picker, multi-checkpoint semantics confirmed with backend.

---

Part of #3871.

DonOmalVindula:
### Purpose
Adds an **Enable SSO** toggle to the flow builder for AUTHENTICATION (login) flows, so builders can add or remove SSO session reuse on an existing flow without rebuilding it from the "Basic with SSO" template or hand-wiring the `SSOCheckExecutor` / `SessionExecutor` pairing contract.

This implements #4066: the toggle with graph-derived state, single-checkpoint auto-insert with the join heuristic, the interactive placement fallback, confirmed removal, the `checkpointRef` picker in the SSO check's properties panel, and mirrored client-side validation. 

### Approach

**Toggle (resource panel footer, AUTHENTICATION flows only)**
- Toggle state is derived from the graph (ON iff an `SSOCheckExecutor` node exists), so template-created SSO flows read as enabled with no migration or stored flag.
- **Enable** runs a pure graph transformation (`ssoGraphTransforms.ts`): splices `sso_check_<suffix>` after Start (success skips to the checkpoint, failure falls back to the entry view) and `session_<suffix>` in front of the join node, retargeting all fresh-path feeders through it. The join is found heuristically (the node feeding `AuthAssertExecutor`, or the `AuthorizationExecutor` immediately before it). The pair shares an id suffix, and `properties.checkpointRef` is set by construction so the result always passes backend pairing validation.
- When the heuristic is ambiguous (multiple completion points), the builder enters a lightweight placement mode: candidate edges highlight, the rest dim, and clicking a candidate places the checkpoint (Esc or Cancel exits).
- When enabling is impossible (no Start wiring, entry step is not a view, no auth assert step), the toggle is disabled with a reason tooltip instead of producing an invalid graph.
- **Disable** shows a confirmation dialog, then removes every SSO check and session node (orphaned halves included) and splices the edges back. If the properties panel is showing a removed node, it closes.
- New nodes are placed relative to their anchors and de-overlapped with the existing collision resolver; the whole-graph layout is never touched. Enabling forces verbose mode on so the change is never invisible.

**Validation**
- New graph-level validation rules mirror the backend pairing contract (missing/dangling `checkpointRef`, session node unreferenced by any SSO check) so hand-deleting half a pair surfaces immediately in the notifications panel and blocks Save. The same checks were added to the save-time `validateFlowGraph` backstop.

**Canvas UX (came out of testing this feature)**
- Hovering an edge now highlights it and spotlights the node it leads into, using the preview mode's visual language. Implemented with `data-id` CSS selectors so hovering never mutates node or edge objects.
- Edges no longer overlap on shared corridors: `BaseEdge` previously routed each edge independently, so edges taking the same corridor stacked exactly. A new `EdgePathsProvider` collects the endpoint geometry each edge already receives and routes all edges together through the existing (previously unused) `calculateAllEdgePaths`, separating shared corridors into parallel lanes. Fixed a latent bug in that algorithm where offsetting terminal segments detached paths from their handles (regression tests added).

**Core / feature boundary**: all SSO knowledge lives in the `login-flow` feature; the shared `flows` engine only gained generic extensions (resource panel `footer` slot, `onEdgeClick` pass-through, edge path provider).

### Related Issues
- Fixes #4066

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added SSO enable/disable workflow in the login-flow editor, including join-placement guidance, session checkpoint selection, and confirmation + success messages.
  * Introduced cross-node SSO pairing validation (missing/invalid/orphaned checkpoint/session references).
  * Improved flow visuals with centralized edge-path routing and non-simulation edge-hover spotlighting.
  * Added an optional bottom **footer** area to the Resource Panel.
* **Bug Fixes**
  * Fixed edge routing to better preserve handle anchoring and separation behavior.
  * Restored forwarding of edge click/hover handlers.
* **Tests**
  * Expanded coverage for SSO transforms, graph validation, and edge-path routing.

### DIFF
--- a/frontend/apps/console/src/features/flows/components/react-flow-overrides/__tests__/EdgePathsProvider.test.tsx
+++ b/frontend/apps/console/src/features/flows/components/react-flow-overrides/__tests__/EdgePathsProvider.test.tsx
@@ -1,0 +1,134 @@
+/**
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import {render, screen, waitFor} from '@testing-library/react';
+import {Position} from '@xyflow/react';
+import {useContext, useEffect} from 'react';
+import {beforeEach, describe, expect, it, vi} from 'vitest';
+import EdgeGeometryContext from '../../../context/EdgeGeometryContext';
+import EdgePathsContext from '../../../context/EdgePathsContext';
+import type {EdgeInput} from '../../../utils/calculateEdgePath';
+import EdgePathsProvider from '../EdgePathsProvider';
+
+const {mockState} = vi.hoisted(() => ({
+  mockState: {nodes: [] as unknown[]},
+}));
+
+vi.mock('@xyflow/react', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('@xyflow/react')>();
+  return {
+    ...actual,
+    useStore: (selector: (state: unknown) => unknown) => selector(mockState),
+    useStoreApi: () => ({getState: () => mockState}),
+  };
+});
+
+vi.mock('../../../hooks/useFlowConfig', () => ({
+  default: () => ({edgeStyle: 'step'}),
+}));
+
+function edgeInput(id: string, sourceX: number, sourceY: number, targetX: number, targetY: number): EdgeInput {
+  return {
+    id,
+    sourcePosition: Position.Right,
+    sourceX,
+    sourceY,
+    targetPosition: Position.Left,
+    targetX,
+    targetY,
+  };
+}
+
+function Probe({inputs}: {inputs: EdgeInput[]}) {
+  const registry = useContext(EdgeGeometryContext);
+  const paths = useContext(EdgePathsContext);
+
+  useEffect(() => {
+    inputs.forEach((input) => registry?.register(input));
+  }, [registry, inputs]);
+
+  return (
+    <div data-testid="paths">
+      {paths ? [...paths.entries()].map(([id, result]) => `${id}=${result.path}`).join(';') : 'none'}
+    </div>
+  );
+}
+
+describe('EdgePathsProvider', () => {
+  beforeEach(() => {
+    mockState.nodes = [];
+  });
+
+  it('should publish combined paths once edges register their geometry', async () => {
+    render(
+      <EdgePathsProvider>
+        <Probe inputs={[edgeInput('e1', 0, 0, 300, 200), edgeInput('e2', 0, 400, 300, 250)]} />
+      </EdgePathsProvider>,
+    );
+
+    await waitFor(() => {
+      const text = screen.getByTestId('paths').textContent ?? '';
+      expect(text).toContain('e1=M 0,0 ');
+      expect(text).toContain('e2=M 0,400 ');
+    });
+  });
+
+  it('should keep each combined path anchored to its endpoints', async () => {
+    render(
+      <EdgePathsProvider>
+        <Probe inputs={[edgeInput('e1', 0, 0, 300, 200), edgeInput('e2', 0, 0, 300, 200)]} />
+      </EdgePathsProvider>,
+    );
+
+    await waitFor(() => {
+      const text = screen.getByTestId('paths').textContent ?? '';
+      const paths = text.split(';');
+      expect(paths).toHaveLength(2);
+      paths.forEach((entry) => {
+        expect(entry).toMatch(/=M 0,0 /);
+        expect(entry.endsWith('L 300,200')).toBe(true);
+      });
+    });
+  });
+
+  it('should publish no paths while a node is being dragged', async () => {
+    mockState.nodes = [{dragging: true, id: 'a', position: {x: 0, y: 0}}];
+
+    render(
+      <EdgePathsProvider>
+        <Probe inputs={[edgeInput('e1', 0, 0, 300, 200), edgeInput('e2', 0, 400, 300, 250)]} />
+      </EdgePathsProvider>,
+    );
+
+    await waitFor(() => {
+      expect(screen.getByTestId('paths').textContent).toBe('none');
+    });
+  });
+
+  it('should publish no paths while fewer than two edges are registered', async () => {
+    render(
+      <EdgePathsProvider>
+        <Probe inputs={[edgeInput('e1', 0, 0, 300, 200)]} />
+      </EdgePathsProvider>,
+    );
+
+    await waitFor(() => {
+      expect(screen.getByTestId('paths').textContent).toBe('none');
+    });
+  });
+});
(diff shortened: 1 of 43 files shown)


frontend/apps/console/src/features/flows/validation/computeValidationNotifications.ts
```diff
@@ -150,12 +151,14 @@ function walkElements(
...
  * @param rules - The validation rule registry.
  * @param t     - i18next translate function.
+ * @param graphRules - Cross-node rules applied against the whole node set.
  * @returns A map of notification ID → Notification.
  */
 export function computeValidationNotifications(
   nodes: Node[],
   rules: ValidationRuleDefinition[],
   t: TFunction,
+  graphRules: GraphValidationRule[] = GRAPH_VALIDATION_RULES,
```
JayaShakthi97: As I understood SSO rule is for AUTHENTICATION-only. Shouldn't this default to [], pass the SSO rules explicitly from the AUTHENTICATION path?

frontend/apps/console/src/features/login-flow/hooks/useSsoToggle.ts
```diff
@@ -0,0 +1,308 @@
...
+  // stays stable across node-drag ticks. The derived state below is likewise
+  // keyed on its serialized value, keeping object identity stable so the
+  // toggle element (and the memoized resource panel hosting it) doesn't
+  // re-render on position-only graph changes.
+  const graphRef = useRef<{nodes: Node[]; edges: Edge[]}>({edges, nodes});
+  useEffect(() => {
+    graphRef.current = {edges, nodes};
+  }, [nodes, edges]);
+
+  const ssoStateKey = useMemo(() => JSON.stringify(deriveSsoState(nodes)), [nodes]);
```
JayaShakthi97: `deriveSsoState(nodes)` and `findJoinCandidates(nodes, edges)` re-run on every `nodes`/`edges` change, including position-only drag ticks. The `JSON.stringify`/`parse` guard keeps downstream identity stable, but the O(n) scan + serialize still runs each tick. Fine at typical flow sizes; could get costly on large graphs.

frontend/apps/console/src/features/flows/components/react-flow-overrides/__tests__/EdgePathsProvider.test.tsx
```diff
@@ -0,0 +1,134 @@
+/**
+ * Copyright (c) 2025, WSO2 LLC. (https://www.wso2.com).
```
brionmario:
```suggestion
 * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
```

brionmario (dismissed):

frontend/apps/console/src/features/flows/validation/computeValidationNotifications.ts
DonOmalVindula: Done. `computeValidationNotifications` now defaults `graphRules` to `[]`. The rules are registered per flow type through a new `graphValidationRules` field on `FlowConfigContext` (same pattern as `flowNodes`): `LoginFlowBuilder` registers the SSO pairing rules only when `flowType === 'AUTHENTICATION'`, and `ValidationProvider` consumes whatever the host registered. Added a test pinning that the default runs no graph rules.

frontend/apps/console/src/features/login-flow/hooks/useSsoToggle.ts
DonOmalVindula: the derivation is now keyed on a structural signature (node ids/types/executors + edge endpoints), which is the full input set of the computation; positions are excluded. A drag tick now costs one string concatenation instead of derive + serialize, and the derived objects keep identity across position-only changes. Same treatment applied to `EdgePathsProvider`: edges no longer register geometry during drags (the obstacles-key change on drop re-fires registration with settled values), so the provider no longer updates state per tick.

brionmario (approved):

PR #4253, opened 2026-07-22
Allow editing entities after schema tightening by filtering undeclared attributes

### LINKED ISSUE #3648: Changing an entity schema strands existing non-conforming entities in a non-editable state
## Summary

Changing an agent schema (e.g. change existing attribute ) can strand existing non-conforming agents in a non-editable state. An admin can effectively lock themselves out of editing or rotating credentials for existing agents, and the resulting error gives no hint that the schema — not the current edit — is the cause.

The same behaviour is observed for users as well. Changing any of the schemas for users as well.

## Details

- Modifications to the schema, such as changing an attribute name or deleting an existing attribute, are **not** validated against existing data at save time — there is no warning such as "N agents violate this" when changing the schema.
- As a result, pre-existing agents that lack the newly added `required` attribute will not work for cases such as **token issuance**. Also, there'll be stale attributes.
- However, every edit of an entity goes through the full-object `PUT /agents/{id}`. 
- Any such update now fails with:

```
{
    "code": "USR-1019",
    "message": {
        "key": "error.userservice.schema_validation_failed",
        "defaultValue": "Schema validation failed"
    },
    "description": {
        "key": "error.userservice.schema_validation_failed_description",
        "defaultValue": "User attributes do not conform to the required schema"
    }
}
```

Until the missing attribute is backfilled.

## How to reproduce

1. Create a new User schema with a set of attributes
2. Create a user of that type, filling in the attributes
3. Make changes to the schema such as changing the attribute or deleting an attribute key
4. Go back to the User created in step 2 and try updating the user from the console

## Net effect

An admin can lock themselves out of editing or rotating credentials for existing agents simply by tightening the schema. Because token issuance still works while edits fail, the failure is confusing — and the error message gives no indication that the schema (rather than the current edit) is the root cause.

## Suggested improvements

- Decide the lifecycle and acceptable changes for a given schema
- Handle the UX as needed to accommodate accepted changes to the schema to ensure existing entities are compliant to the new schema, and no stale data.

PasinduYeshan:
### Purpose
Editing a schema (entity type) after entities exist could strand those entities: once the schema was tightened, existing users whose stored attributes no longer matched could not be updated, and the UI showed only a generic error.

### Approach
- Surface backend validation errors on user create, update (map the error code instead of a static toast).
- On entity update, drop top-level attributes not declared in the current schema before validation, so a schema tightened after creation no longer blocks edits. Create still rejects undeclared attributes. Missing-required and tightened type/pattern/enum still reject by design.
- Warn in the user/agent type schema editor before saving any schema change, since it can affect existing entities of that type.
- Show removed attributes in the application token page, so admins can remove them from UI.

### Related Issues
- Fixes #3648

### Checklist
- [x] Tests provided.
    - [x] Unit Tests

## Summary by CodeRabbit

- **New Features**
  - Added a schema-change confirmation dialog when saving User Type and Agent Type edits that may impact existing records.
- **Bug Fixes**
  - Prevented entity update failures by stripping undeclared/stale top-level attributes before validation and uniqueness checks.
  - Improved error toasts for user create/update/credential updates with error-specific messaging.
  - Sanitized optional attribute updates to drop non-conforming values while preserving required fields.
- **Tests**
  - Expanded backend and frontend test coverage for attribute cleanup and the warning dialog (Cancel/Continue).
- **Documentation**
  - Added new English translation strings for backend errors and schema-change warning UI.


frontend/packages/configure-users/src/utils/dropNonConformingAttributes.ts
```diff
@@ -0,0 +1,69 @@
...
+      return true;
+    }
+    case 'number':
+      return typeof value === 'number' && Number.isFinite(value);
+    case 'boolean':
+      return typeof value === 'boolean';
+    case 'array':
+      return Array.isArray(value);
+    case 'object':
+      return typeof value === 'object' && value !== null && !Array.isArray(value);
```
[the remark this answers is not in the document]
PasinduYeshan: The schema editor cannot create array or object attributes. Both SchemaPropertyEditor (user types) and EditSchemaSettings (agent types) render a Type select with exactly four options: string, number, boolean, enum. There is no array/object menu item, and no sub-editor for items or nested properties. The items/properties fields exist on the input model only for round-trip fidelity (a schema authored elsewhere is preserved on save), never for editing.

Kept the shallow guards (value must be an array / a plain object). If nested attribute types become editable in the schema UI later, we'll add the recursive diff and value checks together, matched against the backend validators. Per our contributing guidance (avoid premature abstraction / speculative code), holding off until then.

pnpm-lock.yaml
```diff
@@ -1560,6 +1563,9 @@ importers:
       '@thunderid/types':
         specifier: workspace:^
         version: link:../types
+      '@thunderid/utils':
```
rajithacharith, line 1783: Do we need this 
cc: @brionmario @jeradrutnam @JayaShakthi97

PasinduYeshan: @rajithacharith this import is required since we have imported `getErrorMessage` from utils in configure-users files (useCreateUser, useUpdateUser, useUpdateUserCredentials).

backend/internal/entity/service.go
```diff
@@ -240,6 +240,13 @@ func (s *entityService) UpdateEntity(
 	}
 	s.logger.Debug(ctx, "Updating entity", log.MaskedString("id", entityID))
 
+	// Drop stale attributes no longer in the schema so they don't block the update.
+	cleanedAttrs, err := s.stripUndeclaredAttributes(ctx, entity.Category, entity.Type, entity.Attributes)
+	if err != nil {
+		return nil, err
+	}
+	entity.Attributes = cleanedAttrs
+
```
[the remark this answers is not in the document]
PasinduYeshan: This is already documented, under manage users and manage agents.

ThaminduDilshan (dismissed):

frontend/packages/configure-user-types/src/index.ts
```diff
@@ -86,3 +86,6 @@ export {default as ViewUserTypePage} from './pages/ViewUserTypePage';
 // Routes
 export type {UserTypeRoutePaths} from './hooks/useUserTypeRoutes';
 export {defaultUserTypeRoutePaths, default as useUserTypeRoutes} from './hooks/useUserTypeRoutes';
+
+// Utils
+export {default as getBreakingSchemaChanges} from './utils/getBreakingSchemaChanges';
```
DonOmalVindula, line 91: Do not add `index.ts` imports as it could lead to circular dependencies. Import directly from file

frontend/packages/configure-users/src/index.ts
```diff
@@ -68,3 +68,4 @@ export {defaultUserRoutePaths, default as useUserRoutes} from './hooks/useUserRo
 
 // Utils
 export {default as renderSchemaField} from './utils/renderSchemaField';
+export * from './utils/dropNonConformingAttributes';
```
DonOmalVindula, line 71: Remove `index.ts` exports

frontend/packages/configure-user-types/src/index.ts
DonOmalVindula: This is fine as it is an `index.ts` in the package root

frontend/packages/configure-users/src/index.ts
DonOmalVindula: This is fine as it is an `index.ts` in the package root

DonOmalVindula (dismissed):

ThaminduDilshan (approved):

PR #4296, opened 2026-07-23
Moved supported OAuth grantTypes, responseTypes and AuthMethods to configuration

anushasunkada:
### Purpose

DCR — dynamic client registration endpoint + discovery metadata is based on the "oauth.dcr.enabled" flag. By default value is true.
Token revocation — /oauth2/revoke endpoint + deny-list enforcement on the token-validation hot path is now enabled on the "oauth.token_revocation.enabled" flag. By default value is true.
Logout — RP-initiated logout endpoint + end_session_endpoint discovery metadata depends on "oauth.logout.enabled" flag. By default value is true.
CIBA — gated by allowed_grant_types, now also skips initializing the CIBA service entirely when the grant type isn't allowed. if "allowed_grant_types" is empty all the available grant types are supported.

-->

### Related Issues
- #3037

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added configurable allowlists for permitted OAuth grant types, response types, and token endpoint authentication methods.
  * Added feature toggles for Dynamic Client Registration, token revocation, and logout.
  * Added support for the Token Exchange grant handler.
* **Bug Fixes**
  * Disabled OAuth features no longer expose related endpoints or handlers.
  * Improved resilience by skipping revocation enforcement when not configured; CIBA callbacks now return a controlled `400` when CIBA isn’t enabled.
* **Documentation**
  * Updated OAuth configuration docs for DCR and registration constraints.
* **Tests**
  * Expanded coverage for discovery, CIBA callbacks, and allowlist/endpoint behavior.


backend/internal/application/tools_test.go
```diff
@@ -41,6 +43,31 @@ func TestApplicationToolsTestSuite(t *testing.T) {
...
 }
 
+func (suite *ApplicationToolsTestSuite) SetupTest() {
+	config.ResetServerRuntime()
+	cfg := &config.Config{
+		Server: engineconfig.ServerConfig{
+			Identifier: "test-dep",
+			Hostname:   "thunder.io",
+			Port:       443,
+			PublicURL:  "https://thunder.io",
```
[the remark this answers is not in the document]
anushasunkada: fixed

backend/tests/resources/deployment.yaml
```diff
@@ -27,6 +27,12 @@ oauth:
   jwt:
     issuer: thunderid
     validity_period: 3600
+  dcr:
+    enabled: true
+  logout:
+    enabled: true
+  token_revocation:
+    enabled: true
```
senthalan, line 35: Shall we remove this from the deployment.yaml. let's keep this simple. As these are already defined in the default.json it will be picked

senthalan:
I don't see the configurations like allowed_auth_methods, allowed_response_types and allowed_grant_types have been used to validate the oauth client details during the creation and update paths. @anushasunkada

backend/tests/resources/deployment.yaml
senthalan: This is a test file content. so we can ignore this comment

backend/internal/application/service.go
```diff
@@ -1334,7 +1340,10 @@ func translateOAuthValidationError(err error) *tidcommon.ServiceError {
 
 	// OAuth: token endpoint auth method
 	case errors.Is(err, inboundclient.ErrOAuthInvalidTokenEndpointAuthMethod):
-		return &ErrorInvalidTokenEndpointAuthMethod
+		return tidcommon.CustomServiceError(ErrorInvalidOAuthConfiguration, tidcommon.I18nMessage{
+			Key:          "error.applicationservice.invalid_token_endpoint_auth_method",
+			DefaultValue: "Invalid token endpoint authentication method",
+		})
```
senthalan: Shall we keep this as it is. With the Application level error codes defined in backend/internal/application/error_constants.go without this custom error

backend/pkg/thunderidengine/config/config.go
```diff
@@ -179,6 +179,7 @@ type AuthorizationCodeConfig struct {
 
 // DCRConfig holds the Dynamic Client Registration configuration.
 type DCRConfig struct {
+	Enabled  bool `yaml:"enabled" json:"enabled"`
```
thiva-k: Shouldn't this be `*bool` instead? We are setting the default as true, so if we override this to false via deployment.yaml, the merge logic will ignore zero values(false is the zero value for bool).

backend/internal/application/service.go
anushasunkada: For all 3 checks?

senthalan (approved):

PR #4299, opened 2026-07-23
Make id_token_hint optional for RP-initiated logout

madurangasiriwardena:
### Purpose
Makes id_token_hint optional for OIDC RP-Initiated Logout. Previously the logout endpoint rejected any request that supplied post_logout_redirect_uri without an id_token_hint.

Per the RP-Initiated Logout spec, the OP may perform post-logout redirection without an id_token_hint as long as it confirms the legitimacy of the redirection target by other means, and per RPLogout it MUST ask the End-User to confirm the logout when no id_token_hint was provided. This PR implements both: a logout carrying only client_id now resolves (the redirect target is still validated against the client's registered allow-list), and when no valid hint is present the sign-out flow can prompt the user to confirm before the session is terminated.

A new "Confirm Sign Out Without a Hint" flow-builder template wires up this conditional-confirmation behavior.

### Approach
- Relax the requirement (internal/oauth/oauth2/logout): dropped the "id_token_hint required when post_logout_redirect_uri is present" check. A request with only client_id (with or without a registered post_logout_redirect_uri) now resolves; at least one of id_token_hint/client_id is still required, and post_logout_redirect_uri is still validated against the client's registered list, which serves as the spec's "other means of confirming the redirection target."
- Signal the prompt requirement: when no valid id_token_hint was supplied, Resolve marks the resolution PromptRequired, and InitiateSignOutFlow passes logoutPromptRequired into the sign-out flow via RuntimeData.
- Confirm in the executor (SessionSignOutExecutor): when the sign-out node opts in via the promptOnSignOut property and logoutPromptRequired is set, the executor returns the incomplete state so the engine routes to the node's onIncomplete confirmation prompt, terminating the session only after the user confirms (guarded to prompt once per run). Nodes that do not opt in terminate as before, so the existing always-prompt default sign-out flow is unchanged.
- Key design decision: the conditional behavior is expressed with builder-native constructs (an onIncomplete edge plus a node property) rather than a node-level execution condition, because those already render and round-trip in the flow editor. This is why the confirmation logic lives in the executor instead of a graph condition.
- Template: added a SIGNOUT flow-builder template with this wiring so developers can adopt it from the Console flow builder.

Prompt logout consent always
image

Prompt logout consent when id_token_hint is not provided
image

image

Available flow templates
image

### Related Issues
- https://github.com/thunder-id/thunderid/issues/2036

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Added optional end-user confirmation for RP-initiated sign-out when a valid ID token hint is unavailable.
  - Introduced a configurable “Prompt for Confirmation” setting on the session sign-out executor (default off).
  - Added a conditional sign-out flow template that prompts only when needed.
  - Kept redirect validation intact for logout requests.
- **Bug Fixes**
  - Avoided repeated confirmation prompts within the same sign-out flow run.
  - Ensured sign-out proceeds correctly after confirmation or when prompting is disabled.
- **Tests**
  - Added/updated coverage for prompt routing and sign-out behavior across new scenarios.

### DIFF
--- a/backend/internal/flow/common/constants.go
+++ b/backend/internal/flow/common/constants.go
@@ -244,6 +244,14 @@ const (
 	// from the SSO checkpoint snapshot so each flow execution mints a fresh tfid rather than reusing a
 	// prior one on SSO reuse.
 	RuntimeKeyTokenFamilyID = "tokenFamilyId"
+	// RuntimeKeyLogoutPromptRequired is set by the OAuth RP-initiated logout layer when a logout was
+	// requested without a valid id_token_hint. A sign-out flow's session sign-out node reads it to
+	// decide whether the End-User must confirm the logout before the session is terminated.
+	RuntimeKeyLogoutPromptRequired = "logoutPromptRequired"
+	// RuntimeKeyLogoutPromptShown is the session sign-out node's own guard, set when it routes to the
+	// confirmation prompt so that on re-run (after the user confirms) it terminates instead of
+	// prompting again.
+	RuntimeKeyLogoutPromptShown = "logoutPromptShown"
 )
 
 // SSOCheckpointKey scopes a per-checkpoint SSO control key (RuntimeKeySSOSessionPresent,
--- a/backend/internal/flow/executor/constants.go
+++ b/backend/internal/flow/executor/constants.go
@@ -115,6 +115,10 @@ const (
 	propertyKeyCallbackType                            = "callbackType"
 	propertyKeyLoginHintAttribute                      = "loginHintAttribute"
 	propertyKeyMaxOTPAttempts                          = "maxAttempts"
+	// propertyKeyPromptOnSignOut, when set to boolean true on a session sign-out node, makes the executor
+	// confirm the logout with the End-User (via the node's onIncomplete prompt) whenever the RP-initiated
+	// logout was not accompanied by a valid id_token_hint (RuntimeKeyLogoutPromptRequired).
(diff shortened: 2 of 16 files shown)


backend/internal/flow/common/constants.go
```diff
@@ -233,6 +233,14 @@ const (
...
 	// concepts off the reusable engine contract.
 	RuntimeKeySSOSessionCleared = "ssoSessionCleared"
+	// RuntimeKeyLogoutPromptRequired is set by the OAuth RP-initiated logout layer when a logout was
+	// requested without a valid id_token_hint. A sign-out flow's session sign-out node reads it to
+	// decide whether the End-User must confirm the logout before the session is terminated.
+	RuntimeKeyLogoutPromptRequired = "logoutPromptRequired"
+	// RuntimeKeyLogoutPromptShown is the session sign-out node's own guard, set when it routes to the
+	// confirmation prompt so that on re-run (after the user confirms) it terminates instead of
+	// prompting again.
+	RuntimeKeyLogoutPromptShown = "logoutPromptShown"
```
ThaminduDilshan, line 254: We have a `type` field in the prompt node `action` that can be used to indicate a metadata about the action edge. We should be able to improve this logic and get rid of this runtime data key by utilizing that.
Let's do it as a followup

ThaminduDilshan (dismissed):

rajithacharith (approved):

PR #4317, opened 2026-07-24
Fixes current implemented E2E tests including MFA login

### LINKED ISSUE #4276: Fix the current E2E Tests
### Description

Currently the E2E tests on the console and admin workflows are broken.

### Steps to Reproduce

Build the product. 
Run `make test_e2e`

### Version

0.48.0

_No response_

Sadeesha-Sath:
### Purpose

Fix the currently implemented E2E test suite, both locally and in CI, and un-skip the MFA registration + login test that was previously marked fixme.

The suite had several issues currently:

- Login/logout page checks took a single instant DOM snapshot, which can fire ahead of the post-login OAuth redirect on Firefox/WebKit and report false "not logged in" results.
- Local .env defaults, `run-e2e.sh`, and the CI workflow each carried their own copy of the test dataset, so the data can be inconsistent.
- The `run-e2e.sh` admin token request was missing the resource parameter (RFC 8707), and the sample app ID was hardcoded, so a config change to the sample could break the script.
- MFA setup changed the sample application's flow bindings without ever restoring them, leaving stale state for future tests.
- The CORS test reused `SAMPLE_APP_URL` as its test origin, and its cleanup would strip that origin from future tests
- `run-e2e.sh`'s admin token bootstrap didn't account for the console login flow's new SSO check, so it broke as soon as SSO reuse landed on the console app
- client-initiated OIDC logout was enabled on the console app but never configured for the react-sample-app so logout test broke 

#### Config consolidation (single source of truth):

- Added `tests/e2e/defaults.env` as the canonical fixed dataset. Both `run-e2e.sh` (local) and the PR workflow now load it, so defaults are consistent.
- Reworked the CI step in `pr-builder.yaml` to load `defaults.env` and fall back to it, keeping the existing variable priority.
- Changed `tests/e2e/.env.example` and `tests/e2e/README.md` to show that `.env` is auto-generated and only needed for local overrides.

#### Test stability fixes

- Replaced instant snapshot checks with auto-retrying `waitFor` (using a combined `locator.or()`) for both the logged-in indicators and the logout affordance, fixing the redirect race condition on Firefox/WebKit.
- Extracted `getOTPErrorMessage()` as a reusable helper.

##### `run-e2e.sh`:

- Added the `resource` parameter to the admin authorize request (tokens are bound to a single resource server per RFC 8707).
- Look up the "Sample App ID" dynamically via the API instead of hardcoding a UUID.
- Fixed the admin-token bootstrap for the console login flow's new SSO check

##### MFA test fixes

- Un-skipped `TC003`. Since the React SDK's SignUp component does not redirect after sign-up (unlike SignIn), the test now verifies registration by asserting on the `/flow/execute` response body, rather than waiting for the missing URL redirect
- `updateApplicationFlows` now returns the original flow bindings, and a `revertApplicationFlows` cleanup step restores them in afterAll.
- Tightened OTP assertions to hard fail instead of warning.
- Extracted `captureCreatedUserId()` for cleanup.

#### Sample-App client initiated logout flow support

- With client-initiated OIDC logout enabled, the React SDK sample now sends `post_logout_redirect_uri` + `id_token_hint` to `/oauth2/logout `on sign-out. 
- Added a SIGNOUT flow resource to `thunderid-config.yaml` of the sample app and wired the application to it
- The sign-out flow surfaces a confirmation page from the gate before completing the redirect back to the app. Updated logout to click through that confirmation when present, and `verifyLoggedOut()` to assert the app's home page

##### Other:

- CORS test now uses a dedicated fake origin instead of `SAMPLE_APP_URL`, so its cleanup won't affect other functions.
- `verifyAuthState` now confirms the stored token with a real `/oauth2/userinfo` call, catching server-rejected tokens that client-side checks previously missed.

### Related Issues
- Closes #4276 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Added E2E coverage for sample app authentication (login/logout) and MFA.
  - Added a sign-out flow for the React SDK sample, including post-logout redirect support.

- **Bug Fixes**
  - Improved auth validation by combining client checks with server-side token verification.
  - Made login/logout and OTP/error handling more reliable; strengthened MFA flow assertions and cleanup.

- **Documentation**
  - Refreshed E2E setup guide, including canonical defaults and clear environment override precedence.

- **Tests**
  - Improved E2E runner credential resolution and flow bootstrapping robustness; made CORS origin checks more deterministic.

### DIFF
--- a/tests/e2e/tsconfig.json
+++ b/tests/e2e/tsconfig.json
@@ -1,10 +1,10 @@
 {
   "compilerOptions": {
-    "target": "ES2020",
-    "module": "commonjs",
-    "lib": ["ES2020", "DOM"],
+    "target": "ES2023",
+    "module": "nodenext",
+    "lib": ["ES2023", "DOM"],
     "types": ["node", "@playwright/test"],
-    "moduleResolution": "bundler",
+    "moduleResolution": "nodenext",
     "esModuleInterop": true,
     "skipLibCheck": true,
     "strict": true,
--- a/.github/workflows/pr-builder.yml
+++ b/.github/workflows/pr-builder.yml
@@ -906,12 +906,25 @@ jobs:
           fi
           chmod +x setup.sh start.sh
 
-      - name: 🚀 Start the Server
+      - name: 🔑 Resolve Admin Credentials
         env:
-          # Pin the admin credentials so setup.sh seeds a known password instead of
-          # generating a random one. The token step and Playwright below expect admin/admin.
-          ADMIN_USERNAME: admin
-          ADMIN_PASSWORD: admin
+          SECRET_ADMIN_USERNAME: ${{ secrets.PLAYWRIGHT_ADMIN_USERNAME }}
+          VAR_ADMIN_USERNAME: ${{ vars.PLAYWRIGHT_ADMIN_USERNAME }}
+          SECRET_ADMIN_PASSWORD: ${{ secrets.PLAYWRIGHT_ADMIN_PASSWORD }}
+        run: |
+          # Resolve a single admin username/password pair (Secret > Variable > defaults.env) once,
+          # up front, and reuse it for server bootstrap, token acquisition, and the Playwright run
+          # below, instead of each resolving (or hardcoding) its own copy.
+          DEFAULT_ADMIN_USERNAME=$(grep '^ADMIN_USERNAME=' tests/e2e/defaults.env | cut -d= -f2-)
+          DEFAULT_ADMIN_PASSWORD=$(grep '^ADMIN_PASSWORD=' tests/e2e/defaults.env | cut -d= -f2-)
+          ADMIN_USERNAME="${SECRET_ADMIN_USERNAME:-${VAR_ADMIN_USERNAME:-$DEFAULT_ADMIN_USERNAME}}"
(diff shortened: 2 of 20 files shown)


tests/e2e/tsconfig.json
DonOmalVindula, line 1: Is this change intentional? why have we changed these?

Sadeesha-Sath: Yes. I noticed some ts errors in my vs code. Upon investigating, I found the following tsconfig enforcement.

https://www.typescriptlang.org/docs/handbook/modules/reference.html#implied-and-enforced-options-3

Apparently, the "bundler" moduleResolution is only permitted with module "esnext" or "preserve". 

To support "commonJS" the only moduleResolution supported is "node"/"node10", which is now deprecated and not functioning in TS 7.0. 

The "nodenext" moduleResolution support both legacy and modern TS exports. That's why I have used it. However it is a bit stricter than "node" or "bundler"

The playwright doesn't throw any error on this cause it uses esbuild, which doesn't do the module/moduleResolution verification. 

Kindly give your input to what to proceed

DonOmalVindula: Okay. This approach is fine. +1 to proceed

rajithacharith (approved):

PR #4356, opened 2026-07-26
Restructure agent onboarding completion screen

Dilusha-Madushan:
### Purpose

Improves the AI agent onboarding completion screen (the final step after creating an agent, `ShowClientSecret`) to make the generated identity and credentials easier to read and act on, and to bring it in line with the MCP client onboarding experience.

Previously the screen showed the agent name, client ID, and client secret stacked in a single card, with no reference to the OAuth/OIDC endpoints a developer needs to actually use the agent. It also had two ways to copy the same secret (an inline icon and a footer button), and the "secret won't be shown again" warning sat below an unrelated section.

localhost_5191_console_agents_create

### Approach

Reworked `ShowClientSecret` into three cards, reusing the primitives introduced for the MCP client onboarding screen so the two flows stay visually consistent:

- **Agent card** shows the agent name and agent ID (a new `agentId` prop, passed from
  `AgentCreatePage` as `createdAgent.id`).
- **Client credentials card** shows the client ID and the masked client secret.
- **Endpoints card** renders a curated set of discovery endpoints via the shared
  `CopyableField` component. The rows come from a new `getAgentDiscoveryEndpointRows`
  util, sourced from `useThunderID().discovery.wellKnown`. It mirrors the MCP util but
  drops the MCP-specific "Authorization server metadata" row and scopes the labels to the
  `agents` i18n namespace. Rows with missing values are omitted, and the whole card is
  hidden when discovery is unavailable.

UX refinements on the same screen:
- The "you won't be able to see this secret again" warning now sits directly below the
  credentials card, adjacent to the secret.
- The redundant footer "Copy client secret" button was removed, leaving a single inline
  copy affordance on the secret field (now with a copied confirmation icon) and a single
  primary **Continue** action. Added aria-labels to the visibility and copy controls.
- The client secret field is given a light highlight so it still stands out.

The shared `CopyableField` and the `McpDiscoveryEndpoints`/`McpDiscoveryEndpointRow` types are imported directly from the `applications` feature, consistent with the existing cross-feature reuse (the agent OAuth model already imports from `applications/models`). Moving those primitives to a truly shared location was intentionally left out of scope.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3492

## Summary by CodeRabbit

* **New Features**
  * Added optional Agent ID display to the agent creation completion screen.
  * Added an “Endpoints” card with localized, copyable discovery details (issuer, OIDC discovery URL, authorization, token) when discovery data is available.
  * Replaced inline client-secret controls with a “View client secret” dialog (copy, show/hide, done).

* **Bug Fixes**
  * Improved resilience by omitting missing or incomplete discovery fields.

* **Tests**
  * Added/updated tests for the new discovery-row utility and for both the completion screen and the client-secret dialog.


Dilusha-Madushan:
Resolve these changes with this PR - https://github.com/thunder-id/thunderid/pull/4598

Now the important OIDC end-points and the client id and agent id information are exist in overview page.

So this PR can close.

cc: @sahandilshan

PR #4358, opened 2026-07-26
Strip undeclared user attributes on inbound client update

PasinduYeshan:
### Purpose
When an application is updated, user attributes that are no longer declared in the allowed user types' schema were kept in the token allow-lists, and the update failed validation. This strips such attributes on update instead of rejecting it.

### Approach
In `UpdateInboundClient`, before validation, remove from the four token allow-lists (native assertion, access token, id token, userinfo) any attribute not declared by the allowed user types' schema. Computed attributes (groups, roles, ou*, userType) are kept. The create and non-persisting validate paths are unchanged and still reject genuinely undeclared attributes.

### Related Issues
- Refs #4240

### Related PRs
- N/A

### Checklist
- [x] Tests provided.
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Inbound client updates now strip undeclared, non-computed user attributes from configured assertion and OAuth profile attribute allow-lists (access token, ID token, and userinfo).
  - Computed attributes are preserved.
  - Validation now explicitly ignores computed attributes and continues to reject undeclared attributes when configured sources are present.

- **Tests**
  - Added coverage for attribute stripping (including no-op scenarios) and updated validation test placement to match the new behavior.

### DIFF
--- a/backend/internal/inboundclient/service.go
+++ b/backend/internal/inboundclient/service.go
@@ -227,7 +227,8 @@ func (s *inboundClientService) UpdateInboundClient(ctx context.Context, client *
 	if fkErr := s.validateFKs(ctx, client); fkErr != nil {
 		return fkErr
 	}
-	if err := s.validateUserAttributesAgainstAllowedTypes(
+	// Stripping leaves only attributes the validator accepts, so no separate validation is needed here.
+	if err := s.stripUndeclaredUserAttributes(
 		ctx, client.AllowedUserTypes, client.Assertion, oauthProfile); err != nil {
 		return err
 	}
@@ -1315,19 +1316,9 @@ func (s *inboundClientService) validateUserAttributesAgainstAllowedTypes(
 		return nil
 	}
 
-	validAttrs := make(map[string]bool)
-	for _, entityTypeName := range allowedEntityTypes {
-		attrInfos, svcErr := s.entityType.GetAttributes(
-			security.WithRuntimeContext(ctx), entitytype.TypeCategoryUser, entityTypeName, false, true, false)
-		if svcErr != nil {
-			if svcErr.Type == tidcommon.ServerErrorType {
-				return ErrUserSchemaLookupFailed
-			}
-			return ErrFKInvalidUserType
-		}
-		for _, info := range attrInfos {
-			validAttrs[info.Attribute] = true
-		}
+	validAttrs, err := s.resolveValidUserAttributes(ctx, allowedEntityTypes)
+	if err != nil {
+		return err
 	}
 
 	for attr := range attrs {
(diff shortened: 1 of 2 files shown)


backend/internal/inboundclient/service.go
```diff
@@ -1315,32 +1319,127 @@ func (s *inboundClientService) validateUserAttributesAgainstAllowedTypes(
...
+			return ErrInvalidUserAttribute
+		}
+	}
+	return nil
+}
+
+// resolveValidUserAttributes returns the union of non-credential attribute names declared in the
+// schemas of the given allowed user types. Returns (nil, nil) when there are no allowed types or the
+// entity-type service is unavailable, signaling callers to skip attribute reconciliation.
+func (s *inboundClientService) resolveValidUserAttributes(
```
ThaminduDilshan, line 1338: Isn't this method invoked twice during inbound client creation/ update, once via `stripUndeclaredUserAttributes` method and again via `validateUserAttributesAgainstAllowedTypes` method?

backend/internal/inboundclient/service.go
```diff
@@ -1315,32 +1319,127 @@ func (s *inboundClientService) validateUserAttributesAgainstAllowedTypes(
...
 		}
 	}
+
+	if len(dropped) > 0 {
+		names := make([]string, 0, len(dropped))
+		for a := range dropped {
+			names = append(names, a)
+		}
+		slices.Sort(names)
+		s.logger.Info(ctx, "Stripped user attributes no longer declared in the entity type schema",
```
ThaminduDilshan: We shouldn't add unnecessary Info logs. Let's convert this to a Debug log.
But since there's a computation, let's also guard the entire section behind a `logger.IsDebugEnabled()` check

backend/internal/inboundclient/service.go
```diff
@@ -1315,32 +1319,127 @@ func (s *inboundClientService) validateUserAttributesAgainstAllowedTypes(
...
+}
 
-	for attr := range attrs {
-		if isComputedAttribute(attr) {
-			continue
+// stripUndeclaredUserAttributes removes from the application's token allow-lists any user attribute
+// no longer declared in the schema of its allowed user types, keeping computed attributes. Mutates
+// the same allow-lists collectConfiguredUserAttributes reads. No-op when there are no allowed types
+// or the entity-type service is unavailable (matches the validator's skip).
+func (s *inboundClientService) stripUndeclaredUserAttributes(
```
ThaminduDilshan, line 1366: I feel like stripping logic is bit overcomplicated with multiple anonymous functions. Some of these functions (both anonymous and named functions) are invoked more than once as well.
Shall we check whether this logic can be simplified?

Maybe we should be able to simplify it by combining `collectConfiguredUserAttributes` and `validateUserAttributesAgainstAllowedTypes` functions.

@coderabbit WDYT?

backend/internal/inboundclient/service.go
PasinduYeshan: Removed the validateUserAttributesAgainstAllowedTypes call from UpdateInboundClient. After stripping, every attribute left in the four allow-lists is either computed or declared in the allowed user types' schema, which is exactly what the validator checks, so on that path it could never reject anything, it only repeated the schema lookup. Both helpers also surface the same errors (ErrUserSchemaLookupFailed / ErrFKInvalidUserType) from the shared resolve, so error behaviour is unchanged.

backend/internal/inboundclient/service.go
PasinduYeshan: I simplified the logic, check @CodeRabbit.

ThaminduDilshan (approved):

PR #4388, opened 2026-07-27
Adding initiator request param during PAR and flow

anushasunkada:
### Purpose
Adding initiator request param during PAR and flow

### Related Issues
- #3037 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/internal/oauth/oauth2/par/model.go
+++ b/backend/internal/oauth/oauth2/par/model.go
@@ -20,12 +20,14 @@ package par
 
 import (
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/model"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
 // pushedAuthorizationRequest holds the stored PAR data.
 type pushedAuthorizationRequest struct {
-	ClientID        string
-	OAuthParameters model.OAuthParameters
+	ClientID         string
+	OAuthParameters  model.OAuthParameters
+	InitiatorRequest providers.InitiatorRequest
 }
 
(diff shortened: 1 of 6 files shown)


backend/internal/oauth/oauth2/par/model.go
```diff
@@ -20,12 +20,14 @@ package par
...
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
 // pushedAuthorizationRequest holds the stored PAR data.
 type pushedAuthorizationRequest struct {
-	ClientID        string
-	OAuthParameters model.OAuthParameters
+	ClientID         string
+	OAuthParameters  model.OAuthParameters
+	InitiatorRequest providers.InitiatorRequest
```
ThaminduDilshan, line 30: Prefer if we can have generic fields rather than having a flow related field here. But InitiatorRequest is also kind of generic since it doesn't carry any flow specific terms.

ThaminduDilshan: Okay to proceed with this for now. We can revisit later if required since this is a un-exported struct

ThaminduDilshan (approved):

backend/internal/oauth/oauth2/par/model.go
anushasunkada: providers.InitiatorRequest is still a generic struct. But the places from where this should be captured will differ. In CIBA also we may need to capture request params. But for now it should be fine.

PR #4402, opened 2026-07-28
Fix(scim): required-attribute and core schema validation, correct attribute mapping bugs

ravindu439:
### Purpose
Fixes several SCIM 2.0 correctness bugs found during API testing: missing required-attribute validation on CreateUser/ReplaceUser, missing core schema URN validation on Group PUT/replace, and two attribute-mapping bugs — fabricated `type` values on emails/phoneNumbers/photos/addresses that were never provided, and non-canonical address sub-attributes leaking into the core `addresses` response instead of the RFC 7643 canonical names.

### Approach
- `missingRequiredAttrs` in `schema_builder.go` checks entity-type schema `required` properties against merged attributes post reverse-mapping; `CreateUser`/`ReplaceUser` now return a per-attribute error via `newMissingRequiredAttributesError` instead of a generic schema-validation failure.
- `ValidateSCIMUserRequest` now requires the core User schema URN whenever the request carries core attributes (previously ungated), and the extension object is optional when the request has no extension attributes. Group PUT/replace gained the equivalent core Group schema URN check.
- `core_attr_mapper.go`: removed the `defaultType` fabrication for `kindMultiComplex` fields (email/phone/photo/address) — `type` is now only set when the source data actually has one; `primary:true` is still applied to a lone value. Added `filterCanonicalAddrKeys` so only RFC 7643 `streetAddress`/`locality`/`region`/`postalCode`/`country`/`formatted` sub-attributes (plus `type`/`primary`) survive into the core `addresses` response, anything else is dropped rather than leaked.

### Related Issues
- N/A

### Related PRs
- #4184 

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/internal/scim/core_attr_mapper.go
+++ b/backend/internal/scim/core_attr_mapper.go
@@ -41,7 +41,12 @@ const (
 	scimFieldProfileURL    scimCoreField = "profileUrl"
 )
 
-const scimValueKey = "value"
+const (
+	scimValueKey     = "value"
+	scimTypeKey      = "type"
+	scimPrimaryKey   = "primary"
+	scimFormattedKey = "formatted"
+)
 
 type attrKind string
 
@@ -58,7 +63,6 @@ type coreAttrRule struct {
 	kind        attrKind
 	parentField scimCoreField // only for kindSubAttr; top-level complex object this rolls into, e.g. "name"
 	subAttr     string        // only for kindSubAttr/kindAddrPart; key within the parent/addresses object
-	defaultType string        // only for kindMultiComplex; e.g. "work"
 	valueKey    string        // e.g. "value" (default)
 }
 
@@ -70,11 +74,10 @@ var coreAttrRules = []coreAttrRule{
 		kind:      kindSimpleString,
 	},
 	{
-		candidate:   "email",
-		scimField:   scimFieldEmails,
-		kind:        kindMultiComplex,
-		defaultType: "work",
-		valueKey:    scimValueKey,
+		candidate: "email",
+		scimField: scimFieldEmails,
+		kind:      kindMultiComplex,
+		valueKey:  scimValueKey,
 	},
 	{
 		candidate:   "given_name",
@@ -89,11 +92,10 @@ var coreAttrRules = []coreAttrRule{
 		subAttr:     "familyName",
 	},
 	{
-		candidate:   "phone_number",
-		scimField:   scimFieldPhoneNumbers,
-		kind:        kindMultiComplex,
-		defaultType: "work",
-		valueKey:    scimValueKey,
+		candidate: "phone_number",
+		scimField: scimFieldPhoneNumbers,
+		kind:      kindMultiComplex,
+		valueKey:  scimValueKey,
 	},
 	{
 		candidate: "display_name",
@@ -118,11 +120,10 @@ var coreAttrRules = []coreAttrRule{
 		kind:      kindSimpleString,
 	},
 	{
-		candidate:   "picture",
-		scimField:   scimFieldPhotos,
-		kind:        kindMultiComplex,
-		defaultType: "photo",
-		valueKey:    scimValueKey,
+		candidate: "picture",
+		scimField: scimFieldPhotos,
+		kind:      kindMultiComplex,
+		valueKey:  scimValueKey,
 	},
 	{
 		candidate: "locale",
@@ -148,6 +149,12 @@ var coreAttrRules = []coreAttrRule{
 		candidate: "title",
 		scimField: scimFieldTitle,
 		kind:      kindSimpleString,
+	},
+	{
+		candidate: "address",
+		scimField: scimFieldAddresses,
+		kind:      kindMultiComplex,
+		valueKey:  scimFormattedKey,
 	},
 	{
 		candidate: "street_address",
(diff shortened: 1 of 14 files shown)


backend/internal/scim/schema_builder.go
```diff
@@ -57,7 +58,6 @@ func mapEntityTypeToSCIMSchema(et entitytype.EntityType, baseURL string) (SCIMSc
 		Meta: SCIMMeta{
 			ResourceType: "Schema",
 			Location:     location,
```
sadilchamishka, line 60: This drops `Meta.Version` (the weak ETag) from dynamically generated extension schemas, and `computeSchemaVersion` is removed below. The PR description doesn't mention this. Given `ETag.Supported` is advertised and other resources still return weak ETags, the `/Schemas` endpoint now being version-less is inconsistent. Is this an intentional part of the fix, or unrelated cleanup that slipped in? If intentional, worth a line in the PR description; if not, please restore it.

backend/internal/scim/schema_builder.go
```diff
@@ -641,6 +645,35 @@ type rawPropertyDef struct {
...
+
+	var missing []string
+	for name, def := range rawProps {
+		if !def.Required {
+			continue
+		}
+		if skipCredential && def.Credential {
+			continue
+		}
+		if _, ok := attrs[name]; !ok {
```
sadilchamishka, line 669: This only checks that the key is present, not that it carries a value. A required attribute supplied as `""`, `null`, or `[]` would pass here. If `entitytype.Schema.Validate` already enforces non-empty required values, this is redundant (fine); if not, empty-but-present required attrs slip through. Worth confirming the two layers agree.

backend/internal/scim/core_attr_mapper.go
```diff
@@ -513,74 +607,95 @@ func normalizeToMultiComplex(raw json.RawMessage, defaultType string, valueKey s
 	// Case 1: Plain string -> wrap in array
 	var s string
 	if err := json.Unmarshal(raw, &s); err == nil {
-		return []map[string]interface{}{{valueKey: s, "type": defaultType, "primary": true}}
+		return []map[string]interface{}{{valueKey: s, "primary": true}}
```
sadilchamishka, line 610: For the new `address` multi-complex rule, a scalar address value produces `{value: ..., primary: true}`. Downstream `normalizeAndTranslateAddrOutbound` filters to canonical address sub-attributes, and `value` is not one, so the whole address entry gets dropped and never reaches the response. RFC 7643 represents the free-form address string as `formatted`. Consider mapping a plain-string address to `formatted` (or confirm scalar addresses are never expected here).

backend/internal/scim/schema_builder.go
ravindu439: Intentional, and thanks for catching the missing context. The `Meta.Version` (weak ETag) field was dropped from `mapEntityTypeToSCIMSchema` to align with the core User and Group schema builders (`buildCoreUserSchema` and `buildCoreGroupSchema`), neither of which populate `Version` either. The `/Schemas` endpoint is entirely read-only — schemas cannot be mutated via SCIM, so there is no concurrent-write scenario that requires optimistic locking via ETags. Carrying a version field would create a misleading signal that clients should honour it for conditional requests, when no SCIM-standard write path exists for these resources. Consistency across all three schema types (core User, core Group, and dynamic extension schemas) is the driving reason here. Will add a note to the PR description to make the rationale visible.

backend/internal/scim/core_attr_mapper.go
```diff
@@ -219,7 +226,7 @@ func translateSCIMFilterAttr(attr string) string {
 // scimUnsupportedMultiComplexSubAttrs: sub-attrs ThunderID never stores as a flat
 // field, regardless of scalar/array schema. "value" excluded — wrong only for
 // array schemas, undetectable at filter-parse time.
-var scimUnsupportedMultiComplexSubAttrs = []string{"type", "primary"}
+var scimUnsupportedMultiComplexSubAttrs = []string{scimTypeKey, "primary"}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/group_handler.go
```diff
@@ -121,13 +121,25 @@ func (h *scimGroupsHandler) HandleGroupsReplaceRequest(w http.ResponseWriter, r
...
+	for _, s := range payload.Schemas {
+		if strings.EqualFold(s, SCIMCoreGroupSchemaURN) {
+			hasGroupSchema = true
+			break
+		}
+	}
+	if !hasGroupSchema {
+		h.handleSCIMError(w, r, &ErrorMissingSchemas)
+		return
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/schema_builder.go
ravindu439: Good catch on the question. The two layers are intentionally complementary and together they do cover empty/null values — `missingRequiredAttrs` is intentionally key-presence-only, and `entitytype.Schema.Validate` enforces non-empty values downstream.
 Specifically, `Schema.Validate` (called immediately after `missingRequiredAttrs` in the write path) runs `prop.validateValue` on every present key. Each type's validator rejects the empty-but-present cases:
 - **`null`** — `json.Unmarshal` into `map[string]interface{}` produces a Go `nil` for a JSON `null`; all type validators do a type assertion (e.g. `value.(string)`) which returns `!ok` on `nil`, failing validation.
 - **`""`** (empty string) — the string validator accepts `""` unless a `regex`/`pattern` or `enum` constraint is defined; with no constraint, an empty string is technically a valid string value. This is by design: the schema language has no "non-empty" keyword, and enforcing non-empty at the SCIM layer would create a stricter policy than the underlying entity model itself applies. If a user type wants to forbid empty strings for a required field it should define a `regex` constraint.
 - **`[]`** (empty array) — `array.validateValue` explicitly checks `p.required && len(arrayValue) == 0` and returns `false` (line 64–67 in `array.go`), so an empty array on a required array field is caught.

 So `missingRequiredAttrs` is not redundant: it runs *before* the entity-layer validation and produces a specific, actionable SCIM error (`SCIM-1018`) naming the missing fields, rather than a generic schema validation failure. The two layers do agree.

backend/internal/scim/core_attr_mapper.go
ravindu439: fixed

sadilchamishka (approved):

Issue #4408, opened 2026-07-28
[SCIM] Implement SCIM `.search` Endpoint (RFC 7644 §3.4.3)


ravindu439:
### What problem are we solving?

When a SCIM client wants to perform a complex multi-attribute or cross-resource query, they cannot use a simple `GET /Users?filter=...` because HTTP GET does not support a request body. The SCIM RFC 7644 §3.4.3 defines `POST /.search` specifically for this case. Without it, provisioning tools that rely on body-based filter queries either fail silently or fall back to full resource fetches, causing unnecessary load and interoperability failures.

### Who are we solving this for?

**Primary: System administrators / enterprise IT integrators**
- Goal: Configure automated user provisioning from IdPs into ThunderID without manual intervention.
- Pain: Tools that issue `POST /.search` get a `404 Not Found` or `501 Method Not Implemented`, blocking the integration entirely.
**Secondary: ThunderID backend contributors**
- Goal: Achieve full RFC 7644 compliance for the SCIM implementation.
- Pain: Missing endpoint leaves the SCIM surface incomplete.

### Why should we solve this now?

- Value to Persona: Enterprise admins can complete their IdP integrations without workarounds; Okta/Azure AD provisioning works out of the box.
- Value to Org: Unlocks enterprise customer segments that gate on RFC 7644 compliance certifications.
- Urgency: The SCIM Users and Groups resources, filters, and core attribute mapping are now complete. `.search` is the primary remaining RFC gap before the SCIM surface is production-ready.

### Proposed Solution

Implement `POST /scim/v2/.search` (global) and optionally `POST /scim/v2/Users/.search` and `POST /scim/v2/Groups/.search` per RFC 7644 §3.4.3.
**Request body** (`application/scim+json`):
```json
{
  "schemas": ["urn:ietf:params:scim:api:messages:2.0:SearchRequest"],
  "filter": "userName eq \"bjensen\"",
  "startIndex": 1,
  "count": 10,
  "attributes": ["userName", "emails"]
}
```
**Key implementation points:**
- New handler: `POST /{resource}/.search` — parses `SearchRequest` schema, delegates filter/pagination to the existing `ListUsers` / `ListGroups` store layer (reuse current filter parsing).
- Response: standard `ListResponse` (`urn:ietf:params:scim:api:messages:2.0:ListResponse`).
- Reject unsupported operators with `400 scimError` (consistent with existing filter behaviour).
[...]

PR #4424, opened 2026-07-30
feat(scim): Add SCIM search request handling and improve pagination defaults

ravindu439:
### Purpose
Adds the SCIM `.search` endpoint (RFC 7644 §3.4.3) for Users, and along the way fixes a pagination-default inconsistency between the SCIM layer and the native ThunderID APIs, adds `attributes`/`excludedAttributes` response projection (RFC 7644 §3.9) to both `GET /Users` and the new search endpoint, and fixes an outbound attribute-mapping bug where multi-valued core attributes were given a fabricated `type`.

### Approach

**1. `POST /scim/v2/Users/.search`**
- New route registered in `init.go`, backed by `HandleUsersSearchRequest` in `users_handler.go`.
- Request body (`SCIMSearchRequest` in `model.go`) supports `filter`, `startIndex`, `count`, `attributes`, `excludedAttributes`. `sortBy`/`sortOrder` are accepted (no 400 if present) but intentionally not acted on — the underlying store has no `ORDER BY` support today, so wiring sort now would either be a no-op or require a real query-builder change; out of scope for this PR.
- `filter` reuses the existing `parseSCIMFilterForEq` (single `eq` expression only — same constraint as `GET /Users?filter=`).
- Pagination/filtering both route through the same `scimUsersService.ListUsers` the GET list handler already uses — no service-layer changes, no new query paths.
- New permission entry added: `POST /scim/v2/Users/.search` → `p.UserView` (read/list semantics, same tier as `GET /scim/v2/Users`), since a named path like this isn't covered by the existing exact-match/wildcard permission patterns.
- Request `schemas` is validated to include the `urn:ietf:params:scim:api:messages:2.0:SearchRequest` URN (not just non-empty), matching the schema-URN check other core SCIM endpoints already do.

**2. Pagination default fix**
- Found while wiring the search endpoint's default page size: the SCIM layer hardcoded `20` as the default `count` in five places (`users_handler.go` GET+search, `users_service.go`, `group_handler.go`, `group_service.go`), instead of reusing `serverconst.DefaultPageSize` (30) that the native `/users` and `/groups` APIs already use.
- Fixed all five spots to use `serverconst.DefaultPageSize`. `serverconst.MaxPageSize` (100) was already enforced downstream via the native service's `validatePaginationParams` — unchanged.
- Note: this changes the default `itemsPerPage` a SCIM client gets back when it omits `count`, from 20 to 30. Not a breaking change per SCIM spec (default page size is implementation-defined), but flagging since it's an observable behavior change for existing SCIM clients that don't specify `count`.

**3. `attributes` / `excludedAttributes` projection (RFC 7644 §3.9)**
- Added to both `GET /Users` (new `attributes=`/`excludedAttributes=` comma-separated query params) and `POST /Users/.search` (already-parsed body fields, previously accepted but unused).
- Scope: **top-level attribute names only** (e.g. `userName`, `emails`) — no dotted sub-attribute paths (e.g. `name.givenName`). `SCIMUser` is a typed struct with a custom `MarshalJSON` merging fixed core fields, a schema-driven `CoreAttrs` map, and an extension-schema blob; sub-path pruning would need a path-walking pruner over nested maps/slices, a much larger change deferred until there's real demand for it.
- Implementation: post-fetch, JSON-level projection (`projectSCIMAttributes`/`projectSCIMUserListResponse` in `users_resource.go`) — marshal each resource, prune top-level keys, done before the final response write. `id`/`schemas`/`meta` are always returned regardless of what's requested, per spec. `attributes` takes precedence over `excludedAttributes` when both are supplied.
- Comparison against requested attribute names is case-insensitive per RFC 7644 (e.g. `attributes=username` matches `userName`); original resource key casing is preserved in the response.
- Deliberately **not** done at the DB/store layer: the store is schema-driven (EAV-style dynamic attribute keys, not fixed columns) and already fetches the full record regardless of requested fields, so column-level projection would need to thread a field-selection param through the user service, store query builder, and schema-aware mapper for marginal gain. Runtime projection cost is bound by page size (`≤ MaxPageSize` = 100 rows), not table size, since pagination is already applied before this step — cheap either way.

**4. Attribute mapping fix: stop fabricating `type` on multi-valued core attributes**
- `core_attr_mapper.go` previously hardcoded a `type` sub-attribute (`"work"` for emails/phoneNumbers, `"photo"` for photos) on outbound multi-valued complex attributes whenever the underlying ThunderID user data had no type information at all.
- This is a response-shape change for `emails`, `phoneNumbers`, `photos`, and `addresses`: SCIM clients that read `type` off these entries will now see it absent instead of a fabricated value when ThunderID has no real type for that entry.
- Considered safe because `type` is an OPTIONAL sub-attribute per RFC 7643 §4.1.2 — omitting unknown metadata is spec-conformant, whereas synthesizing `"work"`/`"photo"` values ThunderID never actually recorded misrepresents the data to any client that trusts `type` for routing/display logic.
- Same commit also stops leaking non-canonical `addresses` sub-attributes (fields with no defined mapping to RFC 7643 `streetAddress`/`locality`/`region`/`postalCode`/`country`) into the response instead of silently passing them through.
- Covered by `core_attr_mapper_test.go`.

### Related Issues
- #4408 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/scim/users_handler.go
```diff
@@ -64,16 +68,76 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
...
+	if err != nil || len(body) == 0 {
+		h.handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
+	var searchReq SCIMSearchRequest
+	if err := json.Unmarshal(body, &searchReq); err != nil {
+		h.handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
+	if len(searchReq.Schemas) == 0 {
```
sadilchamishka: This only checks that `schemas` is non-empty, it never verifies the SearchRequest URN. `SCIMSearchSchemaURN` was added in constants.go for this but is unused outside tests, so any arbitrary schema string (or a totally unrelated URN) is accepted here. RFC 7644 3.4.3 expects `urn:ietf:params:scim:api:messages:2.0:SearchRequest`. Suggest validating that the array contains `SCIMSearchSchemaURN` (mirroring the Core User/Group schema checks elsewhere in this package), otherwise the constant is dead production code.

backend/internal/scim/users_resource.go
```diff
@@ -85,6 +85,72 @@ func buildSCIMUserListResponse(users []SCIMUser, totalResults, startIndex, items
...
+// projectSCIMAttributes prunes a marshaled SCIM resource's top-level keys per
+// RFC 7644 §3.9. attributes takes precedence over excludedAttributes when both
+// are supplied. id/schemas/meta are always kept.
+func projectSCIMAttributes(
+	resource map[string]interface{}, attributes, excludedAttributes []string,
+) map[string]interface{} {
+	projected := make(map[string]interface{}, len(resource))
+	if len(attributes) > 0 {
+		keep := make(map[string]struct{}, len(attributes))
+		for _, attr := range attributes {
```
sadilchamishka: Attribute-name matching here is case-sensitive (exact map key lookup), but SCIM attribute names are case-insensitive per RFC 7643 2.1. A client requesting `attributes=username` or `excludedAttributes=USERNAME` would silently miss `userName`, returning only id/schemas/meta. Consider lower-casing both the requested names and the resource keys before comparing (the parsed keys come from `parseCSVQueryParam`, which preserves case). Worth a test with a differently-cased attribute name.

backend/internal/scim/users_resource.go
```diff
@@ -85,6 +85,72 @@ func buildSCIMUserListResponse(users []SCIMUser, totalResults, startIndex, items
...
+			return nil, fmt.Errorf("projectSCIMUserListResponse: failed to unmarshal resource: %w", err)
+		}
+		resources = append(resources, projectSCIMAttributes(m, attributes, excludedAttributes))
+	}
+	return map[string]interface{}{
+		"schemas":      listResp.Schemas,
+		"totalResults": listResp.TotalResults,
+		"startIndex":   listResp.StartIndex,
+		"itemsPerPage": listResp.ItemsPerPage,
+		"Resources":    resources,
```
sadilchamishka: Minor/maintainability: the projected response hardcodes the list-response field set (`schemas`/`totalResults`/`startIndex`/`itemsPerPage`/`Resources`). If `SCIMUserListResponse` ever gains a field, the projected (attributes/excludedAttributes) path will silently drop it while the non-projected path keeps it, so the two response shapes drift. Consider marshaling `listResp` to a map once and pruning the `Resources` entries in place, so envelope fields stay in sync automatically.

backend/internal/scim/users_resource.go
```diff
@@ -85,6 +85,72 @@ func buildSCIMUserListResponse(users []SCIMUser, totalResults, startIndex, items
...
+		for _, attr := range attributes {
+			keep[attr] = struct{}{}
+		}
+		for k, v := range resource {
+			_, always := alwaysReturnedUserAttrs[k]
+			_, wanted := keep[k]
+			if always || wanted {
+				projected[k] = v
+			}
+		}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/users_handler.go
```diff
@@ -64,16 +68,76 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
...
+		h.handleSCIMError(w, r, &ErrorMissingSchemas)
+		return
+	}
+
```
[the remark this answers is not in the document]

backend/internal/scim/users_handler.go
ravindu439: fixed it already.

backend/internal/scim/core_attr_mapper.go
```diff
@@ -513,74 +607,95 @@ func normalizeToMultiComplex(raw json.RawMessage, defaultType string, valueKey s
 	// Case 1: Plain string -> wrap in array
 	var s string
 	if err := json.Unmarshal(raw, &s); err == nil {
-		return []map[string]interface{}{{valueKey: s, "type": defaultType, "primary": true}}
+		return []map[string]interface{}{{valueKey: s, "primary": true}}
 	}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/users_resource.go
ravindu439: fixed it

backend/internal/scim/users_resource.go
ravindu439: fixed

sadilchamishka (approved):

ThaminduDilshan (approved):

PR #4440, opened 2026-07-30
Show specific reasons for group update and create errors

JayaShakthi97:
### Purpose
The groups feature was not wired to the shared error catalog at all: every mutation (create, update, delete, add/remove member) surfaced either a generic fallback string or the raw backend error message (e.g. "Request failed with status code 409"), instead of a specific, actionable message.

### Approach
Mapped the backend `GRP-*` error codes that are actually reachable through the console UI (verified against the live create/edit/delete forms, client-side validation, and the group service) to specific messages via the existing shared `getErrorMessage` util — no new mapping mechanism was introduced.

**Codes mapped** (groups i18n namespace):

| Code | Reachability | Message |
|---|---|---|
| GRP-1004 | Guaranteed on create (duplicate name); also update rename | "A group with this name already exists in this organization unit. Choose a different name." |
| GRP-1015 | Race past the `isReadOnly` client guard (update/delete/members) | "This group is managed declaratively and cannot be edited or deleted." |
| GRP-1016 | Server declarative-only mode, not client-gated (create) | "New groups cannot be created while the server is running in declarative-only mode." |
| GRP-1003 | Race: group deleted by another session (update/delete/members) | "This group no longer exists. It may have already been deleted." |
| GRP-1007 | Race: selected member entity deleted before add completes | "One or more selected members no longer exist. Refresh and try again." |

**Full decision matrix** (all `GRP-*` codes, selected and rejected, with reasoning):

| Code | Selected? | Reason |
|---|---|---|
| GRP-1004 name conflict | ✅ | Guaranteed reachable on create; also update rename. No client uniqueness check. |
| GRP-1015 immutable declarative | ✅ | Race past `isReadOnly` client guard; server re-checks on update/delete/members. |
| GRP-1016 declarative-only create | ✅ | Not client-gated; every create fails when server is declarative-only. |
| GRP-1003 not found | ✅ | Race: group deleted by another session; hits update/delete/add/remove. |
| GRP-1007 invalid member ID | ✅ | Race: selected user/app/agent deleted before add completes. |
| GRP-1005 invalid OU | ❌ | Race-only, low frequency; OU chosen from server tree picker, edit page never changes OU. |
| GRP-1001 invalid format | ❌ | Body always well-formed JSON; empty-name triggers blocked client-side. |
| GRP-1002 missing group ID | ❌ | Edit/delete/member routes always carry an id. |
| GRP-1006 cannot delete | ❌ (removed) | No live service path emits it; dead constant. Deleted in this PR. |
| GRP-1008 invalid group member ID | ❌ | Group members can't be added via the console (dialog offers only user/app/agent). |
| GRP-1011 invalid limit | ❌ | DataGrid drives pagination; always integer. |
| GRP-1012 invalid offset | ❌ | Same. |
| GRP-1013 empty members | ❌ | Add button disabled at 0 selection; remove always sends one. |
| GRP-1014 invalid member type | ❌ | UI only sends user/app/agent, all valid types. |

**Other changes:**
- Consolidated each action to a single mapped error surface. Previously several actions (update, delete, add/remove member) showed the same failure twice, once via a generic hook-level toast and once via a raw-message inline display. The hooks no longer show an error toast; the owning component (dialog/page) is now the single mapped surface.
- Refactored the four error-surfacing components (`CreateGroupPage`, `GroupDeleteDialog`, `EditMembersSettings`, `GroupEditPage`) to `useTranslation('groups')` instead of the default namespace with manual `groups:` prefixes, so `getErrorMessage` resolves against a namespace-scoped `t` directly.
- Removed the dead backend constant `ErrorCannotDeleteGroup` (GRP-1006), which no live service path in `groupService` produces, only referenced in the handler's status-code switch and a test mock. Removed its handler case, its test case, and its stale example from `api/group.yaml`. Regenerated `backend/internal/system/i18n/core/defaults.go` accordingly.

No new translation keys duplicate existing ones, the generic per-action fallback keys (`create.error`, `update.error`, etc.) already existed and remain the `getErrorMessage` fallback for unmapped codes.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4441

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Bug Fixes

* Improved group creation, editing, deletion, and member-management error messages with clearer, user-friendly guidance.
* Added localized messages for duplicate names, missing members or groups, and declaratively managed groups.
* Updated group dialogs and member-management screens to display standardized errors.
* Clarified restrictions for groups managed declaratively, including editing and deletion.

## Documentation

* Updated API error examples and group-related translations to reflect supported error scenarios.

### DIFF
--- a/api/group.yaml
+++ b/api/group.yaml
@@ -416,16 +416,6 @@ paths:
                     description:
                       key: "error.groupservice.missing_group_id_description"
                       defaultValue: "Group ID is required"
-                cannot-delete-group:
-                  summary: Cannot delete group
-                  value:
-                    code: "GRP-1006"
-                    message:
-                      key: "error.groupservice.cannot_delete_group"
-                      defaultValue: "Cannot delete group"
-                    description:
-                      key: "error.groupservice.cannot_delete_group_description"
-                      defaultValue: "Cannot delete group with child groups"
         "404":
           description: Group not found
           content:
--- a/backend/internal/group/handler.go
+++ b/backend/internal/group/handler.go
@@ -415,7 +415,7 @@ func (gh *groupHandler) handleError(ctx context.Context, w http.ResponseWriter, 
 			statusCode = http.StatusNotFound
 		case ErrorGroupNameConflict.Code:
 			statusCode = http.StatusConflict
-		case ErrorInvalidOUID.Code, ErrorCannotDeleteGroup.Code,
+		case ErrorInvalidOUID.Code,
 			ErrorInvalidRequestFormat.Code, ErrorMissingGroupID.Code,
 			ErrorInvalidLimit.Code, ErrorInvalidOffset.Code,
 			ErrorEmptyMembers.Code, ErrorInvalidMemberType.Code,
--- a/frontend/apps/console/src/features/groups/components/GroupDeleteDialog.tsx
+++ b/frontend/apps/console/src/features/groups/components/GroupDeleteDialog.tsx
@@ -16,6 +16,7 @@
  * under the License.
  */
 
+import {getErrorMessage} from '@thunderid/utils';
 import {Dialog, DialogTitle, DialogContent, DialogContentText, DialogActions, Button, Alert} from '@wso2/oxygen-ui';
 import {useState, type JSX} from 'react';
 import {useTranslation} from 'react-i18next';
@@ -37,7 +38,7 @@ export default function GroupDeleteDialog({
   onClose,
   onSuccess = undefined,
 }: GroupDeleteDialogProps): JSX.Element {
-  const {t} = useTranslation();
+  const {t} = useTranslation('groups');
   const deleteGroup = useDeleteGroup();
   const [error, setError] = useState<string | null>(null);
 
@@ -58,18 +59,20 @@ export default function GroupDeleteDialog({
         onSuccess?.();
       },
       onError: (err: Error) => {
-        setError(err.message ?? t('groups:delete.error'));
+        setError(getErrorMessage(err, t, 'delete.error'));
       },
     });
   };
 
   return (
     <Dialog open={open} onClose={handleCancel} maxWidth="sm" fullWidth>
-      <DialogTitle>{t('groups:delete.title')}</DialogTitle>
+      <DialogTitle>{t('delete.title', 'Delete Group')}</DialogTitle>
       <DialogContent>
-        <DialogContentText sx={{mb: 2}}>{t('groups:delete.message')}</DialogContentText>
+        <DialogContentText sx={{mb: 2}}>
+          {t('delete.message', 'Are you sure you want to delete this group?')}
+        </DialogContentText>
         <Alert severity="warning" sx={{mb: 2}}>
-          {t('groups:delete.disclaimer')}
+          {t('delete.disclaimer', 'This action cannot be undone. All group associations will be permanently removed.')}
         </Alert>
         {error && (
           <Alert severity="error" sx={{mt: 2}}>
--- a/frontend/apps/console/src/features/groups/components/__tests__/EditMembersSettings.test.tsx
+++ b/frontend/apps/console/src/features/groups/components/__tests__/EditMembersSettings.test.tsx
@@ -181,7 +181,31 @@ describe('EditMembersSettings', () => {
     await user.click(screen.getByText('Add'));
 
     await waitFor(() => {
-      expect(screen.getByText('Add failed')).toBeInTheDocument();
+      expect(screen.getByText('Failed to add member. Please try again.')).toBeInTheDocument();
+    });
+  });
+
+  it('should show mapped error message when a selected member no longer exists', async () => {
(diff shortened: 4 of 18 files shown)


backend/internal/group/handler.go
```diff
@@ -415,7 +415,7 @@ func (gh *groupHandler) handleError(ctx context.Context, w http.ResponseWriter,
 			statusCode = http.StatusNotFound
 		case ErrorGroupNameConflict.Code:
 			statusCode = http.StatusConflict
-		case ErrorInvalidOUID.Code, ErrorCannotDeleteGroup.Code,
+		case ErrorInvalidOUID.Code,
```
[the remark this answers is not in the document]
JayaShakthi97: Checked: `docs/content/apis.mdx` is a 12-line stub that renders `<ApiVersionReference />`, which pulls a combined OpenAPI spec generated from `api/*.yaml` (including `api/group.yaml`, which this PR already updates to drop the GRP-1006 example). There's no hand-maintained error-code documentation in this file to edit. No doc change needed here.

frontend/apps/console/src/features/groups/components/__tests__/EditMembersSettings.test.tsx
```diff
@@ -181,7 +181,31 @@ describe('EditMembersSettings', () => {
...
 
     await waitFor(() => {
-      expect(screen.getByText('Add failed')).toBeInTheDocument();
+      expect(screen.getByText('Failed to add member. Please try again.')).toBeInTheDocument();
+    });
+  });
+
+  it('should show mapped error message when a selected member no longer exists', async () => {
+    mockAddMutate.mockImplementation((_data: unknown, opts: {onError: (err: Error) => void}) => {
+      const error = new Error('Request failed') as Error & {response?: {data?: {code: string}}};
```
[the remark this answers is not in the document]
JayaShakthi97: Checked: each file has exactly one `const error` declaration, scoped inside its own separate `it()` callback (`EditMembersSettings.test.tsx` L190, `GroupDeleteDialog.test.tsx` L117). There's no shared scope where they collide, and both test suites compile and pass. No change needed.

frontend/apps/console/src/features/groups/components/GroupDeleteDialog.tsx
```diff
@@ -58,18 +59,18 @@ export default function GroupDeleteDialog({
...
   return (
     <Dialog open={open} onClose={handleCancel} maxWidth="sm" fullWidth>
-      <DialogTitle>{t('groups:delete.title')}</DialogTitle>
+      <DialogTitle>{t('delete.title')}</DialogTitle>
       <DialogContent>
-        <DialogContentText sx={{mb: 2}}>{t('groups:delete.message')}</DialogContentText>
+        <DialogContentText sx={{mb: 2}}>{t('delete.message')}</DialogContentText>
         <Alert severity="warning" sx={{mb: 2}}>
-          {t('groups:delete.disclaimer')}
+          {t('delete.disclaimer')}
```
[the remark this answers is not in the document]
JayaShakthi97: Fixed: added a fallback default string to every t() call flagged here, per frontend/AGENTS.md's i18n Fallback Values convention. Pushed.

api/group.yaml
```diff
@@ -416,16 +416,6 @@ paths:
                     description:
                       key: "error.groupservice.missing_group_id_description"
                       defaultValue: "Group ID is required"
-                cannot-delete-group:
```
Malith-19, line 419: Aren't we including the specific errors in the spec? May be another one or two?

frontend/packages/i18n/src/locales/en-US.ts
```diff
@@ -1546,6 +1506,11 @@ const translations = {
     'delete.success': 'Group deleted successfully.',
     'addMember.success': 'Member added successfully.',
     'removeMember.success': 'Member removed successfully.',
+    'errors.GRP-1003': 'This group no longer exists. It may have already been deleted.',
```
Malith-19, line 1549:
```suggestion
    'error.GRP-1003': 'This group no longer exists. It may have already been deleted.',
```

Since we mentioning about single error I feel we should go with singular phrases here. WDYT? Is this the pattern we follow in other places as well?

JayaShakthi97: This is the pattern we have followed so far.

api/group.yaml
JayaShakthi97: This error code (GRP-1006) is deleted as dead code.

JayaShakthi97: @coderabbitai check whether there are error codes not documented in the API doc. This is only for the groups feature. If there are missing ones, create a doc issue mentioning to add them as well. Note that some error codes are removed via this PR.

DonOmalVindula (dismissed):

DonOmalVindula (approved):

PR #4463, opened 2026-07-31
Moved RuntimeCryptoProvider interface to pkg/thunderidengine

anushasunkada:
### Purpose

1. Moved RuntimeCryptoProvider interface and related models to pkg/thunderidengine.
2. Required code restructuring to use RuntimeCryptoProvider instead of cryptolib in flow and oauth packages. 

### Related Issues
- #3037 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Security**
  * Strengthened signing, verification, encryption, and decryption across JWT, JWE, DPoP, and credential workflows.
  * Expanded support for RSA, elliptic-curve, Ed25519, and post-quantum keys.
  * HTTPS startup now validates required TLS capabilities and stops safely when unavailable.

* **Bug Fixes**
  * Improved encrypted-token key selection and public-key discovery.
  * Added clearer handling for missing keys, unsupported algorithms, and invalid key data.

* **Reliability**
  * Expanded cryptographic validation and round-trip coverage across authentication and data-protection flows.


backend/pkg/thunderidengine/providers/interface.go
```diff
@@ -274,3 +274,27 @@ type Transactioner interface {
...
+	Sign(ctx context.Context, keyRef KeyRef, alg string, content []byte) ([]byte, error)
+
+	// Verify verifies the signature of the given content using the specified key reference and algorithm.
+	Verify(ctx context.Context, keyRef KeyRef, alg string, content, signature []byte) error
+
+	// GetPublicKeys retrieves public keys based on the provided filter criteria.
+	GetPublicKeys(ctx context.Context, filter PublicKeyFilter) ([]PublicKeyInfo, error)
+
+	// IsSupportedSigningAlgorithm checks if the specified signing algorithm is supported.
+	IsSupportedSigningAlgorithm(alg string) bool
```
senthalan: can't this be part of the Sign function implementation ?

If the underlying crypto provider don't support the provided algorithm, sign function cab return `ErrUnsupportedAlgorithm`

backend/internal/system/jose/sdjwt/sdjwt.go
```diff
@@ -508,68 +507,19 @@ func decodeJWTPayload(token string) (map[string]interface{}, error) {
...
 }
+
+// mapAlgorithmToSignAlg maps JWS alg header values to internal SignAlgorithm.
+func mapAlgorithmToSignAlg(jwsAlg string) (cryptolib.SignAlgorithm, error) {
+	signAlg, err := cryptolib.SignAlgorithmFor(cryptolib.Algorithm(defaultkm.JWSAlgorithm(jwsAlg)))
+	if err != nil {
+		return "", fmt.Errorf("unsupported JWS alg: %s", jwsAlg)
+	}
+	return signAlg, nil
+}
```
senthalan, line 525: Seems like we have the cryptolib reference here, shouldn't this also handled through the provider interface

backend/pkg/thunderidengine/providers/interface.go
anushasunkada: in openid4vc and opid4vp init.go files signing algorithm validations are done. is that ok to remove "fail fast" checks?

backend/internal/system/jose/jwe/utils.go
```diff
@@ -260,63 +262,22 @@ func pkcs7Unpad(data []byte) ([]byte, error) {
...
-	ecdhPub, err := jwkToECPublicKey(epkMap)
+	pub, err := defaultkm.JWKToPublicKey(epkMap)
+	if err != nil {
+		return nil, fmt.Errorf("invalid epk in header: %w", err)
+	}
+	ecdsaPub, ok := pub.(*ecdsa.PublicKey)
+	if !ok {
+		return nil, fmt.Errorf("invalid epk in header: not an EC public key")
+	}
+	ecdhPub, err := ecdsaPub.ECDH()
```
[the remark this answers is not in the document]
anushasunkada: @senthalan  I was also in the same thought of moving JWKToPublicKey to runtimeCryptoProvider contract. but what do you think?

backend/internal/system/jose/sdjwt/sdjwt.go
```diff
@@ -183,7 +182,7 @@ func VerifyKeyBinding(p *Presentation, cred *VerifiedCredential, opts VerifyOpti
 		return ErrMissingConfirmationKey
 	}
 
-	holderKey, err := jwkToPublicKey(cred.ConfirmationKey)
+	holderKey, err := defaultkm.JWKToPublicKey(cred.ConfirmationKey)
```
senthalan, line 185: Seems like, we have the defaultkm references here. I thought this effort is to remove this and always go though the provider interface

backend/pkg/thunderidengine/providers/interface.go
thiva-k: It is added to provide a meaningful error at startup, rather than a generic 500 in runtime. `newJWTService` also needs this check anyways right? So better keep it unless we want to further simplify the interface

backend/internal/system/jose/sdjwt/sdjwt.go
anushasunkada: This needs sdjwt restructuring, so currently ignoring this change @senthalan @thiva-k

backend/pkg/thunderidengine/providers/interface.go
anushasunkada: OK, I will replace this with the new GetSupportedSigningAlgorithms() method check

backend/internal/oauth/oauth2/jwksresolver/resolver.go
```diff
@@ -171,25 +170,35 @@ func (r *Resolver) parseEncryptionKeyFromJWKS(ctx context.Context,
...
 }
 
+// keyTypeForEncryptionAlg returns the JWK "kty" required by the given JWE key encryption
+// algorithm: "RSA" for RSA-OAEP variants, "EC" for ECDH-ES variants.
+func keyTypeForEncryptionAlg(encryptionAlg string) string {
+	if strings.HasPrefix(encryptionAlg, "ECDH-ES") {
+		return "EC"
+	}
+	return "RSA"
+}
```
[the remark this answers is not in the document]
anushasunkada: This currently is not required to be documented.

backend/internal/system/jose/jwe/service.go
```diff
@@ -156,23 +161,40 @@ func (js *jweService) Decrypt(ctx context.Context, jweToken string) ([]byte, *ti
...
+		params[providers.ParamAPV] = apv
+	}
+
 	// Decrypt CEK via the runtime crypto provider (uses server's private key).
-	cek, err := js.cryptoProvider.Decrypt(ctx, &js.keyRef, params, encryptedKey)
+	cek, err := js.cryptoProvider.Decrypt(ctx, &js.keyRef, algStr, params, encryptedKey)
 	if err != nil {
 		js.logger.Error(ctx, "Failed to decrypt CEK: "+err.Error())
 		return nil, &ErrorJWEDecryptionFailed
 	}
```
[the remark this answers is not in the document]
anushasunkada: Fixed

backend/internal/system/jose/jwe/utils.go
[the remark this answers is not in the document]
anushasunkada: Removed from backend/internal/system/jose/jwe/utils.go, sdjwt clean up is left open for now.

backend/pkg/thunderidengine/providers/constants.go
```diff
@@ -534,3 +534,35 @@ var (
...
+// Runtime crypto provider errors
+var (
+	// ErrKeyNotFound indicates that no key managed by the provider matches the
+	// requested identifier.
+	ErrKeyNotFound = errors.New("kmprovider: no key found matching the requested identifier")
+
+	// ErrUnsupportedAlgorithm indicates the requested signature algorithm is not
+	// supported by the provider or is incompatible with the resolved key.
+	ErrUnsupportedAlgorithm = errors.New("kmprovider: unsupported signature algorithm")
+)
```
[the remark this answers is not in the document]
anushasunkada: Fixed

backend/internal/system/jose/jwt/service.go
```diff
@@ -334,16 +331,19 @@ func (js *jwtService) VerifyJWTSignatureWithPublicKey(jwtToken string,
...
-	alg, err := jws.MapAlgorithmToSignAlg(jws.Algorithm(algStr))
-	if err != nil {
-		return &ErrorUnsupportedJWSAlgorithm
-	}
 
-	// Verify the signature
-	err = cryptolib.Verify([]byte(signingInput), signature, alg, jwtPublicKey)
-	if err != nil {
+	// Verify the signature through the provider (uses provided public key).
+	if err = js.cryptoProvider.Verify(context.Background(), keyRef,
```
senthalan: The context needs to be passed to the VerifyJWTSignatureWithPublicKey function from the callers and then should be passed to the cryptoProvider.Verify(..) without passing the background.

backend/internal/system/kmprovider/defaultkm/jwk_test.go
senthalan, line 1: why we have this as a separate test file ?

I assume this test the new JWKToPublicKey function added to the runtime crypto provider, so can't we add these tests to the existing runtime_crypto_provider_test.go

backend/internal/system/kmprovider/defaultkm/runtime_crypto_provider.go
```diff
@@ -340,3 +484,156 @@ func (s *runtimeCryptoService) getECPrivateKey(
 	}
 	return ecPriv, nil
 }
+
+// JWKToPublicKey converts a JWK map to a crypto.PublicKey supporting RSA, EC, and Ed25519.
+func JWKToPublicKey(jwk map[string]interface{}) (crypto.PublicKey, error) {
```
senthalan, line 489: Why this function is public, I assume it can be private

backend/internal/oauth/jwks/service.go
```diff
@@ -148,8 +148,9 @@ func getRSAPublicKeyJWKS(pub *rsa.PublicKey, kid string, x5c []string, x5t, x5tS
 // getECDSAPublicKeyJWKS converts an ECDSA public key to JWKS format.
 func getECDSAPublicKeyJWKS(pub *ecdsa.PublicKey, kid string, x5c []string, x5t, x5tS256 string) JWKS {
 	crv := pub.Curve.Params().Name
-	x := encodeBase64URL(pub.X.Bytes())
-	y := encodeBase64URL(pub.Y.Bytes())
+	coordLen := (pub.Curve.Params().BitSize + 7) / 8
+	x := encodeBase64URL(padCoordinate(pub.X.Bytes(), coordLen))
+	y := encodeBase64URL(padCoordinate(pub.Y.Bytes(), coordLen))
```
senthalan, line 153: looks like this change is not related to the crypto provider moving. What is the reason for this change?

senthalan:
**Ownership of the `alg` / `enc` supported-algorithm lists**

Following on the `validateUserInfoConfig` / discovery change: right now `alg` comes from `cryptoProvider.GetSupportedEncryptionAlgorithms()` while `enc` comes from `jwe.SupportedContentEncryptionAlgorithms()`. Rather than making both come from the provider, I think the cleaner split is the other way around, keeping the provider a pure crypto primitive:

- **Provider = capability only.** `GetSupportedEncryptionAlgorithms()` returns everything it can encrypt with, asymmetric and symmetric alike (RSA-OAEP, ECDH-ES, A128GCM, A128CBC-HS256, and so on), with no notion of JWE roles. Whether an algorithm is "key management" or "content encryption" is an RFC 7516 concept, so the crypto layer shouldn't own it.
- **jwe owns the classification.** The jwe service (which already has the provider injected) buckets the provider's flat capability into the two JWE roles using its existing `KeyEncAlgorithm` / `ContentEncAlgorithm` constants, and exposes:

```go
SupportedKeyEncryptionAlgorithms()     []string // alg
SupportedContentEncryptionAlgorithms() []string // enc (evolves today's package func)
```

Each can return `jwe-known-role ∩ provider-capability`, so we never advertise an algorithm the provider can't actually perform.

- **Consumers read both from jwe.** `validateUserInfoConfig`, `validateIDTokenConfig`, and the discovery metadata get both lists from the jwe service, so they're symmetric. The `cryptoProvider` this PR threads into the validation path for the alg check would be replaced by the jwe service.

Net effect: the provider stays a JOSE-free primitive, and the alg/enc taxonomy lives in the one place that owns JWE semantics. This does keep `jose/jwe` as a dependency of the OAuth-layer packages, which I think is correct: discovery and client validation are JOSE concerns, so depending on the JOSE layer for algorithm metadata is appropriate.

It's a bigger reshape than the original point (it changes the provider method's contract, adds a jwe classifier, and points the two consumers at the jwe service), so it might be cleaner as a follow-up rather than blocking this PR. But since the provider abstraction is exactly what this PR is establishing, it could be worth getting the ownership boundary right now. What do you think?

backend/internal/oauth/jwks/service.go
anushasunkada: This fixed - https://github.com/thunder-id/thunderid/pull/4463#discussion_r3689352623

backend/internal/system/kmprovider/defaultkm/runtime_crypto_provider.go
anushasunkada: Nope @senthalan it is still used in sdjwt.go

anushasunkada:
> **Ownership of the `alg` / `enc` supported-algorithm lists**
> 
> Following on the `validateUserInfoConfig` / discovery change: right now `alg` comes from `cryptoProvider.GetSupportedEncryptionAlgorithms()` while `enc` comes from `jwe.SupportedContentEncryptionAlgorithms()`. Rather than making both come from the provider, I think the cleaner split is the other way around, keeping the provider a pure crypto primitive:
> 
> * **Provider = capability only.** `GetSupportedEncryptionAlgorithms()` returns everything it can encrypt with, asymmetric and symmetric alike (RSA-OAEP, ECDH-ES, A128GCM, A128CBC-HS256, and so on), with no notion of JWE roles. Whether an algorithm is "key management" or "content encryption" is an RFC 7516 concept, so the crypto layer shouldn't own it.
> * **jwe owns the classification.** The jwe service (which already has the provider injected) buckets the provider's flat capability into the two JWE roles using its existing `KeyEncAlgorithm` / `ContentEncAlgorithm` constants, and exposes:
> 
> ```go
> SupportedKeyEncryptionAlgorithms()     []string // alg
> SupportedContentEncryptionAlgorithms() []string // enc (evolves today's package func)
> ```
> 
> Each can return `jwe-known-role ∩ provider-capability`, so we never advertise an algorithm the provider can't actually perform.
> 
> * **Consumers read both from jwe.** `validateUserInfoConfig`, `validateIDTokenConfig`, and the discovery metadata get both lists from the jwe service, so they're symmetric. The `cryptoProvider` this PR threads into the validation path for the alg check would be replaced by the jwe service.
> 
> Net effect: the provider stays a JOSE-free primitive, and the alg/enc taxonomy lives in the one place that owns JWE semantics. This does keep `jose/jwe` as a dependency of the OAuth-layer packages, which I think is correct: discovery and client validation are JOSE concerns, so depending on the JOSE layer for algorithm metadata is appropriate.
> 
> It's a bigger reshape than the original point (it changes the provider method's contract, adds a jwe classifier, and points the two consumers at the jwe service), so it might be cleaner as a follow-up rather than blocking this PR. But since the provider abstraction is exactly what this PR is establishing, it could be worth getting the ownership boundary right now. What do you think?

Makes sense, they're both JWE concepts. The new classifier methods on jweService and rewiring the 3 consumers doesn't touch the RuntimeCryptoProvider interface itself, so there's no coupling forcing it into this PR — We can do it as a separate follow-up.

defaultkm's implementation — it currently under-reports. It only lists RSA-OAEP/RSA-OAEP-256/ECDH-ES variants, but cryptolib already defines and Encrypt already handles A128KW/A192KW/A256KW/A128GCMKW/etc. Broadening the returned list to match real capability is a body-only change, no signature change. In JWE service, we should filter out symmetric algorithms.

backend/internal/system/kmprovider/defaultkm/jwk_test.go
anushasunkada: done

backend/internal/system/jose/jwt/service.go
anushasunkada: Done

senthalan (approved):

PR #4479, opened 2026-07-31
Replace nested token tabs with an audience selector

Dilusha-Madushan:
### Purpose

The Token tab on the application edit page moves from nested Application/User sub-tabs to a side audience selector, per the finalized design. The same selector is applied to the agent edit Token tab, which shared the old layout.

This also fixes a bug the restructure surfaced: a mobile application registered with app-native (Bring Your Own UI) sign-in had **both** token sub-tabs frozen, with no way to unfreeze either. Such an application is created with no OAuth
configuration, so the grant-type check behind both locks could never be satisfied, and the lock notice pointed at the Advanced tab, which hides its OAuth section in exactly that state. The user pane is the one place its claims
and validity can be configured, so those settings were unreachable.

Two smaller fixes in the same area:

- Opening the user pane for an app-native application reported an unsaved change the moment it rendered, before anything was edited.
- Token copy referenced grant names and protocol vocabulary, and the shared card described ID tokens, user info responses, and response formats even in assertion mode, where none of them exist.

### Approach

**One shared selector, two lock presentations.** `TokenAudienceSelector` lives in
`applications/components/common/` and is consumed by both edit pages, following
the existing dependency direction (agents import from applications, never the
reverse). It renders the audience list and the content pane but takes no
position on what a locked audience shows, because the application and agent
lock conditions are different things: a missing grant type versus Delegated mode
being off. Each caller supplies its own notice.

**Locked audiences stay selectable.** Disabling the entry meant its explanation
had to be displayed somewhere else, which put a notice about one audience in
front of someone working on the other. Each audience now renders its own notice
in place of its settings, so the explanation is only visible to someone who
opened that audience. This replaces the previous frozen-card treatment on both
pages, including the agent's `SettingsLockNotice` usage.

**Gating on OAuth-config existence, not application type.** The user audience
unlocks when `!isOAuthTokenMode(oauth2Config) || hasUserAccess(grantTypes)`.
Application type was considered and rejected: the backend never validates grant
types against type, so a `mobile`-typed application can legitimately be
reconfigured as a confidential client, and the type is immutable. Grants and
config shape stay the authoritative signal. `isOAuthTokenMode` was extracted as
a shared helper so the lock decision and the render branch inside
`EditTokenSettings` cannot drift apart.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4435
- https://github.com/thunder-id/thunderid/issues/4357

### Related PRs
- https://github.com/thunder-id/thunderid/pull/4333

## Summary by CodeRabbit

* **New Features**
  * Added audience-based token settings for agent and application configurations.
  * Added selectable audience panels with descriptions, accessibility labels, lock indicators, and keyboard navigation.

* **Improvements**
  * Clarified token attribute and profile descriptions across native, OAuth, agent, and application flows.
  * Improved responsive layouts for larger screens.
  * Added clearer notices and Advanced-tab navigation for unavailable settings.

* **Bug Fixes**
  * Locked audience settings no longer affect validation or save-state behavior.
  * Prevented unnecessary updates when assertion validity remains unchanged.

### DIFF
--- a/frontend/apps/console/src/features/agents/components/edit-agent/tokens/AgentAccessTokenSection.tsx
+++ b/frontend/apps/console/src/features/agents/components/edit-agent/tokens/AgentAccessTokenSection.tsx
@@ -64,7 +64,7 @@ export default function AgentAccessTokenSection({
         attributesTitle: t('agents:edit.tokens.agent.attributes.title', 'Access Token Attributes'),
         attributesDescription: t(
           'agents:edit.tokens.agent.attributes.description',
-          'Attributes included in the access token this agent receives for its own requests (client_credentials grant).',
+          'Extra attributes to add to the access token this agent receives for itself.',
         ),
         attributesLabel: t('agents:edit.tokens.agent.attributes.label', 'Add or Remove Attributes'),
         attributesHint: t(
--- a/frontend/apps/console/src/features/agents/components/edit-agent/tokens/EditTokensSettings.tsx
+++ b/frontend/apps/console/src/features/agents/components/edit-agent/tokens/EditTokensSettings.tsx
@@ -3,13 +3,19 @@
 
 import {OAuth2GrantTypes} from '@thunderid/configure-applications';
 import type {Application} from '@thunderid/configure-applications';
-import {Box, Stack, Tab, Tabs} from '@wso2/oxygen-ui';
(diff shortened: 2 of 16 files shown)


Dilusha-Madushan:
New view for the Application/Agent Token tab:

image
image
image
image

jeradrutnam (approved):

PR #4480, opened 2026-07-31
Signal sign-out confirmation with the prompt action type

madurangasiriwardena:
### Purpose
The session sign-out executor told a confirmed sign-out apart from the initial request by persisting a `logoutPromptShown` marker in RuntimeData. This addresses @ThaminduDilshan's review comment on #4299: the confirmation prompt already knows which button the End-User pressed, so it can forward that as the prompt action's type and the executor can read it straight off the flow definition. No marker has to be stored.

This also restructures the executor so the confirmation prompt can grow further actions (a sign-out cancel button is planned), and stops the API documentation from enumerating action-type values.

### Approach
**Read the decision from the action type.** `common.ActionTypeSignOutConfirm` (`SIGN_OUT_CONFIRM`) marks the confirmation prompt's action edge. The prompt node already forwards a matched action's type on `ForwardedData[actionType]` (`prompt_node.go:154`), so on the re-run after the End-User confirms, the executor sees it and terminates. `RuntimeKeyLogoutPromptShown` is gone.

**Dispatch on the type rather than compare against one value.** The previous guard was `actionType != SIGN_OUT_CONFIRM`, which collapsed three distinct cases (no type, the confirm type, some other type) into one boolean and would have sent any future action type down the prompt-again path. `decide` now returns a named `signOutOutcome` from a switch keyed on the action type:

```go
switch common.ActionType(actionType) {
case common.ActionTypeSignOutConfirm:
    return signOutTerminate
default:
    // The initial request forwards no action type at all, and a type this executor does not
    // recognize is not consent to end the session. Both ask the End-User to confirm.
    return signOutPrompt
}
```

Adding the planned cancel action is then one constant, one `case`, one `signOutOutcome`, and its handling in `Execute`. The `default` branch is deliberately fail closed: an unrecognized type never terminates a session, so adding a type to a flow before the executor handles it cannot sign a user out unintentionally.

**Mark the template's confirm button.** The Conditional Sign Out Flow template carries the type on the prompt action (which the engine reads) and on the button element (which the flow builder's Action selector reads and re-emits on save). Without the element field, opening that template in the builder and saving it would silently drop the type and regress the flow into an endless confirmation loop.

**Stop enumerating action types in the docs.** The `prompts[].action.type` description listed `SUBMIT` and `REJECT` in three places, while the same description also said arbitrary strings are accepted. The enumerations were therefore never the contract, and every new action type would need edits in all three. They are replaced with a structural description: the value is opaque, and the executor that consumes it owns the vocabulary.

### Related Issues
- N/A

### Related PRs
- https://github.com/thunder-id/thunderid/pull/4299
- https://github.com/thunder-id/thunderid/pull/4456

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Improved sign-out confirmation handling to reliably distinguish confirmed actions from unrecognized requests.
  * Added support for executor-defined action types in flow configurations.

* **Bug Fixes**
  * Unrecognized sign-out actions now prompt for confirmation again instead of ending the session.

* **Documentation**
  * Updated flow action documentation to describe action types as executor-defined values rather than fixed options.

### DIFF
--- a/backend/internal/flow/common/constants.go
+++ b/backend/internal/flow/common/constants.go
@@ -248,10 +248,6 @@ const (
 	// requested without a valid id_token_hint. A sign-out flow's session sign-out node reads it to
 	// decide whether the End-User must confirm the logout before the session is terminated.
 	RuntimeKeyLogoutPromptRequired = "logoutPromptRequired"
-	// RuntimeKeyLogoutPromptShown is the session sign-out node's own guard, set when it routes to the
-	// confirmation prompt so that on re-run (after the user confirms) it terminates instead of
-	// prompting again.
-	RuntimeKeyLogoutPromptShown = "logoutPromptShown"
 )
 
 // SSOCheckpointKey scopes a per-checkpoint SSO control key (RuntimeKeySSOSessionPresent,
@@ -291,6 +287,11 @@ const (
 	ActionTypeSubmit ActionType = "SUBMIT"
 	// ActionTypeReject represents a reject/deny action
 	ActionTypeReject ActionType = "REJECT"
+	// ActionTypeSignOutConfirm marks the confirmation prompt's action edge in a sign-out flow. When the
+	// End-User confirms, the prompt node forwards this type to the session sign-out node (as the action
+	// type in ForwardedData), which reads it to tell a confirmed re-run apart from the initial request,
+	// so no runtime flag has to be persisted.
+	ActionTypeSignOutConfirm ActionType = "SIGN_OUT_CONFIRM"
 )
 
 // ForwardedData key constants define keys used in the ForwardedData map.
--- a/backend/internal/flow/executor/session_signout_executor.go
+++ b/backend/internal/flow/executor/session_signout_executor.go
@@ -65,7 +65,7 @@ func newSessionSignOutExecutor(flowFactory core.FlowFactoryInterface, sso sessio
 // a valid id_token_hint (RuntimeKeyLogoutPromptRequired), the executor first routes to the node's
 // onIncomplete confirmation prompt and only terminates the session once the End-User confirms. This
 // keeps the confirmation logic in the executor rather than a node condition the flow editor cannot
-// represent.
+// represent. See decide for how the prompt's action types map to outcomes.
 func (e *sessionSignOutExecutor) Execute(ctx *providers.NodeContext) (*providers.ExecutorResponse, error) {
 	logger := e.logger.With(log.String(log.LoggerKeyExecutionID, ctx.ExecutionID))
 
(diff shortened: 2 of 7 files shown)


backend/internal/flow/executor/session_signout_executor_test.go
```diff
@@ -111,8 +111,8 @@ func (suite *SessionSignOutExecutorTestSuite) TestTerminateError() {
 }
 
 // TestPromptsWhenConfirmationRequired covers a prompt-enabled node whose logout arrived without a
-// valid id_token_hint: the executor routes to the confirmation prompt (incomplete), marks the prompt
-// as shown, and does not terminate the session.
+// valid id_token_hint and no confirmation yet: the executor routes to the confirmation prompt
+// (incomplete) and does not terminate the session.
```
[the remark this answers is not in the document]
madurangasiriwardena: The sentence wraps across the line break: line 113 ends with "arrived without a" and line 114 continues "valid id_token_hint". The comment already reads "arrived without a valid id_token_hint"

backend/internal/flow/common/constants.go
```diff
@@ -291,6 +287,11 @@ const (
 	ActionTypeSubmit ActionType = "SUBMIT"
 	// ActionTypeReject represents a reject/deny action
 	ActionTypeReject ActionType = "REJECT"
+	// ActionTypeSignOutConfirm marks the confirmation prompt's action edge in a sign-out flow. When the
+	// End-User confirms, the prompt node forwards this type to the session sign-out node (as the action
+	// type in ForwardedData), which reads it to tell a confirmed re-run apart from the initial request,
+	// so no runtime flag has to be persisted.
+	ActionTypeSignOutConfirm ActionType = "SIGN_OUT_CONFIRM"
```
ThaminduDilshan, line 294: What if we keep this generic something like "CONFIRM" so that it won't be tied up to the signout use case?

backend/internal/flow/executor/session_signout_executor.go
```diff
@@ -99,17 +99,42 @@ func (e *sessionSignOutExecutor) Execute(ctx *providers.NodeContext) (*providers
 	return execResp, nil
 }
 
-// confirmationRequired reports whether the executor should route to its onIncomplete confirmation
-// prompt before terminating the session. It is true only when the node opts in (promptOnSignOut),
-// the RP-initiated logout requires a prompt (no valid id_token_hint), and the prompt has not already
-// been shown in this flow run.
-func (e *sessionSignOutExecutor) confirmationRequired(ctx *providers.NodeContext) bool {
+// signOutOutcome is what the executor does with the current request.
+type signOutOutcome int
```
ThaminduDilshan, line 103: Can there be a case where there are more than 2 signout outcomes? If not we could simply make `decide()` function to return a boolean instead of defining s dedicated type.
WDYT?

ThaminduDilshan (approved):

backend/internal/flow/common/constants.go
ThaminduDilshan: Fine to go ahead with this

backend/internal/flow/executor/session_signout_executor.go
madurangasiriwardena: There are 3 outcomes

- Prompt for confirmation
- Terminate the session (Sign out button in the prompt)
- Cancel the termination request (future planned cancel button in the prompt)

Hence keeping the current implementation.

PR #4481, opened 2026-07-31
Round-trip the prompt action type through the flow builder

madurangasiriwardena:
### Purpose
`prompts[].action.type` is write-only in the console. Serialization emits it from the button element's `actionType`, but the load path never reads it back, and `FlowNodeAction` does not even declare the field. A flow definition authored outside the flow builder therefore loses its action type the first time the flow is saved, and the Action selector shows the wrong option for it in the meantime.

That makes a hand-authored sign-out flow silently regress into an endless confirmation loop after an unrelated edit in the builder, and it discards the `SUBMIT` and `REJECT` action types the API documents.

### Approach
**Restore the type on load.** `restoreButtonAction` already receives the matching prompt action, so it backfills `actionType` onto the element, and `FlowNodeAction` now declares `type`. The element wins when it already carries a type: an unwired button holds the only copy, because a prompt action is emitted only once the button has a `nextNode`.

**Clear only the type the selector owns.** Picking a plain action previously wiped any `actionType`, so an action type with no option in the dropdown was discarded on the next save. It is now cleared only when it was the sign-out value.

**Key the option on the persisted value.** The Sign out option used a UI-only `SIGN_OUT` sentinel that never left the component, while the flow definition stored `SIGN_OUT_CONFIRM`. The option is now keyed on `PromptActionTypes.SignOutConfirm`, so the value selected in the editor and the value in the flow definition are one vocabulary.

**Align the option label.** `Trigger Signout` becomes `Sign Out Action`, matching the `Submit Form` and `Trigger Action` options beside it.

### Related Issues
- N/A

### Related PRs
- https://github.com/thunder-id/thunderid/pull/4299
- https://github.com/thunder-id/thunderid/pull/4456
- https://github.com/thunder-id/thunderid/pull/4480

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing button actions when changing trigger types.
  * Restored button action types correctly when loading and serializing flows.
  * Ensured sign-out actions retain their correct value across flow edits and reloads.

* **UI Updates**
  * Renamed the button action label from “Trigger Signout” to “Sign Out Action.”

* **Tests**
  * Added coverage for action preservation and flow round-trip behavior.

### DIFF
--- a/frontend/apps/console/src/features/flows/utils/flowToCanvasTransformer.ts
+++ b/frontend/apps/console/src/features/flows/utils/flowToCanvasTransformer.ts
@@ -123,6 +123,13 @@ function restoreButtonAction(
   if (matchingAction) {
     return {
       ...component,
+      // Backfill the prompt action's type onto the element, which is where the property panel reads
+      // it from and where serialization projects it back out of. Only when the element does not
+      // already carry one: an unwired button keeps its type on the element alone, since a prompt
+      // action is only emitted once the button has a nextNode. Clearing the selector writes an empty
+      // string rather than dropping the key, and that empty string is persisted, so treat it as no
+      // type rather than as a type worth preserving.
+      ...(matchingAction.type && !component.actionType ? {actionType: matchingAction.type} : {}),
       action: {
         type: matchingAction.executor ? 'EXECUTOR' : 'NEXT',
         onSuccess: matchingAction.nextNode,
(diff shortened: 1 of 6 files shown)


frontend/apps/console/src/features/flows/utils/flowToCanvasTransformer.ts
```diff
@@ -123,6 +123,11 @@ function restoreButtonAction(
   if (matchingAction) {
     return {
       ...component,
+      // Backfill the prompt action's type onto the element, which is where the property panel reads
+      // it from and where serialization projects it back out of. Only when the element does not
+      // already carry one: an unwired button keeps its type on the element alone, since a prompt
+      // action is only emitted once the button has a nextNode.
+      ...(matchingAction.type && component.actionType === undefined ? {actionType: matchingAction.type} : {}),
```
DonOmalVindula:
```suggestion
      ...(matchingAction.type && !component.actionType ? {actionType: matchingAction.type} : {}),
```

The selector clears with` onChange('actionType', '', resource)`, and since the component cleaner here doesn't strip actionType, the `''` gets persisted into `meta.components`. The new guard here checks `component.actionType === undefined`, so a component loaded with `actionType: ''` blocks the backfill even though the empty string means no type.

DonOmalVindula (approved):

PR #4515, opened 2026-08-03
Reject invalid SCIM attributes and support compound AND filters

ravindu439:
### Purpose
This PR hardens SCIM `Users` request handling in two ways:

1. **Undeclared attributes / conflicting core-extension values (CreateUser, ReplaceUser):**
   - The API was failing to reject requests that included custom extension attributes not defined in the user's entity type schema. This could lead to silent data dropping or generic "invalid value" errors downstream.
   - When a client supplied different values for a core-mapped attribute in both the core SCIM fields (e.g., `userName`) and the custom extension object (e.g., `username`), one value would silently overwrite the other.

   Both are now explicitly rejected with descriptive SCIM error responses (SCIM-1018 Schema Validation Failed and SCIM-1029 Conflicting Attribute Value).

2. **Compound AND filters (List Users, Users/.search):**
   - RFC 7644 §3.4.2.2 allows joining multiple `eq` comparisons with `and`, but the filter parser rejected any such expression outright, even though the underlying entity store already ANDs multi-key filter maps together.
   - `parseSCIMFilterForEq` now splits a filter string into its `and`-joined clauses (quote-aware, so a quoted value containing the literal word "and" is left intact), parses each clause, and merges them into one filter map. `or`, `not`, grouping, and non-`eq` operators remain unsupported and are still rejected with 400 `invalidFilter`. Duplicate attributes across clauses are also rejected rather than silently overwritten.

### Approach
- Added `undeclaredAttrs` validation in `schema_builder.go` to strictly check for attributes in the request payload that aren't declared in the target entity schema.
- Added a `mergeReverseMappedCoreAttrs` step in `users_service.go` to merge core-mapped fields back into the extension object. If an attribute already exists in the extension object with a semantically different JSON value (using a deep equal comparison via `jsonRawValuesEqual`), the operation rejects the request.
- Added granular error constructs `newUndeclaredAttributesError` and `newConflictingAttributeValueError` in `error_constants.go` to provide meaningful error descriptions to clients.
- Updated HTTP status mapping in `handler.go` to properly map `ErrorConflictingAttributeValue.Code` to `400 Bad Request`.
- Split `parseSCIMFilterForEq` (`users_handler.go`) into `splitSCIMAndClauses` (quote-aware split on `and`) + `parseSCIMEqClause` (single-clause parse), merging clause results into one map; rejects repeated attributes across clauses.
- Added unit tests for the new merging/validation logic and for compound-AND parsing (two/three clauses, case-insensitive keyword, quoted value containing literal "and", duplicate attribute, malformed clause, and regression coverage confirming `or`/`not`/grouping stay rejected).

### Related Issues
- N/A

### Related PRs
- #4184

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/scim/users_handler.go
```diff
@@ -64,16 +68,83 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
...
+	for _, urn := range searchReq.Schemas {
+		if strings.EqualFold(strings.TrimSpace(urn), SCIMSearchSchemaURN) {
+			hasSearchSchema = true
+			break
+		}
+	}
+	if !hasSearchSchema {
+		h.handleSCIMError(w, r, &ErrorMissingSchemas)
+		return
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/init.go
```diff
@@ -161,6 +161,19 @@ func registerRoutes(mux *http.ServeMux, h *scimHandler, uh *scimUsersHandler, gh
...
+		uh.HandleUsersSearchRequest,
+		optsCRUD,
+	))
+	mux.HandleFunc(middleware.WithCORS(
+		"OPTIONS "+SCIMBasePath+"/Users/.search",
+		func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusNoContent) },
+		optsCRUD,
+	))
+
+	// groups CRUD oparations
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/error_constants_test.go
```diff
@@ -32,3 +32,12 @@ func TestNewMissingRequiredAttributesError(t *testing.T) {
 	require.Contains(t, svcErr.ErrorDescription.DefaultValue, `"employee"`)
 	require.Contains(t, svcErr.ErrorDescription.DefaultValue, "department, employee_id")
 }
+
+func TestNewUndeclaredAttributesError(t *testing.T) {
+	svcErr := newUndeclaredAttributesError("employee", []string{"extra1", "extra2"})
+
+	require.Equal(t, ErrorSchemaValidationFailed.Code, svcErr.Code)
+	require.Equal(t, ErrorSchemaValidationFailed.Type, svcErr.Type)
```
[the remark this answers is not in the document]
ravindu439: current test coverage is enough for this .

backend/internal/scim/users_handler.go
```diff
@@ -64,16 +68,83 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
...
 	logger.Debug(ctx, "SCIM Users list sent", log.Int("totalResults", listResp.TotalResults))
 }
 
+// HandleUsersSearchRequest handles POST /scim/v2/Users/.search (RFC 7644 §3.4.3).
+func (h *scimUsersHandler) HandleUsersSearchRequest(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
+
+	if svcErr := validateSCIMContentType(r); svcErr != nil {
+		h.handleSCIMError(w, r, svcErr)
```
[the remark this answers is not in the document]
ravindu439: there is separate PR for this .

sadilchamishka (approved):

sadilchamishka (approved):

backend/internal/scim/error_constants.go
```diff
@@ -428,6 +429,53 @@ var (
...
+			DefaultValue: "Sorting via sortBy or sortOrder is not supported in this implementation",
+		},
+	}
+
+	// ErrorConflictingAttributeValue is returned when a core-mapped attribute is supplied
+	// with different values through both the top-level SCIM core fields and the ThunderID
+	// custom extension object.
+	ErrorConflictingAttributeValue = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "SCIM-1029",
```
sadilchamishka: `ErrorConflictingAttributeValue` reuses `SCIM-1029`, which `ErrorConflictingAttributesParams` already claims 30 lines above. Two distinct client errors sharing a code makes them indistinguishable to callers, and `mapSCIMError` now lists both constants in the same `case` arm, so the duplicate is invisible at the switch. The next free code is `SCIM-1031`:

```go
Code: "SCIM-1031",
```

backend/internal/scim/schema_builder.go
```diff
@@ -671,7 +685,42 @@ func missingRequiredAttrs(attrs map[string]json.RawMessage, schema json.RawMessa
...
+
+	var undeclared []string
+	for name := range extensionAttrs {
+		if _, ok := rawProps[name]; !ok {
+			undeclared = append(undeclared, name)
+		}
+	}
+
+	for name := range coreAttrs {
+		if strings.EqualFold(name, "id") || strings.EqualFold(name, "meta") {
```
sadilchamishka: Only `id` and `meta` are exempted, but `CoreAttrs` (from `parseSCIMUserPayload`) holds every top-level field that isn't `schemas` or the extension URN, and `coreAttrRules` covers just 13 of them. So `externalId`, `active`, `password`, `userType`, `groups`, `roles`, `entitlements`, `ims`, and `x509Certificates` are never in `consumedCoreKeys` and now fail with `400 does not declare the following attribute(s): active, externalId`.

`externalId` is a required-to-tolerate core attribute (RFC 7643 §3.1) and is sent by nearly every provisioning client, so this is likely to break real integrations. Suggest either exempting the known SCIM core attribute names here, or scoping the undeclared check to `extensionAttrs` and leaving core-field handling to the existing mapper.

backend/internal/scim/schema_builder.go
```diff
@@ -671,7 +685,42 @@ func missingRequiredAttrs(attrs map[string]json.RawMessage, schema json.RawMessa
...
+	schema json.RawMessage,
+) ([]string, error) {
+	rawProps, hasSchema, err := parseSchemaRawProps(schema)
+	if err != nil || !hasSchema {
+		return nil, err
+	}
+
+	var undeclared []string
+	for name := range extensionAttrs {
+		if _, ok := rawProps[name]; !ok {
```
sadilchamishka: This lookup is case-sensitive, but `reverseMapCoreAttrsForSchema` matches schema properties case-insensitively (see `TestReverseMapCoreAttrsForSchema_CaseInsensitivePropName`). An extension attribute sent as `Given_Name` against a schema declaring `given_name` is accepted by one path and rejected as undeclared by the other. Worth making both use the same comparison.

backend/internal/scim/users_service.go
```diff
@@ -117,19 +120,40 @@ func (s *scimUsersService) CreateUser(
...
+	var consumedCoreKeys map[string]struct{}
 	if len(payload.CoreAttrs) > 0 {
-		reverseMapped := reverseMapCoreAttrsForSchema(payload.CoreAttrs, et.Schema)
-		for k, v := range reverseMapped {
-			if _, exists := payload.ExtensionAttrs[k]; !exists {
-				payload.ExtensionAttrs[k] = v
-			}
+		reverseMapped, consumed, err := reverseMapCoreAttrsForSchema(payload.CoreAttrs, et.Schema)
+		if err != nil {
+			logger.Error(ctx, "SCIM CreateUser: failed to parse entity type schema", log.Error(err))
```
sadilchamishka, line 127: A schema that fails to parse is stored server-side data, not something the client sent, so `ErrorInvalidRequestBody` (400) points the blame in the wrong direction and gives the caller no actionable fix. `ErrorInternalServer` (500) matches what actually went wrong. Same at lines 139, 149 and in the three corresponding `ReplaceUser` branches.

backend/internal/scim/users_handler.go
```diff
@@ -37,7 +42,12 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
 	ctx := r.Context()
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
 
-	// Parse optional SCIM filter — only single "eq" expressions are supported.
+	if r.URL.Query().Get("sortBy") != "" || r.URL.Query().Get("sortOrder") != "" {
```
sadilchamishka: This hardcodes "sorting is unsupported", but `scimconfig.SortSupported` already exists and is what `/ServiceProviderConfig` advertises (`service.go:152`). If that flag is ever flipped on, the endpoint will keep returning 400 while the config claims support. Gate on the flag instead:

```go
if !scimconfig.SortSupported && (r.URL.Query().Get("sortBy") != "" || r.URL.Query().Get("sortOrder") != "") {
```

Same in `HandleUsersSearchRequest` (line 117) and `HandleGroupsListRequest`.

backend/internal/scim/error_constants.go
ravindu439: fixed

backend/internal/scim/schema_builder.go
ravindu439: kept this fix for future implementations

backend/internal/scim/schema_builder.go
ravindu439: fixed to be case insensitive

backend/internal/scim/users_service.go
ravindu439: fixed

backend/internal/scim/users_handler.go

sadilchamishka (approved):

PR #4520, opened 2026-08-03
Validate client jwt assertion's aud as string

thiva-k:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OAuth client assertion validation by requiring a single audience value matching the configured issuer.
  * Rejected assertions with array audiences or mismatched audience values.
  * Rejected assertions with malformed or undecodable payloads.
  * Preserved successful authentication for valid issuer-matching private-key assertions.

### DIFF
--- a/backend/internal/oauth/oauth2/clientauth/clientauth.go
+++ b/backend/internal/oauth/oauth2/clientauth/clientauth.go
@@ -225,6 +225,20 @@ func validateClientAssertion(ctx context.Context,
 		return fmt.Errorf("no certificate configured for client assertion validation")
 	}
 
+	// FAPI 2.0 Security Profile Section 5.3.2.1: the client assertion's 'aud' claim must be the
+	// authorization server's issuer identifier as a single string.
+	payload, err := jwt.DecodeJWTPayload(clientAssertion)
+	if err != nil {
+		return fmt.Errorf("failed to decode client assertion payload: %w", err)
+	}
+	aud, ok := payload[constants.ClaimAud].(string)
+	if !ok {
+		return fmt.Errorf("client assertion 'aud' claim must be a single string")
+	}
+	if aud != issuer {
+		return fmt.Errorf("client assertion 'aud' claim %q does not match the issuer", aud)
+	}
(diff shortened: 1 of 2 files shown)


backend/internal/oauth/oauth2/clientauth/clientauth.go
```diff
@@ -225,6 +225,20 @@ func validateClientAssertion(ctx context.Context,
 		return fmt.Errorf("no certificate configured for client assertion validation")
 	}
 
+	// FAPI 2.0 Security Profile Section 5.3.2.1: the client assertion's 'aud' claim must be the
+	// authorization server's issuer identifier as a single string.
+	payload, err := jwt.DecodeJWTPayload(clientAssertion)
+	if err != nil {
+		return fmt.Errorf("failed to decode client assertion payload: %w", err)
+	}
+	aud, ok := payload["aud"].(string)
```
ThaminduDilshan: Shall we use the already existing constant for the aud claim?

backend/internal/oauth/oauth2/clientauth/clientauth.go
```diff
@@ -225,6 +225,20 @@ func validateClientAssertion(ctx context.Context,
 		return fmt.Errorf("no certificate configured for client assertion validation")
 	}
 
+	// FAPI 2.0 Security Profile Section 5.3.2.1: the client assertion's 'aud' claim must be the
+	// authorization server's issuer identifier as a single string.
+	payload, err := jwt.DecodeJWTPayload(clientAssertion)
```
ThaminduDilshan, line 230: @coderabbitai hope we're not double decoding the jwt here and later somewhere else which could affect the performance. Can you check and provide a short reply?

thiva-k: We are double decoding this. Our current validation function check the aud claim in string as well as array. For this case we need to reject arrays

thiva-k:
Duplicate of https://github.com/thunder-id/thunderid/pull/4544

PR #4536, opened 2026-08-04
Fix the DSN string of SQLite to use modernc pragma syntax so busy_timeout applies

### LINKED ISSUE #4535: SQLite DSN options use different syntax, so busy_timeout is never applied
### Description

The shipped SQLite `options` strings use `mattn/go-sqlite3` DSN syntax, but the driver in use is `modernc.org/sqlite`. `modernc.org/sqlite` reads only the `_pragma=name(value)` form; the keys `_journal_mode` and `_busy_timeout` pass URL parsing and are then silently dropped.

See https://tessl.io/registry/tessl-labs/sqlite-go-best-practices/0.2.0

The practical effect is that **`busy_timeout` stays at 0**. Any writer that hits a lock fails immediately with `SQLITE_BUSY` instead of waiting, and since `isRetryableDBError` has no SQLite busy/locked case, the error surfaces to the caller as `SSE-5000 Internal server error`.

The shipped string:

```
_journal_mode=WAL&_busy_timeout=5000&_pragma=foreign_keys(1)
```

Opening a database with it and reading back the pragmas:

```
busy_timeout = 0
journal_mode = delete
```

versus the `_pragma` form:

```
busy_timeout = 5000
journal_mode = wal
```

WAL is unaffected in production only because `init_script.sh` runs `PRAGMA journal_mode=WAL` out of band and the mode persists in the database file. `busy_timeout` is per-connection, so nothing compensates for it.

#### Impact

Any concurrent write load against a SQLite deployment produces intermittent 500s with `SSE-5000 Internal server error`. It shows up most visibly in the e2e suite, which runs 6 Playwright workers. If a lot of workers try to get the same resource, the server is more likely to have a write collision.

### Steps to Reproduce

To confirm the pragmas are dropped in the current string, we can run the string with the current go SQL module.

```go
package main

import (
	"database/sql"
	"fmt"
	"os"
	"path/filepath"

	_ "modernc.org/sqlite"
)

func probe(dsn string) {
	db, _ := sql.Open("sqlite", dsn)
	defer db.Close()
	for _, p := range []string{"busy_timeout", "journal_mode"} {
		var v string
		_ = db.QueryRow("PRAGMA " + p).Scan(&v)
		fmt.Printf("  %s = %s\n", p, v)
	}
}

func main() {
	dir, _ := os.MkdirTemp("", "probe")
	p := filepath.Join(dir, "t.db")
	fmt.Println("as shipped:")
	probe(p + "?_journal_mode=WAL&_busy_timeout=5000&_pragma=foreign_keys(1)")
	fmt.Println("with _pragma form:")
	probe(p + "?_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)&_pragma=foreign_keys(1)")
}
```

Prints `busy_timeout = 0` / `journal_mode = delete` for the current string and `busy_timeout = 5000` / `journal_mode = wal` with the fixed string.

2. To check for write contention, Running E2E suite with high concurrency (Workers > 6) will usually result a flaky test or more due to write contention with the `SSE-5000 Internal server error`

### Version

1.0.0-alpha2

_No response_

Sadeesha-Sath:
### Purpose

The shipped SQLite `options` strings use `mattn/go-sqlite3` DSN syntax, but the driver is `modernc.org/sqlite`, which reads only the `_pragma=name(value)` form. The keys `_journal_mode` and `_busy_timeout` pass URL parsing and are then silently dropped, leaving `busy_timeout` at 0.

With no busy timeout, a writer that hits a lock fails immediately with `SQLITE_BUSY` rather than waiting. `isRetryableDBError` has no SQLite busy/locked case, so the failure reaches the caller as `SSE-5000 Internal server error`. Any concurrent write load against a SQLite deployment therefore produces intermittent 500s.

Opening a database with the shipped string and reading the pragmas back:

| DSN | `busy_timeout` | `journal_mode` |
|---|---|---|
| `_journal_mode=WAL&_busy_timeout=5000&...` (shipped) | `0` | `delete` |
| `_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)&...` | `5000` | `wal` |

`journal_mode` is dropped too, but it does not surface in practice as the `init_script.sh` runs `PRAGMA journal_mode=WAL` at database creation and SQLite persists that setting in the file header, so later connections inherit it. `busy_timeout` is per-connection and has no such fallback. WAL does not mitigate the problem either, since it removes reader-vs-writer blocking but still serializes writers against each other.

This surfaced as e2e flakiness. At `PLAYWRIGHT_WORKERS=6`, `POST /flow/execute` failed roughly half the time when several workers fetched an admin token at once, and `retries: 1` masked it as flaky tests.

### Approach

Rewrite every shipped `options` string to the `_pragma` form, preserving each file's configured timeout (the integration test config keeps its 60000). 31 occurrences across 8 files:

| File | Occurrences |
|---|---|
| `backend/cmd/server/deployment.yaml` | 4 |
| `backend/cmd/server/config/default.json` | 4 |
| `tests/integration/resources/deployment.yaml` | 4 |
| `install/helm/values.yaml` | 4 |
| `install/openchoreo/thunderid-oc-resourcetype/templates/thunderid-resourcetype.yaml` | 4 |
| `install/helm/README.md` | 4 |
| `docs/content/deployment/configuration.mdx` | 4 |
| `docs/content/deployment/deployment-paths/kubernetes.mdx` | 4 |

Config changes only. No production Go code changes

The failure mode here is silent: a wrong dialect parses fine but drops so a string comparison in a test would not have caught the original bug. Wrote a test `dbprovider_sqlite_options_test.go` that reads the options out of the shipped `deployment.yaml` and `config/default.json`, opens a throwaway database with each, and asserts the pragmas actually took effect.

### Related Issues
- Fixes #4535 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQLite database reliability by applying consistent busy-timeout, WAL journaling, and foreign-key settings across deployments.
  * Reduced potential database locking and concurrency issues.

* **Documentation**
  * Updated configuration and deployment guidance to reflect the standardized SQLite options across supported installation methods.


rajithacharith (dismissed):

senthalan (dismissed):

backend/internal/system/database/provider/dbprovider_sqlite_options_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"gopkg.in/yaml.v3"
+
+	"github.com/thunder-id/thunderid/internal/system/config"
+)
+
+// The SQLite driver is modernc.org/sqlite, which reads only the `_pragma=name(value)` DSN form.
+// mattn/go-sqlite3 spellings such as `_busy_timeout=5000` are accepted by url parsing and then
+// silently dropped, leaving busy_timeout at 0 - so a concurrent write fails immediately with
+// SQLITE_BUSY instead of waiting. Assert the shipped options take effect, not just that they parse.
+func TestShippedSQLiteOptionsTakeEffect(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/system/database/provider/dbprovider_sqlite_options_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+}
+
+// openWithOptions opens a throwaway database with the given DSN options and reports the pragma
+// values the driver actually ended up with.
+func openWithOptions(t *testing.T, options string) (busyTimeout, journalMode string) {
+	t.Helper()
+	require.NotEmpty(t, options, "no SQLite options configured")
+
+	db, err := sql.Open("sqlite", filepath.Join(t.TempDir(), "probe.db")+"?"+options)
```
ThaminduDilshan: Do we need this test?

IMO this is not a unit test. We're testing a integration scenario by reading files outside this component/package and opening a actual DB connection. If we're to have this, I prefer to add it as a integration test. However there we should avoid running it for postgress

Sadeesha-Sath: We can remove it tbh. It just checks that the DSN string works. But since we are changing this string to work, we can safely remove it.

Sadeesha-Sath: Shall I remove this then?

ThaminduDilshan (approved):

PR #4543, opened 2026-08-04
Provide seamless login for OIDC/github connections by defualt.

### LINKED ISSUE #4101: Google/ github doesn't work by default
### Description

Google, github needs a linking attribute to be configured in order to work. Either needs to;
- Create a `sub` attribute in the user type (schema)
- Configure a linking attribute

Suggestion is to make it work seamlessly with the default configurations. Attributes may not be synced to the provisioned user profile, but login/ registration should work seemlessly

### Steps to Reproduce

1. Create google connection
2. Configure google login flow and attach to an application
3. Try to login

### Version

v0.48.0

_No response_

ImalshaD:
### Purpose

  Fixes the default Google, GitHub, and OIDC sign-in experience so users can authenticate and be provisioned without requiring administrators to manually configure account
  linking, attribute mappings, scopes, and provisioning behavior.

  This PR:

  - Enables self-registration for the default Person user type.
  - Applies provider-specific scope and account-linking defaults.
  - Adds schema-aware username mappings where required for provisioning.
  - Enables authentication without an existing local user for Google, GitHub, and OIDC executors.
  - Adds provisioning to the corresponding flow-builder widgets and default flow templates.
  - Reuses an existing provisioning executor when applicable.

  ### Approach

  The implementation applies defaults in the backend so connections created through the Console, REST APIs, and declarative resources behave consistently.

  Key changes include:

  - Default GitHub connections to the user:email scope when scopes are not provided.
  - Default account linking to a unique email attribute when the effective provider scopes support email.
  - Inspect eligible user-type schemas and create provider-specific username mappings when username is required.
  - Preserve explicitly configured scopes, account-linking attributes, and attribute mappings.
  - Enable allowAuthenticationWithoutLocalUser by default for Google, GitHub, and OIDC flow executors.
  - Bundle a provisioning executor and missing-attribute prompt with Google and GitHub flow-builder widgets.
  - Update affected flow templates so federated authentication paths reach provisioning before assertion generation.
  - Reuse an existing provisioning executor and remove the redundant bundled provisioning prompt when multiple federated widgets converge on the same provisioning path.
  - Preserve generic OAuth behavior because ThunderID cannot infer arbitrary OAuth provider scope and claim semantics.

  ### Related Issues

  - Fixes #4101 (https://github.com/thunder-id/thunderid/issues/4101)

  ### Related PRs

  - N/A

  ### Checklist
  - [x] Tests provided.
      - [x] Unit Tests
      - [ ] Integration Tests
  - [x] Breaking changes not applicable.
      - [ ] breaking change label added.

  ### Security checks
    (https://security.docs.wso2.com/en/latest/security-guidelines/secure-engineering-guidelines/secure-coding-guidlines/introduction/)

## Summary by CodeRabbit

* **New Features**
  * Enabled self-registration for default users.
  * Added automatic identity-provider defaults for email linking, usernames, and provider scopes.
  * Improved Google and GitHub login flows with local-user provisioning and prompts for missing profile information.
  * Preserved original authentication claims alongside mapped local attributes.
  * Added English prompts for required schema attributes.
* **Bug Fixes**
  * Improved social-login flow wiring and duplicate provisioning-step handling.
  * Added default GitHub email scopes when none are configured.

### DIFF
--- a/backend/internal/idp/constants.go
+++ b/backend/internal/idp/constants.go
@@ -24,8 +24,21 @@ const (
 	PropIDJagEnabled          = "id_jag_enabled"
 )
 
-// defaultOIDCScopes is seeded for OIDC and Google connections when no scopes are supplied.
-const defaultOIDCScopes = "openid,email,profile"
+// Claims and scopes shared by the OIDC-style providers. A claim is what the provider emits; a scope
+// only asks for it.
+const (
+	emailClaim        = "email"
+	emailScope        = "email"
+	defaultOIDCScopes = "openid,email,profile"
+)
+
+// Local attribute names, as declared by user type schemas.
+// TODO: Drop these once schemas can declare attribute types (EMAIL, MOBILE, ...) as a first class
+// concept, so the attribute carrying an email can be found rather than named here.
+const (
+	defaultAccountLinkingAttribute = "email"
+	localUsernameAttribute         = "username"
+)
 
 // Known endpoints for Google OAuth2/OIDC.
 const (
@@ -35,12 +48,17 @@ const (
 	googleJwksEndpoint          = "https://www.googleapis.com/oauth2/v3/certs"
 )
 
-// Known endpoints for GitHub OAuth.
+// Known endpoints, claims and scopes for GitHub OAuth.
 const (
 	gitHubAuthorizationEndpoint = "https://github.com/login/oauth/authorize"
 	gitHubTokenEndpoint         = "https://github.com/login/oauth/access_token" // #nosec G101
 	gitHubUserInfoEndpoint      = "https://api.github.com/user"
 	gitHubUserEmailEndpoint     = "https://api.github.com/user/emails"
+
+	gitHubLoginClaim     = "login"
+	gitHubUserScope      = "user"
+	gitHubUserEmailScope = "user:email"
+	defaultGitHubScopes  = gitHubUserEmailScope
 )
 
 // idpPropertyConfig defines the required and optional properties for an IDP type,
--- a/backend/internal/idp/service.go
+++ b/backend/internal/idp/service.go
@@ -8,11 +8,13 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
 
 	"github.com/thunder-id/thunderid/internal/entitytype"
+	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	declarativeresource "github.com/thunder-id/thunderid/internal/system/declarative_resource"
 	"github.com/thunder-id/thunderid/internal/system/log"
 	"github.com/thunder-id/thunderid/internal/system/resourcedependency"
@@ -36,6 +38,7 @@ type IDPServiceInterface interface {
 	DeleteIdentityProvider(ctx context.Context, idpID string) *tidcommon.ServiceError
 	GetIDPUsages(ctx context.Context, idpID string) (*resourcedependency.DependenciesResponse, *tidcommon.ServiceError)
 	SetDependencyRegistry(r resourcedependency.Registry)
+	ApplySchemaAwareDefaults(ctx context.Context, idp *providers.IDPDTO)
 }
 
 // idpService is the default implementation of the IdPServiceInterface.
@@ -46,6 +49,26 @@ type idpService struct {
 	dependencyRegistry resourcedependency.Registry
 	logger             *log.Logger
 	uuidGenerator      func() (string, error)
+}
+
+// userTypeAttributes holds a user type's non-credential schema attributes.
+type userTypeAttributes struct {
+	name       string
+	attributes []entitytype.AttributeInfo
+}
+
+// isUnique reports whether the user type declares attr as unique.
+func (u userTypeAttributes) isUnique(attr string) bool {
+	return slices.ContainsFunc(u.attributes, func(a entitytype.AttributeInfo) bool {
+		return a.Attribute == attr && a.Unique
+	})
+}
+
+// isRequired reports whether the user type declares attr as required.
+func (u userTypeAttributes) isRequired(attr string) bool {
+	return slices.ContainsFunc(u.attributes, func(a entitytype.AttributeInfo) bool {
+		return a.Attribute == attr && a.Required
+	})
 }
 
 // newIDPService creates a new instance of IdPService.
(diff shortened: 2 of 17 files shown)


frontend/apps/console/src/features/flows/data/templates.json
```diff
@@ -1713,6 +1714,76 @@
...
+                    "id": "dynamic_inputs"
+                  },
+                  {
+                    "type": "ACTION",
+                    "id": "action_schema_attrs",
+                    "label": "Continue",
+                    "variant": "PRIMARY",
+                    "eventType": "SUBMIT"
+                  }
+                ]
```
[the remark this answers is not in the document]
ImalshaD: This is fixed please check, the places where translation tokens are used are fixed, some templates by default uses literals those are kept matching with the existing pattern.

backend/internal/idp/service.go
```diff
@@ -323,6 +328,173 @@ func (is *idpService) DeleteIdentityProvider(ctx context.Context, idpID string)
...
+	}
+
+	selfRegistrationEnabled := make([]string, 0, len(userTypes))
+	for _, userType := range userTypes {
+		if userType.AllowSelfRegistration {
+			selfRegistrationEnabled = append(selfRegistrationEnabled, userType.Name)
+		}
+	}
+	return selfRegistrationEnabled
+}
```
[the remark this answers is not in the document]
ImalshaD: Not changing this one. I don't think the single-type branch needs the filter.
AllowSelfRegistration is a runtime provisioning gate, not an eligibility signal for connection defaults, and it is enforced in the flow rather than here:- internal/flow/executor/user_type_resolver.go:410 (resolveUserTypeFromSingleAllowed) fails with ErrSelfRegDisabledForUserType when the flag is off. That is exactly the single-user-type deployment this finding describes.
- user_type_resolver.go:380 does the same on the explicit
- -user-type path, and provisioning_executor.go:833 checks it again.

So a deployment with one user type and self-registration disabled cannot provision, whatever the connection carries. The seeded linking attribute and username mapping are inert configuration in that case, not a bypass.

The filter in the multi-type branch is doing a different job: choosing which of several types the defaults are derived from. With one type there is nothing to choose, and the flag would only decide whether to seed configuration that is already unreachable.
The proposed fix would also break a legitimate case. Account linking is not provisioning: AllowSelfRegistration is read nowhere in authn or entityprovider. A deployment that deliberately disables self-registration but still wants federated sign-in to link to existing local users by email is a normal setup. Filtering the single-type branch would withhold the email linking default from exactly that deployment and force it to be configured by hand, which is the problem this PR set out to remove.
The test you cite asserts the intended behaviour rather than exposing a gap: one user type with the flag unset should still receive the defaults.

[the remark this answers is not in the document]
ImalshaD: @coderabbitai try again and resolve this comment if you are satisfied.

ImalshaD:
@coderabbitai review

backend/internal/idp/constants.go
```diff
@@ -24,8 +24,30 @@ const (
...
+// Claim names the providers emit. Distinct from the scopes below, which only ask for them.
+const (
+	emailClaim       = "email"
+	gitHubLoginClaim = "login"
+)
+
+// Scopes that grant access to a user's email address, per provider.
+const (
+	emailScope           = "email"
+	gitHubUserScope      = "user"
```
ThaminduDilshan: We already have constant blocks defined for google and github. Shall we move these to the appropriate blocks to increase readability?
Applies to most of the new constants added

backend/internal/idp/constants.go
```diff
@@ -24,8 +24,30 @@ const (
...
+	gitHubUserEmailScope = "user:email"
+)
+
+// Local attribute names, as declared by user type schemas.
+const (
+	defaultAccountLinkingAttribute = "email"
```
ThaminduDilshan, line 39: Hope we're handling the case when such a attribute is not present in any schema. Applies to both email and username.

Note that default Person schema gets created during bootstrap (setup script). It's possible to start thunderid even without any default resources

ThaminduDilshan: I think we can get rid of having hard coded attribute names if the schema has first class support to define attribute types as EMAIL, MOBILE, etc. 
Shall we add a TODO comment here to revisit this with such a capability in schema?

backend/internal/idp/service.go
```diff
@@ -323,6 +330,206 @@ func (is *idpService) DeleteIdentityProvider(ctx context.Context, idpID string)
 	return nil
 }
 
+// ApplySchemaAwareDefaults seeds account-linking and username-mapping defaults derived from user-type
+// schemas, on create only: an update replaces the whole connection, so re-seeding would silently
+// restore a section the administrator removed. Explicit configuration wins, and schema lookup failures
+// leave the connection unchanged rather than blocking the operation. Entity-type reads are authorized
+// against ctx and never elevated, so a caller with no subject must pass a runtime context of its own.
+func (is *idpService) ApplySchemaAwareDefaults(ctx context.Context, idp *providers.IDPDTO) {
```
ThaminduDilshan, line 358: Shall we update the method comment to explain just the method functionality? Not the usage specific context. That is already mentioned in the usages

backend/internal/idp/utils.go
```diff
@@ -435,6 +444,64 @@ func ensureOpenIDScope(ctx context.Context, propertyMap map[string]cmodels.Prope
...
+		if scope != "" {
+			return nil
+		}
+	}
+
+	return createAndAppendProperty(ctx, propertyMap, PropScopes, defaultGitHubScopes, false, logger)
+}
+
+// scopesGrantEmail reports whether the effective scopes let the connection read an email address.
+// Generic OAuth is absent by design: ThunderID cannot infer an arbitrary provider's scope semantics.
```
ThaminduDilshan:
```suggestion
// Generic OAuth is absent by design as we cannot infer an arbitrary provider's scope semantics.
```

backend/internal/idp/service.go
```diff
@@ -323,6 +330,206 @@ func (is *idpService) DeleteIdentityProvider(ctx context.Context, idpID string)
...
+}
+
+// candidateUserTypesForDefaults returns self-registration-enabled user types. A deployment with one
+// user type uses that type because there is no alternative target.
+func candidateUserTypesForDefaults(userTypes []entitytype.EntityTypeListItem) []string {
+	if len(userTypes) == 1 {
+		return []string{userTypes[0].Name}
+	}
+
+	selfRegistrationEnabled := make([]string, 0, len(userTypes))
```
ThaminduDilshan: Do we need to filter only the self registration enabled user types? Even if we have a mapping for self registration disabled type, it's not going to break anything right?

Even attribute mappings are applied during login flows too. If someone create the users manually, having default mappings will help to provide a seamless federated login experience right?

backend/internal/idp/constants.go
ImalshaD: Resolved.

backend/internal/idp/constants.go
ImalshaD: Resolved and added the TODO comment.

backend/internal/idp/service.go
ImalshaD: Resolved.

backend/internal/idp/service.go
ImalshaD: Removed the filter as agreed offline.

ImalshaD:
@coderabbitai review

ThaminduDilshan (approved):

PR #4554, opened 2026-08-04
Include ID-token-only scope claims in issued ID tokens for code and CIBA flows

### LINKED ISSUE #4553: Requested OIDC claims are missing from ID tokens in the authorization code flow
### Description

When an OIDC application uses the authorization code flow, claims requested through standard OIDC scopes may be missing from the issued ID token.

For example, requesting the openid email scopes does not include the authenticated user’s email claim in the ID token, even when:
  - The user has an email address.
  - The application allows email in the ID token.
  - The email scope is mapped to the email claim.

  The claim may still appear in the access token when configured there, but remains missing from the ID token.

### Steps to Reproduce

1. Create an OIDC application that uses the authorization code flow.
2. Configure email as an allowed ID-token user attribute.
  3. Configure the email scope to include the email claim.
  4. Authenticate a user who has an email address using an authorization request similar to:
     response_type=code
     scope=openid email
  5. Exchange the authorization code for tokens.
  6. Decode the issued ID token.

#### Actual result

  The issued ID token does not contain the requested email claim.

  #### Expected result

  The issued ID token should contain the authenticated user’s email claim because:

  - The authorization request includes the email scope.
  - The application permits email in the ID token.
  - The authenticated user has an email value.

  ### Possible cause and fix

  During the authorization code flow, the attributes needed for later token generation are determined before the authentication flow completes. Scope-derived attributes
  allowed exclusively in the ID token are not included in that required set.

  The required attributes for an authorization code flow should cover all token outputs that can be generated during the subsequent token exchange, including both ID-token
  and UserInfo attributes.

  A regression test should verify that an attribute requested through an OIDC scope and allowed only in the ID token is included in the resulting ID token

### Version

ThunderID 1.0.0-alpha2

_No response_

thiva-k:
### Purpose
Scope-derived OIDC claims allow-listed only for the ID token (e.g. `email` via the `email` scope) were never resolved into the attribute cache in the authorization code and CIBA flows, so they never appeared in the issued ID token. Attribute selection filtered scope claims against the UserInfo allow-list only.

### Approach
When an ID token is issued, cache scope-derived attributes allow-listed for either the ID token or UserInfo (previously UserInfo only). Since only `response_type=code` is supported, the now-dead response-type branch and its `responseType` parameter were removed from the attribute-selection chain. Applied the same fix to the CIBA mirror.

### DIFF
--- a/backend/internal/oauth/oauth2/authz/service.go
+++ b/backend/internal/oauth/oauth2/authz/service.go
@@ -386,7 +386,7 @@ func (as *authorizeService) initiateFlowAndStoreRequest(
 
 	effectiveAcrValues := requestvalidator.ResolveACRValues(oauthParams.AcrValues, app.AcrValues)
 	essentialAttributes, optionalAttributes := getRequiredAttributes(
-		oauthParams.StandardScopes, oauthParams.ClaimsRequest, oauthParams.ResponseType, app)
+		oauthParams.StandardScopes, oauthParams.ClaimsRequest, app)
 
 	authRequestCtx := authRequestContext{
 		OAuthParameters: *oauthParams,
@@ -805,8 +805,8 @@ func createAuthorizationCode(
 }
 
 // getRequiredAttributes determines the essential and optional user attributes required based on OIDC scopes,
-// claims parameter, response type, and app configuration.
-func getRequiredAttributes(oidcScopes []string, claimsRequest *oauth2model.ClaimsRequest, responseType string,
+// claims parameter, and app configuration.
+func getRequiredAttributes(oidcScopes []string, claimsRequest *oauth2model.ClaimsRequest,
 	app *providers.OAuthClient) (essentialAttributes, optionalAttributes string) {
 	if app == nil {
 		return "", ""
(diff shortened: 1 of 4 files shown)


ImalshaD (approved):

ThaminduDilshan (approved):

ThaminduDilshan (changes requested):
Need to cross check this against the spec

ThaminduDilshan (approved):

thiva-k:
> Need to cross check this against the spec

```
The Claims requested by the profile, email, address, and phone scope values are returned from the UserInfo Endpoint, as described in [Section 5.3.2](https://openid.net/specs/openid-connect-core-1_0.html#UserInfoResponse), when a response_type value is used that results in an Access Token being issued. However, when no Access Token is issued (which is the case for the response_type value id_token), the resulting Claims are returned in the ID Token.
```

The common interpretation to the above clause seems not to restrict the claims being added to id token, when access token is issued. This was concluded on an offline discussion.

CC: @ThaminduDilshan @ThumulaPerera

PR #4574, opened 2026-08-05
feat(scim): Implement SCIM /Me endpoint for GET and PUT requests

ravindu439:
### Purpose
Adds support for the SCIM `/Me` endpoint (RFC 7644 §3.11), letting an authenticated SCIM client retrieve and update their own user profile without needing to know their own user ID upfront. Previously SCIM clients could only self-serve via `/Users/{id}`, requiring the ID to already be known.

### Approach
- Registered `GET /scim/v2/Me`, `PUT /scim/v2/Me`, and `OPTIONS /scim/v2/Me` routes in `backend/internal/scim/init.go`, alongside the existing SCIM Users routes.
- Implemented `HandleMeGetRequest` and `HandleMeReplaceRequest` in `backend/internal/scim/users_handler.go`. Per RFC 7644 §3.11 option 3 ("process the request directly"), both handlers resolve the authenticated subject from the security context (`security.GetSubject(ctx)`) and delegate straight to the existing `GetUser` / `ReplaceUser` service methods keyed by that subject, the same service methods `/scim/v2/Users/{id}` uses, rather than aliasing to a redirect or to the native `/users/me` (`UpdateUserAttributes`) path.
- Authorization is scoped correctly for this: `sysauthz.IsActionAllowed` grants self-service access whenever the caller's subject matches the target resource ID (`isResourceOwner`), so no elevated or bypassed context is needed. Both handlers pass the plain authenticated `ctx` through.
- The full `ReplaceUser` call surface doesn't grant more than intended for self-service: user type is immutable on PUT (`ErrorImmutableUserType` if the payload's type doesn't match the existing user's), OUID is derived from the (unchangeable) type's schema rather than the request, and credential fields are rejected by `UpdateUser` itself regardless of caller.
- `HandleMeGetRequest`: resolves subject, fetches the user, returns 200 with `Location`/`ETag` headers set.
- `HandleMeReplaceRequest`: resolves subject, validates content type and request body, validates the SCIM payload, then replaces the user using `If-Match` for optimistic concurrency, returning 200 with updated `Location`/`ETag`.
- Both return 401 (new `ErrorUnauthenticated`, SCIM-1030) if no subject is present in context, distinct from the existing `ErrorUnauthorized` (403) used for authenticated-but-not-authorized failures elsewhere in the package.
- Added permission entries `GET /scim/v2/Me` and `PUT /scim/v2/Me` in `backend/internal/system/security/permissions.go` (empty permission string, consistent with other self-service paths, since access is scoped by the caller's own identity rather than a role).
- Added unit tests in `backend/internal/scim/users_handler_test.go` covering GET/PUT success and error paths (unauthenticated, not found, invalid body, validation failures).

### Related Issues
- #4575 

### Related PRs
- #4424 (SCIM search endpoint, pagination defaults, attribute projection)
- #4515 (undeclared/conflicting SCIM attribute validation)

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks


backend/internal/scim/users_handler.go
```diff
@@ -162,6 +239,65 @@ func (h *scimUsersHandler) HandleUsersReplaceRequest(w http.ResponseWriter, r *h
...
+	if err != nil || len(body) == 0 {
+		h.handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
+	payload, svcErr := ValidateSCIMUserRequest(body)
+	if svcErr != nil {
+		h.handleSCIMError(w, r, svcErr)
+		return
+	}
+	runtimeCtx := security.WithRuntimeContext(ctx)
```
sadilchamishka: Blocking: this disables authorization for the whole request.

`security.IsRuntimeContext` is the internal-caller escape hatch — `sysauthz.IsActionAllowed` returns `true, nil` unconditionally for a runtime context (system/sysauthz/service.go:98), and `GetAccessibleResources` returns `AllAllowed: true` (:191). Wrapping the ctx here means an ordinary end user's `PUT /scim/v2/Me` runs the entire write path (`GetUser`, `UpdateUser`, and everything nested) with authorization switched off, not just the entity-type lookup.

Three things suggest this isn't intended:

- The admin route `HandleUsersReplaceRequest` passes plain `ctx` (line 136).
- `ReplaceUser` already creates its own narrowly-scoped `runtimeCtx` for exactly the lookups that need it (users_service.go:155, 183).
- `HandleMeGetRequest` doesn't wrap, and works.

Please pass `ctx`. If this was added because the call was being denied, the fix belongs in the authz layer (a self-service action for the subject's own resource), not a blanket bypass on a user-facing route.

Separately, the description says these handlers mirror the native `/users/me` endpoint. They don't: `HandleSelfUserPutRequest` (internal/user/handler.go:402) calls `UpdateUserAttributes` — attributes only, credential fields rejected, no runtime context — whereas this calls the full admin `ReplaceUser`, which also sets `OUID` and `Type`. Worth confirming that the wider surface is what you want for self-service.

backend/internal/scim/users_resource.go
```diff
@@ -85,6 +85,75 @@ func buildSCIMUserListResponse(users []SCIMUser, totalResults, startIndex, items
 	}
 }
 
+var alwaysReturnedUserAttrs = map[string]struct{}{"id": {}, "schemas": {}, "meta": {}}
+
+// projectSCIMAttributes prunes a marshaled SCIM resource's top-level keys per
+// RFC 7644 §3.9. attributes takes precedence over excludedAttributes when both
+// are supplied. id/schemas/meta are always kept.
+func projectSCIMAttributes(
```
sadilchamishka, line 108: Projection only prunes **top-level** keys, but every ThunderID user attribute lives nested under the extension URN object (see `SCIMUser.MarshalJSON`). Two consequences:

- `?attributes=given_name` returns a resource containing only `id`/`schemas`/`meta` — the requested attribute can never match a top-level key.
- `?attributes=userName` silently drops the **entire** extension object, since the URN key is neither in `alwaysReturnedUserAttrs` nor in `keep`.

RFC 7644 §3.9 also requires sub-attribute paths (`name.givenName`) and URN-qualified names (`urn:...:User:department`), neither of which is handled. Silently returning near-empty resources is worse than not supporting the parameter — please either make projection URN- and sub-attribute-aware, or reject paths you can't honor with 400 `invalidValue`.

Two smaller §3.9 points:

- `attributes` and `excludedAttributes` are mutually exclusive; supplying both should be 400 `invalidRequest` rather than silently preferring `attributes`.
- Projection is wired only into List and `.search`. `GET /Users/{id}`, `GET /Me`, POST and PUT ignore both parameters, but §3.9 applies to every response that returns a resource.

backend/internal/scim/users_handler.go
```diff
@@ -64,16 +69,88 @@ func (h *scimUsersHandler) HandleUsersListRequest(w http.ResponseWriter, r *http
...
+				Schemas:  []string{SCIMErrorSchemaURN},
+				Status:   "400",
+				ScimType: "invalidFilter",
+				Detail:   err.Error(),
+			})
+			return
+		}
+	}
+
+	startIndex := searchReq.StartIndex
```
sadilchamishka, line 157: Two ServiceProviderConfig mismatches on this endpoint:

1. `count` is not capped against the advertised `filter.maxResults = 200` (scim/config/config.go:54), so `"count": 100000` is passed straight through to `ListUsers`. Same gap exists in `HandleUsersListRequest`; worth fixing both while you're here.
2. `sortBy`/`sortOrder` are parsed into `SCIMSearchRequest` and then silently ignored, while the config advertises `sort.supported = false` (config.go:62). A 200 with arbitrarily-ordered results is a trap for clients — return 400 `invalidValue` for a sort request until sorting is implemented.

And a third, pre-existing but now on a second endpoint: `filter.supported` is `false` in the config (config.go:50) while both this handler and the GET list accept filters. The config should be corrected.

backend/internal/scim/users_service.go
```diff
@@ -119,17 +122,22 @@ func (s *scimUsersService) CreateUser(
...
+				log.Any("error", svcErr))
+			return nil, svcErr
 		}
 	}
 	if missing := missingRequiredAttrs(payload.ExtensionAttrs, et.Schema, false); len(missing) > 0 {
 		logger.Debug(ctx, "SCIM CreateUser: missing required attributes for user type",
 			log.String("userType", canonicalName), log.Any("missing", missing))
 		return nil, newMissingRequiredAttributesError(canonicalName, missing)
 	}
+	if undeclared := undeclaredAttrs(payload.ExtensionAttrs, et.Schema); len(undeclared) > 0 {
```
sadilchamishka: This is a breaking behavior change — requests carrying extra attributes used to succeed and will now get a 400. It needs calling out in the description (and arguably belongs in its own PR alongside the conflicting-value check).

Two gaps in the check itself:

- It's asymmetric: an unknown key in the extension object is a hard 400, but an unmapped **core** attribute (e.g. `active` or `nickName` with no schema mapping) is still silently dropped by `reverseMapCoreAttrsForSchema`. From a client's point of view, one channel is strict and the other quietly discards data.
- `undeclaredAttrs` returns `nil` when `json.Unmarshal(schema, ...)` fails, so for a user type with an empty or malformed schema the validation is silently skipped rather than erroring.

backend/internal/system/security/permissions.go
```diff
@@ -282,6 +284,7 @@ func InitSystemPermissions(handle string) {
 		{"PUT /scim/v2/Groups/**", p.Group},
 		{"PATCH /scim/v2/Groups/**", p.Group},
 		{"DELETE /scim/v2/Groups/**", p.Group},
+		{"POST /scim/v2/Users/.search", p.UserView},
```
sadilchamishka: Nit: this belongs in the SCIM Users block above (next to `POST /scim/v2/Users`), not after the Groups entries. `apiPermissionEntries` is documented as ordered first-match-wins, so if anyone later adds a `POST /scim/v2/Users/**` entry in the Users block, it would silently take precedence over this one and hand `.search` the stronger `p.User` permission.

`p.UserView` is the right choice for a read-only search, though.

backend/internal/scim/users_handler.go
```diff
@@ -204,6 +340,29 @@ func (h *scimUsersHandler) handleSCIMError(w http.ResponseWriter, r *http.Reques
...
 
+// writeUserListResponse writes listResp, applying attribute projection
+// (RFC 7644 §3.9) when attributes/excludedAttributes were requested.
+func (h *scimUsersHandler) writeUserListResponse(
+	ctx context.Context, w http.ResponseWriter, listResp SCIMUserListResponse,
+	attributes, excludedAttributes []string,
+) {
+	projected, err := projectSCIMUserListResponse(listResp, attributes, excludedAttributes)
+	if err != nil {
+		log.GetLogger().Error(ctx, "SCIM attribute projection failed", log.Any("error", err))
```
sadilchamishka, line 403: Nit: bare `log.GetLogger()` here, whereas every other function in this file uses `log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))` — this error will land without the component field.

backend/internal/scim/users_resource.go
```diff
@@ -85,6 +85,75 @@ func buildSCIMUserListResponse(users []SCIMUser, totalResults, startIndex, items
...
+
+// projectSCIMUserListResponse rebuilds a list response with each resource's
+// attributes pruned. Returns nil, nil if no projection was requested.
+func projectSCIMUserListResponse(
+	listResp SCIMUserListResponse, attributes, excludedAttributes []string,
+) (map[string]interface{}, error) {
+	if len(attributes) == 0 && len(excludedAttributes) == 0 {
+		return nil, nil
+	}
+	raw, err := json.Marshal(listResp)
```
sadilchamishka: Nit: the whole list (including every resource) is marshaled here just to build the envelope, and then each resource is marshaled a second time in the loop below. Marshaling a shallow copy with `Resources` set to `nil` for the envelope, or reusing the resources already decoded from `raw`, avoids the duplicate work.

sadilchamishka (changes requested):
Reviewed the full diff. One blocking issue (authorization bypass on `PUT /Me`), a set of RFC 7644 §3.9 / §3.4.3 gaps in the attribute-projection and `.search` additions, and a scope point.

**Scope:** the title and description cover `/Me` GET+PUT only, but the diff also adds `POST /Users/.search`, §3.9 attribute projection, conflicting core/custom-value validation, undeclared-attribute rejection, and changes the SCIM default page size from 20 to `serverconst.DefaultPageSize` = 30 — which alters existing `GET /scim/v2/Users` and `GET /scim/v2/Groups` responses. There is also unrelated `"employee"` → `testUserTypeEmployee` churn across the test files, which AGENTS.md discourages ("keep changes minimal and focused on the task requested"). Please either split this up or, at minimum, update the description: as it stands a reviewer reading it would not know the page size changed.

**Description mismatches the code:** it says both `/Me` handlers "return 401 (`ErrorUnauthorized`) if no subject is present". `tidcommon.ErrorUnauthorized` maps to **403** in `mapSCIMError` (handler.go:195), which is what the new tests assert (`..._Returns403`). It also says the handlers "mirror the pattern already used by the native `/users/me` self-service endpoint" — they don't; see the inline comment on `HandleMeReplaceRequest`.

**Also missing:** integration tests and docs for two new public endpoints (both unchecked in the PR checklist).

backend/internal/scim/users_handler.go
```diff
@@ -162,6 +239,65 @@ func (h *scimUsersHandler) HandleUsersReplaceRequest(w http.ResponseWriter, r *h
...
+		return
+	}
+	payload, svcErr := ValidateSCIMUserRequest(body)
+	if svcErr != nil {
+		h.handleSCIMError(w, r, svcErr)
+		return
+	}
+	runtimeCtx := security.WithRuntimeContext(ctx)
+	replaced, svcErr := h.svc.ReplaceUser(runtimeCtx, userID, payload, r.Header.Get("If-Match"), h.baseURL)
+	if svcErr != nil {
```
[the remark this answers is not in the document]

backend/internal/scim/init.go
```diff
@@ -160,7 +160,35 @@ func registerRoutes(mux *http.ServeMux, h *scimHandler, uh *scimUsersHandler, gh
...
+		"OPTIONS "+SCIMBasePath+"/Me",
+		func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusNoContent) },
+		optsCRUD,
+	))
+	// users/.search endpoint
+	mux.HandleFunc(middleware.WithCORS(
+		"POST "+SCIMBasePath+"/Users/.search",
+		uh.HandleUsersSearchRequest,
+		optsCRUD,
+	))
```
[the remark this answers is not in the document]
ravindu439: it is in separate PR

backend/internal/scim/users_handler.go
```diff
@@ -162,6 +239,65 @@ func (h *scimUsersHandler) HandleUsersReplaceRequest(w http.ResponseWriter, r *h
...
+// Resolves the authenticated subject and processes the request directly
+// against the aliased User resource, per RFC 7644 §3.11 option 3.
+func (h *scimUsersHandler) HandleMeGetRequest(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
+
+	userID := strings.TrimSpace(security.GetSubject(ctx))
+	if userID == "" {
+		h.handleSCIMError(w, r, &tidcommon.ErrorUnauthorized)
+		return
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/users_handler.go

backend/internal/scim/users_resource.go
ravindu439: fixed

backend/internal/scim/users_handler.go
ravindu439: fixed

backend/internal/scim/users_service.go
ravindu439: fixed

backend/internal/system/security/permissions.go
ravindu439: fixed

backend/internal/scim/users_resource.go
ravindu439: fixed

sadilchamishka (approved):

PR #4586, opened 2026-08-05
Replace Wayfinder sample SQLite storage with in-memory data

PasinduYeshan:
### Purpose

Installing the Wayfinder sample fails on Node.js 26. `better-sqlite3` 11.x ships no prebuild for the Node 26 ABI and its sources no longer compile against Node 26 V8 headers, so the install ends in `node-gyp` errors.

The package was only the Node 18/20 fallback for `node:sqlite`, yet declared as a required dependency, so every install built it. Following the maintainer suggestion, the sample data now lives in memory and SQLite is gone.

### Approach

- New `backend/src/data.js` holds the sample flights, hotels, and trips in API shape.
- `backend/src/db.js` is now an in-memory store. Every export keeps its name and signature, so `server.js` and `mcp.js` are unchanged.
- Removed `scripts/seed.js`, `src/sqlite.js`, the `better-sqlite3` dependency, and the matching `allowBuilds` entry. `npm run seed` is gone, so the docs and the `thunderid` CLI sample runner drop that step.
- Removed the CI step that installed `build-essential`. `allowBuilds` is a strict allowlist and the only remaining entry, `sharp`, installs prebuilt binaries.
- Bookings and upgrade requests are in memory too. Cold-start behaviour is unchanged, since `seed.js` already dropped every table on each start.

Also fixed while here: an absent `maxNightlyRate` was coerced to `0`, so `GET /api/hotels` filtered on `nightlyRate <= 0` and always returned an empty list. `openapi.yaml` marks the parameter optional.

No tests added, per review feedback that the sample does not need them. Verified manually against a running server: `/health`, `/api/flights`, `/api/hotels` (uncapped, capped, and empty parameter), `/api/trips`, `/api/locations`, booking create and its 409 duplicate, `/api/bookings/flights`, delete, and both `/api/demo/*-business-class` toggles.

<details>
<summary><code>pnpm-lock.yaml</code>: 158 deletions, 0 insertions, no version bumps</summary>

The only hand edit is the `better-sqlite3` specifier under the `samples/apps/wayfinder-sample/backend` importer (3 lines). pnpm regenerated the rest: 20 packages that nothing else in the workspace reaches, each appearing once in `packages` (66 lines) and once in `snapshots` (89 lines).

```
better-sqlite3@11.10.0
├── bindings@1.5.0
│   └── file-uri-to-path@1.0.0
└── prebuild-install@7.1.3
    ├── expand-template@2.0.3
    ├── github-from-package@0.0.0
    ├── mkdirp-classic@0.5.3
    ├── napi-build-utils@2.0.0
    ├── node-abi@3.94.0
    ├── pump@3.0.4
    │   └── end-of-stream@1.4.5
    ├── simple-get@4.0.1
    │   └── simple-concat@1.0.1
    ├── tar-fs@2.1.5
    │   ├── chownr@1.1.4
    │   └── tar-stream@2.2.0
    │       ├── bl@4.1.0
    │       │   └── buffer@5.7.1
    │       └── fs-constants@1.0.0
    └── tunnel-agent@0.6.0
```

That subtree is the native-addon install machinery itself: `prebuild-install` fetches a prebuilt `.node` for the current ABI (`node-abi` picks the tag, `simple-get` and `tunnel-agent` download it, `tar-fs`, `tar-stream`, `bl` and `chownr` unpack it), and `bindings` loads it at runtime. It is exactly the chain that failed on Node 26.

Shared dependencies keep their entries and lose only their reference lines under the deleted snapshots: `semver`, `inherits`, `safe-buffer`, `detect-libc`, `minimist`, `readable-stream`, `base64-js`, `rc`, `once`, `ieee754`, `decompress-response`. Each still has at least one live referrer. `buffer` is deduped at two versions and only `5.7.1` goes, since `bl` was its sole consumer.

</details>

### Related Issues
- Fixes #4585

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * The Wayfinder sample app now uses in-memory sample data for flights, hotels, and trips.
  * Bookings and upgrade requests are available during runtime and reset when the server restarts or reloads.

* **Bug Fixes**
  * Hotel searches without a nightly-rate limit now return results correctly instead of applying a zero limit.

* **Documentation**
  * Updated setup instructions and architecture documentation to reflect the simplified startup process and in-memory data behavior.


samples/apps/wayfinder-sample/backend/src/db.js
```diff
@@ -299,324 +105,261 @@ export function createBookingRecord({
...
+    toFlightId: null,
+    priceDifference: 0,
+    idToken: idToken ?? null,
+    status: "pending",
+    createdAt,
+    updatedAt: createdAt
+  });
 
   return { id, userId, email, bookingId, fromFlightId, status: "pending", createdAt };
 }
```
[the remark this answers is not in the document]
PasinduYeshan: Confirmed that these fields are unused, so the fix removes them instead of populating them.

`getUpgradeRequestById` has only one caller: the `process_upgrade` MCP tool. It uses only `userId`, `status`, `fromFlightId`, and `bookingId`. The tool resolves the target flight using `findMatchingBusinessFlight(upgradeRequest.fromFlightId)` and calculates the price difference from that result, so it does not need the persisted `toFlightId` or `priceDifference`.

The values read by `ai-agent/agent.ts` come from `get_pending_upgrades`, backed by `getOnePendingUpgrade`, which calculates both values dynamically and remains unchanged.

Populating these fields during a status update would introduce new write behavior that did not exist in the SQLite implementation, where `to_flight_id` remained `NULL` and `price_difference` remained `0` throughout the request lifecycle. It would also create a second, potentially stale source for values already calculated by the consumer. Removing the fields from `createUpgradeRequest` and the `getUpgradeRequestById` response eliminates this stale data path entirely.

samples/apps/wayfinder-sample/backend/src/db.test.js
```diff
@@ -0,0 +1,367 @@
+// Copyright 2026 The ThunderID Authors
```
ThaminduDilshan: We don't need unit tests for the sample

PasinduYeshan: Removed.

ThaminduDilshan (approved):

PR #4587, opened 2026-08-05
Add the agent identity playground sample

Dilusha-Madushan:
### Purpose

Adds a new sample, `agent-identity-playground-sample`, covering the three ways an AI
agent can hold identity in ThunderID. Each posture gets its own tab, and every tab runs
real OAuth calls against a local instance and decodes the tokens that come back, so the
`sub`, `act`, and `scope` claims are shown from actual responses rather than described.

| Tab | Grant | `sub` | `act` | `scope` |
|---|---|---|---|---|
| Acts as itself | `client_credentials` | the agent | none | `booking:read booking:recommend` |
| Acts for a customer | `authorization_code` | the user | the agent | `booking:read booking:create` |
| Delegates to agents | RFC 8693 exchange | the user | worker, then agent | `booking:read` |

One travel booking cast is shared across all three tabs (a Concierge agent, a worker
agent it delegates to, and a customer), so the only thing that changes between tabs is
who the token names and what it is allowed to do.

Also wires the sample into `build.sh` and `build.ps1` so it is packaged from source
alongside the existing four samples.

### Approach

**Zero dependencies.** The app is Node built-ins only (`node:http`, `node:https`,
`node:crypto`), so there is no install step and nothing to keep patched. `package.json`
declares no dependencies at all, which keeps its lockfile footprint to a single empty
importer entry.

**One resource file, sliced per tab.** `thunderid-config/thunderid-config.yaml` is the
single source of truth. Each tab's download is produced at request time by
`lib/resources.mjs`, which splits the multi document YAML and selects by each document's
`id`. Slices also resolve their own dependencies: any slice containing an agent with an
`authFlowHandle` automatically carries the flow declaring that handle, so a per tab
import cannot be incomplete. This avoids duplicating agent definitions across four files.

**Self contained resources.** Every id is prefixed `agent-identity-`, the resource servers
use identifiers under `travel.example`, and the user is `johndoe`, so importing cannot
collide with anything already in an instance. Import is upsert by id, so re importing, or
importing overlapping per tab slices, is safe.

**One module per posture.** A scenario exports its tab metadata, an `explain` function, a
`render` function, the agents it `requires`, and optionally extra `routes` and a `check`.
Adding a fourth posture means adding one file and listing it in `server.mjs`. Startup
validates that every agent a scenario needs has credentials in `config.json` and that
every document it offers exists in the resource file, so a renamed resource fails fast
with the missing name rather than at request time.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4069

## Summary by CodeRabbit

* **New Features**
  * Added an Agent Identity Playground demonstrating self-acting agents, delegated agent chains, and acting on behalf of a user.
  * Added interactive OAuth sign-in, token exchange, refresh flows, scope narrowing, and delegation-chain visualizations.
  * Added readiness checks, token claim displays, error guidance, theme support, and clipboard copying.
  * Added a command-line flow for inspecting issued token claims.
* **Documentation**
  * Added setup instructions, configuration guidance, scenario explanations, and extension conventions.
* **Configuration**
  * Added a complete local travel-booking demonstration environment with agents, resources, roles, and permissions.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2673,6 +2673,8 @@ importers:
         specifier: '>=4.1.0'
         version: 4.1.10(@opentelemetry/api@1.9.0)(@types/node@24.7.2)(@vitest/browser-playwright@4.1.10)(@vitest/coverage-istanbul@4.1.10)(jsdom@27.0.1)(vite@8.0.16(@types/node@24.7.2)(esbuild@0.28.1)(jiti@2.7.0)(sass-embedded@1.98.0)(sass@1.98.0)(terser@5.47.1)(tsx@4.22.4)(yaml@2.8.3))
 
+  samples/apps/agentid-playground: {}
+
   samples/apps/react-api-based-sample:
     dependencies:
       '@wso2/oxygen-ui':
--- a/samples/apps/agentid-playground/.env.example
+++ b/samples/apps/agentid-playground/.env.example
@@ -1,0 +1,12 @@
+# Copy to .env. Loaded automatically by npm start.
+
+# ThunderID backend. Overrides thunderidBaseUrl in config.json.
+THUNDERID_BASE_URL=https://localhost:8090
+
+# Which scenarios to enable, comma separated: self, obo, chain. Defaults to all three.
+# SCENARIOS=self,chain
+
+# Port for this app. Defaults to the port in config.json appBaseUrl.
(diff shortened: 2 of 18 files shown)


brionmario:
Should we integrate with the CLI?

brionmario (approved):

PR #4591, opened 2026-08-05
Add criteria-based revocation administration flows

indeewari:
### Purpose

Add criteria-based token revocation and reusable administrative flow execution to ThunderID. This enables administrators to design and invoke a secure user deletion flow that validates the operation, revokes all tokens issued to the user, terminates the user's sessions, and then deletes the user.

The Console flow builder now exposes an `ADMINISTRATION` flow type, reusable revocation and user deletion executors, a secure blank administrative starter, and a complete User Deletion Flow template. Native `DELETE /users/{id}` behavior remains unchanged.

### Approach

- Extend the existing `REVOCATION_CRITERIA` write and enforcement paths with typed criteria and a generic `RevocationIdentity` contract.
- Enforce token revocation by JTI, token family, and subject on OAuth and native API hot paths using fail-closed behavior.
- Add `flowId` execution through `POST /flow/execute` for `ADMINISTRATION` flows, including continuation support and trusted runtime data across call nodes.
- Require an authenticated caller at the administrative flow service boundary for both initiation and continuation. Keep `PermissionValidator` in the starter templates as defense in depth.
- Add reusable `UserDeletionValidationExecutor`, `CriteriaRevocationExecutor`, `SessionRevocationExecutor`, and `UserDeleteExecutor` building blocks.
- Provide the secure default sequence: permission validation, deletion validation, token revocation, session termination, then physical user deletion.
- Keep revocation writes idempotent and prevent later boundary revocations from weakening an existing terminal revocation.
- Preserve the native user deletion endpoint so adopting flow-based administration is additive in this phase.

### Validation

- `go test ./...` passed.
- Focused Console flow tests passed: 2 files, 29 tests.
- Prettier checks passed for all changed Console files.
- Generated mocks were refreshed successfully.
- `make test_integration` passed in the developer environment.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added administrative flows, including user deletion and blank flow templates.
  * Added flow execution by flow ID with system-permission authorization.
  * Added configurable revocation modes and console setup.
  * Added criteria-based revocation for tokens, sessions, and users.
  * Added subject-based token revocation and session termination.
* **Bug Fixes**
  * Improved revocation handling across token, session, and subject criteria.
  * Added clearer authentication, authorization, and deletion error responses.
* **Documentation**
  * Updated the Flow Execution API documentation for administrative flows.


backend/cmd/server/servicemanager.go
```diff
@@ -362,6 +361,8 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 			OpenID4VPVerifierSvc:  openid4vpSvc,
 			SessionService:        sessionService,
 			ResourceService:       resourceServerProvider,
+			UserService:           userService,
```
ThaminduDilshan, line 362: Need to think whether we can plug user service as a executor dependency. This already has `EntityProvider` and components in gate should use that.
But seem like we're going to miss some validations we're performing at user service when invoking entity provider directly. Need to check this further.

cc: @senthalan @ThumulaPerera @thiva-k

backend/cmd/server/servicemanager.go
```diff
@@ -549,14 +550,27 @@ func initSessionService(ctx context.Context, svc serverconfig.ServerConfigServic
...
-// interface, fixing the revocation reason to session sign-out.
-type sessionCriteriaRevoker struct {
+// criteriaRevokerAdapter adapts the OAuth criteria revoker to the narrower flow-executor and SSO
+// session consumer interfaces. Token-family revocation initiated by session sign-out uses the
+// session-logout reason.
+type criteriaRevokerAdapter struct {
 	revoker revocation.CriteriaRevokerInterface
 }
 
+func (a criteriaRevokerAdapter) RevokeByCriteria(ctx context.Context, request executor.CriteriaRevocation) error {
```
ThaminduDilshan: Shall we add a method comment?

backend/cmd/server/servicemanager.go
```diff
@@ -549,14 +550,27 @@ func initSessionService(ctx context.Context, svc serverconfig.ServerConfigServic
 	return sessionService, cfg
 }
 
-// sessionCriteriaRevoker adapts the OAuth criteria revoker to the SSO session service's consumer
-// interface, fixing the revocation reason to session sign-out.
-type sessionCriteriaRevoker struct {
+// criteriaRevokerAdapter adapts the OAuth criteria revoker to the narrower flow-executor and SSO
+// session consumer interfaces. Token-family revocation initiated by session sign-out uses the
+// session-logout reason.
+type criteriaRevokerAdapter struct {
```
ThaminduDilshan: This looks like an existing problem, but I feel like it's wrong to have it defined here. Service manager just do the service initializations and wire the dependencies. But here we're kind of wiring a business logic.

Additionally I see an issue with the revocation service initialization. At [1] and [2], we're creating two service objects which kind of goes against the dependency injection pattern. Both these locations are invoked from the service manager code.

[1] https://github.com/thunder-id/thunderid/blob/d80cadcbd4cd4a048393c684ae86cb32e73af884/backend/internal/oauth/oauth2/revocation/init.go#L40
[2] https://github.com/thunder-id/thunderid/blob/d80cadcbd4cd4a048393c684ae86cb32e73af884/backend/internal/oauth/oauth2/revocation/init.go#L55

backend/internal/flow/executor/constants.go
```diff
@@ -109,6 +113,12 @@ const (
 	propertyKeyPromptOnSignOut = "promptOnSignOut"
 )
 
+const (
```
ThaminduDilshan: These are taken as flow execution inputs right? Shall we move these to the proper grouping? We already have a constant block defined for such fields

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type UserDeletionService interface {
+	ValidateDeleteUser(ctx context.Context, userID string) *common.ServiceError
+	DeleteUser(ctx context.Context, userID string) *common.ServiceError
+}
+
+type userDeletionValidationExecutor struct {
```
ThaminduDilshan: Let's add struct and method comments. Check all the places

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type UserDeletionService interface {
```
ThaminduDilshan:
```suggestion
type userDeletionProvider interface {
```

We can make this private. Also let's follow the provider naming convention.

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	ValidateDeleteUser(ctx context.Context, userID string) *common.ServiceError
+	DeleteUser(ctx context.Context, userID string) *common.ServiceError
+}
+
+type userDeletionValidationExecutor struct {
+	providers.Executor
+	users UserDeletionService
+}
+
+func newUserDeletionValidationExecutor(factory core.FlowFactoryInterface,
```
ThaminduDilshan: Since constructor method is returning a pointer of the specific object, if there's a interface mismatch, this will fail at runtime. In order to avoid it either;

1. We can set return type to `providers.Executor` (common interface)
2. Have below line to fail it at compilation time
    `var _ providers.Executor = (*provisioningExecutor)(nil)`

Let's follow approach 2. It's the pattern followed in executors

ThaminduDilshan: Check this in other executor codes too

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	users UserDeletionService) *userDeletionValidationExecutor {
+	base := factory.CreateExecutor(ExecutorNameUserDeletionValidation, providers.ExecutorTypeUtility,
+		[]providers.Input{
+			{Identifier: revocationInputSubject, Type: providers.InputTypeText, Required: true},
+			{Identifier: revocationInputMode, Type: providers.InputTypeText, Required: true},
+		}, nil, &providers.ExecutorMeta{})
+	return &userDeletionValidationExecutor{Executor: base, users: users}
+}
+
+func (e *userDeletionValidationExecutor) Execute(ctx *providers.NodeContext) (*providers.ExecutorResponse, error) {
```
ThaminduDilshan: On a side question, why do we need a separate executor to perform this validation? Can't we do this as a pre check in the same user deletion executor?

This is just going to increase the no of executor we're going to have when onboarding other deletion/ revocation scenarios

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	if !e.HasRequiredInputs(ctx, execResp) {
+		execResp.Status = providers.ExecUserInputRequired
+		return execResp, nil
+	}
+	subject, ok := ctx.ConsumeInput(revocationInputSubject)
+	if !ok || subject == "" {
+		return nil, fmt.Errorf("subject is required")
+	}
+	mode, ok := ctx.ConsumeInput(revocationInputMode)
+	if !ok || mode != "revoke_all" {
```
ThaminduDilshan: Shall we define proper constants for these? We should already have a constant list for the valid modes.
Check the same if you're adding anymore modes as well

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type UserDeletionService interface {
+	ValidateDeleteUser(ctx context.Context, userID string) *common.ServiceError
+	DeleteUser(ctx context.Context, userID string) *common.ServiceError
+}
+
+type userDeletionValidationExecutor struct {
+	providers.Executor
+	users UserDeletionService
```
ThaminduDilshan:
```suggestion
	userProvider UserDeletionService
```

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	if e.users == nil {
+		return nil, fmt.Errorf("user service is not configured")
+	}
+	if svcErr := e.users.ValidateDeleteUser(ctx.Context, subject); svcErr != nil {
+		execResp.Status = providers.ExecFailure
+		execResp.Error = svcErr
+		return execResp, nil
+	}
+	plan, err := encodeRevocationPlan(revocationPlan{
+		Criteria: []RevocationCriterion{{Type: "subject", Value: subject}},
```
ThaminduDilshan: Better to define constants with proper go types for revocation criteria types, modes, reason, etc. Preferably in the revocation package

backend/internal/flow/executor/revocation_plan.go
```diff
@@ -0,0 +1,57 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package executor
```
ThaminduDilshan: Shall we not introduce this file and decompose models and methods in to existing files? Can use executor/utils.go for the utility methods.

Also make them private if not required to be exposed outside executor package. 

If required to be exposed, I feel like this is not the proper package to add them. This doesn't belong to executor. Even though these are in a separate file, for a package importing this, it's going to come from executor package

backend/internal/flow/executor/revocation_plan.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+type revocationPlan struct {
+	Criteria []RevocationCriterion `json:"criteria"`
+	Mode     string                `json:"mode"`
+	Cutoff   time.Time             `json:"cutoff,omitempty"`
+	Reason   string                `json:"reason"`
+}
+
+func encodeRevocationPlan(plan revocationPlan) (string, error) {
```
ThaminduDilshan: Add comments to all the applicable structs and methods

backend/internal/flow/executor/revocation_workflow_executors_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/system/cache"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+	"github.com/thunder-id/thunderid/tests/mocks/flow/sessionmock"
+)
+
+func TestUserDeletionFlowExecutors(t *testing.T) {
```
ThaminduDilshan: Let's follow test suite pattern. Check all the places where you've added new tests

backend/internal/flow/executor/user_deletion_validation_executor.go
ThaminduDilshan: Okay, maybe a separate executor was added because we first need to check the conditions, and if satisfied only, then revoke tokens, revoke sessions and finally delete the user.
But I still have a concern on the no of executors we're going to add. If no other solution, fine to keep it. But let's think through and decide

backend/internal/flow/executor/session_revocation_executor.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/flow/session"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type sessionRevocationExecutor struct {
+	providers.Executor
+	sessions session.Service
```
ThaminduDilshan:
```suggestion
	sessionSvc session.Service
```

backend/internal/flow/executor/user_delete_executor.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type userDeleteExecutor struct {
+	providers.Executor
+	users UserDeletionService
```
ThaminduDilshan:
```suggestion
	userProvider UserDeletionService
```

backend/internal/flow/executor/user_delete_executor.go
```diff
@@ -0,0 +1,46 @@
...
+		if criterion.Type == "subject" {
+			subject = criterion.Value
+			break
+		}
+	}
+	if subject == "" {
+		return nil, fmt.Errorf("trusted revocation plan has no subject")
+	}
+	if svcErr := e.users.DeleteUser(ctx.Context, subject); svcErr != nil {
+		return &providers.ExecutorResponse{Status: providers.ExecFailure, Error: svcErr}, nil
```
ThaminduDilshan: If this needs to be handled as a server error, we should return a go error instead. With this pattern, it will be handled as a client error.

Also we shouldn't return external service error codes/structs from a different service. For this, we should define executor error struct with a proper code. We can keep a single error constant and use it with customServiceError() method to modify the error description and details.

Check other places as well. Other executors also could be handling the same way

backend/internal/flow/executor/user_deletion_validation_executor.go
```diff
@@ -0,0 +1,67 @@
...
+	users UserDeletionService
+}
+
+func newUserDeletionValidationExecutor(factory core.FlowFactoryInterface,
+	users UserDeletionService) *userDeletionValidationExecutor {
+	base := factory.CreateExecutor(ExecutorNameUserDeletionValidation, providers.ExecutorTypeUtility,
+		[]providers.Input{
+			{Identifier: revocationInputSubject, Type: providers.InputTypeText, Required: true},
+			{Identifier: revocationInputMode, Type: providers.InputTypeText, Required: true},
+		}, nil, &providers.ExecutorMeta{})
```
ThaminduDilshan: Let's also define `ExecutorMeta`. This allows flow creation validator to validate for the applicable inputs, modes, etc. 

Check this for other executors too

backend/internal/flow/flowexec/handler.go
```diff
@@ -59,9 +60,16 @@ func (h *flowExecutionHandler) HandleFlowExecutionRequest(w http.ResponseWriter,
 	// them available to the flow service, which selects the handle once the flow is known.
 	ctx := session.WithInbound(r.Context(), h.ssoTransport.Read(r))
 
-	flowStep, flowErr := h.flowExecService.Execute(
-		ctx, appID, executionID, flowTypeStr, verbose, action, inputs, challengeToken,
-		flowSecret, attestationToken)
+	var flowStep *FlowStep
+	var flowErr *tidcommon.ServiceError
+	if flowID != "" {
```
ThaminduDilshan, line 65: Shall we consider this path (execution by providing flowid) only for the administrative flows for now? Otherwise this is going to allow flow execution by providing flow id for all the types, even authentication and registration.

Up to discussion.

cc: @senthalan

backend/cmd/server/servicemanager.go
```diff
@@ -325,15 +325,14 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	csp.InitializeConfigReader(serverConfigService)
```
ThaminduDilshan, line 325: Two comments for the PR;

1. Shall we add bootstrap flow definitions for the user deletion flow (and revocation too if applicable)
2. Shall we have server configs (server config API) to specify handle of the user deletion flow (and any other applicable flows we need to recognize)?

backend/internal/flow/session/interface.go
```diff
@@ -18,6 +18,7 @@ type sessionStore interface {
 	// GetByExecutionID fetches the session established by the given flow execution, or (nil, nil)
 	// when that execution has not established one.
 	GetByExecutionID(ctx context.Context, flowExecutionID string) (*Session, error)
+	ListBySubject(ctx context.Context, subjectID string) ([]Session, error)
```
ThaminduDilshan, line 22: Shall we add a comment?

backend/internal/flow/session/service.go
```diff
@@ -58,6 +58,34 @@ type Service interface {
 	// against ending a session grouped under a different flow. It is idempotent, returning (nil, nil)
 	// when no session matches the handle, and returns the deleted session on success.
 	Terminate(ctx context.Context, handle, flowID string) (*Session, error)
+	// TerminateBySubject ends every SSO session belonging to the subject.
```
ThaminduDilshan:
```suggestion

	// TerminateBySubject ends every SSO session belonging to the subject.
```

backend/internal/flow/session/store_constants.go
```diff
@@ -40,6 +40,13 @@ var (
 			`FROM "SSO_SESSION" WHERE FLOW_EXECUTION_ID = $1 AND DEPLOYMENT_ID = $2`,
 	}
 
+	queryListSessionsBySubject = model.DBQuery{
+		ID: "SSO-SESS-13",
```
ThaminduDilshan: Shall we move this to the end of file so that query constants are in ascending order?

backend/internal/oauth/oauth2/revocation/model.go
```diff
@@ -34,8 +34,67 @@ const (
...
 	// revocation of a token that carries a token family id.
 	RevocationReasonExplicitTokenFamily RevocationReason = "explicit_token_family"
+	// RevocationReasonApplicationDeleted permanently revokes artifacts issued to a deleted application.
+	RevocationReasonApplicationDeleted RevocationReason = "application_deleted"
+	// RevocationReasonApplicationSecretRegenerated revokes artifacts established before secret regeneration.
+	RevocationReasonApplicationSecretRegenerated RevocationReason = "application_secret_regenerated"
+	// RevocationReasonRoleRemoved revokes artifacts established before a role was removed.
+	RevocationReasonRoleRemoved RevocationReason = "role_removed"
+	// RevocationReasonGroupMembershipRemoved revokes artifacts established before group membership was removed.
+	RevocationReasonGroupMembershipRemoved RevocationReason = "group_membership_removed"
```
ThaminduDilshan: Two constants are added for `RevocationReasonRoleRemoved` and `RevocationReasonGroupMembershipRemoved`. Should we also consider role assignment removal and group removal too?

Or does this `RevocationReasonRoleRemoved` capture both membership removal and role deletion? If so better to keep these two (role and group revocation reasons) consistent

backend/internal/oauth/oauth2/revocation/service.go
```diff
@@ -53,6 +53,9 @@ type RefreshTokenRevokerInterface interface {
 // consumed by the refresh grant (reuse), the RFC 7009 endpoint (explicit), the authorization service
 // (code replay), and session sign-out (logout).
 type CriteriaRevokerInterface interface {
```
ThaminduDilshan, line 56: Not related to this PR, but as per the project conventions, all services should return a service error instead of an go error.
Returning a go error does not support client/ server error type separation as well in build i18n capability for all errors. Could apply to multiple other services as well.

Better to address this in a followup PR.

backend/internal/oauth/oauth2/revocation/service_test.go
```diff
@@ -240,6 +240,27 @@ func TestRevokeTokenFamily_EmptyIDIsNoOp(t *testing.T) {
 	store.AssertNotCalled(t, "insertCriterion", mock.Anything, mock.Anything)
 }
 
+func TestRevokeByCriteria_UsesCutoffAsRevokedAt(t *testing.T) {
```
ThaminduDilshan, line 243: Again not related to this PR. But we should have followed test suite pattern

backend/internal/oauth/oauth2/revocation/store_constants.go
```diff
@@ -27,13 +27,24 @@ var queryInsertRevocationCriterion = dbmodel.DBQuery{
 	Query: `INSERT INTO "REVOCATION_CRITERIA" (ID, CRITERION_TYPE, CRITERION_VALUE, REASON, ` +
 		`REVOKED_AT, EXPIRY_TIME, DEPLOYMENT_ID) VALUES ($1, $2, $3, $4, $5, $6, $7) ` +
 		`ON CONFLICT (DEPLOYMENT_ID, CRITERION_TYPE, CRITERION_VALUE) DO UPDATE SET ` +
-		`REASON = excluded.REASON, REVOKED_AT = excluded.REVOKED_AT, EXPIRY_TIME = excluded.EXPIRY_TIME`,
+		`REASON = CASE WHEN "REVOCATION_CRITERIA".REASON NOT IN (` +
```
ThaminduDilshan: Why do we need this special handling based on criteria reason? Having this list hard coded in query may not scale well. 

If this is needed maybe we can build a query utilizing a private array with criteria reason constants.
Same applies to queryIsCriterionRevoked

backend/internal/oauth/oauth2/revocation/store_constants_integration_test.go
```diff
@@ -0,0 +1,74 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package revocation
```
ThaminduDilshan: Shall we move this to the existing store test instead of adding a new test file?
Also we should follow test suite pattern

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -703,9 +703,33 @@ func (tv *tokenValidator) isAuthAssertion(
...
+func (tv *tokenValidator) ensureNotRevoked(ctx context.Context,
+	identity revocation.RevocationIdentity) error {
 	if tv.enforcementService != nil {
-		return tv.enforcementService.EnsureNotRevoked(ctx, jti, tokenFamilyID)
+		return tv.enforcementService.EnsureNotRevoked(ctx, identity)
 	}
 	return nil
 }
+
+func revocationIdentity(claims map[string]interface{}, jti, tokenFamilyID string) revocation.RevocationIdentity {
```
ThaminduDilshan, line 722: Shall we add a method comment?

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -703,9 +703,33 @@ func (tv *tokenValidator) isAuthAssertion(
...
+	criteria := make([]revocation.Criterion, 0, 3)
+	if tokenFamilyID != "" {
+		criteria = append(criteria, revocation.Criterion{Type: revocation.CriterionTypeTokenFamily, Value: tokenFamilyID})
+	}
+	subject, _ := extractStringClaim(claims, "sub")
```
ThaminduDilshan: Let's use constants wherever possible. We should have defined constants already for these

backend/internal/oauth/oauth2/tokenservice/validator_test.go
```diff
@@ -2052,7 +2071,7 @@ func (suite *TokenValidatorTestSuite) TestValidateAccessToken_EnforcementUnavail
 	suite.mockJWTService.On("VerifyJWT", mock.Anything, token, "", "https://example.com").Return(nil)
 
 	enforcement := revocationmock.NewEnforcementServiceInterfaceMock(suite.T())
-	enforcement.On("EnsureNotRevoked", mock.Anything, "at-jti-unknown", mock.Anything).
+	enforcement.On("EnsureNotRevoked", mock.Anything, mock.MatchedBy(func(identity revocation.RevocationIdentity) bool { return identity.JTI == "at-jti-unknown" })).
```
ThaminduDilshan: Line length should be exceeding 120 chars

backend/internal/system/revocationcache/model.go
```diff
@@ -14,11 +14,12 @@ type revokedEntry struct {
...
 
-// revokedSnapshot is one source read: the revoked single-token jtis and the revoked token-family ids
-// for a deployment, held in separate cache dimensions so a jti is never matched against a tfid.
+// revokedSnapshot is one source read. Dimensions are separate so values of different types cannot collide.
 type revokedSnapshot struct {
 	// Tokens holds the revoked single-token entries (keyed by jti).
 	Tokens []revokedEntry
 	// Families holds the revoked token-family entries (keyed by tfid).
 	Families []revokedEntry
+	// Subjects holds revoked ThunderID user-subject entries.
```
ThaminduDilshan:
```suggestion
	// Subjects holds revoked user-subject entries.
```

backend/internal/system/revocationcache/query_constants.go
```diff
@@ -19,3 +19,9 @@ var querySnapshotRevokedTokenFamilies = dbmodel.DBQuery{
 	Query: `SELECT CRITERION_VALUE, EXPIRY_TIME FROM "REVOCATION_CRITERIA" ` +
 		`WHERE CRITERION_TYPE = $1 AND EXPIRY_TIME > $2 AND DEPLOYMENT_ID = $3`,
 }
+
+var querySnapshotRevokedSubjects = dbmodel.DBQuery{
```
ThaminduDilshan, line 25: Is this query correct? There's no difference between this and `querySnapshotRevokedTokenFamilies`.

Also let's add a comment

backend/internal/user/service.go
```diff
@@ -48,12 +48,40 @@ type UserServiceInterface interface {
...
 		credentials json.RawMessage) *tidcommon.ServiceError
 	DeleteUser(ctx context.Context, userID string) *tidcommon.ServiceError
+	ValidateDeleteUser(ctx context.Context, userID string) *tidcommon.ServiceError
 	ResolveUserOUHandle(ctx context.Context, user *User) *tidcommon.ServiceError
 	SetDependencyRegistry(r resourcedependency.Registry)
 	GetUserUsages(ctx context.Context, userID string) (
 		*resourcedependency.DependenciesResponse, *tidcommon.ServiceError)
 }
 
+// ValidateDeleteUser checks whether the caller may delete the user and whether the target is deletable,
```
ThaminduDilshan: Can we add new methods to the appropriate place rather than adding to random places?

frontend/apps/console/src/features/flows/components/create-flow/SelectFlowType.tsx
```diff
@@ -58,6 +58,14 @@ export default function SelectFlowType({selectedType, onTypeChange, onReadyChang
       descriptionDefault: 'Confirm and terminate an established SSO session',
       icon: <LogOut size={28} />,
     },
+    {
+      type: FlowType.ADMINISTRATION,
+      labelKey: 'flows:create.type.administration.label',
+      labelDefault: 'Administrative Flow',
+      descriptionKey: 'flows:create.type.administration.description',
+      descriptionDefault: 'Orchestrate authenticated administrative and security operations',
```
ThaminduDilshan: Let's make sure label and description are consistent with existing ones

backend/internal/flow/flowexec/service.go
```diff
@@ -188,20 +213,33 @@ func applyInboundSSO(engineCtx *EngineContext, ctx context.Context) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		flow, svcErr := s.flowProvider.GetFlow(ctx, flowID)
+		if svcErr != nil {
+			return nil, svcErr
+		}
+		if flow.FlowType != providers.FlowTypeAdministration {
+			return nil, &ErrorInvalidFlowType
```
ThaminduDilshan: Maybe better to return a specific error to give an indication that this action is prohibited/ not allowed

backend/internal/flow/flowexec/service.go
```diff
@@ -188,20 +213,33 @@ func applyInboundSSO(engineCtx *EngineContext, ctx context.Context) {
...
-		return nil, err
+	var flowType providers.FlowType
+	if flowID == "" {
+		var err *tidcommon.ServiceError
+		flowType, err = validateFlowType(flowTypeStr)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		flow, svcErr := s.flowProvider.GetFlow(ctx, flowID)
```
ThaminduDilshan, line 324: With this approach, when flow id is provided, we're going to fetch the flow twice; once here and again in `initContext()` method.
Shall we try to optimize the logic?

backend/internal/flow/flowexec/handler.go
ThaminduDilshan: Looks like implementation already follows the suggestion

backend/cmd/server/servicemanager.go
indeewari: Fixed

backend/internal/flow/executor/constants.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/executor/revocation_plan.go
indeewari: Fixed

backend/internal/flow/executor/session_revocation_executor.go
indeewari: Fixed

backend/internal/flow/executor/revocation_workflow_executors_test.go
indeewari: Fixed

backend/internal/flow/executor/user_delete_executor.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/session/store_constants.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/model.go
indeewari: Fixed

backend/internal/flow/session/interface.go
indeewari: Fixed

backend/internal/oauth/oauth2/tokenservice/validator.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/store_constants_integration_test.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/store_constants.go
indeewari: Fixed

backend/internal/flow/flowexec/handler.go
indeewari: Fixed

backend/cmd/server/servicemanager.go
indeewari: Fixed

backend/internal/flow/flowexec/service.go
indeewari: Fixed

backend/internal/oauth/oauth2/tokenservice/validator.go
indeewari: Fixed

backend/cmd/server/servicemanager.go
indeewari: Fixed

backend/internal/system/revocationcache/query_constants.go
indeewari: Fixed

backend/internal/user/service.go
indeewari: Fixed

backend/internal/flow/flowexec/service.go
indeewari: Fixed

backend/internal/flow/executor/revocation_plan.go
indeewari: Fixed

backend/internal/flow/executor/user_delete_executor.go
indeewari: Fixed

frontend/apps/console/src/features/flows/components/create-flow/SelectFlowType.tsx
indeewari: Fixed

backend/internal/oauth/oauth2/tokenservice/validator_test.go
indeewari: Fixed

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/system/revocationcache/model.go
indeewari: Fixed

backend/internal/flow/executor/user_deletion_validation_executor.go
indeewari: Fixed

backend/cmd/server/servicemanager.go
indeewari: Will be added in a separate PR.

backend/internal/oauth/oauth2/revocation/service_test.go
indeewari: Fixed

api/flow-execution.yaml
```diff
@@ -27,14 +29,23 @@ paths:
     post:
       summary: Execute a flow step
       description: >-
-        Execute a step in an authentication flow. Backend/server-side applications must present
+        Execute a step in a flow. Backend/server-side applications must present
         their Flow Secret in the `Flow-Secret` header (see the `FlowSecret` security scheme) when
         initiating a new flow; the header is ignored for other application types.
+
+
```
ThaminduDilshan:
```suggestion
```

backend/cmd/server/servicemanager.go
```diff
@@ -325,15 +325,12 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
...
 
 	flowConfig := flowconfig.FromServerRuntime()
-	// The SSO session service revokes a session's token families on sign-out. The criteria revoker is
-	// built here (the OAuth engine's own revoker is created later, so it cannot be shared) and adapted
-	// to the session service's consumer interface with the sign-out reason fixed.
 	tokenFamilyRevocationTTL := time.Duration(runtime.Config.OAuth.RefreshToken.ValidityPeriod) * time.Second
-	sessionCriteriaRev := sessionCriteriaRevoker{
-		revoker: revocation.InitializeCriteriaRevoker(tokenFamilyRevocationTTL),
-	}
+	revocationEnforcer, revocationSvc := revocation.InitializeServices(jwtService, observabilitySvc,
```
```suggestion
	revocationEnforcer, revocationSvc := revocation.Initialize(jwtService, observabilitySvc,
```

This is a pattern we follow. Let's not introduce other methods

backend/internal/flow/executor/constants.go
```diff
@@ -109,6 +114,8 @@ const (
 	propertyKeyPromptOnSignOut = "promptOnSignOut"
 )
 
+const sharedRevocationPlan = "revocation.plan"
```
ThaminduDilshan: This is a runtime data key right? We already keep runtime data keys in a common location (flow/common/constants.go).
Can we move it there? Also follow the key name pattern we have used for other runtime data keys. No need to scope it as shared (maybe can add a comment)

[1] https://github.com/indeewari/thunder/blob/1ce7e5afe4c24633df82e39178b83d940513e930/backend/internal/flow/common/constants.go#L115

backend/internal/flow/executor/criteria_revocation_executor.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/revocation"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type criteriaRevocationExecutor struct {
```
ThaminduDilshan, line 15: Shall we add comment?

backend/internal/flow/executor/revocation_workflow_executors_test.go
```diff
@@ -0,0 +1,229 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package executor
```
ThaminduDilshan, line 4: [CAN ADDRESS AS FOLLOWUP] Go convention is to keep unit tests in a file name suffixed by `_test`. 
Ex: `criteria_revocation_executor.go` -> `criteria_revocation_executor_test.go`. 
Combined test is a responsibility of integration tests, not unit.

Shall we move it like this rather than having a single unit test file for all revocation executors?

backend/internal/flow/executor/session_revocation_executor.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/flow/session"
+	"github.com/thunder-id/thunderid/internal/revocation"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type sessionRevocationExecutor struct {
```
ThaminduDilshan, line 17: Shall we add comment?

backend/internal/flow/executor/user_delete_executor.go
```diff
@@ -0,0 +1,61 @@
...
+import (
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/flow/core"
+	"github.com/thunder-id/thunderid/internal/revocation"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+type userDeleteExecutor struct {
```
ThaminduDilshan, line 17: Shall we add comment?

backend/internal/flow/executor/utils.go
```diff
@@ -8,13 +8,45 @@ import (
...
+	"time"
 
 	authncm "github.com/thunder-id/thunderid/internal/authn/common"
 	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/revocation"
 	systemutils "github.com/thunder-id/thunderid/internal/system/utils"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
+type revocationPlan struct {
```
ThaminduDilshan, line 23: Shall we add comment?

backend/internal/flow/executor/utils.go
```diff
@@ -8,13 +8,45 @@ import (
...
+	Criteria []revocation.Criterion `json:"criteria"`
+	Mode     revocation.Mode        `json:"mode"`
+	Cutoff   time.Time              `json:"cutoff,omitempty"`
+	Reason   revocation.Reason      `json:"reason"`
+}
+
+func encodeRevocationPlan(plan revocationPlan) (string, error) {
```
ThaminduDilshan, line 31: Shall we add comment?

backend/internal/flow/executor/utils.go
```diff
@@ -8,13 +8,45 @@ import (
...
+	encoded, err := json.Marshal(plan)
+	if err != nil {
+		return "", fmt.Errorf("failed to encode revocation plan: %w", err)
+	}
+	return string(encoded), nil
+}
+
+func decodeRevocationPlan(data map[string]string) (revocationPlan, error) {
```
ThaminduDilshan, line 41: Shall we add comment?

backend/internal/flow/flowexec/error_constants.go
```diff
@@ -239,3 +239,47 @@ var ErrorAttestationNotConfigured = tidcommon.ServiceError{
...
+		DefaultValue: "Insufficient permissions",
+	},
+	ErrorDescription: tidcommon.I18nMessage{
+		Key:          "error.flowexecservice.administration_permission_required_description",
+		DefaultValue: "Administrative flows require the system permission",
+	},
+}
+
+// ErrorFlowIDExecutionNotPermitted is returned when direct execution by ID targets a non-administrative flow.
+var ErrorFlowIDExecutionNotPermitted = tidcommon.ServiceError{
```
ThaminduDilshan: Shall we move this before FES-1019?

backend/internal/flow/flowexec/service.go
```diff
@@ -188,20 +234,51 @@ func applyInboundSSO(engineCtx *EngineContext, ctx context.Context) {
...
+		if flow.FlowType != providers.FlowTypeAdministration {
+			return nil, &ErrorFlowIDExecutionNotPermitted
+		}
+		flowType = flow.FlowType
+		resolvedFlow = flow
+	}
+
+	// Gate administration flows before any further resolution so an unauthenticated caller cannot
+	// drive application lookups or context creation.
+	if svcErr := validateAdministrationCaller(ctx, flowType); svcErr != nil {
```
ThaminduDilshan, line 340: We can try to optimize this later, but for administrator flows, this check is done twice, once at line 253 and again here

backend/internal/flow/executor/administrative_flow_pre_executor.go
```diff
@@ -0,0 +1,107 @@
...
+
+// newAdministrativeFlowPreExecutor creates the preparatory executor for administrative flows.
+func newAdministrativeFlowPreExecutor(factory core.FlowFactoryInterface,
+	userProvider userDeletionProvider) *administrativeFlowPreExecutor {
+	base := factory.CreateExecutor(ExecutorNameAdministrativeFlowPre, providers.ExecutorTypeUtility,
+		[]providers.Input{
+			{Identifier: revocationInputSubject, Type: providers.InputTypeText, Required: true},
+		}, nil, &providers.ExecutorMeta{
+			SupportedFlowTypes: []providers.FlowType{providers.FlowTypeAdministration},
+			// The revocation breadth is part of the flow's design, not something the caller chooses per
```
ThaminduDilshan: We don't need this comment. This is a implementation time input

backend/internal/flow/executor/pre_delete_executor.go
```diff
@@ -0,0 +1,107 @@
...
+			SupportedModes: []string{string(revocation.ModeAll)},
+		})
+	return &administrativeFlowPreExecutor{Executor: base, userProvider: userProvider}
+}
+
+// resolveRevocationMode returns the revocation breadth configured on the node, falling back to the
+// executor's default when the node omits it. An unsupported value cannot normally reach here because
+// flow creation validates the mode, so this is the runtime backstop for a flow persisted before the
+// mode was constrained.
+func resolveRevocationMode(ctx *providers.NodeContext) (revocation.Mode, bool) {
```
ThaminduDilshan, line 57: [CAN DO AS A FOLLOWUP] We follow a switch-case based pattern in other executors with multiple modes. Let's modify this to follow the same convention. Not via a resolver method.
Also let's add a private method to execute particular mode

backend/internal/flow/executor/administrative_flow_pre_executor.go
```diff
@@ -0,0 +1,107 @@
...
+
+// administrativeFlowPreExecutor is the preparatory node of an administrative flow. It runs ahead of
+// the executors that act, validates the target of the operation, and publishes the trusted
+// revocation plan those executors consume.
+//
+// Nothing downstream re-derives the plan or reads it from the request: the plan travels on
+// SharedRuntimeData, which is unexported on the engine context, never serialized to the client, and
+// only writable by an executor. That is what makes the breadth recorded here binding on every node
+// that follows, rather than something the caller can influence per request.
+type administrativeFlowPreExecutor struct {
```
ThaminduDilshan: Prefer if we can rename this, constructor method and executor name to something else. This is just a pre delete executor right? Yes it's generic for multiple delete criterias, but this is not a pre executor for all administrative flows that we may introduce in future.

Also I don't like having flow part here and it doesn't tally with other existing executor names.

backend/internal/oauth/oauth2/revocation/init.go
```diff
@@ -19,40 +19,29 @@ import (
...
-// root that cannot receive the revocation service from Initialize (which is created inside the OAuth
-// engine after those consumers are constructed) — notably the SSO session service, which revokes a
-// session's families on sign-out. It returns a revocation service narrowed to CriteriaRevokerInterface;
-// only RevokeTokenFamily is reachable, which needs just the store and TTL, so the unused jwt and
-// observability dependencies are nil and the family write path shares no mutable state (it is a
-// stateless, idempotent writer). tokenFamilyRevocationTTL bounds each entry.
-func InitializeCriteriaRevoker(tokenFamilyRevocationTTL time.Duration) CriteriaRevokerInterface {
-	return newRevocationService(nil, newRevocationStore(), tokenFamilyRevocationTTL, false, nil)
+// RegisterRoutes registers the RFC 7009 revocation endpoint using the shared revocation service.
+func RegisterRoutes(
```
ThaminduDilshan, line 35: Why do we need this public method? Can't we create the revocation svc inside Initialize() method and pass to the private registerRoutes()?

This doen't adhere to the pattern we follow

backend/internal/oauth/oauth2/revocation/model.go
```diff
@@ -17,25 +21,77 @@ const (
 )
 
 // RevocationReason enumerates why a single token was revoked.
-type RevocationReason string
+type RevocationReason = sharedrevocation.Reason
 
 const (
```
ThaminduDilshan, line 26: Not clear why we need to duplicate everything in two places, but not a major issue

backend/internal/oauth/oauth2/revocation/service.go
ThaminduDilshan: @coderabbitai create a followup issue and assign @indeewari

backend/internal/oauth/init.go
```diff
@@ -63,15 +64,11 @@ func Initialize(
...
-	var enforcementService revocation.EnforcementServiceInterface
-	var revocationSvc revocation.RevocationServiceInterface
 	if cfg.OAuth.TokenRevocation.Enabled {
-		// The enforcement service (revocation read path) is built before the token service so it can be
-		// injected into the validator, which enforces the deny list as the final step of every validation.
-		tokenFamilyRevocationTTL := time.Duration(cfg.OAuth.RefreshToken.ValidityPeriod) * time.Second
-		enforcementService, revocationSvc = revocation.Initialize(
-			mux, jwtService, actorProvider, authnProvider, discoveryService, observabilitySvc,
-			tokenFamilyRevocationTTL, cfg.OAuth.Revocation.TokenFamily.OnExplicitRevoke)
+		revocation.RegisterRoutes(mux, jwtService, actorProvider, authnProvider, discoveryService, revocationSvc)
```
ThaminduDilshan, line 71: Not the correct pattern

backend/internal/revocation/model.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package revocation defines dependency-light contracts shared by revocation producers and consumers.
```
ThaminduDilshan: Let's only add package comment to a single file. Existing pattern is to keep this in service file or init file

backend/internal/revocation/model.go
```diff
@@ -0,0 +1,114 @@
...
+// CriteriaRevocation describes a criteria-based revocation write.
+type CriteriaRevocation struct {
+	Criterion Criterion
+	Mode      Mode
+	Cutoff    time.Time
+	Reason    Reason
+}
+
+// CriteriaRevoker persists criteria-based revocations.
+type CriteriaRevoker interface {
```
ThaminduDilshan: Interfaces shouldn't be in model.go file. Let's move it to related implementation file (if exists) or interface.go file

ThaminduDilshan: Why this is added now? Is it something required?

backend/internal/system/revocationcache/query_constants.go
```diff
@@ -19,3 +19,11 @@ var querySnapshotRevokedTokenFamilies = dbmodel.DBQuery{
 	Query: `SELECT CRITERION_VALUE, EXPIRY_TIME FROM "REVOCATION_CRITERIA" ` +
 		`WHERE CRITERION_TYPE = $1 AND EXPIRY_TIME > $2 AND DEPLOYMENT_ID = $3`,
 }
+
+// querySnapshotRevokedSubjects reads subject criteria together with the reason and action boundary
+// needed to distinguish permanent revocations from time-bounded revocations.
+var querySnapshotRevokedSubjects = dbmodel.DBQuery{
```
ThaminduDilshan, line 25: Can optimize this later, but I don't see a clear difference between querySnapshotRevokedTokenFamilies and querySnapshotRevokedSubjects except retrieving reason and revoked at fields

frontend/apps/console/src/features/flows/components/resource-property-panel/extended-properties/execution-properties/AdministrativeFlowPreProperties.tsx
```diff
@@ -0,0 +1,80 @@
...
+      },
+    };
+
+    onChange('data', updatedData, resource);
+  };
+
+  return (
+    <Stack gap={2}>
+      <Typography variant="body2" color="text.secondary">
+        {t('flows:core.executions.administrativeFlowPre.description')}
```
ThaminduDilshan: Let's add fallback translations for all the places

api/flow-execution.yaml
indeewari: Fixed. Removed the duplicated blank line from the folded description block.

backend/cmd/server/servicemanager.go
indeewari: Fixed. Renamed `InitializeServices` to `Initialize` across all call sites (servicemanager, engine, tests).

backend/internal/flow/executor/constants.go
indeewari: Fixed. Moved to `flow/common/constants.go` as `RuntimeKeyRevocationPlan` ("revocationPlan"), following the existing `RuntimeKey*` naming. Dropped the "shared" scoping from the name and explained the cross-frame carriage in a comment instead.

backend/internal/flow/executor/criteria_revocation_executor.go
indeewari: Fixed. Added a struct comment.

backend/internal/flow/executor/session_revocation_executor.go
indeewari: Fixed. Added a struct comment.

backend/internal/flow/executor/user_delete_executor.go
indeewari: Fixed. Added a struct comment.

backend/internal/flow/executor/utils.go
indeewari: Fixed. Added a comment on `revocationPlan`.

backend/internal/flow/executor/utils.go
indeewari: Fixed. Added a comment on `encodeRevocationPlan`.

backend/internal/flow/executor/utils.go
indeewari: Fixed. Added a comment on `decodeRevocationPlan`.

backend/internal/flow/flowexec/error_constants.go
indeewari: Fixed. `ErrorFlowIDExecutionNotPermitted` (FES-1018) now sits before FES-1019, so the codes read in order.

backend/internal/flow/executor/administrative_flow_pre_executor.go
indeewari: Fixed. Removed the comment.

backend/internal/flow/executor/administrative_flow_pre_executor.go
indeewari: Fixed. Renamed to `PreDeleteExecutor` throughout: type, constructor, `ExecutorNamePreDelete`, file (`pre_delete_executor.go`), the console resource entry and property panel, the template node id (`pre-delete`), and the i18n keys.

backend/internal/revocation/model.go
indeewari: Fixed. Package comment now lives in `interface.go` only; `model.go` just declares the package.

backend/internal/revocation/model.go
indeewari: Fixed. Moved `CriteriaRevoker` into a new `interface.go`. It is required: it is the seam the flow executors write revocation intent through, so they do not have to depend on the OAuth revocation implementation.

frontend/apps/console/src/features/flows/components/resource-property-panel/extended-properties/execution-properties/AdministrativeFlowPreProperties.tsx
indeewari: Fixed. All five `t()` calls now pass fallback text.

backend/internal/oauth/oauth2/revocation/init.go
indeewari: Agreed on the pattern, but collapsing these hits a circular dependency rather than an ordering preference.

`revocationSvc` has two consumers wired before the OAuth engine: `sessionCriteriaRevoker` for SSO sign-out (servicemanager.go:331) and `ExecutorDependencies.CriteriaRevoker` for `CriteriaRevocationExecutor` (:363). The second feeds the executor registry -> `flowExecService`, and `flowExecService` is an input parameter to `oauth.Initialize` (:471). So creating the service inside `oauth.Initialize` would mean something it consumes needs something it has not produced yet.

The session dependency alone could be reordered; the executor one cannot. Breaking it would mean handing the executors a holder populated after the fact, trading a compile-time guarantee for a runtime one on the revocation enforcement path.

Separately, `RegisterRoutes` needs `discoveryService`, which is created inside `oauth.Initialize` and consumed by six other subsystems there (par, ciba, token, introspect, userinfo).

Fully adopting the pattern means hoisting `discovery.Initialize` and the revocation wiring up to `servicemanager` so a single `Initialize` can call a private `registerRoutes`. Happy to do that, but it is an OAuth-wiring refactor touching all seven `discoveryService` call sites, so I would prefer it as a follow-up. Your servicemanager.go:329 suggestion is applied: it is `revocation.Initialize(...)` now.

Leaving this open for your call.

backend/internal/oauth/init.go
indeewari: Same root cause as the thread on `revocation/init.go` — answering there so the discussion stays in one place rather than duplicated.

Short version: the revocation service cannot be constructed inside `oauth.Initialize`, because `flowExecService` is an input parameter to it and the flow executors need the criteria revoker. Adopting the pattern fully means hoisting `discovery.Initialize` and the revocation wiring up to `servicemanager`, which I would prefer to do as a follow-up.

backend/internal/oauth/oauth2/revocation/model.go
indeewari: Not fixed, leaving open for your call. The aliases in the OAuth revocation package re-export the contracts from `internal/revocation` so existing OAuth call sites keep using `revocation.CriterionTypeSubject` etc. rather than importing a second package. The shared package exists because the flow executors and the Resource Server cache need those contracts without depending on the OAuth implementation. Happy to drop the aliases and import the shared package directly at each call site if you prefer, it just touches more files.

ThaminduDilshan: I think we can keep revocation package one only. I feel like it's okay to import revocation into oauth. Hope no circular dependency.
Since this is going to affect internal/ packages only and going to change multiple files, we can do this later

backend/internal/oauth/oauth2/revocation/init.go
ThaminduDilshan: >  So creating the service inside oauth.Initialize would mean something it consumes needs something it has not produced yet.

I think we already follow some setter pattern for this. Let's use a static setter method in init.go file to set the service from service manager.

ThaminduDilshan: Let's take this to a followup PR

ThaminduDilshan (dismissed):

ThaminduDilshan (approved):

PR #4631, opened 2026-08-06
Improve OIDC scope-claim mapping and console token UI

thiva-k:
### Purpose

https://github.com/user-attachments/assets/b83fa7b9-85a0-4ac4-baa8-1c3fc2c817e1

OIDC scopes were resolved from two places: the built-in `StandardOIDCScopes` table and the application's `scope_claims` mapping, with a separate `scopes` allow-list filtering the result. An application could not narrow or  override what `profile` or `email` released, and the Console showed a scope list and a claim mapping that could disagree.

This makes the application's `scope_claims` mapping the single source of truth. A scope the application maps is OIDC and releases exactly the claims it lists; a standard OIDC scope it does not map is dropped; anything else is a permission scope for the resource servers. `openid` is exempt and always granted, since it carries no claims of its own.

### Approach

- `SeparateOIDCAndNonOIDCScopes` classifies from the mapping alone. `FilterOIDCScopesByAllowedScopes`
  and the `StandardOIDCScopes` fallbacks in the authorize and token paths are removed, so claim
  release no longer depends on the built-in table at runtime.
- `ResolveEffectiveScopeClaims` is replaced by `DefaultScopeClaims`, applied once at client creation. New clients are seeded with the standard mapping, pruned to the attributes their allowed user types actually declare, and their ID token attribute list is derived from the pruned mapping when the caller sends none.
- Console: the separate scope selector is gone. `ScopeMapper` is now the one editor, shared by the ID Token and User Info tabs, and edits on either side keep the two attribute lists in sync.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/3573
- https://github.com/thunder-id/thunderid/issues/3710

## Summary by CodeRabbit

* **New Features**
  * Scope-to-claim mappings now control OIDC claims in tokens and UserInfo responses.
  * Default OIDC mappings are applied when no mappings are provided; unmapped custom scopes are omitted.
  * The Console supports adding, validating, removing, and synchronizing custom scope mappings.
  * Scope mappings synchronize ID Token and UserInfo attributes while preserving manual changes.
  * Single-audience settings now display directly without unnecessary tabs.

* **Bug Fixes**
  * Improved claim validation, attribute derivation, and JWT preview sizing and scrolling.


backend/internal/inboundclient/service.go
```diff
@@ -154,6 +158,10 @@ func (s *inboundClientService) CreateInboundClient(ctx context.Context, client *
 		ctx, client.SubjectAttribute, client.AllowedUserTypes); err != nil {
 		return err
 	}
+	// Must run before applyInboundDefaults: UserInfo inherits the ID token list there.
```
ThaminduDilshan, line 161: Hope this is handled for update path as well. @coderabbitai check

docs/content/guides/protocols/oauth-oidc/claims-and-scopes.mdx
```diff
@@ -45,11 +51,17 @@ Each application can define its own scope-to-claim bundles. Asking for the scope
...
 | Custom scope names | Allowed, `roles` here is an application-defined scope |
-| Scope not in `scopeClaims` | Returns no extra claims (but the scope still passes through to access tokens) |
+| Scope not in `scopeClaims` | Not granted, an OIDC scope the client does not map is dropped from the request |
 | Bypasses resource-server filtering | Yes, see "Filtering rules" below |
 
+**Create and update differ.** On create, omitting `scopeClaims` seeds the standard scopes, while sending one stores exactly what you sent, so `{"profile": [...]}` yields a client with only `profile`. On update the mapping is replaced by what you send, so a scope you remove stays removed, and omitting the field clears the mapping.
+
+**Declarative resource files are never seeded.** A client loaded from a file under the declarative resources directory is taken exactly as written, so a file that declares no `scopeClaims` grants no OIDC scopes beyond `openid`. List the scopes the client needs. Configuration applied through the API, Dynamic Client Registration, or the import endpoint follows the create rule above and is seeded.
```
ThaminduDilshan: Runtime code removals will affect declarative paths. Hope we're fine with it.
cc: @senthalan @rajithacharith

ThaminduDilshan (dismissed):

backend/internal/inboundclient/service.go
thiva-k: This is intentional

docs/content/guides/protocols/oauth-oidc/claims-and-scopes.mdx
ThaminduDilshan: @coderabbitai we decided to preserve the previous runtime behaviour so there won't be any behavioral change with this. Can you verify it's addressed correctly?

ThaminduDilshan (approved):

docs/content/guides/protocols/oauth-oidc/claims-and-scopes.mdx
ThaminduDilshan: @thiva-k can you address this documentation issue with a followup PR?

PR #4632, opened 2026-08-06
Add per request CSP nonces for script and style src

### LINKED ISSUE #4475: Make the Console and Gate applications compliant with the deny-first CSP baseline
### Current Limitation

The Console and Gate applications depend on runtime style injection from Oxygen UI, MUI, Emotion, and shared design components. Under the deny-first baseline these injected styles are blocked, so style-src cannot be enforced without a temporary unsafe-inline relaxation, which is exactly what the baseline is intended to avoid. Some resources, such as application logos, user avatars, and externally hosted stylesheets, also load from operator or tenant controlled origins that the baseline does not cover.

### Suggested Improvement

- Make both applications compliant so the policy can be enforced without relaxing style-src.

- Apply the per-request nonce to every runtime injection point: the Oxygen UI theme provider, the shared GoogleFontLoader and StylesheetInjector components, and the Console gate-preview iframe.
- Remove inline style attributes from sanitized rich-text output in the flow simulation preview.
- Define and implement a strategy for tenant-controlled external resources (a configurable allowlist)
- Remove the temporary style-src unsafe-inline relaxation once the applications are verified compliant.

### LINKED ISSUE #4477: Add CSP nonce support for dynamic script and style injection
### Current Limitation

SDK integrations that dynamically inject scripts or styles can force host applications toward unsafe-inline or unsafe-eval, undermining a strict CSP. 

### Suggested Improvement

Provide nonce support for dynamic script and style injection. Introduce and use nonce support for emotion css to the React SDK

- Add nonce support for dynamically injected scripts and styles, and remove unsafe-inline and unsafe-eval dependencies 
- Update the JavaScript SDK with nonce support. 
- Expose an optional `nonce` prop/config option on top-level SDK providers (e.g., `<ThunderProvider nonce={...}>`).

Osara-B:
### Purpose

Adds per-request CSP nonces for `script-src` and `style-src-elem`, closing the remaining `'unsafe-inline'` gaps in the deny-first CSP baseline for both the Console and Gate apps.

- The backend generates a cryptographically random nonce per request (`crypto/rand`, base64 URL-encoded), stores it on the request context, and injects it into both the CSP response header and a substituted `<meta property="csp-nonce">` tag in the served HTML.
- The frontend reads the nonce via a new `getCspNonce()` utility and tags every runtime-created `<style>` element (`StylesheetInjector`, `FontImporter`, `JwtPreview`'s token viewer, `GatePreview`'s iframe bootstrap) and the oxygen-ui/emotion cache accordingly.
- Console keeps `'unsafe-inline'` on `style-src-elem` because the embedded Monaco editor has no nonce support; Gate has no such dependency and stays nonce-only.

### Approach

`sourcesWithNonce` appends the nonce to `script-src`/`style-src-elem` sources, but skips a directive whenever `'unsafe-inline'` is already configured on it, since per the CSP spec a nonce-source present in a directive causes browsers to ignore `'unsafe-inline'` in that same directive. This lets Console's `style-src-elem` keep `'unsafe-inline'` (for Monaco) while Gate's same directive gets the nonce, without a special case per app, since the two apps' generated `csp.yaml` configs simply differ in whether `'unsafe-inline'` is present on that directive to begin with. Gate is treated as the app that must stay maximally strict, since it is the one most commonly embedded and exposed to end users.

### Related Issues
- Closes #4477
- Closes #4475

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Security Enhancements**
  * Added per-request CSP nonces for scripts and styles in Console and Gate.
  * Applied nonce support to previews, themes, fonts, and dynamically injected styles.
  * Improved handling of inline style elements and attributes.

* **Bug Fixes**
  * Static HTML responses now correctly substitute CSP nonces and use the proper content type.
  * React Query Devtools are excluded from production builds.

* **Tests**
  * Added coverage for nonce generation, propagation, substitution, and fallback behavior.

### DIFF
--- a/backend/cmd/server/config/resources/server_configs/csp.yaml
+++ b/backend/cmd/server/config/resources/server_configs/csp.yaml
@@ -1,59 +1,56 @@
 resource_type: server_config
 name: csp
 value:
-  # Enforcing. This policy applies only to the Console and Gate applications (/console/ and /gate/);
-  # all other responses (APIs, OAuth2 endpoints) receive no Content-Security-Policy header at all.
-  # Each path policy's directives replace the matching baseline directives for that path.
+  # Enforcing. Applies only to /console/ and /gate/; all other responses (APIs, OAuth2) get no CSP
+  # header. Each path policy's directives replace the matching baseline directive for that path.
   reportOnly: false
   paths:
     # Console application.
     - location: "/console/"
       directives:
-        # 'unsafe-inline' is TEMPORARY: it covers oxygen-ui/MUI/Emotion inline style injection until
-        # nonce-based styling is introduced. Remove 'unsafe-inline' once inline styles are nonce-tagged.
-        style-src:
+        # Governs <style> elements and stylesheet links.
+        style-src-elem:
           - "'self'"
-          - "'unsafe-inline'"
           - "https://fonts.googleapis.com"
-        # 'data:' covers inline library images; 'https:' is a TEMPORARY broad allowance for arbitrary
-        # tenant-supplied images. Tighten to a proxy or operator allowlist later.
+          - "'unsafe-inline'" # Needed for the embedded Monaco editor (Custom CSS, JWT claims, Import/Export viewers), which has no CSP nonce support (monaco-editor#271).
+        # Governs inline style="" attributes; these can't carry a nonce at all per spec.
+        style-src-attr:
+          - "'self'"
+          - "'unsafe-inline'" # Needed for React/MUI's style={{}} usage.
         img-src:
           - "'self'"
-          - "data:"
-          - "https:"
-        # 'data:' covers a bundled library font self-hosted as a base64 URI.
-        # fonts.gstatic.com serves the actual font files referenced by the fonts.googleapis.com stylesheet.
+          - "data:" # Inline library images.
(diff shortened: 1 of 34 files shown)


backend/cmd/server/config/resources/server_configs/csp.yaml
```diff
@@ -1,58 +1,71 @@
...
+        # Import/Export viewers), which has no CSP nonce support at all (open since 2016:
+        # microsoft/monaco-editor#271). A nonce and 'unsafe-inline' can't coexist on the same
+        # directive per spec (a nonce makes browsers ignore 'unsafe-inline'), so
+        # internal/system/csp/policy.go's sourcesWithNonce skips appending the nonce here, letting
+        # this relaxation actually take effect.
+        style-src-elem:
           - "'self'"
-          - "'unsafe-inline'"
           - "https://fonts.googleapis.com"
+          - "'unsafe-inline'"
```
brionmario: Lets add the reason line.

Osara-B: Fixed in [Add per request CSP nonces for script and style src](https://github.com/thunder-id/thunderid/pull/4632/commits/d7ce41e1f188d41a18786020441e585f9691549c)

brionmario (approved):

PR #4673, opened 2026-08-07
Add replay protection to private_key_jwt client assertions

thiva-k:
### Purpose

`private_key_jwt` client assertions had no replay protection: `jti` was never tracked,
unlike DPoP proofs and ID-JAG assertions which both use the shared `oauth2/jti` deny-list.
A captured, still-valid `client_assertion` could be replayed to mint tokens as that client
without its private key. Affects all endpoints behind `ClientAuthMiddleware`. (CWE-294.)

### Approach

Wire the existing `jti` deny-list into the `private_key_jwt` path, mirroring DPoP/ID-JAG:
after verification, record the assertion's `jti` (now required) and reject replays. The
entry is bounded by the assertion's `exp` plus clock-skew leeway. Threaded the shared
`jtiStore` and leeway through `ClientAuthMiddleware` into all five consuming endpoints.

## Summary by CodeRabbit

* **Security Enhancements**
  * Added replay protection for OAuth client assertions.
  * Assertions now require valid, one-time-use identifiers and expiration claims.
  * Reused, incomplete, or invalid assertions are rejected.
  * Consistent replay protection and configurable JWT clock tolerance are applied across OAuth authentication flows.

### DIFF
--- a/backend/internal/oauth/oauth2/clientauth/clientauth.go
+++ b/backend/internal/oauth/oauth2/clientauth/clientauth.go
@@ -12,15 +12,20 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/thunder-id/thunderid/internal/cert"
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/jti"
 	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	"github.com/thunder-id/thunderid/internal/system/jose/jwt"
 	"github.com/thunder-id/thunderid/internal/system/log"
 	"github.com/thunder-id/thunderid/internal/system/utils"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
+
+// jtiNamespace identifies private_key_jwt client assertions in the shared JTI replay store.
+const jtiNamespace = "client_assertion"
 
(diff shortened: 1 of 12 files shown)


backend/internal/oauth/oauth2/clientauth/clientauth.go
```diff
@@ -12,16 +12,21 @@ import (
...
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/jti"
 	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	"github.com/thunder-id/thunderid/internal/system/jose/jwt"
 	"github.com/thunder-id/thunderid/internal/system/log"
 	"github.com/thunder-id/thunderid/internal/system/utils"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
+// jtiNamespace identifies private_key_jwt client assertions in the shared JTI replay store.
+const jtiNamespace = "client_assertion"
```
ThaminduDilshan, line 28: Let's move this to backend/pkg/thunderidengine/providers/constants.go

ThaminduDilshan (commented):
LGTM

backend/internal/oauth/oauth2/clientauth/clientauth.go
thiva-k: This is the domain specific namespace. The generic jti store namespace form provider is already used

ThaminduDilshan (approved):

PR #4676, opened 2026-08-07
Restrict OpenID4VCI credential issuance to wallet applications

thiva-k:
### Purpose

The OpenID4VCI credential endpoint accepted any JWT signed by the server, including
authentication assertions that carry no `client_id` and no scope. Any authenticated user
could mint a Verifiable Credential of any configured type.

`POST /openid4vci/credential` now requires an OAuth 2.0 access token (`typ: at+jwt`) carrying
a `client_id`, and that client must resolve to an application created from the **Digital
Wallet** template (`wallet` or `wallet-embedded`).

### Approach

`IssueCredential` validates the `typ` header and `client_id` claim up front, then
`verifyWalletClient` resolves the application behind `client_id` through the application
service and checks its template.

`openid4vci.Initialize` now takes `application.ApplicationServiceInterface`, so it moved in
`servicemanager.go` to after that service is constructed. It cannot be hoisted the other way:
`application` → `inboundClient` → `flowMgt` → `authnProvider` → `openid4vpSvc` cycles back
into `initializeVCServices`. OpenID4VP and the credential-configuration service stay put.

The Wayfinder sample and both bundled wallet clients already carry `template: wallet`, so the
VC walkthrough is unaffected. The integration test app is now registered as a wallet.

## Summary by CodeRabbit

* **Security Enhancements**
  * Credential issuance now requires a valid OAuth 2.0 access token and client ID.
  * Issuance is restricted to registered wallet applications.
  * Unknown, unresolved, or non-wallet clients are rejected.
  * Mobile and embedded wallet applications are supported.
  * Access tokens are checked for the required permissions before issuance.
  * Startup now fails safely when required authentication services cannot initialize.

### DIFF
--- a/backend/internal/openid4vci/init.go
+++ b/backend/internal/openid4vci/init.go
@@ -13,8 +13,8 @@ import (
 	"time"
 
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/dpop"
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/tokenservice"
 	"github.com/thunder-id/thunderid/internal/system/config"
-	"github.com/thunder-id/thunderid/internal/system/jose/jwt"
 	"github.com/thunder-id/thunderid/internal/system/middleware"
 	"github.com/thunder-id/thunderid/internal/user"
 	"github.com/thunder-id/thunderid/internal/vc/credential"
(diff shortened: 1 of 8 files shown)


backend/internal/openid4vci/init.go
```diff
@@ -23,11 +24,14 @@ import (
...
 // credSvc is the already-initialized credential-configuration service (owned by the caller).
+// tokenValidator validates the OAuth 2.0 access token presented at the credential endpoint.
+// appSvc resolves the wallet application behind an issuance request's access token.
 // When no signing key is configured, issuance is disabled and Initialize returns nil without error.
 func Initialize(
 	mux *http.ServeMux, cryptoProvider providers.RuntimeCryptoProvider,
-	jwtService jwt.JWTServiceInterface, userService user.UserServiceInterface,
+	tokenValidator tokenservice.TokenValidatorInterface, userService user.UserServiceInterface,
 	dpopVerifier dpop.VerifierInterface, credSvc credential.CredentialConfigurationServiceInterface,
+	appSvc application.ApplicationServiceInterface,
```
senthalan: Shall we use the actor provider hear?

thiva-k: Updated

senthalan (approved):

PR #4683, opened 2026-08-07
Add SSO support to the Wayfinder sample authentication flow

### LINKED ISSUE #4237: Add SSO support for the Wayfinder Sample to avoid double login (app + chat)
### Current Limitation

In the Wayfinder sample, the user is forced to authenticate twice: once to access the application and again to access the chat. This produces a poor login experience where the user appears to "re-login" even though they already have an active session.

This happens because of how access tokens are issued today. When requesting a token, the sample passes a resource server (audience) in the token request, so ThunderID issues an access token that is bound to a single resource server. As a result:

One access token is valid for exactly one resource server.
The app and the chat are treated as separate resource servers, so the token obtained for the app cannot be reused for the chat.
To call the chat, the sample initiates a fresh token request, which triggers a full login flow again.

### Suggested Improvement

Add SSO support to the Wayfinder sample so that the second (and subsequent) token acquisitions reuse the existing authenticated session instead of prompting the user to log in again.

The security model of resource-server-bound tokens stays intact (each resource server still gets its own scoped token), but token acquisition for additional resource servers becomes silent:

- When the app has already established a session with ThunderID, requesting an access token for the chat resource server should leverage that session and skip the interactive login prompt.
- The user logs in once; tokens for the app and the chat are then obtained transparently behind the scenes, each still audience-restricted to its own resource server.

Dilusha-Madushan:
### Purpose

The Wayfinder sample's authentication flow (`wayfinder-app-auth-flow`) had no SSO
handling, so every authorization request landed on the credentials prompt even when
the user already had a valid session. This PR adds SSO support to that flow, bringing
the sample in line with the default Console app authentication flow.

Scope is limited to `wayfinder-app-auth-flow` in the two Wayfinder sample
configurations (`redirect` and `app-native`). The registration, recovery, CIBA email,
CIBA SMS, and agent authentication flows are unchanged.

### Approach

Two nodes were added to the flow:

* `sso_check` (`SSOCheckExecutor`), inserted between `start` and the credentials
  prompt. It declares `checkpointRef: session`, so on a session hit the flow jumps
  straight to the session join node, and on a miss `onFailure` falls through to
  `prompt_credentials`.
* `session` (`SessionExecutor`), inserted as the join node before
  `authorization_check`. `credentials_auth.onSuccess` now targets it instead of
  `authorization_check`, so both the SSO path and the interactive login path
  converge on the same node before authorization.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4237

## Summary by CodeRabbit

* **New Features**
  * Added an existing-session check before requesting authentication credentials.
  * Successful session checks now proceed directly through authentication.
  * Credential-based authentication includes an intermediate session step before authorization.

* **Bug Fixes**
  * Users with incomplete or failed authentication are returned to the credential prompt.

### DIFF
--- a/samples/apps/wayfinder-sample/thunderid-config/app-native/thunderid-config.yaml
+++ b/samples/apps/wayfinder-sample/thunderid-config/app-native/thunderid-config.yaml
@@ -995,7 +995,15 @@ flowType: AUTHENTICATION
 nodes:
   - id: start
     type: START
-    onSuccess: prompt_credentials
+    onSuccess: sso_check
+  - id: sso_check
+    type: TASK_EXECUTION
+    executor:
+      name: SSOCheckExecutor
+    properties:
+      checkpointRef: session
+    onSuccess: session
+    onFailure: prompt_credentials
   - id: prompt_credentials
     type: PROMPT
     meta:
@@ -1062,8 +1070,13 @@ nodes:
     type: TASK_EXECUTION
     executor:
       name: CredentialsAuthExecutor
-    onSuccess: authorization_check
+    onSuccess: session
     onIncomplete: prompt_credentials
+  - id: session
+    type: TASK_EXECUTION
+    executor:
+      name: SessionExecutor
+    onSuccess: authorization_check
   - id: authorization_check
     type: TASK_EXECUTION
     executor:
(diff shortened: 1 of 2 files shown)


ThaminduDilshan (approved):

ThaminduDilshan:
Use case content will be added with https://github.com/thunder-id/thunderid/pull/5315

PR #4769, opened 2026-08-10
feat(scim): Default SCIM user creation to sole configured user type, advertise pagination support

ravindu439:
### Purpose
POST /Users required the ThunderID extension schema URN even when the request only carried core SCIM attributes, forcing clients to know a custom type URN just to create a user. This PR lets CreateUser omit the extension URN and fall back to the sole configured user type. It also advertises RFC 9865 pagination capabilities on GET /ServiceProviderConfig, which was previously missing from the response.

### Approach
- `scim_validator.go`: relaxed validation so zero ThunderID extension URNs is allowed, but only when the payload carries core attributes (empty payloads still rejected via `ErrorMissingCustomSchema`). More than one extension URN remains rejected as before.
- `service.go`: added `resolveDefaultEntityTypeName`, which looks up configured user entity types and returns the sole one's canonical name. Errors with `ErrorMissingCustomSchema` if zero or more than one type is configured (default would be ambiguous).
- `users_service.go`: `CreateUser` now calls `resolveDefaultEntityTypeName` when `UserTypeName` is empty, instead of failing. `ReplaceUser` (and `/Me` PUT) keep the extension URN mandatory, no fallback on update, since an update should be explicit about which type it targets.
- `config.go` / `model.go`: added `PaginationCursorSupported` (false), `PaginationIndexSupported` (true), default method/page size constants, and the `SCIMPaginationConfig` response struct, wired into `GetServiceProviderConfig`.
- Added unit tests for core-only create with single/multiple/zero configured types, and for replace still rejecting a missing extension URN.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
- [x] Tests provided.
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/internal/scim/config/config.go
+++ b/backend/internal/scim/config/config.go
@@ -21,6 +21,7 @@ package scimconfig
 
 import (
 	"github.com/thunder-id/thunderid/internal/system/config"
+	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	engineconfig "github.com/thunder-id/thunderid/pkg/thunderidengine/config"
 )
 
@@ -51,7 +52,7 @@ const (
 
 	// FilterMaxResults caps the number of resources returned in a single
 	// filtered query, guarding against excessively large result sets.
-	FilterMaxResults = 200
+	FilterMaxResults = serverconst.MaxPageSize
 
 	// ChangePasswordSupported indicates that the SCIM change-password
 	// operation is not yet supported.
@@ -64,6 +65,27 @@ const (
 	// ETagSupported indicates that ETag / versioning is supported
 	// per RFC 7644 §3.14.
 	ETagSupported = true
+
+	// PaginationCursorSupported indicates whether cursor-based pagination
+	// (RFC 9865) is supported. Not implemented; this server only supports
+	// index-based pagination.
+	PaginationCursorSupported = false
+
+	// PaginationIndexSupported indicates whether index-based pagination
+	// (startIndex/count, RFC 7644 §3.4.2.4) is supported.
+	PaginationIndexSupported = true
+
+	// PaginationDefaultMethod is the pagination method used when a client
+	// does not specify a preference. Must be "cursor" or "index" per RFC 9865.
+	PaginationDefaultMethod = "index"
+
+	// PaginationDefaultPageSize is the default number of resources returned
+	// per page when a client does not specify "count".
+	PaginationDefaultPageSize = serverconst.DefaultPageSize
+
+	// PaginationMaxPageSize is the maximum number of resources returned
+	// per page, regardless of the requested "count".
+	PaginationMaxPageSize = serverconst.MaxPageSize
 )
 
 // SCIMConfig holds the SCIM service configuration resolved from the
--- a/backend/internal/scim/scim_validator.go
+++ b/backend/internal/scim/scim_validator.go
@@ -21,13 +21,46 @@ type SCIMUserPayload struct {
 
 // ValidateSCIMUserRequest parses and validates a SCIM user payload.
 func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceError) {
-	// Step 1: Parse the request body as JSON.
 	var raw map[string]json.RawMessage
 	if err := json.Unmarshal(body, &raw); err != nil {
 		return nil, &ErrorInvalidRequestBody
 	}
 
-	// Step 2: Validate the presence of the "schemas" array and extract URNs.
+	schemas, svcErr := parseSCIMSchemas(raw)
+	if svcErr != nil {
+		return nil, svcErr
+	}
+
+	extensionURN, userTypeName, svcErr := resolveThunderExtensionURN(raw, schemas)
+	if svcErr != nil {
+		return nil, svcErr
+	}
+
+	extAttrs, svcErr := extractExtensionAttrs(raw, extensionURN)
+	if svcErr != nil {
+		return nil, svcErr
+	}
+
+	coreAttrs := extractCoreAttrs(raw, extensionURN)
+	if extensionURN == "" && len(coreAttrs) == 0 {
+		return nil, &ErrorMissingCustomSchema
+	}
+
+	if svcErr := validateCoreUserSchemaDeclared(schemas, coreAttrs); svcErr != nil {
+		return nil, svcErr
+	}
+
+	return &SCIMUserPayload{
+		ExtensionURN:   extensionURN,
+		UserTypeName:   userTypeName,
+		CoreAttrs:      coreAttrs,
+		ExtensionAttrs: extAttrs,
+	}, nil
+}
+
+// parseSCIMSchemas validates the presence of the "schemas" array, unmarshals it,
+// and rejects duplicate URNs within it.
+func parseSCIMSchemas(raw map[string]json.RawMessage) ([]string, *tidcommon.ServiceError) {
 	schemasRaw, ok := raw["schemas"]
 	if !ok {
 		return nil, &ErrorMissingSchemas
@@ -37,7 +70,6 @@ func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceE
 		return nil, &ErrorMissingSchemas
 	}
 
-	// Step 3: Check for duplicate URNs in the "schemas" array.
 	seen := make(map[string]struct{}, len(schemas))
 	for _, urn := range schemas {
 		lower := strings.ToLower(strings.TrimSpace(urn))
@@ -46,8 +78,18 @@ func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceE
 		}
 		seen[lower] = struct{}{}
 	}
+	return schemas, nil
+}
 
-	// Step 4: exactly one ThunderID extension URN .
+// resolveThunderExtensionURN finds the single ThunderID extension URN declared in
+// "schemas", if any. At most one is allowed; zero is allowed only when the payload
+// carries SCIM core attributes only (no custom type declared), in which case the
+// service layer falls back to the sole configured user type. When zero URNs are
+// declared, any body key that is itself a ThunderID extension URN is rejected,
+// since SCIM requires extension objects to be declared in "schemas".
+func resolveThunderExtensionURN(
+	raw map[string]json.RawMessage, schemas []string,
+) (extensionURN, userTypeName string, svcErr *tidcommon.ServiceError) {
 	thunderPrefix := strings.ToLower(ThunderIDURNPrefix)
 	var thunderURNs []string
 	for _, urn := range schemas {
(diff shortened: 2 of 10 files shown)


backend/internal/scim/config/config.go
```diff
@@ -64,6 +65,27 @@ const (
...
+	// does not specify a preference. Must be "cursor" or "index" per RFC 9865.
+	PaginationDefaultMethod = "index"
+
+	// PaginationDefaultPageSize is the default number of resources returned
+	// per page when a client does not specify "count".
+	PaginationDefaultPageSize = serverconst.DefaultPageSize
+
+	// PaginationMaxPageSize is the maximum number of resources returned
+	// per page, regardless of the requested "count".
+	PaginationMaxPageSize = serverconst.MaxPageSize
```
sadilchamishka, line 88: `PaginationMaxPageSize` is `serverconst.MaxPageSize` (100), but the list handlers don't cap at 100 — they clamp `count` to `scimconfig.FilterMaxResults` (200):

- `users_handler.go:79` — `if count > scimconfig.FilterMaxResults { count = scimconfig.FilterMaxResults }`
- `users_handler.go:165` (`/.search`) and `group_handler.go:47` do the same.

So `GET /Users?count=150` returns 150 resources while `GET /ServiceProviderConfig` advertises `pagination.maxPageSize: 100`. The same response also advertises `filter.maxResults: 200`, so the document contradicts itself.

Either advertise the value that is actually enforced, or change the handlers to clamp at `MaxPageSize`.

```suggestion
	PaginationMaxPageSize = FilterMaxResults
```

backend/internal/scim/users_service.go
```diff
@@ -182,6 +192,11 @@ func (s *scimUsersService) ReplaceUser(
 ) (*SCIMUser, *tidcommon.ServiceError) {
 	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, loggerComponentName))
 
+	if payload.UserTypeName == "" {
```
sadilchamishka: Worth reconsidering this asymmetry. The target user's type is not actually ambiguous on update — `ReplaceUser` loads `existingUser` a few lines below and then *enforces* that the requested type equals `existingUser.Type` (`ErrorImmutableUserType`, line ~206). So on PUT the extension URN can only ever be the type the user already has; requiring it is ceremony rather than disambiguation.

Concrete cost: a client that created a user via the new core-only POST cannot do the natural GET → modify → PUT round trip with a core-only body, and `PUT /Me` is worse — a `/Me` client generally has no way to know its own type URN, yet it now must send one.

Suggest resolving the type from `existingUser.Type` when `payload.UserTypeName == ""` (keeping the mismatch check for the case where a URN *is* supplied), which keeps POST and PUT consistent.

backend/internal/scim/users_service.go
```diff
@@ -182,6 +192,11 @@ func (s *scimUsersService) ReplaceUser(
...
+		logger.Error(ctx, "SCIM ReplaceUser: custom schema URN required")
```
sadilchamishka: Logging a plain client-input error at `Error` level is inconsistent with the other 400 paths in this same function, which use `logger.Debug` (e.g. the conflicting core/custom value branch). Any client sending a core-only PUT will now generate error-level entries, which degrades error-rate alerting.

backend/internal/scim/service.go
```diff
@@ -463,3 +481,24 @@ func resolveEntityTypeNameForSchemaURN(
...
+	page, svcErr := entityTypeService.GetEntityTypeList(
+		ctx, entitytype.TypeCategoryUser, serverconst.MaxPageSize, 0, false)
+	if svcErr != nil {
+		if svcErr.Type == tidcommon.ServerErrorType {
+			return "", &ErrorInternalServer
+		}
+		return "", &ErrorMissingCustomSchema
+	}
+	if page.TotalResults != 1 || len(page.Types) != 1 {
+		return "", &ErrorMissingCustomSchema
```
sadilchamishka, line 501: The zero-types and multiple-types cases are collapsed into one client error, but they are quite different:

- **More than one type configured** — 400 is right; the client genuinely has to disambiguate.
- **Zero types configured** — this is a server configuration state. Returning 400 with "The request must include exactly one ThunderID custom user schema URN" tells the client to add a URN that does not exist, so a well-behaved client will retry indefinitely instead of surfacing a deployment problem. A server-side error (or at minimum a distinct, accurate message) would be more actionable.

Also worth noting these two branches are indistinguishable in the logs, since `resolveDefaultEntityTypeName` logs nothing.

backend/internal/scim/service.go
```diff
@@ -463,3 +481,24 @@ func resolveEntityTypeNameForSchemaURN(
...
+func resolveDefaultEntityTypeName(
+	ctx context.Context, entityTypeService entitytype.EntityTypeServiceInterface,
+) (string, *tidcommon.ServiceError) {
+	page, svcErr := entityTypeService.GetEntityTypeList(
+		ctx, entitytype.TypeCategoryUser, serverconst.MaxPageSize, 0, false)
+	if svcErr != nil {
+		if svcErr.Type == tidcommon.ServerErrorType {
+			return "", &ErrorInternalServer
+		}
+		return "", &ErrorMissingCustomSchema
```
sadilchamishka, line 498: Every non-server error from `GetEntityTypeList` is flattened into `ErrorMissingCustomSchema` and the original is dropped without logging, so an authorization or validation failure from the entity type service surfaces to the caller as "Missing custom schema" and leaves no trace to diagnose. Consider logging `svcErr` here before mapping.

backend/internal/scim/scim_validator.go
```diff
@@ -91,11 +94,14 @@ func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceE
...
 			continue
 		}
-		if strings.EqualFold(k, extensionURN) {
+		if extensionURN != "" && strings.EqualFold(k, extensionURN) {
 			continue
 		}
 		coreAttrs[k] = v
 	}
+	if extensionURN == "" && len(coreAttrs) == 0 {
+		return nil, &ErrorMissingCustomSchema
```
sadilchamishka: `ErrorMissingCustomSchema`'s description in `error_constants.go` still reads *"The request must include exactly one ThunderID custom user schema URN"*, which this change makes inaccurate — zero URNs is now valid. The same error code is also reused for two further cases (empty payload here, and "multiple user types configured so the default is ambiguous" in `resolveDefaultEntityTypeName`), so clients cannot tell them apart and the message is wrong for all three.

Worth updating the description and/or splitting the ambiguous-default case into its own code.

backend/internal/scim/service_test.go
```diff
@@ -86,6 +86,12 @@ func TestGetServiceProviderConfig_CapabilitiesMatchConstants(t *testing.T) {
 	require.Equal(t, scimconfig.ChangePasswordSupported, result.ChangePassword.Supported)
 	require.Equal(t, scimconfig.SortSupported, result.Sort.Supported)
 	require.Equal(t, scimconfig.ETagSupported, result.ETag.Supported)
+	require.Equal(t, scimconfig.ETagSupported, result.ETag.Supported)
+	require.Equal(t, scimconfig.PaginationCursorSupported, result.Pagination.Cursor)
+	require.Equal(t, scimconfig.PaginationIndexSupported, result.Pagination.Index)
+	require.Equal(t, scimconfig.PaginationDefaultMethod, result.Pagination.DefaultPaginationMethod)
+	require.Equal(t, scimconfig.PaginationDefaultPageSize, result.Pagination.DefaultPageSize)
+	require.Equal(t, scimconfig.PaginationMaxPageSize, result.Pagination.MaxPageSize)
```
sadilchamishka, line 94: Duplicate of the assertion on the previous line (`result.ETag.Supported` is checked twice) — looks like a stray copy-paste; this line can just be deleted.

backend/internal/scim/scim_validator.go
```diff
@@ -47,34 +47,37 @@ func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceE
...
 
-	// Step 4: exactly one ThunderID extension URN .
+	// Step 4: at most one ThunderID extension URN. Zero is allowed only when
+	// the payload carries SCIM core attributes only (no custom type declared);
+	// the service layer then falls back to the sole configured user type.
 	thunderPrefix := strings.ToLower(ThunderIDURNPrefix)
 	var thunderURNs []string
 	for _, urn := range schemas {
 		if strings.HasPrefix(strings.ToLower(strings.TrimSpace(urn)), thunderPrefix) {
 			thunderURNs = append(thunderURNs, urn)
```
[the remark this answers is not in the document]
ravindu439: fixed ,added a error for this senario.

backend/internal/scim/service_test.go
```diff
@@ -86,6 +86,12 @@ func TestGetServiceProviderConfig_CapabilitiesMatchConstants(t *testing.T) {
 	require.Equal(t, scimconfig.ChangePasswordSupported, result.ChangePassword.Supported)
 	require.Equal(t, scimconfig.SortSupported, result.Sort.Supported)
 	require.Equal(t, scimconfig.ETagSupported, result.ETag.Supported)
+	require.Equal(t, scimconfig.ETagSupported, result.ETag.Supported)
+	require.Equal(t, scimconfig.PaginationCursorSupported, result.Pagination.Cursor)
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/config/config.go
ravindu439: fixed

backend/internal/scim/users_service.go
ravindu439: fixed

backend/internal/scim/users_service.go
ravindu439: fixed

backend/internal/scim/service.go
ravindu439: skiped for now.

backend/internal/scim/service.go
ravindu439: resolved

backend/internal/scim/scim_validator.go
ravindu439: skipped

backend/internal/scim/service_test.go
ravindu439: fixed

backend/internal/scim/scim_validator.go
```diff
@@ -21,13 +21,46 @@ type SCIMUserPayload struct {
 
 // ValidateSCIMUserRequest parses and validates a SCIM user payload.
 func ValidateSCIMUserRequest(body []byte) (*SCIMUserPayload, *tidcommon.ServiceError) {
```
PasinduYeshan, line 23: Let's change the method name to mean the extraction as well.

backend/internal/scim/service.go
```diff
@@ -463,3 +481,24 @@ func resolveEntityTypeNameForSchemaURN(
 		}
 	}
 }
+
+// resolveDefaultEntityTypeName returns the sole configured user entity type's
+// canonical name, for SCIM payloads that carry only core attributes and omit
+// the ThunderID extension URN. Errors if zero or more than one user type is
+// configured, since the default type is then ambiguous.
+func resolveDefaultEntityTypeName(
```
PasinduYeshan, line 489: EntityType -> UserType

backend/internal/scim/users_service.go
```diff
@@ -105,11 +105,21 @@ func (s *scimUsersService) CreateUser(
...
 	runtimeCtx := security.WithRuntimeContext(ctx)
-	canonicalName, svcErr := resolveEntityTypeNameForSchemaURN(runtimeCtx, s.entityTypeService, payload.UserTypeName)
-	if svcErr != nil || canonicalName == "" {
-		logger.Error(ctx, "SCIM CreateUser: entity type not found",
-			log.String("userTypeName", payload.UserTypeName), log.Any("error", svcErr))
-		return nil, &ErrorUnknownUserType
+	var canonicalName string
+	var svcErr *tidcommon.ServiceError
+	if payload.UserTypeName == "" {
+		canonicalName, svcErr = resolveDefaultEntityTypeName(runtimeCtx, s.entityTypeService)
```
PasinduYeshan, line 111: canonicalName -> defaultUserType

PasinduYeshan:
Let's address the comments in future PR.

PasinduYeshan (approved):

PR #4771, opened 2026-08-10
feat(scim): Add integration tests for SCIM Users, Groups, filters, search, Me, and Discovery endpoints

ravindu439:
### Purpose
The SCIM implementation (Users, Groups, filtering, search, /Me, discovery) had no integration-level coverage exercising it through real HTTP requests against a running server. This PR adds an integration test suite for the SCIM API surface.

### Approach
Added `tests/integration/scim/` with one suite per concern, each built on `testify/suite`:
- `discovery_test.go` (`TestSCIMDiscoveryTestSuite`): GET /ServiceProviderConfig, /ResourceTypes, /Schemas.
- `users_test.go` (`TestSCIMUsersTestSuite`): Users CRUD, attribute projection, ETag/If-Match handling.
- `groups_test.go` (`TestSCIMGroupsTestSuite`): Groups CRUD and PATCH per RFC 7644 §3.3.
- `scim_filter_test.go` (`TestSCIMFilterTestSuite`): filter grammar coverage (equality operators, "and" conjunction, attribute mapping edge cases).
- `search_test.go` (`TestSCIMSearchTestSuite`): POST /.search.
- `me_test.go` (`TestSCIMMeTestSuite`): /Me GET and PUT (self-service).
- `scim_authz_test.go` (`TestSCIMAuthzTestSuite`): authentication/authorization and OU-scoping across endpoints.
- `helpers.go` / `model.go`: shared request helpers (Bearer token injection, SCIM request wrapper) and response/request models used across suites.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
- [x] Tests provided.
    - [ ] Unit Tests
    - [x] Integration Tests

### Security checks

### DIFF
--- a/tests/integration/scim/discovery_test.go
+++ b/tests/integration/scim/discovery_test.go
@@ -1,0 +1,263 @@
+/*
+ * Copyright (c) 2026, WSO2 LLC. (https://www.wso2.com).
+ *
+ * WSO2 LLC. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package scim
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"github.com/thunder-id/thunderid/tests/integration/testutils"
+)
+
+// SCIMDiscoveryTestSuite covers the endpoints a real SCIM client (an IdP
+// provisioning connector) calls before ever creating a resource: capability
+// discovery, resource types, and schema introspection. It also pins down the
+// two known gaps in this server's SCIM support (no Users PATCH, no ETag-less
+// ServiceProviderConfig drift) so a regression here is caught immediately.
+type SCIMDiscoveryTestSuite struct {
+	suite.Suite
+	ouID           string
+	entityTypeID   string
+	entityTypeName string
+}
+
+func TestSCIMDiscoveryTestSuite(t *testing.T) {
+	suite.Run(t, new(SCIMDiscoveryTestSuite))
+}
+
+// scimPaginationMaxPageSize mirrors scimconfig.PaginationMaxPageSize
+// (backend/internal/scim/config/config.go). Update here whenever that constant
+// changes so the advertisement-to-enforcement assertion remains accurate.
+const scimPaginationMaxPageSize = 100
+
+func (ts *SCIMDiscoveryTestSuite) SetupSuite() {
+	ouID, err := testutils.CreateOrganizationUnit(testutils.OrganizationUnit{
+		Handle:      "scim-it-discovery-ou",
+		Name:        "SCIM Discovery Integration Test OU",
+		Description: "Organization unit for SCIM discovery endpoint tests",
+	})
+	ts.Require().NoError(err, "failed to create test organization unit")
+	ts.ouID = ouID
+
+	ts.entityTypeName = "scim-it-discovery-person"
+	entityTypeID, err := testutils.CreateUserType(testutils.UserType{
+		Name: ts.entityTypeName,
+		OUID: ouID,
+		Schema: map[string]interface{}{
+			"email":      map[string]interface{}{"type": "string", "required": true},
+			"department": map[string]interface{}{"type": "string"},
+		},
+	})
+	ts.Require().NoError(err, "failed to create test entity type")
+	ts.entityTypeID = entityTypeID
+}
+
+func (ts *SCIMDiscoveryTestSuite) TearDownSuite() {
+	if ts.entityTypeID != "" {
+		_ = testutils.DeleteUserType(ts.entityTypeID)
+	}
+	if ts.ouID != "" {
+		_ = testutils.DeleteOrganizationUnit(ts.ouID)
+	}
+}
+
+// TestServiceProviderConfig pins the capability flags a provisioning
+// connector reads during setup. Patch is false because only Groups PATCH is
+// implemented (Users PATCH is 501) — ServiceProviderConfig reports the
+// conservative, accurate value rather than "true because Groups supports it".
+func (ts *SCIMDiscoveryTestSuite) TestServiceProviderConfig() {
+	status, body, err := scimRequest(http.MethodGet, "/ServiceProviderConfig", nil, nil)
+	ts.Require().NoError(err)
+	ts.Require().Equal(http.StatusOK, status)
+
+	var cfg scimServiceProviderConfig
+	ts.Require().NoError(json.Unmarshal(body, &cfg))
+	ts.False(cfg.Patch.Supported, "Patch should report unsupported: only Groups PATCH is implemented")
+	ts.True(cfg.Filter.Supported, "Filter should report supported: Users list supports single eq")
+	ts.True(cfg.ETag.Supported, "ETag should report supported: If-Match/If-None-Match are implemented")
+	ts.False(cfg.Pagination.Cursor, "Pagination.cursor should be false: only index-based pagination is implemented")
+	ts.True(cfg.Pagination.Index, "Pagination.index should be true: startIndex/count pagination is supported")
+	ts.Equal(scimPaginationMaxPageSize, cfg.Pagination.MaxPageSize,
+		"Pagination.maxPageSize must equal the PaginationMaxPageSize constant")
+}
+
+// TestServiceProviderConfigPaginationClampsCount asserts that requesting a count
+// above the advertised pagination.maxPageSize results in the response's
+// itemsPerPage being clamped to maxPageSize. This ties the capability
+// advertisement to the enforcement in the handler so that any future drift
+// between the two constants is caught here.
+func (ts *SCIMDiscoveryTestSuite) TestServiceProviderConfigPaginationClampsCount() {
+	// First confirm what maxPageSize is advertised.
+	status, body, err := scimRequest(http.MethodGet, "/ServiceProviderConfig", nil, nil)
+	ts.Require().NoError(err)
+	ts.Require().Equal(http.StatusOK, status)
+
+	var cfg scimServiceProviderConfig
+	ts.Require().NoError(json.Unmarshal(body, &cfg))
+	maxPageSize := cfg.Pagination.MaxPageSize
+	ts.Require().Greater(maxPageSize, 0, "maxPageSize must be a positive integer")
+
+	// Request a count strictly greater than the advertised maxPageSize.
+	overCount := maxPageSize + 1
+	listURL := fmt.Sprintf("/Users?count=%d", overCount)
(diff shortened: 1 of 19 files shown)


tests/integration/scim/me_test.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// TestReplaceMe is skipped: PUT /Me needs only authentication, but
+// scim.ReplaceUser calls internal/user.UpdateUser, whose
+// validateOrganizationUnitForUserType check requires system:usertype:view
+// with no self-access bypass (unlike checkUserAccess just above it) — so it
+// 500s for any self-service caller. Fix belongs in internal/user/service.go,
+// out of scope here.
+func (ts *SCIMMeTestSuite) TestReplaceMe() {
+	ts.T().Skip("known bug: PUT /Me 500s for self-service callers, see internal/user.UpdateUser")
```
sadilchamishka, line 188: This skip means the PR's stated `/Me GET and PUT (self-service)` coverage doesn't actually cover PUT — the success path is never exercised, only the two request-shape rejections below it.

Two asks:

1. The comment describes a real server-side 500 (`internal/user.UpdateUser`'s `validateOrganizationUnitForUserType` has no self-access bypass). That belongs in a tracked issue, not only in a skip message — please file one and reference it here, e.g. `ts.T().Skip("blocked on #NNNN: PUT /Me 500s for self-service callers")`. Otherwise the suite stays green and nothing ever prompts the fix.
2. Note this would be the first `t.Skip` in `tests/integration/` used to park a product defect; every existing one gates on environment (`SERVER_EXTRACTED_HOME`, sqlite-only). Worth flagging to reviewers so it isn't read as precedent.

Also relevant to the stacked #4769: it makes `PUT /Me` reject core-only payloads outright, so this path has one more reason to need coverage once unblocked.

tests/integration/scim/search_test.go
```diff
@@ -0,0 +1,237 @@
...
+
+	ts.Equal(1, list.StartIndex, "startIndex should default to 1")
+	ts.GreaterOrEqual(list.TotalResults, 2, "should see at least both fixture users")
+	ts.GreaterOrEqual(list.ItemsPerPage, 1)
+
+	ids := make([]string, 0, len(list.Resources))
+	for _, u := range list.Resources {
+		ids = append(ids, u.ID)
+	}
+	ts.Contains(ids, ts.engineeringUserID)
```
sadilchamishka: This asserts the fixture users are present in `list.Resources`, but that's only **page 1** — the search request sends no `count`, so the server applies `DefaultPageSize` (30) and the list is deployment-wide, not scoped to this suite's OU (the admin token sees everything).

So the test silently depends on the whole deployment holding fewer than ~30 users at this exact moment. Other suites' fixtures, bootstrap users, and leftovers from an interrupted run all count. Once that's exceeded, the fixtures land on page 2 and this fails even though `/.search` is working correctly.

`TotalResults` and `StartIndex` assertions above are fine. For the membership check, either send a filter that isolates the fixtures, or send `Count` large enough and assert against `TotalResults`.

tests/integration/scim/discovery_test.go
```diff
@@ -0,0 +1,223 @@
...
+func (ts *SCIMDiscoveryTestSuite) TestServiceProviderConfig() {
+	status, body, err := scimRequest(http.MethodGet, "/ServiceProviderConfig", nil, nil)
+	ts.Require().NoError(err)
+	ts.Require().Equal(http.StatusOK, status)
+
+	var cfg scimServiceProviderConfig
+	ts.Require().NoError(json.Unmarshal(body, &cfg))
+	ts.False(cfg.Patch.Supported, "Patch should report unsupported: only Groups PATCH is implemented")
+	ts.True(cfg.Filter.Supported, "Filter should report supported: Users list supports single eq")
+	ts.True(cfg.ETag.Supported, "ETag should report supported: If-Match/If-None-Match are implemented")
```
sadilchamishka, line 97: This branch also carries the change that adds the RFC 9865 `pagination` block to `GET /ServiceProviderConfig` (see `backend/internal/scim/config/config.go` in this same PR), but `scimServiceProviderConfig` in `model.go` doesn't declare the field and nothing asserts it here.

That gap matters concretely: the advertised `pagination.maxPageSize` is `serverconst.MaxPageSize` (100), while `users_handler.go:79` and `group_handler.go:47` actually clamp `count` to `scimconfig.FilterMaxResults` (200). An assertion like "request `count` above the advertised `maxPageSize` and confirm `itemsPerPage` respects it" would have caught that; right now nothing does.

Since this suite is the integration coverage for discovery, please add the `pagination` sub-object to the model and assert it.

tests/integration/scim/discovery_test.go
```diff
@@ -0,0 +1,223 @@
...
+	ts.True(ids[scimCoreUserSchemaURN], "Schemas list should include the core User schema")
+	ts.True(ids[scimCoreGroupSchemaURN], "Schemas list should include the core Group schema")
+	ts.Require().NotNil(extSchema, "Schemas list should include our test entity type's extension schema")
+
+	required := map[string]bool{}
+	for _, attr := range extSchema.Attributes {
+		required[attr.Name] = attr.Required
+	}
+	ts.True(required["email"], "email should be reported required, matching the entity type definition")
+	ts.False(required["department"], "department was not declared required in the entity type")
```
sadilchamishka, line 180: `required` is populated only from attributes actually present in the response, so a missing key yields `false` and this assertion passes either way — it's green both when `department` is correctly non-required *and* when `department` has vanished from the extension schema entirely.

Assert presence separately so the test can distinguish the two:

```suggestion
	ts.Contains(required, "department", "department should be present in the extension schema")
	ts.False(required["department"], "department was not declared required in the entity type")
```

tests/integration/scim/scim_filter_test.go
```diff
@@ -0,0 +1,247 @@
...
+	id, _ := created["id"].(string)
+	ts.Require().NotEmpty(id)
+	ts.createdUserIDs = append(ts.createdUserIDs, id)
+	return id
+}
+
+// listWithFilter issues GET /Users?filter=<filter>, escaped exactly as a real
+// client would send it, and returns the decoded status and body.
+func (ts *SCIMFilterTestSuite) listWithFilter(filter string) (int, scimUserListResponse) {
+	ts.T().Helper()
```
sadilchamishka, line 148: `department eq "Engineering"` matches across the whole deployment, not just this suite's OU, and `SCIMSearchTestSuite` provisions its own `Engineering` and `Sales` users (`search_test.go` `SetupSuite`). Both suites then assert `TotalResults == 1` on the same values.

Today that passes only because Go runs the two suites sequentially *and* both `TearDownSuite` deletes succeed. If a teardown DELETE fails, or a run is interrupted between the two suites, the leftover Engineering user makes the other suite return 2 and fail with an error that points nowhere near the cause.

Cheap fix: prefix the fixture values per suite (e.g. `"filter-Engineering"` / `"search-Engineering"`) so the exact-count assertions are genuinely self-contained.

tests/integration/scim/helpers.go
```diff
@@ -0,0 +1,132 @@
...
+
+// scimRequestUnauthenticated issues a request against the SCIM API with no
+// token-injecting transport, so an empty or garbage Authorization header
+// (headers) actually reaches the server as-is. testutils.GetHTTPClient's
+// transport unconditionally overwrites Authorization on every call, which
+// makes it unusable for no-token/invalid-token tests.
+func scimRequestUnauthenticated(method, path string, headers map[string]string) (int, []byte, error) {
+	client := &http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
sadilchamishka, line 77: This client has no `Timeout`, unlike `testutils.GetNoRedirectHTTPClient` which sets 30s. If the server hangs on an unauthenticated request, the test blocks until the whole `go test` binary times out and you get a goroutine dump instead of a named failing test.

```suggestion
	client := &http.Client{
		Transport: &http.Transport{
			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
		},
		Timeout: 30 * time.Second,
	}
```
(needs a `"time"` import)

tests/integration/scim/groups_test.go
```diff
@@ -0,0 +1,394 @@
...
+	})
+	ts.Require().NoError(err, "failed to create test entity type")
+	ts.entityTypeID = entityTypeID
+
+	urn, _, err := discoverExtensionSchema(ts.entityTypeName)
+	ts.Require().NoError(err, "failed to discover extension schema via GET /Schemas")
+	ts.extensionURN = urn
+
+	ts.memberUserName = "scim.it.group-member"
+	body, _ := json.Marshal(map[string]interface{}{
```
sadilchamishka: Marshal error discarded here, unlike every other `json.Marshal` in these suites which uses `ts.Require().NoError(err)`. A failure would post an empty body and surface as a puzzling 400 from the create below rather than at the actual fault.

```suggestion
	body, err := json.Marshal(map[string]interface{}{
```
(and add `ts.Require().NoError(err)` after the literal; note `err` is already redeclared by the `scimRequest` call below, so use `=` there)

tests/integration/scim/groups_test.go
```diff
@@ -0,0 +1,394 @@
...
+
+// TestListWithFilterRejected pins a known limitation: unlike Users, Groups
+// rejects any "filter" query param outright, even though
+// ServiceProviderConfig currently advertises Filter.Supported=true for the
+// service as a whole. A provisioning connector cannot rely on filtering
+// Groups here — it must list and match client-side.
+func (ts *SCIMGroupsTestSuite) TestListWithFilterRejected() {
+	status, created := ts.createGroup("scim-it-group-filter", nil)
+	ts.Require().Equal(http.StatusCreated, status)
+	_ = created
```
sadilchamishka: `_ = created` is dead — it reads as though an assertion was dropped. The `createGroup` result isn't needed here, so just discard it at the call site:

```suggestion
	status, _ = ts.createGroup("scim-it-group-filter", nil)
	ts.Require().Equal(http.StatusCreated, status)
```
(placed in the two lines above, replacing the `status, created := ...` / `_ = created` pair)

Separately, worth noting what this test pins: `ServiceProviderConfig` advertises `filter.supported = true` while `/Groups` rejects `filter` outright. Same shape as the `pagination.maxPageSize` mismatch — the discovery document is drifting from what the endpoints do. A tracked issue for the ServiceProviderConfig accuracy as a whole would be better than three tests each documenting one gap.

tests/integration/scim/me_test.go
ravindu439: skipped this fix , since the , fix need in the users package.

tests/integration/scim/discovery_test.go
ravindu439: fixed

tests/integration/scim/discovery_test.go
ravindu439: fixed

tests/integration/scim/scim_filter_test.go
ravindu439: fixed

tests/integration/scim/helpers.go
ravindu439: fixed

tests/integration/scim/groups_test.go
ravindu439: fixed

tests/integration/scim/groups_test.go
ravindu439: fixed

tests/integration/scim/search_test.go
ravindu439: fixed

PasinduYeshan (approved):

PR #4786, opened 2026-08-10
Disable adding and deleting resources for read-only resource servers

### LINKED ISSUE #4775: Read-only (declarative) resource servers are editable from the console
### Description

A resource server loaded from declarative YAML is marked read-only. The list correctly replaces its Edit and Delete actions with a view-only icon, and its detail page shows "This resource is read-only and cannot be modified".

Directly below that banner, the resource tree is fully interactive: add, delete, and save controls are all live and enabled. None of the operations they offer can succeed.

- **Add** a resource or action returns a 500.
- **Delete** a resource or action fails, and the message says "Cannot delete — remove child resources and actions first", which is not the actual reason.
- **Editing** a resource or action name or description offers a Save button that fails.

Clicking the row in the list also navigates to the detail page, so the view-only icon is not a barrier.

### Steps to Reproduce

1. Add a resource server as declarative YAML under `config/resources/resource_servers/`, with at least one resource and action. Restart.
2. Open the resource servers list. The row shows the read-only badge and a view-only icon.
3. Click the row. The detail page opens with the read-only banner and the resource tree below it.
4. Hover a node in the tree. Add and delete controls appear.
5. Add a sub-resource. Observe the 500.
6. Delete a node. Observe the failure and the unrelated message.
7. Select a node, change its name, and save. Observe the failure.

### Version

1.0.0-beta2

_No response_

rajithacharith:
### Purpose

A resource server loaded from declarative YAML is marked read-only, and the console already reflects that in the list (view-only icon) and at the top of the detail page ("This resource is read-only and cannot be modified"). The resource tree directly below that banner was still fully interactive: add, delete, and save controls were all live, and none of the operations they offered could succeed. Adding returned a 500, deleting failed with an unrelated "remove child resources and actions first" message, and renaming offered a Save button that failed.

This makes the resource tree genuinely view-only when the resource server is read-only.

### Approach

`isReadOnly` on the resource server is now threaded into the tree instead of stopping at the page chrome:

- **`ResourceTree`** hides the "Resource Hierarchy" toolbar Add button and its menu, passes `readOnly` down to every node, and shows a `tree.emptyReadOnly` message in place of the empty-state hint that invited adding.
- **`ResourceTreeNode`** takes a `readOnly` prop on both `ResourceNode` and `ActionNode`. The hover/selected control cluster (add + delete) no longer renders when it is set, and the flag propagates through recursive children and nested action nodes. Copy-permission and expand/collapse remain available.
- **`ResourceDetailPanel`** previously derived `isReadOnly` as `selectedNode.type === 'server' && selectedNode.data.isReadOnly`. No server node is selectable in the generic tree, so that condition never tripped and resource/action fields stayed editable with a live Save button. It now derives from `resourceServer.isReadOnly`, which disables name/description/identifier and suppresses the Save/Reset bar for every node type. The inline alert stays scoped to the server node so it does not duplicate the page-level banner.
- **`McpCapabilitiesPanel`** gets the same treatment for MCP servers: the header Add button/menu and the empty-state "Add tool permission" / "Add resource permission" buttons are hidden, and action nodes receive `readOnly`.

Two related points intentionally left alone:

- Clicking a read-only row still navigates to the detail page. The issue notes the view-only icon is not a barrier, but the page is a legitimate inspection surface and is now genuinely read-only, so blocking navigation would remove the only way to view a declarative server's resources.
- The backend still rejects a create against a declarative server with a 500, and `tree.deleteResource.error` still reports the child-resources reason for any delete failure. Both are now unreachable for declarative servers through the console, so this change stays UI-scoped.

### Related Issues
- Fixes https://github.com/thunder-id/thunderid/issues/4775

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

Read-only coverage added to `ResourceTree.test.tsx` (no Add button, read-only empty copy, `readOnly` reaching nodes, for both API and MCP servers), `ResourceTreeNode.test.tsx` (add/delete absent on hover), and `ResourceDetailPanel.test.tsx` (resource and action nodes: fields disabled, no Save bar). Full `configure-resource-servers` suite passes (318 tests), along with lint, prettier, and `tsc --noEmit`.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added read-only mode for resource servers.
  * Prevented adding, editing, and deleting resources or actions in read-only mode.
  * Added dedicated empty-state messaging for read-only servers.
  * Preserved permission-copy controls while limiting other actions.

* **Bug Fixes**
  * Ensured read-only restrictions apply consistently across server, resource, and action views.

* **Tests**
  * Added coverage for read-only controls, messaging, fields, and save behavior.


rajithacharith:
image

brionmario (approved):

PR #4834, opened 2026-08-11
Cascade resource, action and resource server deletions to role permissions

### LINKED ISSUE #4806: Deleting resources/actions does not cascade to role permission assignments; deleting the resource server then triggers "resource server not found" in Console
### Description

When resources or actions belonging to a resource server are deleted, the deletion is not reflected in the permission assignments of roles that reference them. The stale permission references remain assigned to the role instead of being cleaned up.

This referential-integrity gap surfaces two distinct problems, the second being a follow-on of the first:

- Stale permission assignments — Deleting a resource or action leaves a reference in every role it was assigned to.

https://github.com/user-attachments/assets/639b41c5-760f-43e1-a5f8-e2d2c0e5c9c0

- "Resource server not found" error — Once the resource server itself is deleted (along with its previously-deleted resources/actions that were still assigned to a role), viewing that role's permissions in the Console fails with a resource server not found error, because the UI tries to resolve references that no longer exist.

Image

### Steps to Reproduce

1. Create a resource server with one or more resources and actions.
2. Assign some of those resources/actions as permissions to a role.
3. Delete one or more resources or actions(that were assigned to the role) from the resource server.
4. Check the role's permission
5. The permission tree doesn't show those resource existence but at the bottom you can see the earlier assigned permissions
6. After deleteing all resources, delete the resource server too
7. Now you will see  "Resource server not found" error in role's permission assignment tab

### Version

1.0.0-beta2

_No response_

thiva-k:
### Purpose
Deleting a resource, action or resource server left the permissions it contributed assigned to every role that held them. The stale references caused three problems: roles kept permissions that no longer existed, viewing a role's permissions failed with "resource server not found" once the server was deleted, and the role could no longer be updated because the stale permission failed validation.

### Approach
Resource deletions now participate in the existing dependency registry, which already handles cascade cleanup for user, group, application and agent deletions.

- `resourceService.DeleteResourceServer`, `DeleteResource` and `DeleteAction` call `CascadeDelete` inside their delete transaction, after the row is removed, so cleanup failures roll back with the delete.
- `roleService` is registered as a dependency provider and implements `CascadeDeleter`. Role permissions are stored as opaque strings scoped to a resource server rather than as references to the resource defining them, so orphans are found by re-validating referenced permissions through `resourceService.ValidatePermissions` instead of matching the deleted resource ID. This also clears permissions orphaned by earlier deletions, and keeps declarative resource servers correct since validation is composite-aware.
- Added a `ResourceTypeAction` dependency type.

Declarative role permissions are immutable and are never pruned.

### Related Issues
- Fixes #4806

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Deleting resource servers, resources, or actions now automatically removes related role permissions.
  - Unrelated and still-valid permissions are preserved.
  - Cleanup failures now return an error instead of leaving partially cleaned-up permissions.
  - Deletions now fail safely when cleanup services are unavailable.

- **Tests**
  - Added coverage for dependency cleanup, failure handling, and role permission preservation during deletions.

### DIFF
--- a/backend/internal/role/service.go
+++ b/backend/internal/role/service.go
@@ -7,6 +7,7 @@ package role
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
@@ -17,6 +18,7 @@ import (
 	resourcepkg "github.com/thunder-id/thunderid/internal/resource"
 	serverconst "github.com/thunder-id/thunderid/internal/system/constants"
 	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/system/resourcedependency"
 	"github.com/thunder-id/thunderid/internal/system/security"
 	"github.com/thunder-id/thunderid/internal/system/sysauthz"
 	"github.com/thunder-id/thunderid/internal/system/utils"
@@ -46,6 +48,9 @@ type RoleServiceInterface interface {
 	ResolveRoleOUHandle(
 		ctx context.Context, role *RoleWithPermissionsAndAssignments,
 	) *tidcommon.ServiceError
+	GetResourceDependencies(
+		ctx context.Context, resourceType, id string) ([]resourcedependency.ResourceDependency, error)
+	CascadeDeleteDependencies(ctx context.Context, resourceType, id string) (int, error)
 }
 
 // roleService is the default implementation of the RoleServiceInterface.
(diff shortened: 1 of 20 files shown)


backend/internal/role/service.go
```diff
@@ -698,3 +703,53 @@ func (rs *roleService) isRoleDeclarative(ctx context.Context, roleID string) boo
...
+
+// CascadeDeleteDependencies implements resourcedependency.CascadeDeleter. When a resource server,
+// resource or action is deleted, the permissions it contributed can no longer be resolved, so they
+// are removed from every role holding them. Permissions are stored as opaque strings scoped to a
+// resource server rather than as references to the resource that defines them, so the orphans are
+// found by re-validating the referenced permissions against the resource service. This also clears
+// any permission orphaned by an earlier deletion. Must be called after the target has been deleted,
+// so the deleted permissions no longer validate.
+func (rs *roleService) CascadeDeleteDependencies(
+	ctx context.Context, resourceType, _ string) (int, error) {
```
rajithacharith, line 722: We have omited the id? How the resource dependencies were identified?

thiva-k: Yes, `id` is intentionally unused, since it can't identify anything here.

`ROLE_PERMISSION` stores `(ROLE_ID, RESOURCE_SERVER_ID, PERMISSION)` as an opaque permission string, with its only FK on `ROLE_ID`. The resource/action UUID is never stored, so the `id` can't be joined to any role permission.

Instead orphans are found by re-validation: read the distinct permissions referenced by roles, ask `resourceService.ValidatePermissions` which no longer resolve, and delete those. `resourceType` still gates whether the prune runs.

Going through `ValidatePermissions` rather than a SQL anti-join also keeps declarative resource servers correct (their permissions live in the file store, not the DB), and clears rows orphaned before this fix.

rajithacharith (approved):

PR #4846, opened 2026-08-11
Fix retry button label and message pluralization on import summary

### LINKED ISSUE #4768: Import Summary page retry button wraps oddly and messages don't pluralize
### Description

Two small UI polish bugs on the Import Summary page

1. "Retry" button wraps to 3 lines, rendering oversized

On a failed dry-run, the retry button's label is "Retry Import Test"  vs just "Test" for the initial button in the same slot. Same `size="small"` styling in both cases but nothing accommodates the longer text, so it wraps onto three lines and the button balloons vertically, looking oversized next to the rest of the compact layout.

Image

2. Import success message doesn't pluralize ("N resource imported" for any N)

`summary.import.completedSuccessfully_plural` exists in `en-US.ts:3273` but is never used, and silently falls back to the singular form always. The same dead-code pattern affects 4 other keys: `summary.import.completedWithFailures_plural`, `summary.importTest.failedCount_plural`, `summary.import.tooltip.missingVariables_plural`, `quickTest.summary.method.social_plural`.

Image

Cosmetic only, no functional impact.

### Steps to Reproduce

- Go to Import Configuration, upload a config that will fail dry-run (e.g. missing env value), note the "Retry Import Test" button wraps to 3 lines.
- Import  resources successfully, note the success message says "N resource imported".

### Version

1.0.0-beta2

OS: macOS 
Database: default SQLite

ZiyamSanthosh:
### Purpose

Fixes two UI polish bugs on the Import Configuration summary page reported in #4768.

1. On a failed dry-run, the retry button's label was `Retry Import Test` while the initial button in the same narrow `Alert` action slot reads `Test`. Nothing accommodated the longer text, so the label wrapped onto three lines and the button ballooned vertically.
2. Counted messages never pluralized (`N resource imported` for any `N`). The keys used the legacy i18next v3 `_plural` suffix, which i18next 25 (JSON v4 plural resolution) does not resolve, so lookups silently fell back to the singular base key.

### Approach

1. Shortened `summary.importTest.retry` to `Retry`, matching the `Test` label that occupies the same slot, so the button stays on one line without extra styling.
2. Renamed the four dead `_plural` keys to the `_one`/`_other` pairs i18next 25 resolves, matching the convention already used elsewhere in the locale file (e.g. `welcome.wayfinderFolderImport.status.resourcesImported_one`/`_other`):
   - `summary.import.completedSuccessfully`
   - `summary.import.completedWithFailures`
   - `summary.importTest.failedCount`
   - `summary.import.tooltip.missingVariables`

#### Why the keys had to be renamed, and why nothing that references them breaks

i18next 25 resolves plurals through `Intl.PluralRules` categories, which for English are `key_one` and `key_other`. The `_plural` suffix is the i18next v3 format and is never looked up in v4+, so each `_plural` entry was unreachable text and every call fell through to the singular base key. That is the bug in the issue.

Call sites never spell out the suffix. They call `t('summary.import.completedSuccessfully', {count: n})` and i18next appends the suffix at lookup time from `count`, so the key as written in code is unchanged and no consumer needs updating. The only way a rename like this could break a caller is a site that resolves one of these keys *without* `count`, which would now render the raw key. There are none: each of the four keys has exactly one reference, all four pass `count`, none is pulled in through `$t()` nesting in another string, there is no second locale file, and `TranslationResources` is a mapped type over the locale object rather than a hardcoded key list.

Strictly, only the `_plural` to `_other` half of each rename is required; i18next falls back to the base key when `_one` is absent. The base key was renamed to `_one` as well so each pair is explicit and consistent with the rest of the file instead of relying on fallback behaviour.

A fifth key from the issue, `quickTest.summary.method.social`, was left untouched. It has no references anywhere (dead since it was introduced in b65d8cae8), so renaming it would be churn rather than a fix. Its equally unused `quickTest.*` siblings mean that whole namespace deserves a separate cleanup.

Updated the two retry-button assertions in `ImportConfigurationSummaryPage.test.tsx` and added a success-toast test covering a multi-resource import. That test reproduces the bug against the old strings (`2 resource imported`) and passes with the fix (65/65 in the package).

### Related Issues
- Fixes #4768

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated import failure actions to display the shorter **“Retry”** label.
  - Improved import success messages with correct singular and plural wording when importing multiple resources.

- **Localization**
  - Updated English import-summary translations to use standardized singular and plural wording.

### DIFF
--- a/frontend/packages/configure-import-export/src/pages/__tests__/ImportConfigurationSummaryPage.test.tsx
+++ b/frontend/packages/configure-import-export/src/pages/__tests__/ImportConfigurationSummaryPage.test.tsx
@@ -537,7 +537,7 @@ describe('ImportConfigurationSummaryPage', () => {
       render(<ImportConfigurationSummaryPage />);
 
       await waitFor(() => {
-        expect(screen.getByRole('button', {name: /Retry Import Test/})).toBeInTheDocument();
+        expect(screen.getByRole('button', {name: 'Retry'})).toBeInTheDocument();
       });
     });
 
@@ -600,7 +600,7 @@ describe('ImportConfigurationSummaryPage', () => {
       render(<ImportConfigurationSummaryPage />);
 
       await waitFor(() => {
-        expect(screen.getByRole('button', {name: /Retry Import Test/})).toBeInTheDocument();
+        expect(screen.getByRole('button', {name: 'Retry'})).toBeInTheDocument();
       });
     });
   });
@@ -1022,6 +1022,35 @@ describe('ImportConfigurationSummaryPage', () => {
 
       await waitFor(() => {
         expect(mockShowToast).toHaveBeenCalledWith('Import completed with 1 failed resource.', 'warning');
+      });
+    });
+
+    it('pluralizes the success toast when more than one resource is imported', async () => {
+      mockLocationState.configContent = noTemplateState.configContent;
+      mockLocationState.envData = noTemplateState.envData;
+      mockMutateAsync
+        .mockResolvedValueOnce({
+          results: [],
+          summary: {imported: 2, totalDocuments: 2, failed: 0, importedAt: new Date(0).toISOString()},
+        })
+        .mockResolvedValueOnce({
+          results: [],
+          summary: {imported: 2, totalDocuments: 2, failed: 0, importedAt: new Date(0).toISOString()},
+        });
+
+      render(<ImportConfigurationSummaryPage />);
+
+      await userEvent.click(screen.getByText('Test'));
+
+      const importButton = await waitFor(() => {
+        const button = getImportConfigurationButton();
+        expect(button).not.toBeDisabled();
+        return button;
+      });
+      await userEvent.click(importButton);
+
+      await waitFor(() => {
+        expect(mockShowToast).toHaveBeenCalledWith('Import completed successfully. 2 resources imported.', 'success');
       });
     });
 
--- a/frontend/packages/i18n/src/locales/en-US.ts
+++ b/frontend/packages/i18n/src/locales/en-US.ts
@@ -3265,25 +3265,25 @@ const translations = {
     'summary.importTest.running': 'Running pre-flight dry-run...',
     'summary.importTest.runningShort': 'Running...',
     'summary.importTest.test': 'Test',
-    'summary.importTest.retry': 'Retry Import Test',
+    'summary.importTest.retry': 'Retry',
     'summary.importTest.passed': 'Import test passed. {{imported}} of {{totalDocuments}} resources validated.',
-    'summary.importTest.failedCount': 'Import test failed for {{count}} resource',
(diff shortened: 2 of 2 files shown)


frontend/packages/configure-import-export/src/pages/__tests__/ImportConfigurationSummaryPage.test.tsx
```diff
@@ -537,7 +537,7 @@ describe('ImportConfigurationSummaryPage', () => {
       render(<ImportConfigurationSummaryPage />);
 
       await waitFor(() => {
-        expect(screen.getByRole('button', {name: /Retry Import Test/})).toBeInTheDocument();
+        expect(screen.getByRole('button', {name: /Retry/})).toBeInTheDocument();
```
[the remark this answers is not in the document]
ZiyamSanthosh: Good catch, fixed. Both assertions now use the exact accessible name (`{name: 'Retry'}`), so they fail if the longer `Retry Import Test` label returns.

frontend/packages/i18n/src/locales/en-US.ts
```diff
@@ -2545,8 +2545,8 @@ const translations = {
     'quickTest.summary.text': '{{appName}} is configured with {{methods}} sign-in.',
     'quickTest.summary.method.usernamePassword': 'username & password',
     'quickTest.summary.method.passkey': 'passkeys',
-    'quickTest.summary.method.social': '{{count}} social provider',
-    'quickTest.summary.method.social_plural': '{{count}} social providers',
+    'quickTest.summary.method.social_one': '{{count}} social provider',
+    'quickTest.summary.method.social_other': '{{count}} social providers',
```
ThaminduDilshan: Do we have valid usages for these new keys?

ZiyamSanthosh: Good question, and no, there are none. `quickTest.summary.method.social` has zero references in code (no literal usage and no dynamic key construction), so it never rendered. It has been dead since it was added in b65d8cae8, which introduced the strings without a consumer. In fact all 32 `quickTest.*` keys are unreferenced, so the whole namespace is dead.

I have reverted that pair to its original form, since renaming a key nothing reads is churn rather than a fix. The remaining four pairs in this PR are all rendered with `{count}` from `ImportConfigurationSummaryPage.tsx` (`summary.import.completedSuccessfully`, `summary.import.completedWithFailures`, `summary.importTest.failedCount`, `summary.import.tooltip.missingVariables`), so the pluralization fix applies only where it is user visible.

Cleaning up the dead `quickTest.*` namespace is worth doing, but its siblings (`usernamePassword`, `passkey`, and the rest) are equally unused, so I have left that for a separate PR rather than removing one arbitrary pair here.

ThaminduDilshan (approved):

PR #4861, opened 2026-08-11
Add authorization and credential state check on refresh grant

thiva-k:
### Purpose
A refresh re-minted tokens carrying stale scopes: unassigning a role, deleting it, or stripping its permissions had no effect until the refresh token expired. A password reset or a client secret rotation also left existing refresh tokens usable.

### Approach
- Re-evaluate permission scopes against the subject's current role and group assignments on every refresh, dropping the ones they no longer hold. OIDC scopes are not affected.
- Record `credentialUpdatedAt` in the entity's system attributes when a password or client secret changes, and reject a refresh token established at or before that instant. The marker is written in the same transaction as the credential, and carried across by the application, agent, and user paths that rebuild the blob.
- Reject the refresh when the subject no longer resolves to an entity and the client maps no subject attribute.

Scoped to the refresh grant, so other grants are unaffected. Already-issued access tokens still live out their TTL, and applications that map `sub` to a user attribute skip the subject-derived checks.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Refresh-token requests now re-evaluate current permissions and may reduce authorization scopes.
  - Refresh tokens are rejected after password resets, client-secret rotations, or user deletion.
  - Credential changes are tracked to reliably invalidate earlier refresh tokens.

- **Documentation**
  - Updated OAuth, token, and application guides to explain scope reduction and refresh-token invalidation.

- **Tests**
  - Added comprehensive coverage for permission changes, credential updates, deleted users, and client-secret rotations.

### DIFF
--- a/backend/internal/authnprovider/common/constants.go
+++ b/backend/internal/authnprovider/common/constants.go
@@ -11,6 +11,10 @@ const (
 	UserAttributeSub = "sub"
 )
 
+// SystemAttrCredentialUpdatedAt is the entity system-attribute key recording when the entity's
+// authentication credential last changed: a user's password, or a client secret.
+const SystemAttrCredentialUpdatedAt = "credentialUpdatedAt" // #nosec G101 -- attribute key, not a secret
+
 // Credential type keys used in the credentials map passed to the authentication providers.
 const (
 	// CredentialTypeProvisionedEntityID identifies an entity provisioned earlier in the same flow.
--- a/backend/internal/entity/service.go
+++ b/backend/internal/entity/service.go
@@ -9,7 +9,9 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
+	authnprovidercm "github.com/thunder-id/thunderid/internal/authnprovider/common"
 	"github.com/thunder-id/thunderid/internal/entitytype"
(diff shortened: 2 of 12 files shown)


backend/internal/application/service.go
```diff
@@ -463,7 +463,19 @@ func (as *applicationService) updateEntityDataForApplicationUpdate(ctx context.C
...
+	if epErr != nil {
+		if svcErr := mapEntityProviderError(epErr); svcErr != nil {
+			return svcErr
+		}
+		as.logger.Error(ctx, "Failed to get entity for system attribute update",
+			log.String("appID", appID), log.Error(epErr))
+		return &tidcommon.InternalServerError
+	}
+
+	sysAttrsJSON, marshalErr := buildSystemAttributes(app, clientID, existingEntity.SystemAttributes)
```
senthalan: can't this logic in the entity layer? So it will be common for any entity

thiva-k: Moved to entity/service.go

ThaminduDilshan (approved):

PR #4862, opened 2026-08-12
Add integration coverage for the flow engine and SSO sessions

indeewari:
### Purpose

Integration coverage for the flow engine and SSO sessions sat well below the 85% bar, with the gaps concentrated in error, resume, timeout and administration paths that unit tests cover only in isolation.

This adds 50 integration tests across fifteen areas. Measured statement coverage from the instrumented build (`target/coverage_integration.out`):

| Package | Before | After | Δ |
|---|---|---|---|
| `flow/executor` | 50.5% | **63.9%** | **+13.4** |
| `flow/session` | 57.4% | **70.5%** | **+13.1** |
| `flow/flowexec` | 65.2% | **74.8%** | **+9.6** |
| `flow/mgt` | 61.4% | **70.7%** | **+9.3** |
| `flow/core` | 66.3% | 66.8% | +0.5 |

859 statements newly covered.

### Approach

**Flow execution error branches** (`tests/integration/flow/execution/flow_execution_error_test.go`)

Six flow-execution error branches had no integration assertion at all. The three administration gates matter most, since `/flow/execute` is a public path and these checks are the only thing between any caller and administration flow execution:

- unknown and missing flow type, unknown application
- registration and recovery disabled on the application
- unknown and malformed execution id
- unauthenticated execution by flow id, and the same rejection for a flow id that does not exist, so the endpoint does not leak which flows are present
- an administrator executing a non-administration flow by id
- a client credentials token refused at the administration entry point

**Nested call depth** (`tests/integration/flow/execution/call_depth_test.go`)

A chain of flows each calling the next, one longer than the engine allows, must be refused rather than recursing. The limit is what stops a mutually recursive set of flows, which the designer does not prevent an operator authoring, from exhausting the stack.

**Registration flow inference** (`tests/integration/flow/mgt/flow_inference_test.go`)

Creating an authentication flow with `flow.auto_infer_registration` enabled derives a registration flow from it, renamed and carrying the provisioning step that turns collected credentials into a user. The flag is off by default, so these tests patch the deployment configuration and restart, restoring on teardown. Also covers the flow-type executor requirements: a registration flow without a provisioning executor is rejected rather than stored.

**Flow usages** (`tests/integration/flow/mgt/flow_usages_test.go`)

`GET /flows/{flowId}/usages` had no test. Covers an unreferenced flow reporting a known-empty result, an application binding appearing as a usage with the fields the Console renders, and the not-found case.

**Execution lifecycle** (`tests/integration/flow/execution/flow_lifecycle_test.go`)

Resume of an existing execution, refusal to resume without the required input, and context expiry. Expiry is driven by writing `authFlow.expirySeconds` and needs no restart, because the flow section is read from merged server config on every execution. The original writable layer is restored on cleanup.

**Administration flow** (`tests/integration/flow/execution/administration_flow_test.go`)

One execution of the shipped `default-user-deletion-flow` drives the whole chain: permission validation, pre-delete validation publishing the trusted revocation plan, criteria revocation, session termination, and record deletion. Plus the unknown-subject and missing-subject cases. This is the first integration coverage of the criteria-based revocation path.

**SSO session timeouts** (`tests/integration/oauth/sso/session_timeout_test.go`)

Session timeouts are read once when the session service is constructed, so these tests write the session configuration and restart the server, restoring and restarting again on cleanup. The restore is registered before any change, so a mid-test failure cannot leave the run with second-scale session lifetimes.

The absolute-timeout test sets idle equal to absolute and uses the session mid-window. That slides the idle deadline past the absolute one, leaving the absolute cap as the only thing that can end the session, which is what distinguishes the two deadlines.

**Consent** (`tests/integration/flow/authentication/consent_test.go`)

The consent executor had no integration coverage at all. Covers the prompt being raised for the attributes an application requests, approval recording consent so a second sign-in skips the prompt entirely, denial of optional attributes completing the flow with nothing consented, an unparseable decisions payload, a submission carrying the timeout reason (recorded as no decision), and a decision arriving after the configured step timeout.

**Identifying executor modes** (`tests/integration/flow/authentication/identify_modes_test.go`)

`resolve` narrows a candidate set instead of failing on ambiguity: two users sharing an email return the attribute that separates them with the candidate values as options, and answering it resolves the user from the stored candidates rather than a second search. `check_state` records whether zero, one or several users match and lets the flow branch on it; each of the three outcomes is asserted from the prompt the flow lands on, which also exercises the engine's condition-skip path.

**User onboarding** (`tests/integration/flow/execution/user_onboarding_test.go`)

`USER_ONBOARDING` is app-independent: it carries no `applicationId` and is resolved through `flow.userOnboardingFlow.defaultHandle`. The OU is chosen first from the full tree, and the selected OU then decides which user types are on offer, so picking the root auto-selects its only type while picking a child prompts between the two. Also covers an unknown OU, a type outside the node's allowed list, and a type whose OU is not an ancestor of the selected one.

**OU resolver strategies** (`tests/integration/flow/registration/ou_resolver_strategies_test.go`)

`prompt` asks for an OU only when the user type's OU has children and accepts only selections inside that subtree; `caller` fails on a self-service registration flow because there is no authenticated caller to take an OU from; an unrecognized strategy fails the node rather than being ignored.

**Attribute uniqueness** (`tests/integration/flow/registration/attribute_uniqueness_test.go`)

A value already held by another user is reported against the attribute that conflicts and the details are asked for again, before provisioning runs. Covers a conflict on either unique attribute and the free-value path through to provisioning.

**Flow observability events** (`tests/integration/flow/execution/flow_events_test.go`)

Enables the `observability` section with the file sink and asserts what the engine publishes: the full node trail of a successful run correlated by execution id, a node that forwards back to its prompt (published as completed with the forwarding status and the reason), and a node that genuinely fails. The sink flushes on a fixed ticker, so the reads wait on the specific event rather than on any event of the type, which would otherwise return a half-flushed prefix of the run.

**Call frames** (`tests/integration/flow/execution/call_frames_test.go`)

A call that pauses inside the callee has its caller frame written into the stored context and read back on resume, then returns to the caller's success target. A failing callee ends the caller when the call node names no failure target, and resumes at that target when it does.

**Session termination and configuration** (`tests/integration/oauth/sso/`)

Deleting a user through the shipped administration flow tears down the SSO sessions that user holds; the session is proven live beforehand by an authorize that skips the credential prompt, so the assertion afterwards is a change in behaviour rather than an absence. Session configuration validation refuses negative timeouts, an idle deadline beyond the absolute cap, and a refresh interval not below the idle window.

**Default flow fallback** (`tests/integration/flow/execution/flow_lifecycle_test.go`)

Deleting a flow an application references is allowed, so the reference dangles. Rather than failing every sign-in for that application, the engine falls back to the configured default authentication flow.

**Inference prompt meta** (`tests/integration/flow/mgt/flow_inference_test.go`)

The inferred registration flow carries the source flow's UI, so its authentication wording is rewritten to its registration equivalent and the self sign-up link is dropped, because the inferred flow is the sign-up.

### Notes for reviewers

Behaviours worth knowing, each of which cost a test run to discover:

- An `AUTHENTICATION` flow must contain an `AuthAssertExecutor` (`FLM-1023`), so even a fixture flow that is never completed needs one.
- The shared test HTTP clients treat `/flow/execute` as a public endpoint and **skip token injection**. Any test of the administration entry point has to set the bearer header itself on a raw client. This is documented in the helper.
- Resuming a prompt without its required input returns `ERROR` rather than re-presenting the prompt. That is now pinned by its own test.
- A `REGISTRATION` flow must carry both a `UserTypeResolver` and a `ProvisioningExecutor`; the full table is `requiredExecutorsByFlowType` in the validator, alongside a `companionExecutors` map that pairs executors which must appear together.
- **An application binding a registration flow needs an isolated auth flow.** Otherwise the default authentication flow's own registration reference collides with it and application creation fails with `APP-1039`. `testutils.CreateIsolatedAuthFlow` exists for this.
- **A node nothing can reach is rejected at create time** (`FLM-1020`), so a fixture's failure-target node has to exist only in the variant whose call node points at it.
- **A wrong password is not a node failure.** The credentials executor returns `ExecUserInputRequired`, so the published event is a completed node execution carrying the forwarding status and the reason, not a failure.
- **`PatchDeploymentConfig` merges at the top level only.** Patching one key inside a nested block replaces the whole block, silently dropping its siblings. Doing that to `flow` dropped `max_version_history` and broke two unrelated version-history tests in the same package, which no scoped test run could reveal. Both patches here restate the block exactly as `tests/integration/resources/deployment.yaml` sets it.

### Known gaps

- **`FES-1019` (administration permission required) is not covered.** A client credentials token is rejected as unauthenticated before permissions are consulted, because it establishes no user subject. Reaching that branch needs a signed-in non-administrator user. Recorded in the test file.
- **`flow/interceptor` and `flow/graphbuilder` are unchanged**, and deliberately so. `CaptchaValidationProvider` is an engine SDK extension point with no implementation or configuration in the server, so the captcha interceptor cannot execute in this deployment. The graph builder's error branches require a nil or node-less flow, or a structural build failure, both of which create-time validation rejects first. Both files already have unit tests, which is the right vehicle for defensive paths and SDK extension points.
- **`federated_auth_resolver` is not covered.** Reaching it needs stored candidates from the identifying executor together with a verified federated subject, but the OIDC executor writes the federated attributes only on success and returns before them when the local user is ambiguous. Whether that composition is reachable at all is unresolved, so it was left rather than guessed at.
- **85% is not reached, and per-package 85% does not look reachable through integration tests.** Three separate causes, in order of size:
  - `flow/executor`'s remaining gap is dominated by `passkey` (200 statements, needing a WebAuthn virtual authenticator) and `openid4vp` (67, needing a wallet). Both are test-harness components rather than tests, and are separate work.
  - **Some of the uncovered code has no production call site.** In `flow/core`, `graph.ToJSON`, `graph.RemoveEdge`, `factory.CloneNode`, `factory.CloneNodes` and the graph setters are declared on their interfaces and called from nowhere in the server: roughly 111 statements no integration test can execute. In `flow/mgt`, `insertPhoneInputPromptIfNeeded` and its helper `createInputPromptNode` (42 statements) are invoked only by their own unit tests; a test written against the behaviour they describe fails, because nothing calls them. Both look like dead code and are worth a separate issue.
  - The rest is largely defensive: nil-guards on paths that always pre-initialize, row-scan and decode error branches, and the `InterceptorRunnerContext` helpers, which are reachable only through an interceptor and the sole implementation (captcha) cannot execute in this deployment.
- **`publishFlowFailedEvent` never fires on a node failure.** Across 198 events published during a run, `FLOW_FAILED` appeared zero times: the paths that publish it are gated on `flowStep.Status == FlowStatusError`, which a node-level failure does not set. The node itself is published as `FLOW_NODE_EXECUTION_FAILED`, so the failure is observable, but the run-level event is not. Possibly intended, flagging it either way.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Tests**
  - Expanded integration coverage for authentication flows, including consent, user identification, onboarding, registration, nested calls, lifecycle handling, and execution errors.
  - Added validation for administration user deletion, including termination of active SSO sessions.
  - Added coverage for flow usage reporting, observability events, call-depth limits, and fallback behavior.
  - Added tests for SSO idle and absolute session timeouts.
  - Improved test utilities for safely managing writable server configuration.

### DIFF
--- a/tests/integration/flow/execution/call_depth_test.go
+++ b/tests/integration/flow/execution/call_depth_test.go
@@ -1,0 +1,181 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package execution
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"github.com/thunder-id/thunderid/tests/integration/flow/common"
+	"github.com/thunder-id/thunderid/tests/integration/testutils"
+)
+
+// errCodeMaxCallDepth is returned when nested CALL nodes exceed the engine's frame limit.
+const errCodeMaxCallDepth = "FES-1013"
+
+// callChainLength is longer than the engine's maximum call depth of 10 frames, so executing the head
+// of the chain is guaranteed to cross the limit.
+const callChainLength = 12
+
+var callDepthTestOU = testutils.OrganizationUnit{
+	Handle:      "call_depth_test_ou",
+	Name:        "Test OU for Call Depth",
+	Description: "Organization unit created for call depth testing",
+	Parent:      nil,
+}
+
+type CallDepthTestSuite struct {
+	suite.Suite
+	ouID     string
+	appID    string
+	flowIDs  []string
+	headFlow string
+}
+
+func TestCallDepthTestSuite(t *testing.T) {
+	suite.Run(t, new(CallDepthTestSuite))
+}
+
+// SetupSuite builds a chain of authentication flows where each one calls the next, then binds an
+// application to the head of the chain. The chain is built from the tail backwards because a CALL
+// node must reference a flow that already exists.
+func (ts *CallDepthTestSuite) SetupSuite() {
+	ouID, err := testutils.CreateOrganizationUnit(callDepthTestOU)
+	ts.Require().NoError(err, "Failed to create test organization unit")
+	ts.ouID = ouID
+
+	// The tail is a plain authentication flow that calls nothing.
+	tailID, err := testutils.CreateFlow(ts.buildLeafFlow())
+	ts.Require().NoError(err, "Failed to create leaf flow")
+	ts.flowIDs = append(ts.flowIDs, tailID)
+
+	next := tailID
+	for i := 0; i < callChainLength; i++ {
+		flowID, err := testutils.CreateFlow(ts.buildCallingFlow(i, next))
+		ts.Require().NoError(err, "Failed to create calling flow %d", i)
+		ts.flowIDs = append(ts.flowIDs, flowID)
+		next = flowID
+	}
+	ts.headFlow = next
+
+	app := testutils.Application{
+		Name:         "Call Depth Test Application",
+		Description:  "Application bound to a deeply nested call chain",
+		ClientID:     "call_depth_test_client",
+		ClientSecret: "call_depth_test_secret", //nolint:gosec // test credential
+		RedirectURIs: []string{"http://localhost:3000/callback"},
+		OUID:         ts.ouID,
+		AuthFlowID:   ts.headFlow,
+	}
+	appID, err := testutils.CreateApplication(app)
+	ts.Require().NoError(err, "Failed to create test application")
+	ts.appID = appID
+}
+
+func (ts *CallDepthTestSuite) TearDownSuite() {
+	if ts.appID != "" {
+		if err := testutils.DeleteApplication(ts.appID); err != nil {
+			ts.T().Logf("Failed to delete test application during teardown: %v", err)
+		}
+	}
+	// Delete from the head backwards so a flow is never removed while another still references it.
+	for i := len(ts.flowIDs) - 1; i >= 0; i-- {
+		if err := testutils.DeleteFlow(ts.flowIDs[i]); err != nil {
+			ts.T().Logf("Failed to delete flow %s during teardown: %v", ts.flowIDs[i], err)
+		}
+	}
+	if ts.ouID != "" {
+		if err := testutils.DeleteOrganizationUnit(ts.ouID); err != nil {
+			ts.T().Logf("Failed to delete test organization unit during teardown: %v", err)
+		}
+	}
+}
+
+// credentialsAuthNodes returns the credential prompt, the authentication executor and the assertion
+// that make up a complete authentication flow, ending at the given node.
+//
+// Every flow in the chain carries these so each one is a valid authentication flow in its own right
+// rather than a graph that only happens to be accepted today, which keeps the fixture standing as
+// flow validation gains rules.
+func credentialsAuthNodes(endNode string) []map[string]interface{} {
+	return []map[string]interface{}{
+		{
+			"id":   "credentials_prompt",
+			"type": "PROMPT",
+			"prompts": []map[string]interface{}{
+				{
+					"inputs": []map[string]interface{}{
+						{"ref": "input_001", "identifier": "username", "type": "TEXT_INPUT", "required": true},
+						{"ref": "input_002", "identifier": "password", "type": "PASSWORD_INPUT", "required": true},
+					},
+					"action": map[string]interface{}{
+						"ref":      "action_credentials",
+						"nextNode": "credentials_auth",
+					},
+				},
+			},
+		},
+		{
+			"id":        "credentials_auth",
+			"type":      "TASK_EXECUTION",
+			"executor":  map[string]interface{}{"name": "CredentialsAuthExecutor"},
+			"onSuccess": "auth_assert",
+		},
+		{
+			"id":        "auth_assert",
+			"type":      "TASK_EXECUTION",
+			"executor":  map[string]interface{}{"name": "AuthAssertExecutor"},
+			"onSuccess": endNode,
+		},
+		{"id": endNode, "type": "END"},
+	}
+}
+
+// buildLeafFlow returns the flow at the end of the chain, which calls nothing and authenticates the
+// user itself.
+func (ts *CallDepthTestSuite) buildLeafFlow() testutils.Flow {
+	nodes := []map[string]interface{}{
+		{"id": "start", "type": "START", "onSuccess": "credentials_prompt"},
+	}
+	return testutils.Flow{
+		Name:     "Call Depth Leaf Flow",
+		FlowType: "AUTHENTICATION",
+		Handle:   "auth_flow_call_depth_leaf",
+		Nodes:    append(nodes, credentialsAuthNodes("end")...),
+	}
+}
+
+// buildCallingFlow returns a flow whose only work is to call the given target flow.
+//
+// Its own authentication section sits after the CALL, so the chain still descends on the first node
+// and trips the depth limit before any prompt is reached.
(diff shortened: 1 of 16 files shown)


tests/integration/flow/execution/call_depth_test.go
```diff
@@ -0,0 +1,165 @@
...
+			{"id": "end", "type": "END"},
+		},
+	}
+}
+
+// buildCallingFlow returns a flow whose only work is to call the given target flow.
+func (ts *CallDepthTestSuite) buildCallingFlow(index int, targetFlowID string) testutils.Flow {
+	return testutils.Flow{
+		Name:     fmt.Sprintf("Call Depth Link %d Flow", index),
+		FlowType: "AUTHENTICATION",
```
ThaminduDilshan, line 166: Shall we have a valid flow with an auth executor? Soon this is going to fail when we onboard more flow validation rules

tests/integration/flow/execution/call_depth_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ts.T().Logf("Failed to delete test organization unit during teardown: %v", err)
+		}
+	}
+}
+
+// buildLeafFlow returns the flow at the end of the chain, which calls nothing.
+func (ts *CallDepthTestSuite) buildLeafFlow() testutils.Flow {
+	return testutils.Flow{
+		Name:     "Call Depth Leaf Flow",
+		FlowType: "AUTHENTICATION",
```
ThaminduDilshan, line 144: Shall we have a valid flow with some auth executor too?

tests/integration/flow/mgt/flow_inference_test.go
```diff
@@ -0,0 +1,511 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+	"github.com/thunder-id/thunderid/tests/integration/testutils"
+)
+
+// Registration-flow inference is off by default, so it needs the deployment flag enabled and a
+// restart before any of it runs. It lives in its own suite rather than the main flow management
+// suite so the restart is paid once and cannot disturb the other tests.
+type FlowInferenceTestSuite struct {
```
ThaminduDilshan: Let's remove this test suite. This is something we have deprecated and will be removed in next release

tests/integration/flow/execution/call_depth_test.go
indeewari: Fixed

tests/integration/flow/execution/call_depth_test.go
indeewari: Fixed

tests/integration/flow/execution/call_depth_test.go
```diff
@@ -0,0 +1,165 @@
...
+		return
+	}
+
+	ts.Require().NotNil(step, "expected a flow step for a rejected call chain")
+	ts.NotEqual("COMPLETE", step.FlowStatus,
+		"a call chain past the depth limit must not complete")
+	if step.Error != nil {
+		ts.Equal(errCodeMaxCallDepth, step.Error.Code,
+			"the failure should name the call depth limit")
+	}
```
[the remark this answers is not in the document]
indeewari: Fixed

senthalan (approved):

PR #4875, opened 2026-08-12
Fix intermittent failures in FlowsListView and useDeleteRole tests

ThaminduDilshan:
### Purpose
Fixes two intermittently failing frontend tests observed in CI.

1. `FlowsListView.test.tsx` > `should call onClearSelection when selection is cleared` occasionally failed with `Error: The element to be cleared could not be focused.` from `user.clear()` on a MUI Autocomplete combobox.
2. `useDeleteRole.test.tsx` > `should set pending state during deletion` occasionally failed with `expected false to be true` when asserting `isPending`.

### Approach
1. `user.clear()` on a MUI Autocomplete's combobox input is a known source of flakiness: it needs to focus, select all text, and delete it on an input whose displayed value is really driven by the selected option, not free text, and the refocus step can intermittently fail. The codebase already has an established, reliable pattern for clearing an Autocomplete's selection elsewhere (`AppearanceSection.test.tsx`), clicking the Autocomplete's own `[aria-label="Clear"]` icon button. Applied the same pattern here instead of `user.clear()` + tab.

2. The pending-state test raced a real `setTimeout(..., 100)` mock against `waitFor`'s polling inside a Playwright-driven browser test environment. Under CI load, that 100ms pending window could be missed entirely, causing `isPending` to already be `false` by the time it was checked. Replaced the fixed timer with a manually-controlled promise so the mutation stays pending indefinitely until the test explicitly resolves it, removing the wall-clock race.

Both fixes were verified locally by running the affected test files directly (21/21 and 15/15 passing respectively) and confirming `tsc --noEmit` is clean.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Tests**
  * Improved test reliability for clearing selections in sign-in configuration.
  * Replaced timing-based role deletion test behavior with explicit request completion handling.

### DIFF
--- a/frontend/apps/console/src/features/applications/components/create-application/configure-signin-options/__tests__/FlowsListView.test.tsx
+++ b/frontend/apps/console/src/features/applications/components/create-application/configure-signin-options/__tests__/FlowsListView.test.tsx
@@ -209,16 +209,17 @@ describe('FlowsListView', () => {
 
     it('should call onClearSelection when selection is cleared', async () => {
       const user = userEvent.setup();
-      renderExpanded({
+      // The card is already expanded because selectedAuthFlow is set, so use renderComponent
+      // directly; renderExpanded would toggle it back closed.
+      renderComponent({
         selectedAuthFlow: mockFlows[0],
       });
 
       const autocomplete = screen.getByRole('combobox');
-      await user.click(autocomplete);
+      const clearButton = autocomplete.parentElement?.querySelector('[aria-label="Clear"]');
 
-      // Clear the selection by clicking outside or selecting null
-      await user.clear(autocomplete);
-      await user.tab(); // blur to trigger onChange with null
+      expect(clearButton).not.toBeNull();
+      await user.click(clearButton!);
 
       expect(mockOnClearSelection).toHaveBeenCalled();
     });
--- a/frontend/packages/configure-roles/src/api/__tests__/useDeleteRole.test.tsx
+++ b/frontend/packages/configure-roles/src/api/__tests__/useDeleteRole.test.tsx
@@ -103,9 +103,10 @@ describe('useDeleteRole', () => {
   });
 
   it('should set pending state during deletion', async () => {
+    let resolveRequest: () => void;
     mockHttpRequest.mockReturnValue(
       new Promise((resolve) => {
-        setTimeout(() => resolve(undefined), 100);
+        resolveRequest = () => resolve(undefined);
       }),
     );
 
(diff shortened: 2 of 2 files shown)


frontend/apps/console/src/features/applications/components/create-application/configure-signin-options/__tests__/FlowsListView.test.tsx
```diff
@@ -214,11 +214,10 @@ describe('FlowsListView', () => {
       });
 
       const autocomplete = screen.getByRole('combobox');
-      await user.click(autocomplete);
+      const clearButton = autocomplete.parentElement?.querySelector('[aria-label="Clear"]');
```
jeradrutnam, line 219: Shall we use a different HTML attribute for query selector. E.g., `[data-testid]`

Because `[aria-label]` is for accessibility

jeradrutnam (dismissed):

frontend/apps/console/src/features/applications/components/create-application/configure-signin-options/__tests__/FlowsListView.test.tsx
ThaminduDilshan: Good catch on this looking like an accessibility-only hook, but `getByRole` unfortunately can't reach this button at all: MUI's Autocomplete clear icon is `visibility: hidden` by default (only shown on hover/focus via CSS), and per the ARIA accessible-name spec, a hidden element's computed name is always `""` regardless of its `aria-label` — so `getByRole('button', {name: 'Clear'})` (even with `hidden: true`) never matches it. I verified this directly against the rendered DOM.

`aria-label="Clear"` is also the identifier MUI itself assigns to this button (not something we're adding for tests), so it's already the codebase's precedent for this exact case: `AppearanceSection.test.tsx` targets the same clear button the same way. A `data-testid` would need production code changes (passing `slotProps`/`componentsProps` into the Autocomplete purely for test hooks), which felt like more than this one-line assertion warrants.

While digging into this I also found the actual root cause of the original flake: this test called `renderExpanded()` with `selectedAuthFlow` already set, but the component defaults `expanded` to `true` whenever `selectedAuthFlow` isn't null — so the click in `renderExpanded` was toggling an already-open panel back **closed**. Because MUI's `Collapse unmountOnExit` doesn't remove the DOM node instantly (it exits on a real ~195ms timer), the Autocomplete stayed briefly queryable mid-collapse, which is what caused the intermittent "could not be focused" failure. Fixed by rendering with the panel already expanded instead of double-toggling it. Pushed as an update to the same commit.

thiva-k (approved):

PR #4921, opened 2026-08-13
Add integration tests for RP initiated logout negative paths

sahandilshan:
### Purpose

The RP initiated logout suite covered only the happy path: a valid `id_token_hint` plus a
registered `post_logout_redirect_uri` producing a 302. None of the rejection paths were
tested, including the open redirect guard, and the `GET /oauth2/logout` route was never
exercised by any test at all.

This PR adds 17 tests covering the negative and edge paths of the end session endpoint and
its callback.

### Approach

The existing helpers could not express a failure: `initiateLogout` hardcodes
`Require().Equal(http.StatusFound, ...)` and `completeLogout` hardcodes `StatusOK`. This PR
adds raw variants that return the status and body instead of asserting them, leaving the
existing helpers and their two tests untouched.

Coverage falls into three groups.

**Hint validation.** A malformed hint, a hint signed by a foreign key, and a hint carrying a
foreign issuer are each rejected with `invalid id_token_hint`. An **expired** hint must be
accepted, because RP initiated logout permits an expired ID token and the implementation
verifies the signature only. That is a positive case with real regression risk: swapping
`VerifyJWTSignature` for `VerifyJWT` would silently break it.

**Client and redirect resolution.** A request with neither a hint nor a `client_id`, a
`client_id` that does not match the hint, and an unknown `client_id` are each rejected with
their specific message. A registered `post_logout_redirect_uri` is honoured with only a
`client_id` and no hint, while an unregistered one is refused in both modes. The open
redirect case asserts positively that no `Location` header is emitted and the attacker
supplied URI is not echoed back in the body.

**Callback.** An absent redirect yields an empty `redirect_uri`, a replayed `logoutId` is
single use, an unknown `logoutId` yields an empty redirect rather than an error, and an
invalid body is rejected. Both endpoints answer an `OPTIONS` preflight.

#### Two cases that needed care to avoid passing for the wrong reason

The **expired hint** is a genuine product issued token, not a forgery. The test creates a
throwaway application with an ID token validity of one second, drives a real authorization
code login through it, sleeps, and asserts `exp` is in the past before using the hint. A
forged token would have collapsed this case into the foreign key case.

The **foreign issuer** case had the same hazard in reverse: the signature is verified before
`iss` is read, so any foreign signed token dies at the signature step. To reach the issuer
comparison the test signs with the product's own signing key, resolving the key id from the
live JWKS and the issuer from the discovery document rather than hardcoding either. It
carries a control asserting that the same minting routine with the correct issuer yields a
302, so a broken signer cannot make the test pass for the wrong reason.

Reviewers may want to weigh in on that last point: it is the only test here that reads the
deployment's signing key from `SERVER_EXTRACTED_HOME`. The alternative is dropping the
issuer case, since it cannot be expressed black box otherwise.

#### Incidental finding, not changed here

The comment at `backend/internal/oauth/oauth2/logout/handler_test.go:72` states that a
`post_logout_redirect_uri` is only honoured with an `id_token_hint`. That is stale: two
tests in this PR establish the opposite, matching `Resolve`'s own doc comment. Left
unedited to keep this PR to tests only.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for SSO logout scenarios, including GET and POST requests.
  * Added validation tests for expired or invalid tokens, issuer and signature checks, client and redirect URI validation, callback replay, malformed requests, and CORS preflight behavior.
  * Added coverage for logout redirects and OAuth flows involving varied client configurations.

### DIFF
--- a/tests/integration/oauth/sso/logout_negative_test.go
+++ b/tests/integration/oauth/sso/logout_negative_test.go
@@ -1,0 +1,667 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package sso
+
+import (
+	"bytes"
+	"crypto"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/json"
+	"encoding/pem"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/thunder-id/thunderid/tests/integration/testutils"
+)
+
+const (
+	logoutEndpoint         = "/oauth2/logout"
+	logoutCallbackEndpoint = "/oauth2/logout/callback"
+
+	// unregisteredPostLogoutRedirectURI is deliberately absent from the client's
+	// postLogoutRedirectUris list: it stands in for an attacker-supplied landing page and must be
+	// refused by the open-redirect guard.
+	unregisteredPostLogoutRedirectURI = "https://attacker.example.com/logged-out"
+
+	// foreignIssuer is an issuer this server never mints tokens for.
+	foreignIssuer = "https://not-thunderid.example.com"
+
+	// Fixtures for the short-lived ID token application used to obtain a genuinely expired,
+	// server-signed id_token_hint. It is created and deleted inside the single test that needs it.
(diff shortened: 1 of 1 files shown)


sahandilshan:
@coderabbitai, please review

sahandilshan:
@coderabbitai resume

senthalan (approved):

Issue #4918, opened 2026-08-13
Server aborts startup when the runtime persistent database is unreachable, contradicting the fail closed design


sahandilshan:
## Description

The token revocation enforcement path is designed to **fail closed** when the
runtime-persistent deny list cannot be read: `failClosed`
(`backend/internal/oauth/oauth2/revocation/enforcement_service.go:100-107`) logs, trips a
circuit breaker and returns `ErrEnforcementUnavailable`, which surfaces at
`/oauth2/introspect` as `500 server_error` rather than a misleading `{"active": false}`.

That design presumes the server keeps running while the runtime-persistent database is
unavailable. It does not. If the datasource cannot be opened at startup, the process
aborts instead of continuing in a degraded state.

## Evidence

`dbprovider.initializeAllClients`
(`backend/internal/system/database/provider/dbprovider.go:191-197`) only *logs* a
runtime-persistent initialization failure and continues, which is what the fail-closed
design expects. But a later consumer treats it as fatal. Pointing
`database.runtime_persistent` at an unopenable SQLite path and starting the server yields:

```
ERROR Failed to initialize runtime persistent database client ... unable to open database file (14)
ERROR Failed to initialize SSO session service: failed to get runtime persistent DB transactioner ...
```

and the process exits. The server never becomes ready.

## Why it matters

The fail-closed branch is only reachable when the datasource opens successfully but reads
fail afterwards, for example a database that is reachable but has lost its schema, or one
that fails at query time. A genuine outage, which is the scenario fail-closed exists for,
takes the server down instead.

To reach `failClosed` in an integration test at all, the deny list had to be pointed at a
**schema-less but openable** SQLite file: the driver creates it, the ping and the SSO
session transactioner both succeed so the server boots, and every deny-list query then
fails. That is a fairly artificial state compared to the outage the mechanism is meant to
survive.

## Suggested action

Decide which behaviour is intended and make the two consistent:

- if the server should stay up degraded, the SSO session service (and any other
  runtime-persistent consumer that aborts startup) needs to tolerate a missing
  transactioner and let the fail-closed paths do their job; or
- if an unreachable runtime-persistent database should be fatal, the fail-closed handling
  in the revocation enforcement service is dead weight for that scenario and its intent
  should be documented as covering query-time failures only.

## Impact

No incorrect authorization decision results either way: the server is either up and
failing closed, or down. This is a design-consistency gap rather than a security defect.

Issue #4911, opened 2026-08-13
/oauth2/introspect performs no token-ownership check: any authenticated client can introspect any token


sahandilshan:
## Description

`POST /oauth2/introspect` authenticates the caller but never checks whether the presented token
belongs to that caller. Any client holding valid credentials can introspect **any** token issued
by the deployment, including tokens belonging to other clients and to admin users.

## Root cause

`backend/internal/oauth/oauth2/introspect/init.go:56-58` correctly wraps the route in
`clientauth.ClientAuthMiddleware`, so the caller must authenticate. But the handler never reads the
authenticated client back out of the request context — `handler.go:47` calls:

```go
h.service.IntrospectToken(ctx, token, tokenTypeHint)
```

The authenticated client is never passed down, and `service.IntrospectToken` has no client
parameter, so no ownership comparison is possible anywhere in the call chain.

## Reproduction

Two confidential apps A and B, both `client_secret_basic` + `client_credentials`. Mint a token for
A, then introspect it while authenticating as B:

```
POST /oauth2/introspect   -u 'qa-b6-client-B:<secretB>'   -d token=<A's access token>

HTTP 200
{"active":true,"client_id":"qa-b6-client-A","token_type":"Bearer","exp":1786594915,
 "iat":1786591315,"nbf":1786591315,"sub":"019ff923-e7d0-73c5-a137-507823b4e38b",
 "aud":"qa-b6-client-A","iss":"https://localhost:8090","jti":"019ff924-078d-7f28-88a3-f8355678f097"}
```

The response is byte-identical to A introspecting its own token.

A second demonstration: the same throwaway client successfully introspected the **CONSOLE admin
user's** access token, recovering `scope: "openid system"`, `sub` and `jti`.

Controls confirm authentication itself is enforced, so ownership is the only gap:
- unauthenticated → `400 {"error":"invalid_request","error_description":"Missing client_id parameter"}`
- wrong secret → `401 {"error":"invalid_client",...}`

## Expected

RFC 7662 §2.1 requires the endpoint to determine whether the caller is authorized for the token.
The service should compare the authenticated client against the token's `client_id` (or its
audience / resource server), returning `{"active": false}` for tokens the caller does not own.

## Impact

Security-relevant. Any client credential in the deployment becomes an oracle for enumerating
metadata about arbitrary tokens — subject identifiers, scopes, audiences, expiry and `jti` —
including privileged admin tokens.

Thareesha98:
[...]

PR #4922, opened 2026-08-13
Add integration tests for the token introspection endpoint

sahandilshan:
### Purpose

There was no dedicated integration test package for `POST /oauth2/introspect`. The SSO,
tfid and revocation suites each carried a private helper that decoded only the `active`
field and used introspection as a probe for some other feature. The only genuine
introspection assertions in the repository were the two DPoP `cnf` and `token_type` checks.

As a result `iss`, `jti`, `exp`, `iat`, `nbf`, `sub` and `aud` were never asserted anywhere,
none of the client authentication failure modes were covered, and `failClosed` in the
revocation enforcement service had no integration coverage at all.

This PR adds a dedicated package with 18 tests.

### Approach

**Response shape.** The active token case decodes the access token with
`DecodeJWTPayloadMap` and compares the introspection response against the token's own claims
field by field, rather than merely asserting the fields are present.

**Inactive results.** Expired, malformed, foreign signed and revoked tokens must all return
HTTP 200 with `{"active": false}` per RFC 7662, not a 4xx. The expired token is genuinely
product issued: an application registers a one second access token validity, and the test
asserts `exp - iat == 1` before waiting. Note the wait is about 31 seconds, because
`jwt.leeway` defaults to 30 and validation only rejects once `now >= exp + leeway`. Reducing
it would require patching the leeway and restarting the server, which seemed a worse trade
than the wait.

**Refresh tokens.** Token validation is deliberately type agnostic, so a refresh token
introspects as active. The test pins this along with the fact that `sub` is the **client
id**, with the end user in `access_token_sub`, which introspection does not surface.

**Client authentication.** Missing credentials, a `client_id` alone, a wrong secret, an
unknown client and a mismatched authentication method each assert their specific status and
error code. `WWW-Authenticate: Basic` is only added when the request carried an
`Authorization` header, so the `client_id` alone case asserts the header is **absent** and
the challenge assertions live in the wrong secret and unknown client cases.

**Fail closed.** `failClosed` returns `ErrEnforcementUnavailable` when the revocation deny
list cannot be read, which must surface as `500 server_error` rather than
`{"active": false}`. Reporting a token as inactive when its revocation status is unknown
would be the dangerous failure mode, so that distinction is the whole point of the test.

Reaching it took some care. Pointing the datasource at an unopenable target does not leave
the server running: the provider logs and continues, but the SSO session service then aborts
startup, so the server never becomes ready. That mismatch is filed separately as #4918. The
suite instead points `runtime_persistent` at a **schema-less but openable** SQLite file, so
the driver creates it, the ping and the session transactioner succeed and the server boots,
while every deny list query fails. This is a configuration only change with no direct
database manipulation.

The break and its restore sit inside the test method with a `defer`, not in `SetupSuite`.
testify only runs `TearDownSuite` if `SetupSuite` completed, so breaking the configuration
during setup would strand a broken deployment on any setup failure. `TearDownSuite` keeps a
safety net flag in case the deferred restore is bypassed.

Recovery is verified three ways: the deferred restore re-introspects the same token and
requires an active result, the restored deployment configuration was inspected directly, and
the `oauth/revocation` package, which exercises revocation, introspection and token exchange
against the runtime persistent database, passes 9 of 9 afterwards.

#### Out of scope

Cross client introspection is a confirmed defect, filed as #4911, so it is deliberately left
untested and unpinned rather than encoding the current behaviour.

### Related Issues
- Related to #4911, #4918

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive OAuth token introspection coverage for active, expired, malformed, foreign, revoked, and refresh tokens.
  * Added validation for token hints, required parameters, client authentication, HTTP methods, CORS preflight, and RFC 7662 response fields.
  * Added coverage confirming introspection fails safely when revocation data is unavailable and recovers after restoration.

### DIFF
--- a/tests/integration/oauth/introspect/fail_closed_test.go
+++ b/tests/integration/oauth/introspect/fail_closed_test.go
@@ -1,0 +1,285 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package introspect
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+
+	"github.com/thunder-id/thunderid/tests/integration/testutils"
+)
+
+const (
+	failClosedClientID     = "introspect_fail_closed_client"
+	failClosedClientSecret = "introspect_fail_closed_secret"
+	failClosedResourceID   = "https://introspect-fail-closed.example.com"
+
+	// unreadableDenyListPath points database.runtime_persistent at a SQLite file that does not exist
+	// yet. The driver creates it on connect, so the server boots normally, but the file carries none
+	// of the runtime-persistent schema: every deny-list read fails at query time. Pointing at a
+	// genuinely unopenable target instead is not usable here, because the SSO session service aborts
+	// startup when it cannot get a runtime-persistent transactioner. The path has no directory
+	// component so it resolves against the server home, which exists under every database profile.
+	unreadableDenyListPath = "introspect_fail_closed_no_schema.db"
+)
+
+// IntrospectFailClosedTestSuite verifies that token introspection fails closed when the
+// runtime-persistent database, which holds the RFC 7009 revocation deny list, cannot be read. If the
+// deny list cannot be consulted the server must not claim the token is inactive (that would be a
+// silent, unauthenticated denial) and must not claim it is active (that would honor a token whose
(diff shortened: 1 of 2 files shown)


sahandilshan:
@coderabbitai review

sahandilshan:
@coderabbitai resume

rajithacharith (approved):

Issue #4142, opened 2026-07-20
Make auth observability events principal-aware (agent / user / client identity, actor–subject delegation, and correlation)


Dilusha-Madushan:
### Current Limitation

All principal types currently share the same OAuth endpoints and produce shape-identical observability events. `TOKEN_ISSUANCE_STARTED` and `TOKEN_ISSUED` only include fields such as `client_id`, `grant_type`, `scope`, and `duration_ms`.

This was verified against a live run of the agent quickstarts (own-token + OBO).

### Issues

- **No principal-type marker**
  - An agent's `client_credentials` token is indistinguishable from any other machine client's token.
  - An agent OBO (`authorization_code`) token is indistinguishable from a normal user login.
  - Determining whether a token belongs to an agent currently requires an external registry lookup using the raw `client_id`.

- **No subject information**
  - The authenticated end user (human who logged in) is not included in any event.
  - This information is missing even from the `AUTHENTICATION` flow events (`credentials_auth`, `auth_assert`).

- **No actor / delegation relationship**
  - In OBO flows, there is no indication that **agent X acted on behalf of user Y**.
  - There is no `act` / `sub` relationship or delegation information.
  - As a result, the defining characteristic of on-behalf-of authentication is not observable or auditable.

- **Broken correlation**
  - The login flow records the application as `app_id` (internal UUID), while token issuance records it as `client_id` (OAuth client ID), with no cross-reference.
  - `FLOW_STARTED` generates a different `trace_id` from its own child events and `FLOW_COMPLETED`, which instead reuse the `execution_id`.
  - Consequently, the complete flow:
    - Login
    - Authentication
    - Authorization Code
    - Token Issuance

    cannot be stitched together into a single trace.

- **No agent-related identity fields**
  - A keyword scan of the complete logs for:

    ```
    agent
    actor
    subject
    sub
    act
    on_behalf
    delegat
    principal
    identity
    ```

    returned **0 matches**.

As agents become first-class principals in ThunderID, logging only that "a token was issued to client X" is no longer sufficient. Operators must be able to determine **which agent**, **acting for which user**, and **under what delegation** a token was issued.

### Suggested Improvement

Enrich the authentication observability events (`TOKEN_ISSUANCE_STARTED`, `TOKEN_ISSUED`, and the `AUTHENTICATION` flow events) with principal context.

#### Add principal type

Include a field such as:

- `actor_type`: `user | agent | client`

or alternatively:

- `is_agent`

[...]

PR #4931, opened 2026-08-13
Add principal and correlation context on observability events

lashinijay:
### Purpose

Refs #4142.

Authentication observability events are principal-blind. `TOKEN_ISSUANCE_STARTED` and `TOKEN_ISSUED` carry only `client_id`, `grant_type`, `scope` and `duration_ms`, so an agent's `client_credentials` token is indistinguishable from any other machine client's, and an agent acting for a user from a normal login. No event records the authenticated subject or the actor-to-subject delegation that defines on-behalf-of authentication. Correlation is broken separately: the login flow identifies the application by `app_id` while token issuance uses `client_id`, and `FLOW_STARTED` emits a different `trace_id` from its own children, so login → authentication → authorization code → token issuance cannot be stitched into one trace.

This enriches the existing event payloads. No new event types and no pipeline change: the console, file and OpenTelemetry subscribers serialize the `Data` map generically, so the new fields become queryable span attributes with no subscriber change.

Six keys, named after the token claims they mirror so one vocabulary covers tokens and events:

| Key | Value |
|---|---|
| `sub` | resource ID of the principal the token or flow is about |
| `sub_type` | `user`, `agent` or `application` |
| `act_sub` | resource ID of the actor acting for the subject |
| `act_type` | principal type of the acting party |
| `is_delegated` | whether the issuance carries actor-to-subject delegation |
| `correlation_id` | identifier shared by every event of one authentication |

`TOKEN_ISSUED` for an agent acting for a user:

```json
"data": {
  "act_type": "agent",   "act_sub": "019ff53b-63fb-…", "is_delegated": true,
  "sub_type": "user",    "sub":     "019ff975-414a-…",
  "app_id":   "019ff53b-63fb-…", "client_id": "obo_probe_agent_client",
  "correlation_id": "55de9ebc-…", "grant_type": "urn:ietf:params:oauth:grant-type:token-exchange"
}
```

`act_type` alone separates an agent's own token from an ordinary M2M token: same grant, byte-identical requests, different declared entity category. Where `sub_type` diverges from `act_type` the issuance is delegated, and `act_sub` says on whose behalf.

**`sub` is an entity resource ID, never the token's `sub` claim.** An application can map that claim to a schema attribute via `subjectAttribute`, so it may hold a directly identifying value such as an email address, and it varies per application for the same principal. The resource ID is opaque, stable across applications, joins the flow half of an authentication to the token half, and is what `ActorProvider.GetActor` is keyed on. Application logs already log resource IDs rather than attributes; the event pipeline has no masking layer, so what it publishes must be safe unmasked. On a deployment that configures `subjectAttribute`, an event's `sub` and the token's `sub` claim therefore differ — correlate on the resource ID.

Sample Log:

{"trace_id":"797ab4a6-93a1-44d6-badf-dbdc729f886b","event_id":"019ffa4d-2de4-7e1f-8554-ce574ee681d5","type":"TOKEN_ISSUED","timestamp":"2026-08-13T14:16:29.860587+05:30","component":"AuthHandler","status":"success","data":{"act_sub":"019ff53b-63fb-7466-b74a-0300663a4895","actor_type":"agent","app_id":"019ff53b-63fb-7466-b74a-0300663a4895","client_id":"obo_probe_agent_client","correlation_id":"797ab4a6-93a1-44d6-badf-dbdc729f886b","duration_ms":"1","grant_type":"urn:ietf:params:oauth:grant-type:token-exchange","is_delegated":true,"scope":"","sub":"019ffa4d-2c7a-7784-83fa-50cb1b61f1c3","subject_type":"user"}}

---

### Approach

**Principal type comes from the entity, not from inference.** `GetOAuthClientByClientID` already loads the entity behind an OAuth client, so `BuildOAuthClient` stamps `EntityCategory` from that same read and both publish sites get it for free. Nothing infers agent-ness from behaviour, grant type or naming. `Application.EntityCategory` is added to the engine SPI for the flow side, marked `json:"-" yaml:"-"` so it appears on no API response and in no declarative resource.

**`event.PrincipalType` is the one place two spellings are reconciled.** An entity category spells an application `app`; events spell it `application`, matching the `sub_type` claim in #5042. An unrecognized category returns empty, so the field is omitted rather than guessed.

**The subject is resolved once, in `BuildAccessToken`, with an upstream fast path.** Every grant passes through it, so none can omit the fields by forgetting to set them — including one added later. Where an upstream layer already knows the answer it supplies it and the lookup is skipped:

| Grant | Source | Entity read |
|---|---|---|
| `authorization_code` | `sub_id` + `sub_type` claims on the flow assertion | none |
| `client_credentials` | the client already loaded | none |
| `refresh_token` | the entity resolved for the credential check, else the attribute-cache entry written at login | none |
| token exchange / jwt-bearer / ID-JAG | the presented token's subject, gated on resolving | yes |

The login and refresh paths therefore add no latency. Reads that do happen are served by the cache-backed entity store.

**Resolvability is what keeps a mapped attribute off the events.** On the exchange grants the subject arrives as a presented token's `sub` — a resource ID on most deployments, the mapped attribute where the application configured one. A mapped attribute matches no entity, so an unresolvable subject is reported as no subject rather than published as-is. Both fields are omitted together: a field that guesses is worse than one that is absent when it will be read as evidence about who did what.

**Correlation travels on carriers that already cross request boundaries.** An authentication spans several requests, so no request-scoped value can join them. `correlation_id`, `sub_id` and `sub_type` ride the flow assertion (a JWT) into the authorization code (a JSON document in the runtime store) and on to token issuance, following the route `tfid` already takes. For refresh, the identity is recorded on the attribute-cache entry the login created — server-side state the refresh token already points at, so nothing is added to the token. Both carriers are schemaless with respect to this change: **no schema change and no data migration.** Grants with no originating flow fall back to the request trace id.

None of the three new assertion claims is a trust input: the assertion is signed, and none is used to authorize, gate a decision or select a code path. A wrong value produces a mis-stitched trace or an absent field, not a wrong authorization outcome.

**Known limitations, deliberate:**

- `sub`/`sub_type` are omitted on token exchange, jwt-bearer and ID-JAG when the presented subject is a mapped attribute, and on refresh when the attribute-cache entry has expired or the flow was App Native (which caches nothing). Closing these needs the resource ID on the token itself — a wire-format decision with its own blast radius.
- An impersonation exchange (no `actor_token`) reports `is_delegated: false`. `ShouldAppendActorClaim` is applied on `authorization_code` and `refresh_token` but not in `token_exchange.go`, so an agent can obtain a user's token through exchange with no `act` claim. That asymmetry predates this PR; the pattern remains visible as `grant_type=token_exchange` with `is_delegated=false`.
- `servicemanager.go` is unchanged. Neither publish site gains a dependency; the only injection is `ActorProvider` into the token builder.

Design discussion: https://github.com/thunder-id/thunderid/discussions/5018

### Related Issues
- Refs #4142

### Related PRs
- #5042 — adds the `sub_type` claim (`application` | `agent`) to `client_credentials` tokens from the same `EntityCategory` field. This PR adopts its vocabulary rather than inventing a parallel one. Both touch `tokenservice/builder.go` and `oauth2/constants/constants.go`, so whichever lands second rebases, and the two category-mapping helpers should be reconciled into one.

### Checklist
    - `docs/content/community/contributing/contributing-code/backend-development/observability.mdx` — data key reference, the resource-ID-not-`sub` rationale, and the coverage limits
    - [x] Unit Tests — resolution branches (`tokenservice/subject_identity_test.go`), refresh identity sources (`granthandlers/refresh_subject_identity_test.go`), flow principal and correlation fields (`flowexec/engine_principal_event_test.go`), token actor/subject/delegation (`token/events_test.go`), assertion claims beside a mapped `sub` (`flow/executor/auth_assert_executor_test.go`), assertion decode and code carry-through (`authz/handler_test.go`), wire names and mapping (`observability/event/datakeys_test.go`)
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Improved observability for flows and OAuth token issuance with consistent correlation IDs.
  - Events now include principal, subject, actor type, delegation, client, and entity details when available.
  - OAuth tokens and authorization codes preserve relevant subject identity and actor metadata.
  - Refresh-token flows retain subject identity more reliably, including cached identity details.
- **Bug Fixes**
  - Prevented mapped attributes from overwriting security-sensitive identity claims.
  - Improved handling when subject or entity information is unavailable.
- **Documentation**
  - Expanded observability guidance for principal, delegation, and correlation data.

### DIFF
--- a/backend/internal/flow/flowexec/engine_principal_event_test.go
+++ b/backend/internal/flow/flowexec/engine_principal_event_test.go
@@ -1,0 +1,306 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package flowexec
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+
+	"github.com/thunder-id/thunderid/internal/flow/common"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	"github.com/thunder-id/thunderid/internal/system/observability/event"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+	"github.com/thunder-id/thunderid/tests/mocks/flow/coremock"
+	"github.com/thunder-id/thunderid/tests/mocks/observability/observabilitymock"
+)
+
+// setupCapturingObservability returns a mock observability service that records every published
+// event, so tests can assert on the data an event carries rather than only that it was published.
+func setupCapturingObservability(t *testing.T) (
+	*observabilitymock.ObservabilityServiceInterfaceMock, *[]*providers.Event) {
+	t.Helper()
+
+	// setupMockObservability is called for its side effect of initializing the server runtime with
+	// observability enabled; its mock registers a catch-all PublishEvent that would shadow the
+	// capturing expectation below, so a dedicated mock is used here.
+	setupMockObservability(t)
+
+	captured := &[]*providers.Event{}
+	mockObs := &observabilitymock.ObservabilityServiceInterfaceMock{}
+	mockObs.On("IsEnabled").Return(true).Maybe()
+	mockObs.
+		On("PublishEvent", mock.Anything, mock.Anything).
+		Run(func(args mock.Arguments) {
+			*captured = append(*captured, args.Get(1).(*providers.Event))
+		}).
+		Return().Maybe()
+
+	return mockObs, captured
+}
+
+// agentFlowContext returns an engine context for an authentication flow driven by an agent that has
+// authenticated a user. The subject is recorded on AuthUser, which is where the authentication
+// executors put it — EngineContext.AuthenticatedUser is never written during execution, so a test
+// that seeds it would assert against a path production never takes.
+func agentFlowContext() *EngineContext {
+	ctx := &EngineContext{
+		ExecutionID: "flow-exec-1",
+		TraceID:     "request-trace-1",
+		FlowType:    providers.FlowTypeAuthentication,
+		AppID:       "agent-entity-1",
+		Application: providers.Application{
+			ID:             "agent-entity-1",
+			EntityCategory: providers.EntityCategoryAgent,
+			InboundAuthConfig: []providers.InboundAuthConfigWithSecret{
+				{
+					Type:        providers.OAuthInboundAuthType,
+					OAuthConfig: &providers.OAuthConfigWithSecret{ClientID: "agent-client-id"},
+				},
+			},
+		},
+		ExecutionHistory: make(map[string]*providers.NodeExecutionRecord),
+	}
+	ctx.AuthUser.SetStateFor("credentials", providers.AuthState{
+		EntityReference: &providers.EntityReference{
+			EntityID:       "user-123",
+			EntityCategory: string(providers.EntityCategoryUser),
+			EntityType:     "Person",
+		},
+	})
+	return ctx
+}
+
+func TestFlowStartedEvent_ReportsPrincipalAndCorrelation(t *testing.T) {
+	mockObs, captured := setupCapturingObservability(t)
+	defer config.ResetServerRuntime()
+
+	publishFlowStartedEvent(agentFlowContext(), mockObs)
+
+	if len(*captured) != 1 {
+		t.Fatalf("expected 1 published event, got %d", len(*captured))
+	}
+	evt := (*captured)[0]
+
+	// The whole flow must share one trace, keyed on the execution id, so FLOW_STARTED stitches to
+	// its own child events.
+	if evt.TraceID != "flow-exec-1" {
+		t.Errorf("TraceID = %q, want the execution id %q", evt.TraceID, "flow-exec-1")
+	}
+
+	assertEventData(t, evt, map[string]interface{}{
+		event.DataKey.ActorType:     event.PrincipalTypeAgent,
+		event.DataKey.ClientID:      "agent-client-id",
+		event.DataKey.EntityID:      "agent-entity-1",
+		event.DataKey.CorrelationID: "flow-exec-1",
+		event.DataKey.Subject:       "user-123",
+		event.DataKey.SubjectType:   event.PrincipalTypeUser,
+	})
+}
+
+// subjectFlowContext returns a minimal context whose only authenticated subject is the given entity
+// reference, isolating the subject fields from the application-side ones.
+func subjectFlowContext(ref providers.EntityReference) *EngineContext {
+	ctx := &EngineContext{
+		ExecutionID:      "flow-exec-3",
+		FlowType:         providers.FlowTypeAuthentication,
+		AppID:            "app-3",
+		ExecutionHistory: make(map[string]*providers.NodeExecutionRecord),
+	}
+	ctx.AuthUser.SetStateFor("credentials", providers.AuthState{EntityReference: &ref})
+	return ctx
+}
+
+// The authentication executors record the resolved entity on AuthUser; AuthenticatedUser carries no
+// id during execution, so the subject must be read from AuthUser for it to appear at all.
+func TestFlowStartedEvent_SubjectComesFromAuthUser(t *testing.T) {
+	mockObs, captured := setupCapturingObservability(t)
+	defer config.ResetServerRuntime()
+
+	publishFlowStartedEvent(subjectFlowContext(providers.EntityReference{
+		EntityID:       "user-999",
+		EntityCategory: string(providers.EntityCategoryUser),
+		EntityType:     "Person",
+	}), mockObs)
+
+	assertEventData(t, (*captured)[0], map[string]interface{}{
+		event.DataKey.Subject:     "user-999",
+		event.DataKey.SubjectType: event.PrincipalTypeUser,
+	})
+}
+
+// An agent can authenticate through a flow, so the subject category must come from the resolved
+// entity reference rather than being assumed to be a user.
+func TestFlowStartedEvent_AgentSubjectReportsAgentCategory(t *testing.T) {
+	mockObs, captured := setupCapturingObservability(t)
+	defer config.ResetServerRuntime()
+
+	publishFlowStartedEvent(subjectFlowContext(providers.EntityReference{
+		EntityID:       "agent-999",
+		EntityCategory: string(providers.EntityCategoryAgent),
+	}), mockObs)
+
+	assertEventData(t, (*captured)[0], map[string]interface{}{
+		event.DataKey.Subject:     "agent-999",
+		event.DataKey.SubjectType: event.PrincipalTypeAgent,
+	})
+}
+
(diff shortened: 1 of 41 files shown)


backend/internal/flow/flowexec/engine_principal_event_test.go
```diff
@@ -0,0 +1,306 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package flowexec
```
thiva-k: Lets move this into existing test file engine_test.go

backend/internal/oauth/oauth2/granthandlers/refresh_subject_identity_test.go
```diff
@@ -0,0 +1,86 @@
+// Copyright 2026 The ThunderID Authors
```
thiva-k: Let's move this into existing test files if possible. Check other newly created test files too

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -1015,6 +1015,24 @@ type Application struct {
 	InboundAuthProfile `yaml:",inline"`
 	InboundAuthConfig  []InboundAuthConfigWithSecret `yaml:"inboundAuthConfig,omitempty" json:"inboundAuthConfig,omitempty" jsonschema:"Inbound authentication configuration (OAuth2/OIDC settings)."`
 	Metadata           map[string]interface{}        `yaml:"metadata,omitempty" json:"metadata,omitempty" jsonschema:"Generic metadata key-value pairs."`
+
+	// EntityCategory is the category of the entity backing this runtime application view (app or
+	// agent). Runtime-only: never serialized on the application API or in declarative resources.
+	EntityCategory EntityCategory `yaml:"-" json:"-"`
```
thiva-k, line 1051: What is the requirement for adding this? Application itself is an entity category.

lashinijay: `providers.Application` is built for agents too.  So `ctx.Application` in an agent's OBO login is assembled from an agent entity.                                                                                                                                
                                                                                                                                                                                                                 
  That leaves the two cases indistinguishable on flow events:                                                                                                                                                    
                                                                                                                                                                                                                 
  | | Normal login | Agent OBO login |                                                                                                                                                                           
  |---|---|---|                                                                                                                                                                                                  
  | `sub_type` | user | user |                                                                                                                                                                                   
  | `app_id` | a UUID | a UUID |                                                                                                                                                                                 
  | `EntityCategory` | `app` | `agent` |                                                                                                                                                                         
                                                                                                                                                                                                                 
`EntityCategory` is the only field that                                                                                                            separates them.                                                                                                                                                                                            
                                                                                                                                                                                                                 
  `assembleApplication` already reads the entity, so this just stops the category being dropped during                                                                                                           
  the copy rather than re-fetching it per event. Runtime-only (`yaml:"-" json:"-"`), never persisted,                                                                                                            
  absent from the API and declarative resources.

thiva-k:
Let's add integration tests covering the updated scenarios

backend/pkg/thunderidengine/providers/model.go
thiva-k: We may need to rename this to `Client` or `Actor` in flows, since this is used by agents too now. Hope this is tracked @ThaminduDilshan

backend/internal/flow/flowexec/engine_test.go
```diff
@@ -4487,3 +4488,292 @@ func (s *EngineTestSuite) TestHandleCompletedResponse_EndNodeWithFrameStack() {
...
+	evt := (*captured)[0]
+	if got, ok := evt.Data[event.DataKey.SubjectType]; ok {
+		t.Errorf("subject_type should be omitted when the category is unknown, got %v", got)
+	}
+	assertEventData(t, evt, map[string]interface{}{event.DataKey.Subject: "user-999"})
+}
+
+// The entity vocabulary spells an application "app" while the reported principal type spells it
+// "application", matching the token's sub_type claim. The events must carry the claim's spelling.
+func TestFlowStartedEvent_ApplicationReportsApplicationActorType(t *testing.T) {
```
thiva-k, line 4696: Shall we use `testify/suite` like other tests?

backend/internal/oauth/oauth2/model/token.go
```diff
@@ -56,11 +56,28 @@ type TokenDTO struct {
 	// TokenFamilyID is the token family id (tfid) stamped on the token, carried here so the refresh
 	// token issued alongside an access token can be stamped with the same family id.
 	TokenFamilyID string
+	// ActorSub is the resource ID of the principal acting for Subject, mirroring the token's act.sub
+	// claim. Set only for delegated (on-behalf-of) issuance; empty when the subject acts for itself.
+	ActorSub string
```
thiva-k, line 67: Shall we change this to ActorID to be consistent?

backend/internal/oauth/oauth2/tokenservice/builder.go
```diff
@@ -106,6 +115,58 @@ func (tb *tokenBuilder) BuildAccessToken(
...
+
+	// client_credentials: the subject is the authenticated client, already loaded.
+	if tokenCtx.OAuthApp != nil && tokenCtx.Subject == tokenCtx.OAuthApp.ID {
+		return tokenCtx.OAuthApp.ID, string(tokenCtx.OAuthApp.EntityCategory)
+	}
+
+	// Exchange grants: the subject arrives on a presented token, whose sub is a resource ID on most
+	// deployments but the mapped subject attribute where the issuing application configures one.
+	// Resolving it decides which: a mapped attribute matches no entity, so an unresolvable subject is
+	// reported as no subject rather than published as-is. This is what keeps a mapped attribute, which
```
thiva-k, line 148: Don't we need to do this for act claim too?

backend/internal/oauth/oauth2/tokenservice/builder.go
```diff
@@ -85,6 +90,10 @@ func (tb *tokenBuilder) BuildAccessToken(
 		ClaimsLocales:    tokenCtx.ClaimsLocales,
 		TokenFamilyID:    tokenCtx.TokenFamilyID,
 	}
+	if tokenCtx.ActorClaims != nil {
```
thiva-k, line 93: Are we setting the actor claim for token exchange when there is no actor token involved (like we do for auth code for agents)?

lashinijay: No, act claim will only be there in the cases where actor token in presented.

lashinijay: As discussed offline, this is fixed with this PR.

thiva-k: Shall we add this to id-jag grant too to be consistent?

sahandilshan (approved):

PR #4960, opened 2026-08-14
Fix MCP Client tab jumping to Customization when grant type changes

### LINKED ISSUE #4952: Tab automatically change in MCP Client apps
### Description

If we modify the MCP client types of `On its own behalf` to a use facing grant, the Advance tab automatically switched to `Customization` tab

### Steps to Reproduce

1. Go to application
2. Select MCP Client type
3. Choose the `On its own behalf` type. This will only have client credential grant enable
4. Now go to advance tab and try to switch the grant type to authorization code grant

https://github.com/user-attachments/assets/eb7e245c-5a94-41d2-8627-d20a793d4592

### Version

-

_No response_

sahandilshan:
## Purpose
Fixes #4952

## Description
When editing an MCP Client app set to "On its own behalf" (M2M / `client_credentials` only), the visible tabs are `[Overview, General, Token, Advanced]`. If the user navigates to the **Advanced** tab (index 3) and enables the `authorization_code` grant type, `isMcpM2mOnly` flips to `false`, causing the **Flows** and **Customization** tabs to become visible. The tab array becomes `[Overview, General, Flows, Customization, Token, Advanced]`, but `activeTab` stays at index 3 — which now points to **Customization** instead of **Advanced**.

### Root Cause
The `activeTab` state tracked the selected tab by **numeric index**. When the MCP tab list changes dynamically (tabs inserted/removed based on `isMcpM2mOnly`), the index no longer points to the intended tab.

### Fix
Introduced a separate `activeMcpTabKey` state that tracks the active MCP tab by its **string key** (e.g. `'advanced'`, `'overview'`) instead of by index. The numeric index is derived at render time via `findIndex`, so it always resolves to the correct tab regardless of how many tabs are visible.

Changes:
- Added `activeMcpTabKey` state (default `'overview'`)
- Added `handleMcpTabChange` that maps the `<Tabs>` numeric callback to a key
- Changed `safeActiveTab` from `Math.min(activeTab, ...)` to `findIndex` by key
- Updated `onGoToFlows`/`onGoToCustomization` callbacks to use key-based navigation
- Non-MCP (standard app) tabs remain unchanged — they use a fixed tab list with no dynamic show/hide

## Test plan
- [x] Reproduced the original bug: M2M MCP Client → Advanced tab → add `authorization_code` → tab jumps to Customization
- [x] Verified fix: same steps → Advanced tab stays selected, Flows and Customization appear in the tab bar
- [x] Verified reverse direction: remove `authorization_code` → Advanced tab stays, Flows/Customization disappear
- [x] Verified normal tab navigation (Overview, General, Token, Advanced) still works
- [x] Frontend CI gates pass: `format:check`, `typecheck`, `test` (6864 tests), `lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved application tab navigation for standard and MCP applications.
  * Ensured integration-guide links open the correct tab, including Flows and Customization.
  * Added safer handling when the selected tab is unavailable or hidden.
  * Standardized tab behavior for a more consistent editing experience.
  * Improved navigation from MCP overviews to relevant application sections.

### DIFF
--- a/frontend/apps/console/src/features/applications/pages/ApplicationEditPage.tsx
+++ b/frontend/apps/console/src/features/applications/pages/ApplicationEditPage.tsx
@@ -50,7 +50,7 @@ import getTemplateMetadata from '../utils/getTemplateMetadata';
 import isValidRedirectUriFormat from '../utils/isValidRedirectUriFormat';
 import {hasUserAccess} from '../utils/oauth2Rules';
 
-interface McpTabConfig {
+interface TabConfig {
   key: string;
   label: string;
   panel: React.ReactNode;
@@ -106,7 +106,7 @@ export default function ApplicationEditPage() {
   const justCreatedSecret = (location.state as {justCreatedSecret?: JustCreatedSecret} | null)?.justCreatedSecret;
   const [secretDialogOpen, setSecretDialogOpen] = useState(Boolean(justCreatedSecret));
 
-  const [activeTab, setActiveTab] = useState(0);
+  const [activeTabKey, setActiveTabKey] = useState('overview');
   const [editedApp, setEditedApp] = useState<Partial<Application>>({});
   // Bumped on Save/Reset to force AccessSection/McpAccessSection/UrlsSection to remount with a
   // clean form — they keep local state (redirect URI list, react-hook-form defaults) that a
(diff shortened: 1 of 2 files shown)


frontend/apps/console/src/features/applications/pages/ApplicationEditPage.tsx
```diff
@@ -389,7 +390,20 @@ export default function ApplicationEditPage() {
       ]
     : [];
 
-  const safeActiveTab = mcpTabs.length > 0 ? Math.min(activeTab, mcpTabs.length - 1) : 0;
+  const handleMcpTabChange = (_event: SyntheticEvent, newValue: number) => {
```
brionmario: Shall we check whether we can have a generic solution for this? Add a tab id and then push to that instead to an index?

sahandilshan: This is fixed with the latest update

brionmario (dismissed):

brionmario (approved):

PR #4974, opened 2026-08-14
Stop the console asking a generic OIDC issuer for a ThunderID profile

bhagyasakalanka:
## What

A Control Plane console federated to a generic OIDC issuer repeatedly called:

```
https://api.asgardeo.io/t/b1zt/oauth2/token/users/me
```

`/users/me` is a ThunderID endpoint (`backend/internal/user/init.go:158`). The SDK addresses it relative to the single `baseUrl` it is given, and `withConfig` passes the trusted issuer as that base, so the path is glued onto the issuer's token endpoint.

Under the hybrid model the issuer and the server were the same host, so one base URL served both and this never showed. Federating the Control Plane to an external issuer separated them.

The call is also pointless, not just misaddressed: whoever signs in exists at the issuer, not in this deployment, so there is no profile here to read. The SDK already falls back to the ID token claims, which is where those details legitimately come from.

`withConfig` already suppresses the other ThunderID-specific bootstrap call (`{baseUrl}/flow/meta`) for exactly this reason. This adds the profile call to the same branch.

## Note

The SDK offers no separate API base URL, so this cannot be fixed by pointing resource calls at the server. Worth raising against `@thunderid/react` separately: a federated console genuinely has two hosts, and only the SDK can express that.

## Testing

- `tsc -b` clean, lint clean.
- `withConfig` tests updated for the added preference: 50 passed.

## Checklist
- [x] Tested locally
- [x] No new dependencies
- [x] No CI changes


bhagyasakalanka:
Folded into #4902, which now carries all of the Control Plane work as a single commit.

PR #4991, opened 2026-08-15
Require id_token_hint to be an ID token at the logout endpoint

### LINKED ISSUE #4990: Restrict id_token_hint to be an ID token at the logout endpoint
### Description

`GET/POST /oauth2/logout` accepts any server-signed JWT as `id_token_hint` provided its issuer matches and its audience resolves to a registered client. There is no check that the token is actually an ID token.

An access token issued to an application with no configured default audience carries `aud=client_id`, since `ResolveDefaultAudience` falls back to the client id, so it satisfies both conditions. Expiry is intentionally not enforced for `id_token_hint` per OIDC RP-Initiated Logout, so an expired access token is accepted as well.

The sign-out flow suppresses the End-User confirmation whenever a hint is present, so a party holding any access token for the target client can suppress it. That confirmation is the mitigation for forced logout via cross-site navigation. Not currently exploitable because the default sign-out flow ships with `promptOnSignOut: false`, but it must be fixed before that mitigation is relied on.

### Steps to Reproduce

1. Register a confidential client with the `authorization_code` grant and no `defaultAudience` on its access token config.
2. Complete an authorization code flow and keep the `access_token`.
3. Call `GET /oauth2/logout?id_token_hint=<access_token>`.
4. The request is accepted and the client resolves from the access token's `aud`. Setting `promptOnSignOut: true` on the sign-out flow's `SessionSignOutExecutor` node shows the confirmation is skipped, where a request with no hint would prompt.

### Version

1.0.0

_No response_

thiva-k:
### Purpose
`clientIDFromIDTokenHint` verified only the signature and issuer of `id_token_hint`, with no token-class check. An access token issued to an application with no configured default audience carries `aud=client_id`, because `ResolveDefaultAudience` falls back to the client id, so such a token resolved to a valid client and was accepted as a hint. Expiry is intentionally not enforced per OIDC RP-Initiated Logout, so an expired access token worked too.

Prompt suppression keys on the hint being present, so anyone holding an access token for the target client could suppress the End-User sign-out confirmation. That confirmation is the mitigation for forced logout via cross-site navigation, so it must not be bypassable before deployments rely on it.

### Approach
The hint must now be an ID token. Its `typ` header must be the generic `JWT`, which rejects `at+jwt` access tokens, and an `access_token_sub` claim rejects the refresh tokens that share that `typ`. This mirrors the discrimination `ValidateIDJAGSubjectToken` already applies for the same problem.

The refresh-token check is defence in depth: a refresh token's `aud` is the issuer, so it normally fails client resolution, but `client_id` is caller-supplied on the application management API and nothing prevents registering a client whose id equals the issuer.

Signature and issuer verification are unchanged, and expiry stays unenforced so a legitimately expired ID token is still accepted.

### Related Issues
- Fixes #<issue>

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logout security by rejecting access tokens and refresh tokens when provided as ID token hints.
  * Added validation to ensure ID token hints use the correct JWT type before processing.
* **Tests**
  * Added coverage for invalid token types and refresh-token claims.

### DIFF
--- a/backend/internal/oauth/oauth2/logout/service.go
+++ b/backend/internal/oauth/oauth2/logout/service.go
@@ -228,8 +228,22 @@ func (s *logoutService) clientIDFromIDTokenHint(ctx context.Context, idTokenHint
 	if svcErr := s.jwtService.VerifyJWTSignature(ctx, idTokenHint); svcErr != nil {
 		return "", errInvalidIDTokenHint
 	}
+	header, err := jwt.DecodeJWTHeader(idTokenHint)
+	if err != nil {
+		return "", errInvalidIDTokenHint
+	}
+	// The hint must be an ID token, not any other JWT this server signs. An access token issued to an
+	// application with no configured default audience carries aud=client_id, so without this check it
+	// would resolve to a client and be accepted as a hint, which suppresses the End-User sign-out
+	// confirmation.
+	if typ, _ := header["typ"].(string); typ != jwt.TokenTypeJWT {
+		return "", errInvalidIDTokenHint
+	}
 	payload, err := jwt.DecodeJWTPayload(idTokenHint)
 	if err != nil {
+		return "", errInvalidIDTokenHint
+	}
+	if _, isRefreshToken := payload[constants.ClaimAccessTokenSubject]; isRefreshToken {
 		return "", errInvalidIDTokenHint
 	}
 	if iss, _ := payload[constants.ClaimIss].(string); iss != s.issuer {
--- a/backend/internal/oauth/oauth2/logout/service_test.go
+++ b/backend/internal/oauth/oauth2/logout/service_test.go
@@ -300,6 +300,20 @@ func makeIDToken(iss, aud string) string {
 		enc(map[string]interface{}{"iss": iss, "aud": aud}) + ".sig"
 }
 
+// makeTypedToken builds a token with an arbitrary typ header and extra claims, for asserting that
+// only ID tokens are accepted as an id_token_hint.
+func makeTypedToken(typ, iss, aud string, extra map[string]interface{}) string {
+	enc := func(v interface{}) string {
+		b, _ := json.Marshal(v)
+		return base64.RawURLEncoding.EncodeToString(b)
+	}
+	claims := map[string]interface{}{"iss": iss, "aud": aud}
+	for k, v := range extra {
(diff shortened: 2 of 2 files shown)


backend/internal/oauth/oauth2/logout/service.go
```diff
@@ -228,10 +228,24 @@ func (s *logoutService) clientIDFromIDTokenHint(ctx context.Context, idTokenHint
...
+	// confirmation.
+	if typ, _ := header["typ"].(string); typ != jwt.TokenTypeJWT {
+		return "", errInvalidIDTokenHint
+	}
 	payload, err := jwt.DecodeJWTPayload(idTokenHint)
 	if err != nil {
 		return "", errInvalidIDTokenHint
 	}
+	if _, isRefreshToken := payload[constants.ClaimAccessTokenSubject]; isRefreshToken {
+		return "", errInvalidIDTokenHint
```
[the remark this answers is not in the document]
thiva-k: This is the expected behaviour

backend/internal/oauth/oauth2/logout/service.go
```diff
@@ -228,10 +228,24 @@ func (s *logoutService) clientIDFromIDTokenHint(ctx context.Context, idTokenHint
...
 	}
+	header, err := jwt.DecodeJWTHeader(idTokenHint)
+	if err != nil {
+		return "", errInvalidIDTokenHint
+	}
+	// The hint must be an ID token, not any other JWT this server signs. An access token issued to an
+	// application with no configured default audience carries aud=client_id, so without this check it
+	// would resolve to a client and be accepted as a hint, which suppresses the End-User sign-out
+	// confirmation.
+	if typ, _ := header["typ"].(string); typ != jwt.TokenTypeJWT {
```
ThaminduDilshan, line 239: I believe we should add integration tests for these scenarios. Fine to do with a followup PR

ThaminduDilshan (approved):

backend/internal/oauth/oauth2/logout/service.go
thiva-k: Will address in a follow-up for related scenarios

PR #5029, opened 2026-08-17
Validate flow definitions from static executor metadata

bhagyasakalanka:
### Purpose

Flow definition validation asks the executor registry two questions: whether a node names a registered executor (`IsRegistered`), and what that executor's metadata allows (`GetExecutorMeta`). Both answers are static.

The only way to hold a registry, though, was to construct every executor, and that links `internal/flow/executor` together with the data plane's authentication, notification and session packages. The Control Plane paid that cost purely to validate flows it never runs. It also had to build four federated-auth services it never executes, because three executor constructors type-assert their auth service and `panic` on nil:

```go
oauthSvcCast, ok := authService.(authnoauth.OAuthAuthnCoreServiceInterface)
if !ok {
    panic("failed to cast GithubOAuthAuthnService to OAuthAuthnCoreServiceInterface")
}
```

That coupling was documented in place as `TEMPORARY COUPLING` in `cmd/cpserver/servicemanager.go`. This removes it.

### Approach

**One definition of the metadata, not two.** `executormeta` now holds a catalog of every built-in executor's `ExecutorMeta`, keyed by name. The executors read their own metadata from that catalog, so it is the single definition rather than a copy that can fall behind.

**A registry that constructs nothing.** `executormeta.NewRegistry` answers `IsRegistered` and `GetExecutorMeta` from the catalog. It honors the same configured subset as the runtime registry (`flow.executors`), so a flow accepted where it is authored is one the plane that runs it has actually registered, and an unknown name is an error there too.

**A narrower dependency.** `flow/mgt` and `graphbuilder` now depend on `core.ExecutorMetadataProvider` instead of `executor.ExecutorRegistryInterface`. The full registry satisfies that interface structurally, so the Data Plane and the hybrid server are wired exactly as before and behave identically.

**The result, which is the point of the change:**

```
$ go list -deps ./cmd/cpserver | grep -E "internal/authn/(oauth|oidc|google|github)$|internal/flow/executor$"
(nothing)

$ go list -deps ./cmd/dpserver | grep -cE "internal/authn/(oauth|oidc|google|github)$|internal/flow/executor$"
5
```

**Guarding against drift.** Two descriptions of the same thing drift, so `TestCatalogMatchesRegistry` registers the real executors through the real flow factory and compares every one against the catalog. Change an executor's metadata without changing the catalog and it fails, naming the executor. The catalog itself was generated from that same live registry rather than transcribed by hand.

No new validation is introduced, and no validation is removed.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### DIFF
--- a/backend/cmd/cpserver/servicemanager.go
+++ b/backend/cmd/cpserver/servicemanager.go
@@ -35,10 +35,6 @@ import (
 
 	"github.com/thunder-id/thunderid/internal/agent"
 	"github.com/thunder-id/thunderid/internal/application"
-	"github.com/thunder-id/thunderid/internal/authn/github"
-	"github.com/thunder-id/thunderid/internal/authn/google"
-	authnOAuth "github.com/thunder-id/thunderid/internal/authn/oauth"
-	authnOIDC "github.com/thunder-id/thunderid/internal/authn/oidc"
 	"github.com/thunder-id/thunderid/internal/cert"
 	"github.com/thunder-id/thunderid/internal/connection"
 	"github.com/thunder-id/thunderid/internal/dataplanetoken"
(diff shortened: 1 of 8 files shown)


bhagyasakalanka:
Folded into #4902 as its own commit, so the Control Plane work reviews and merges as one change.

PR #5058, opened 2026-08-18
Resolve inbound auth profile handles in agent update process

### LINKED ISSUE #5056: Re-importing an Agent Resource without an AuthFlowID clears its Authentication Flow
### Description

Re-importing (upsert path) an agent resource clears its `AuthFlowID`, breaking any interactive `/oauth2/authorize` request for that agent's client with a `server_error`.

The agent `create` path (`agentService.ValidateAgent`) resolves `authFlowHandle` → `AuthFlowID` via `inboundClientService.ResolveInboundAuthProfileHandles` before persisting the inbound client record. The agent `update` path (`agentService.UpdateAgent` → `reconcileInboundForUpdate` in `backend/internal/agent/service.go`) does not call this resolver. It builds the inbound client record straight from `req.AuthFlowID`, which is empty whenever the request only sets `authFlowHandle` (the normal case for declarative YAML imports that reference a flow by handle rather than by raw ID).

Effect: the first import of an agent resolves and persists the flow correctly. Any subsequent `upsert: true` import of the same agent (e.g. re-running a config import, which the importer routes through `UpdateAgent` once the entity exists) overwrites `AUTH_FLOW_ID` in `INBOUND_CLIENT` with an empty string. From then on, any authorization-code request for that agent's `client_id` fails: the flow engine can't resolve an authentication flow for the entity, and the authorize endpoint returns an error

The application domain does not have this bug. `applicationService`'s shared `validateApplicationFields` helper calls `ResolveInboundAuthProfileHandles` for both create and update, which is the pattern the agent update path should follow.

### Steps to Reproduce

1. Import a config bundle that declares an `agent` resource with `authFlowHandle` set (and no `authFlowId`) like the wayfinder agent.
2. Confirm the agent has authentication flow set.
3. Re-import the same bundle. (Upsert behaviour)
4. Check the Flow tab of the Agent Details page. The authentication flow is empty.

### Version

1.0.0

_No response_

Sadeesha-Sath:
### Purpose

Agent update requests that only supply `authFlowHandle` (e.g. a declarative re-import) were not resolved to an `AuthFlowID` before building the inbound client record, unlike agent creation which resolves handles via `ValidateAgent`. This caused the previously-resolved `AuthFlowID` (and `RegistrationFlowID`) to be silently wiped out on update whenever only the handle was set.

### Approach

`reconcileInboundForUpdate` now builds an `InboundAuthProfile` from the request and calls `ResolveInboundAuthProfileHandles` before constructing the inbound client record, mirroring the resolution already done for creates. FK resolution errors are translated via `translateInboundClientFKError`; any other error is logged and surfaced as an internal server error. Added unit tests covering successful handle resolution and the non-FK error path.

### Related Issues
- Fixes #5056 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent updates now correctly associate inbound authentication and registration flows using configured handles.
  * Improved error handling for unresolved or unexpected flow-resolution failures.
  * Explicit flow IDs take precedence, while handle-only updates preserve existing associations when needed.

* **Documentation**
  * Documented authentication and registration flow handles for agent configurations.

* **Tests**
  * Added coverage for agent updates, imports, flow switching, precedence, and error scenarios.

### DIFF
--- a/backend/internal/agent/service.go
+++ b/backend/internal/agent/service.go
@@ -944,7 +944,25 @@ func (s *agentService) reconcileInboundForUpdate(ctx context.Context, agentID st
 		return inboundmodel.InboundClient{}, nil, nil
 	}
 
-	client := buildInboundClientRecord(agentID, req.AuthFlowID, req.RegistrationFlowID,
+	// Resolve flow handles to IDs when the direct IDs are absent, same as ValidateAgent does for
+	// create - a declarative update (e.g. a re-import) that only sets authFlowHandle must not wipe
+	// out the previously-resolved AuthFlowID.
+	profile := providers.InboundAuthProfile{
+		AuthFlowID:             req.AuthFlowID,
+		AuthFlowHandle:         req.AuthFlowHandle,
+		RegistrationFlowID:     req.RegistrationFlowID,
+		RegistrationFlowHandle: req.RegistrationFlowHandle,
+	}
+	if err := s.inboundClientService.ResolveInboundAuthProfileHandles(ctx, &profile); err != nil {
+		if svcErr := translateInboundClientFKError(err); svcErr != nil {
+			return inboundmodel.InboundClient{}, nil, svcErr
+		}
+		s.logger.Error(ctx, "Failed to resolve inbound auth profile handles",
+			log.Error(err), log.String("agentID", agentID))
+		return inboundmodel.InboundClient{}, nil, &tidcommon.InternalServerError
+	}
+
+	client := buildInboundClientRecord(agentID, profile.AuthFlowID, profile.RegistrationFlowID,
 		req.IsRegistrationFlowEnabled, req.ThemeID, req.LayoutID, req.Assertion,
 		req.LoginConsent, req.AllowedUserTypes, nil)
 	setLogoProperty(&client, req.LogoURL)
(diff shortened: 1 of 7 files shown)


backend/internal/agent/service.go
```diff
@@ -944,7 +944,25 @@ func (s *agentService) reconcileInboundForUpdate(ctx context.Context, agentID st
 		return inboundmodel.InboundClient{}, nil, nil
 	}
 
-	client := buildInboundClientRecord(agentID, req.AuthFlowID, req.RegistrationFlowID,
+	// Resolve flow handles to IDs when the direct IDs are absent, same as ValidateAgent does for
+	// create - a declarative update (e.g. a re-import) that only sets authFlowHandle must not wipe
+	// out the previously-resolved AuthFlowID.
+	profile := providers.InboundAuthProfile{
```
thiva-k, line 950: Will this resolving logic run if we only send the auth flow handle instead of the id?

Sadeesha-Sath: Yes. If the AuthFlowID/RegistrationFlowID is there the ResolveInboundAuthProfileHandles does nothing to change it. It only maps the handles to a ID if the IDs are not there in the request but the handles are there.

thiva-k: updateNeedsInboundClient doesn't check the handle fields right? if so a handle-only request returns early before reaching this block

thiva-k:
Shall we also add integration tests for this behavior?

backend/internal/agent/service.go
Sadeesha-Sath: Yes. The test I used had an InboundAuthConfig (Wayfinder Agent), so it has bypassed this. I'll add the handles to the check

Sadeesha-Sath:
Added Integration Tests

thiva-k (approved):

PR #5085, opened 2026-08-19
added handling of direct jwt response from authnprovider

### LINKED ISSUE #5084: Handle opaque _jwt pseudo-claim returned by authn providers

Today, when an authn provider's backing identity system returns a signed/encrypted token (JWT/JWE), the provider itself decodes that token and hands ThunderID the individual extracted claims. We're changing that: going forward, the provider will stop decoding the token and will instead pass it through as-is, under a single pseudo-claim key, `_jwt`.

ThunderID currently has no handling for this `_jwt` passthrough key. If the provider's attribute map contains only `_jwt`, it is treated like an ordinary claim name and silently dropped everywhere, because it never matches a scope mapping, an allow-list entry, a consent element, or a required input.

Net effect if unaddressed:
- id_token/userinfo carry zero user claims for these users.
- Consent and attribute-collection treat every essential/optional claim as "missing" (since the only claim present is the opaque `_jwt`), which can force re-collection of attributes the identity system already verified, or block consent outright.

Primary: Integrators/implementers of authn providers whose backing identity system issues a signed or encrypted token rather than discrete claims (e.g. the MOSIP/esignet KYC-exchange flow, which is moving from decode-and-forward to raw passthrough).
- Goal: Have ThunderID correctly pass the opaque token through to the userinfo endpoint and route around claim-level filtering that no longer applies.
- Pain: Without this change, these users would get an empty userinfo response and could be repeatedly re-prompted for consent/attribute collection despite already completing KYC with the identity system.

- Value to integrators: unblocks authn providers that are moving to return an opaque JWT/JWE instead of pre-decoded discrete claims (the MOSIP/esignet integration is making this change).
- Value to org: avoids a broken OIDC contract (empty userinfo, spurious re-prompts) once providers stop decoding and start passing the raw token through.

### Proposed Solution

Introduce a shared `providers.RawJWTAttributeKey = "_jwt"` convention. When an authn provider's attribute map contains this key (as its only key), treat the value as an opaque JWT/JWE and route it through as a passthrough rather than through claim-level filtering/validation:

- **Attribute resolution**: pass `_jwt` through to the attribute cache / assertion, bypassing the requested-attributes allow-list (`auth_assert_executor.go`, `resolveUserAttributes`).
- **Token service**: never allow-list-filter `_jwt` out of fetched attributes (`tokenservice.FetchUserAttributes`).
- **UserInfo**: detect JWE (5-part compact serialization) vs JWS (3-part) and either pass the token through as-is, or encrypt it into a nested JWT only when the client's UserInfo config requires encryption (`userinfo/service.go`); never re-sign or decode it.
- **Consent**: skip the profile-presence filter (so essential/optional claims are still evaluated and prompted normally) when only `_jwt` is available (`consent/service.go`, `buildUserAttributeSet`).
- **Attribute collector**: treat `_jwt` presence as "required inputs already satisfied," skipping prompts/re-collection (`attribute_collector.go`, `HasRequiredInputs`).

ThunderID's own authn provider is unaffected — it never emits `_jwt`; this only changes generic engine/OAuth consumer code.

### Alternatives

- Keep decoding the JWT/JWE in the authn provider and forwarding individual claims (the old behavior). Rejected: the provider would need to hold verification/decryption keys and understand the token format just to re-expose claims ThunderID would then re-filter anyway; passing the token through opaquely and letting the relying party/client handle it is simpler.

sacrana0:
### Purpose
Authn providers whose backing identity system issues a signed/encrypted token (e.g. the MOSIP/esignet KYC-exchange flow) are moving from decoding that token into discrete claims to passing it through as-is under a single pseudo-claim key, _jwt. ThunderID had no handling for this — it was silently dropped everywhere, leaving id_token/userinfo empty and forcing spurious re-consent/attribute-collection.

### Approach
Introduced providers.RawJWTAttributeKey = "_jwt", routed as an opaque passthrough instead of through claim-level filtering, in auth_assert_executor.resolveUserAttributes, tokenservice.FetchUserAttributes, userinfo/service.go (JWE vs JWS detection, encrypt-only-if-required), and consent/service.buildUserAttributeSet.

### Related Issues
- #5084 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/backend/internal/flow/executor/auth_assert_executor.go
+++ b/backend/internal/flow/executor/auth_assert_executor.go
@@ -441,6 +441,12 @@ func (a *authAssertExecutor) resolveUserAttributes(
 	fetchedAttributes map[string]interface{},
 	userID, userType, ouID string,
 ) (map[string]interface{}, error) {
+	// An opaque JWT/JWE from the authn provider is passed through as-is, bypassing the
+	// requested-attributes allow-list: it isn't an individual claim to filter, it's the whole payload.
+	if rawJWT, ok := fetchedAttributes[providers.RawJWTAttributeKey]; ok {
+		return map[string]interface{}{providers.RawJWTAttributeKey: rawJWT}, nil
+	}
+
 	if len(requestedAttributes) == 0 {
 		return nil, nil
 	}
--- a/backend/internal/oauth/oauth2/userinfo/service.go
+++ b/backend/internal/oauth/oauth2/userinfo/service.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"slices"
+	"strings"
 
 	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
 
@@ -183,16 +184,22 @@ func (s *userInfoService) buildResponseFromClaims(
 		return nil, &tidcommon.InternalServerError
 	}
 
-	response, svcErr := s.buildUserInfoResponse(ctx, sub, scopes, userAttributes, oauthApp, tokenClaims)
-	if svcErr != nil {
-		return nil, svcErr
-	}
-
 	var userInfoCfg *providers.UserInfoConfig
 	var certificate *providers.Certificate
 	if oauthApp != nil {
 		userInfoCfg = oauthApp.UserInfo
 		certificate = oauthApp.Certificate
+	}
+
+	// The authn provider returned an opaque JWT/JWE instead of individual claims: pass it
+	// through as-is rather than running it through the claims/scope/allow-list pipeline.
+	if rawToken, ok := userAttributes[providers.RawJWTAttributeKey].(string); ok && rawToken != "" {
+		return s.buildRawJWTResponse(ctx, rawToken, userInfoCfg, certificate)
+	}
+
+	response, svcErr := s.buildUserInfoResponse(ctx, sub, scopes, userAttributes, oauthApp, tokenClaims)
+	if svcErr != nil {
+		return nil, svcErr
 	}
 
 	responseType := providers.UserInfoResponseTypeJSON
@@ -260,14 +267,30 @@ func (s *userInfoService) generateNestedJWTUserInfo(
 		return nil, svcErr
 	}
 
-	rpKey, rpKID, svcErr := s.jwksResolver.ResolveEncryptionKey(
-		ctx, certificate, cfg.EncryptionAlg, jwksresolver.KeyUseStrictEnc)
+	compact, svcErr := s.encryptSignedJWT(ctx, jwsResp.JWTBody, cfg, certificate)
 	if svcErr != nil {
 		return nil, svcErr
 	}
 
+	return &UserInfoResponse{Type: providers.UserInfoResponseTypeNESTEDJWT, JWTBody: compact}, nil
+}
+
+// encryptSignedJWT encrypts an already-signed JWT into a nested-JWT (JWE with cty=JWT) compact
+// serialization, using the client's encryption key.
+func (s *userInfoService) encryptSignedJWT(
+	ctx context.Context,
+	signedJWT string,
+	cfg *providers.UserInfoConfig,
+	certificate *providers.Certificate,
+) (string, *tidcommon.ServiceError) {
+	rpKey, rpKID, svcErr := s.jwksResolver.ResolveEncryptionKey(
+		ctx, certificate, cfg.EncryptionAlg, jwksresolver.KeyUseStrictEnc)
+	if svcErr != nil {
+		return "", svcErr
+	}
+
 	compact, svcErr := s.jweService.Encrypt(ctx,
-		[]byte(jwsResp.JWTBody),
+		[]byte(signedJWT),
 		&providers.KeyRef{PublicKeyJWK: rpKey},
 		cfg.EncryptionAlg,
 		jwe.ContentEncAlgorithm(cfg.EncryptionEnc),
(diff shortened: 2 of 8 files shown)


backend/internal/flow/executor/attribute_collector.go
```diff
@@ -130,6 +130,15 @@ func (a *attributeCollector) HasRequiredInputs(ctx *providers.NodeContext,
 	execResp.AuthUser = authUser
 
 	if authnUserAttrs != nil && len(authnUserAttrs.Attributes) > 0 {
+		// An opaque JWT/JWE from the identity system carries no individual attribute names to
+		// match against required inputs; treat them as delivered opaquely and stop prompting.
+		if _, ok := authnUserAttrs.Attributes[providers.RawJWTAttributeKey]; ok {
```
ThaminduDilshan: IMO this is not required for attribute collector. Anyway I feel like this use case will not work with attribute collector as we cannot know what attributes being returned there.

Let's remove this

backend/internal/authn/consent/service.go
```diff
@@ -524,6 +524,12 @@ func buildUserAttributeSet(available *providers.AttributesResponse) map[string]b
 		return nil
 	}
 
+	// An opaque JWT/JWE from the identity system carries no individual claim names, so the
+	// profile-presence filter cannot be applied against it: skip filtering entirely.
+	if _, ok := available.Attributes[providers.RawJWTAttributeKey]; ok {
```
ThaminduDilshan: Won't this lead to skipping consent prompt? In our default provider, we're skipping consent if the user doesn't have any applicable attributes. So here, if it's jwt, we're returning a empty set that leads consent service into thinking there are no user attributes to take consent.

In addition to that, I feel like we don't need to handle this in the default consent provider.

sacrana0: // Skip elements not present in the user profile
		if len(userAttributeSet) > 0 && !userAttributeSet[elem.Name] {
			continue
		}
		
If you see these lines, if the filtering only happens when userAttributeSet is not empty. So the claims will not be skipped.

@ThaminduDilshan if _jwt handling is not needed I can remove it.

https://github.com/thunder-id/thunderid/blob/bceae5bbc1add5bc13f3b1e1b6337920253e610e/backend/internal/authn/consent/service.go#L588-L591

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -441,6 +441,12 @@ func (a *authAssertExecutor) resolveUserAttributes(
 	fetchedAttributes map[string]interface{},
 	userID, userType, ouID string,
 ) (map[string]interface{}, error) {
+	// An opaque JWT/JWE from the identity system is passed through as-is, bypassing the
```
ThumulaPerera: From ThunderID perspective, the term `identity system` is meaningless. Shall we reword the comment to say `authn provider`?

ThumulaPerera: Check other places as well

backend/internal/authn/consent/service.go
ThaminduDilshan: So when the user attributes set is empty, above continue statement will hit and will skip adding the element to the consent purpose. At end of the loop, an empty consent purpose will be returned and hence no consent will be collected from the user. @coderabbitai can you validate?

```go
// Skip elements not in the request's essential/optional sets
		if !slices.Contains(essentialAttributes, elem.Name) &&
			!slices.Contains(optionalAttributes, elem.Name) {
			continue
		}

		// Skip elements not present in the user profile
		if len(userAttributeSet) > 0 && !userAttributeSet[elem.Name] {
			continue
		}

		// Skip elements that already have active consent
		key := purposeElementKey(purpose.Name, elem.Name)
		if consentedElements[key] {
			continue
		}

		// Classify the element as essential or optional for prompting
		if slices.Contains(essentialAttributes, elem.Name) {
			essential = append(essential, providers.PromptElement{Name: elem.Name})
		} else {
			optional = append(optional, providers.PromptElement{Name: elem.Name})
		}
```

backend/internal/flow/executor/auth_assert_executor.go
sacrana0: Resolved in all places.

backend/internal/authn/consent/service.go
ThaminduDilshan: I feel like this is a buggy code. If the intention is to skip consent collection for non existing user attributes, if even none of the attributes exists, we shouldn't collect consent for all. If the strategy is to skip, we should skip for both nil and empty case.
@coderabbitai if you agree this is a bug, can you create a git issue?

@sacrana0 anyway we don't need to touch default consent service for this customization. Hence let's remove this from here

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -331,6 +354,37 @@ func (s *userInfoService) generateJWSUserInfo(
 	}, nil
 }
 
+// buildRawJWTResponse passes through an opaque JWT/JWE returned by the identity system in place
```
ThaminduDilshan:
```suggestion
// buildRawJWTResponse passes through an opaque JWT/JWE returned by the authn provider in place
```

backend/internal/authn/consent/service.go
sacrana0: All changes removed from consent package @ThaminduDilshan

ThaminduDilshan (approved):

PR #5090, opened 2026-08-19
Pick the tool release bump and npm dist-tag from the release branch

brionmario:
### Purpose

Two problems in the tool release path.

**1. `pnpm publish` was called without `--tag`.** npm assigns the `latest` dist-tag to any publish that omits it, **including one whose version carries a prerelease suffix**. So a prerelease published from `main` would have been handed to every existing `npx thunderid` user immediately. A version suffix on its own protects nobody; the `--tag` is the whole mechanism.

**2. `bump_type` was a free-form input.** Nothing stopped a `minor` or `major` bump from a branch that owns a released patch line, and `workflow_dispatch` has no ref restriction, so a release could be cut from any branch.

### Approach

The bump type is not really a choice, so it is no longer an input. It follows from the branch being released:

| Branch | Bump | dist-tag |
|---|---|---|
| `main` | `prerelease` | `next` |
| `N.M.x` | `patch` | `latest` |
| anything else | refused | |

So `1.0.x` can only ever patch the 1.0 line, and `main` can only ever publish prereleases, held off `latest` so existing users are not moved onto them. Trying the development line stays an explicit opt-in via `npx thunderid@next`.

Removing the input also removes the reason for `confirm_major` and the `validate` job that policed it, which is why the workflow is net shorter.

`release-npm` is shared with other packages, so `bump-type` keeps working exactly as before. It gains one optional `dist-tag` input defaulting to `latest`, so no other caller changes behaviour.

### Verification

Branch matching was checked against `main`, `1.0.x`, `1.1.x`, `2.0.x`, `10.24.x` (all resolve as intended) and against `e2e-cli`, `release/1.0`, `1.0.0`, `1.0.0.x`, `1.0.x-hotfix` (all refused).

`pnpm version` output confirmed directly:

- `patch` from `1.0.2` gives `1.0.3`, then `1.0.4`.
- `prerelease` from `1.1.0-m1.0` gives `1.1.0-m1.1`, then `1.1.0-m1.2`.

### One thing to decide before the first release from main

`tools/npx/package.json` on `main` currently reads `1.0.2`, a released stable version. `pnpm version prerelease` from there produces **`1.0.3-0`**, which reads as a prerelease of the very `1.0.3` that `1.0.x` is going to publish. Not an npm collision, but confusing.

The fix is a one-time manual set of `tools/npx/package.json` on `main` to whatever the next line should be, for example `1.1.0-m1.0`, after which `prerelease` increments it cleanly. Deliberately not done in this PR because naming the next line is a product call, not a plumbing one.

### Related Issues
- N/A

### Related PRs
- #5087 (merged) reconciled `tools/npx/package.json` on `1.0.x` with the already-published 1.0.2. Without it a patch release from that branch computed `1.0.1`, which is already on npm. With it merged, `1.0.x` now resolves to `1.0.3`.

### Checklist
- [x] Manual test round performed and verified. Branch resolution and `pnpm version` behaviour verified as described above. The workflow itself can only be exercised by a real `workflow_dispatch`.
- [ ] Tests provided. (Add links if there are any) The change is a branch-to-channel mapping inside a `workflow_dispatch` job, which the repo has no harness for.
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/.github/actions/release-npm/action.yml
+++ b/.github/actions/release-npm/action.yml
@@ -14,8 +14,12 @@ inputs:
     description: 'Path to the package from the repo root (e.g. tools/npx, frontend/packages/eslint-plugin)'
     required: true
   bump-type:
-    description: 'Semver bump type: major, minor, or patch'
+    description: 'Semver bump type: major, minor, patch, or prerelease'
     required: true
+  dist-tag:
+    description: 'npm dist-tag to publish under. Defaults to latest, which is what existing consumers install.'
+    required: false
+    default: 'latest'
   git-user-name:
     description: 'Git committer name'
     required: false
@@ -106,4 +110,7 @@ runs:
       shell: bash
       run: |
         cd ${{ inputs.package-dir }}
-        pnpm publish --no-git-checks --access public --provenance
+        # --tag is load-bearing: npm assigns the 'latest' dist-tag to any publish that
+        # omits it, including a prerelease version, which would move every default
+        # consumer onto it.
+        pnpm publish --no-git-checks --access public --provenance --tag "${{ inputs.dist-tag }}"
--- a/.github/workflows/release-tools.yml
+++ b/.github/workflows/release-tools.yml
@@ -2,25 +2,13 @@
 
 name: 🔧 Release Tools
 
+# The bump type is not an input. It follows from the branch: a maintenance branch owns
+# a released line and can only patch it, and main is ahead of every released line so it
+# can only publish prereleases.
+
 on:
   workflow_dispatch:
     inputs:
-      bump_type:
-        description: 'Version bump type'
-        required: true
-        type: choice
-        options:
-          - patch
-          - minor
-          - major
-        default: patch
-
-      confirm_major:
-        description: 'Check to confirm intentional major bump (required when bump_type is major)'
-        required: false
-        type: boolean
-        default: false
-
       npx:
         description: 'Release npx tool'
         required: false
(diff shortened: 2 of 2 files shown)


brionmario:
Rewrote this considerably smaller after feedback that it was over-engineered. The diff went from 326 insertions across 4 files to 36 across 2, and it is now net-negative on the workflow.

What went away:

- `scripts/resolve-tool-release-version.sh` and its test suite are deleted. The `version.txt` parsing, line pinning, branch/`version.txt` agreement checks and npm/tag preflight are all gone.
- `bump_type` and `confirm_major` are no longer inputs, and the `validate` job that policed them is removed. The bump type was never really a choice, so the branch decides it: `main` publishes a `prerelease` under `next`, an `N.M.x` branch publishes a `patch` under `latest`, anything else is refused.

What stayed is the part that was actually load-bearing: `pnpm publish` now passes `--tag`. npm assigns `latest` to any publish that omits it, prerelease suffix or not, so without that flag a prerelease from `main` still lands on every existing `npx thunderid` user.

The three findings from the automated review were all in the deleted script, so they no longer apply rather than having been individually addressed:

- prerelease identifier validation: the script no longer parses prerelease identifiers.
- prerelease version regression: version arithmetic is `pnpm version` again, not hand-rolled.
- remote checks failing open: the npm and `git ls-remote` preflight is removed, so there is no longer a check to fail open. Worth noting the underlying observation is still true of the pre-existing flow, which commits and pushes a tag before `pnpm publish`, so a failed publish leaves a tag behind and blocks a retry. That predates this PR and is untouched by it.

One open item, called out in the description: `tools/npx/package.json` on `main` reads `1.0.2`, so `pnpm version prerelease` produces `1.0.3-0`, which reads as a prerelease of the `1.0.3` that `1.0.x` will publish. It needs a one-time manual set to whatever the next line should be. Naming that line is a product call so it is deliberately not in here.

brionmario:
Closing this as redundant in favour of #5092.

`forward-port-pr.yml` already forward-ports merged `1.0.x` PRs to `main` automatically, so landing #5092 will produce a `[1.0.x → main]` PR carrying this same commit. The two PRs are the same change: both are 2 files, +36/-31, and both produce byte-identical `release-tools.yml` and `release-npm/action.yml`. The single file works for both branches because the mapping is evaluated at run time from `github.ref_name`, giving `prerelease`/`next` on `main` and `patch`/`latest` on `1.0.x`.

Keeping both open would also be counterproductive. If this one merged first, the forward-port of #5092 would cherry-pick a change `main` already has, and that workflow runs with `--allow-empty --keep-redundant-commits` and commits conflict markers on failure, so it would yield either an empty commit or a draft PR needing manual cleanup.

Review feedback from this PR is carried over: the three findings were all in `scripts/resolve-tool-release-version.sh`, which no longer exists in the simplified change on #5092.

PR #5097, opened 2026-08-19
Apply the configured log format to every log record

UdeshAthukorala:
### Purpose

Setting `log.output.file.format: json` did not apply to all log output. A user reported that some records stayed in the default text format.

The root cause is that `Logger.With()` snapshots the underlying `slog.Logger`, while `Configure()`, the only place that applies the format, replaced it on the singleton. Any logger derived before `Configure()` ran stayed pinned to the boot text handler on stdout forever. Around 470 call sites cache exactly such a logger in their constructor, for example `logger: log.GetLogger().With(log.String(log.LoggerKeyComponentName, "XService"))`.

`cmd/server/main.go` was unaffected by chance, because it configures the logger before registering services. The embedded engine called `Configure()` at the end of `New()`, after every service had been built, so nearly all of its output stayed text even with `Format: "json"`.

Separately, several streams bypassed the configured handler regardless of ordering:

- go-redis wrote pool diagnostics to stderr as plain text, since `redis.SetLogger` was never called.
- OpenTelemetry export failures, its internal logger, and the gRPC transport behind the OTLP exporter all defaulted to stderr.
- The stdout span exporter used `WithPrettyPrint()`, emitting multi line JSON that breaks a line oriented log pipeline.
- Rotation and pruning failures in the rolling file writer were written to stderr as raw text.

### Approach

**Derived loggers now follow the configured handler.** A `Logger` no longer owns an output handler. It records the ordered `WithAttrs` and `WithGroup` derivations applied to it and resolves them against a shared, atomically swapped root. `Configure()` swaps one pointer, so every logger picks up the new output no matter when it was derived, and the ordering of `Configure()` relative to service construction stops mattering. The replayed chain is cached per root state, keeping the hot path at two atomic loads and zero allocations per record.

Ordering is preserved exactly, since slog treats `WithAttrs` and `WithGroup` call order as significant. Sibling derivations copy the slice rather than appending in place, so two loggers derived from the same parent cannot clobber each other.

**The engine configures logging before it builds any service**, so its own boot logging is emitted at the configured level and format.

**Third party diagnostics are routed through the framework logger.** go-redis via `redis.SetLogger` where the clients are constructed, OpenTelemetry via `otel.SetErrorHandler` and `otel.SetLogger`, and the gRPC transport via `grpclog.SetLoggerV2`. gRPC's informational chatter maps to debug, matching what its own default logger discards, so only real warnings surface at the default level. `WithPrettyPrint()` was dropped from the stdout span exporter.

**Rolling file housekeeping failures** are reported through a new `Config.OnError` sink that formats them in the configured format. They stay on stderr and never go through the file writer, because logging a file failure through the writer that just failed would recurse.

Two modules move from indirect to direct in `backend/go.mod`, `github.com/go-logr/logr` and `google.golang.org/grpc`, both required by the logger interfaces above. `go.sum` is unchanged, as both were already in the module graph.

Startup lines emitted before the configuration is read remain in the default text format, because the format itself lives in the file being loaded. This is now documented.

**The embedded engine no longer replaces the host's output configuration.** `engine.New()`
applied `LogConfig.Format` by calling `Configure()` with a console only `OutputOptions`.
`Configure()` rebuilds the sink from the options it is handed and closes the previous file
writer, so passing no file options did three things at once: it discarded a host
application's file sink, closed that host's file writer, and forced console output back on
even when the host had disabled it. A new `Logger.SetFormat()` rebuilds the handler from
the sink already recorded on the shared root, changing the format and nothing else, and
the engine now calls it. The format to handler mapping is centralised in one
`newFormatHandler` helper so `Configure()`, `SetFormat()` and the rolling file error sink
cannot drift apart.

### Issues, Root Causes and Fixes

A user reported log lines still arriving in text format. Investigation separated five
distinct causes. Only the third was a defect in this repository.

| Symptom | Root cause | Resolution |
| --- | --- | --- |
| Flow debug lines absent | Records are emitted at `Debug`; the default level is `info` | Configuration. Set `log.level: debug` |
| Flow lines in text, not JSON | The format defaults to text when unset. Server: `log.output.file.format` omitted. Engine: `LogConfig.Format` empty, so `New()` skips the format change and the boot text handler from `initLogger` stays in place | Configuration. Set the format explicitly. Text remains the documented default |
| A host application's file output stops at `engine.New()` and everything reverts to stdout | `engine.New()` called `Configure()` with a console only `OutputOptions`, rebuilding the sink and closing the host's file writer | **Code fix in this PR.** `engine.New()` now calls `Logger.SetFormat()` |
| `sql: unknown driver ""` from `DBProvider` | `Database.Config.Type` never populated, so the driver name is the empty string | Configuration. Not reachable from `deployment.yaml`: `config/default.json` supplies `sqlite` and `mergeStructs` only overrides a default when the user value is non zero, so this condition is specific to an under configured embedded engine |
| Interleaved JSON lines carrying `appName: esignet` | A separate process. `@timestamp`, `@version` and `level_value` are a Java logback layout | Out of scope. Configure that application's own logback |

### Verification

Unit tests, in addition to those already covering the derived logger behaviour:

- `TestSetFormatPreservesFileOutput` writes to a file sink, changes the format, and asserts
  the first record is text, the second is JSON, and both are in the same file.
- `TestSetFormatDoesNotCloseFileWriter` asserts the writer pointer is unchanged, guarding
  the specific mechanism that broke the embedded path.
- `TestSetFormatAppliesToLoggerDerivedBeforeCall` asserts the swap reaches loggers captured
  in constructors.
- `TestSetFormatRejectsInvalidFormats` covers the empty and unsupported cases, and
  case insensitive matching.

Manual verification against a running server, `level: debug`, `format: json`:

- Console: 133 JSON records, zero text records after JSON begins. The single preceding text
  line is the documented pre configuration exception.
- File: 132 JSON records, zero non JSON lines.
- Nine distinct components present, all as JSON records.
- Access log records are JSON and carry a populated correlation ID.
- With a Redis cache pointed at a closed port, 19 go-redis diagnostics were emitted as JSON
  and nothing leaked to stderr as raw text.
- With the format omitted, output is text and contains no JSON, confirming the documented
  default is intact.
- Size based rotation at a 0.01 MB threshold produced backups capped at `max_backups`.

`go build ./...`, the tests for `internal/system/log/...` and `pkg/thunderidengine/...`, and
`golangci-lint` over both trees all pass.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Unified application, Redis, OpenTelemetry, and gRPC diagnostics under the configured logging format, level, and output.
  - Logger changes now apply consistently to existing and newly created components.
  - Log rotation errors can be routed through a configured error handler.

- **Bug Fixes**
  - Improved logging consistency during startup and runtime reconfiguration.
  - Structured JSON output is now compact and consistent across components.

- **Documentation**
  - Time-based rotation is enabled by default, retaining up to 20 rotated files.
  - Documented logging behavior and early-startup exceptions.

### DIFF
--- a/backend/internal/system/log/log.go
+++ b/backend/internal/system/log/log.go
@@ -7,12 +7,14 @@ package log
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	stdlog "log"
 	"log/slog"
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/thunder-id/thunderid/internal/system/constants"
 	sysContext "github.com/thunder-id/thunderid/internal/system/context"
@@ -26,9 +28,8 @@ var (
 
 // Logger is a wrapper around the slog logger.
 type Logger struct {
-	internal   *slog.Logger
-	levelVar   *slog.LevelVar
-	fileWriter *rollingfile.Writer
+	internal *slog.Logger
+	root     *root
 }
 
 // OutputOptions describes where and how the logger writes. It is a log-package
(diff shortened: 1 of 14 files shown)


backend/internal/system/log/log.go
```diff
@@ -266,6 +446,26 @@ func NewServerErrorLog(logger *Logger) *stdlog.Logger {
 	return stdlog.New(&serverErrorWriter{logger: logger}, "", 0)
 }
 
+// RedisLogger adapts go-redis diagnostics into the framework logger. It satisfies
+// go-redis's logging interface structurally, so its internal package never has to be
+// imported: pass the value NewRedisLogger returns straight to redis.SetLogger.
```
KaveeshaPiumini, line 451:
```suggestion
// It satisfies
// go-redis's logging interface structurally, so its internal package never has to be
// imported: pass the value NewRedisLogger returns straight to redis.SetLogger.
```

are these comments intentional ?

PR #5099, opened 2026-08-19
Provide flexibility to extend allowed oauth configs via pluggable engine

### LINKED ISSUE #5063: [Improvement]: Extend OIDC token/userinfo claim handling: additional scopes, removable roles claim, configurable sub, and custom claims support
### Description
Thunder ID engine owns token issuance (token_endpoint) and claims assembly (userinfo_endpoint) for eSignet's OIDC flows. Following a discussion on evolving eSignet's openid-configuration (well-known), four related changes were identified that all require updates to Thunder's scope-to-claims mapping, since that's where these values are actually resolved and returned.

### Current Limitation

- Scopes: Thunder supports a fixed set of scopes (profile, email, phone, roles) with no supported way to add a new scope and its underlying claim mapping without further code changes.
- Roles claim: The roles scope is currently always included, with no way to disable or remove it.
- sub claim: The subject identifier generation strategy is fixed (not configurable) — there is no option to switch between subject types (e.g., pairwise vs. public) or customize the generation algorithm per deployment.
- Custom claims: There is no general mechanism to include claims beyond the standard/fixed claim set in the ID token or userinfo response — the claim source/resolution layer only handles a hardcoded list.

### Suggested Improvement

- Additional scopes: Add a configurable scope-to-claims mapping so new scopes can be introduced (and their underlying attributes resolved) without a code change per scope, mirroring the flexibility of eSignet's old mosip.esignet.openid.scope.claims config model.
- Removing roles: Make the roles claim inclusion configurable/removable — either via a config flag or by allowing it to be excluded from the scope-to-claims mapping.
- Configurable sub: Add a config option to select the subject identifier strategy (e.g., pairwise vs. public) and/or a pluggable generation algorithm, applied at token-issuance time.
- Custom claims: Extend the claim source/resolution layer to accept an arbitrary, config- or plugin-driven claim-to-attribute mapping, so deployments can expose claims beyond the standard OIDC set in the ID token and userinfo response.

### Acceptance Criteria

- A new scope can be added via configuration and its mapped claims appear correctly in the ID token/userinfo response, without additional code changes.
- The roles claim can be disabled via configuration and no longer appears in the well-knowns response when not configured.
- The sub claim generation strategy can be switched via configuration (e.g., pairwise vs. public) and produces the expected value for each mode.
- A custom claim, configured via the new mechanism, should be used for filtering or identifying the supported claim set.
- Existing default behavior (current scopes, current roles behavior, current sub strategy) is preserved when no new configuration is set, to avoid breaking existing deployments.

ThaminduDilshan:
### Purpose
Introduces the plumbing for [#5063](https://github.com/thunder-id/thunderid/issues/5063): expose the OIDC discovery metadata (supported scopes, claims, subject types, and the default scope→claims mapping) as engine-only configuration. Deployments embedding ThunderID via `thunderidengine` can now override what the well-known endpoint advertises without touching `deployment.yaml` / `default.json`.

### Approach
- Split `OAuthConfig` at the yaml boundary. `internal/system/config` now has its own `OAuthConfig` that mirrors every yaml/json-loaded field of `engineconfig.OAuthConfig` and nothing else. A `ToEngineConfig()` converter copies fields into an `engineconfig.OAuthConfig` value. This structurally prevents `deployment.yaml` / `default.json` from setting the new engine-only fields, because they do not exist on the yaml-loaded struct.
- Add four engine-only fields on `engineconfig.OAuthConfig`: `AllowedScopes`, `AllowedClaims`, `DefaultScopeClaimsMapping`, `AllowedSubjectTypes`. Named to match the existing `AllowedGrantTypes` / `AllowedResponseTypes` / `AllowedAuthMethods` fields already on the struct. Engine integrators populate them via the existing `WithOAuthConfig(...)` option, no new engine options.
- `oauthconfig.FromServerRuntime()` calls `ToEngineConfig()` and seeds the four fields (plus `AllowedGrantTypes` / `AllowedResponseTypes` / `AllowedAuthMethods` when empty) with the built-in defaults derived from `constants.StandardOIDCScopes`, `providers.SupportedGrantTypes`, etc. This keeps stock Thunder behavior byte-identical while giving engine embedders full control.
- Simplify the discovery service: the three `getAllowed*` getters (and the two `getAllowedScopes` / `getAllowedSubjectTypes`) now return the configured slices directly. All fallback logic moved into `applyOIDCDefaults`.
- Remove the now-redundant `constants.GetSupportedResponseTypes` / `GetSupportedGrantTypes` / `GetSupportedTokenEndpointAuthMethods` / `GetSupportedSubjectTypes` helpers along with their tests. Callers under `application/tools.go` and the discovery service read the config directly. This also breaks the previous `constants` → `oauthconfig` import cycle that appeared once `oauthconfig` needed to reach into `constants.StandardOIDCScopes`.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5063

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - OAuth and OpenID Connect discovery metadata now reflects configured allowed scopes, claims, subject types, grant types, response types, and authentication methods.
  - Added standard OIDC defaults when these settings are not explicitly configured.
  - OAuth configuration now supports scope-to-claim mappings and broader OIDC customization.

- **Bug Fixes**
  - Preserved explicitly configured OAuth/OIDC allowed-value lists while preventing unintended overrides.
  - Improved consistency between runtime OAuth settings and published discovery metadata.

### DIFF
--- a/backend/internal/oauth/config/config.go
+++ b/backend/internal/oauth/config/config.go
@@ -5,8 +5,10 @@
 package oauthconfig
 
 import (
+	"github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
 	"github.com/thunder-id/thunderid/internal/system/config"
 	engineconfig "github.com/thunder-id/thunderid/pkg/thunderidengine/config"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
 // Config holds configuration values required by OAuth services.
@@ -19,15 +21,91 @@ type Config struct {
 	GateClient             engineconfig.GateClientConfig
 }
 
-// FromServerRuntime builds OAuth configuration from the global server runtime.
+// FromServerRuntime builds OAuth configuration from the global server runtime, seeding the
+// additional OIDC fields with the server defaults.
 func FromServerRuntime() Config {
 	runtime := config.GetServerRuntime()
+	oauth := runtime.Config.OAuth.ToEngineConfig()
+	applyOIDCDefaults(&oauth)
+
 	return Config{
 		DeploymentID:           runtime.Config.Server.Identifier,
 		RuntimeTransientDBType: runtime.Config.Database.RuntimeTransient.Type,
 		BaseURL:                config.GetServerURL(&runtime.Config.Server),
 		JWT:                    runtime.Config.JWT,
-		OAuth:                  runtime.Config.OAuth,
+		OAuth:                  oauth,
 		GateClient:             runtime.Config.GateClient,
 	}
+}
+
+// applyOIDCDefaults seeds the default values for the additional OIDC fields on the given OAuthConfig.
+func applyOIDCDefaults(oauth *engineconfig.OAuthConfig) {
+	mapping := make(map[string][]string, len(constants.StandardOIDCScopes))
+	scopes := make([]string, 0, len(constants.StandardOIDCScopes))
+
+	claimSet := make(map[string]struct{})
+	for _, c := range constants.GetStandardClaims() {
+		claimSet[c] = struct{}{}
+	}
+	for scope, def := range constants.StandardOIDCScopes {
+		claims := make([]string, len(def.Claims))
+		copy(claims, def.Claims)
+		mapping[scope] = claims
+		scopes = append(scopes, scope)
+		for _, c := range def.Claims {
+			claimSet[c] = struct{}{}
+		}
+	}
+
+	claims := make([]string, 0, len(claimSet))
+	for c := range claimSet {
+		claims = append(claims, c)
+	}
+
+	oauth.DefaultScopeClaimsMapping = mapping
+	oauth.AllowedScopes = scopes
+	oauth.AllowedClaims = claims
+
+	oauth.AllowedSubjectTypes = defaultAllowedSubjectTypes()
+	if len(oauth.AllowedGrantTypes) == 0 {
+		oauth.AllowedGrantTypes = defaultAllowedGrantTypes()
+	}
+	if len(oauth.AllowedResponseTypes) == 0 {
+		oauth.AllowedResponseTypes = defaultAllowedResponseTypes()
+	}
+	if len(oauth.AllowedAuthMethods) == 0 {
+		oauth.AllowedAuthMethods = defaultAllowedAuthMethods()
+	}
+}
+
+// defaultAllowedSubjectTypes returns the default allowed OIDC subject types for the server.
+func defaultAllowedSubjectTypes() []string {
+	return []string{constants.SubjectTypePublic}
+}
+
+// defaultAllowedGrantTypes returns the default allowed grant types for the server.
+func defaultAllowedGrantTypes() []string {
+	result := make([]string, len(providers.SupportedGrantTypes))
+	for i, v := range providers.SupportedGrantTypes {
+		result[i] = string(v)
+	}
+	return result
+}
+
+// defaultAllowedResponseTypes returns the default allowed response types for the server.
+func defaultAllowedResponseTypes() []string {
+	result := make([]string, len(providers.SupportedResponseTypes))
+	for i, v := range providers.SupportedResponseTypes {
+		result[i] = string(v)
+	}
+	return result
+}
+
+// defaultAllowedAuthMethods returns the default allowed token endpoint authentication methods for the server.
+func defaultAllowedAuthMethods() []string {
+	result := make([]string, len(providers.SupportedTokenEndpointAuthMethods))
+	for i, v := range providers.SupportedTokenEndpointAuthMethods {
+		result[i] = string(v)
+	}
+	return result
 }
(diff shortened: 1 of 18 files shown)


ThaminduDilshan:
@coderabbitai review

backend/internal/oauth/config/config.go
```diff
@@ -19,15 +21,91 @@ type Config struct {
...
+	claims := make([]string, 0, len(claimSet))
+	for c := range claimSet {
+		claims = append(claims, c)
+	}
+
+	oauth.DefaultScopeClaimsMapping = mapping
+	oauth.AllowedScopes = scopes
+	oauth.AllowedClaims = claims
+
+	oauth.AllowedSubjectTypes = defaultAllowedSubjectTypes()
```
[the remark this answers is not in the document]
ThaminduDilshan: This isn't correct.

Only default ThunderID distribution uses the server runtime path (via cmd/servicemanager.go) and calls the applyOIDCDefaults() function. Here the input comes from `runtime.Config.OAuth.ToEngineConfig()`. The yaml-loaded system/config.OAuthConfig deliberately does not carry those four fields, that's the enforcement design of this PR. So the input is always zero for those fields, and unconditional assignment is correct.

Apart from the default distribution, thunderid also provides a library where third parties can build their own solution importing the code in `pkg`. It builds configs directly by calling `WithXXX` provider methods and never calls applyOIDCDefaults(). Engine embedders' values are preserved because this code isn't touched.

Can you validate this finding again?

backend/internal/system/config/config_test.go
```diff
@@ -1622,3 +1622,59 @@ func (suite *ConfigTestSuite) TestNotificationConfig_Validate_DelegatesToOTP() {
...
+	// yaml-loaded struct. With KnownFields(true) the decoder would error; here we just
+	// assert that if the yaml keys are absent the produced value matches a zero OAuthConfig
+	// (and ToEngineConfig therefore emits zero OIDC fields).
+	src := OAuthConfig{}
+	dst := src.ToEngineConfig()
+	assert.Empty(suite.T(), dst.AllowedScopes)
+	assert.Empty(suite.T(), dst.AllowedClaims)
+	assert.Empty(suite.T(), dst.DefaultScopeClaimsMapping)
+	assert.Empty(suite.T(), dst.AllowedSubjectTypes)
+}
```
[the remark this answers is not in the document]
ThaminduDilshan: Fixed

backend/internal/system/config/config.go
```diff
@@ -571,6 +571,52 @@ type LogTimeRotationConfig struct {
 	IntervalDays *int  `yaml:"interval_days" json:"interval_days"`
 }
 
+// OAuthConfig is the yaml/json-loaded OAuth section for the server. It mirrors every field
+// of engineconfig.OAuthConfig except the configs which are settable only through thunderidengine.
+type OAuthConfig struct {
```
thiva-k, line 576: Do we need this duplicate config? This may diverge and cause maintainability issues right? Can't we use the engine config only as before and maybe not read the non-overridable configs from deployment.yaml

ThaminduDilshan: No it will not work right? Then it will be configurable through the deployment.yaml, but a undocumented feature.
We don't want to expose this in our configurations, this is only for the lib embedders

backend/pkg/thunderidengine/config/config.go
```diff
@@ -252,6 +252,14 @@ type OAuthConfig struct {
...
 	// AllowedAuthMethods lists allowed client token endpoint auth methods
 	AllowedAuthMethods []string `yaml:"allowed_auth_methods" json:"allowed_auth_methods"`
+	// AllowedScopes lists the OAuth scopes advertised as allowed by the server.
+	AllowedScopes []string
+	// AllowedClaims lists the claims advertised as allowed by the server.
+	AllowedClaims []string
+	// DefaultScopeClaimsMapping maps each allowed scope to the claims it implies.
+	DefaultScopeClaimsMapping map[string][]string
+	// AllowedSubjectTypes lists the OIDC subject types advertised as allowed by the server.
+	AllowedSubjectTypes []string
```
anushasunkada: The four new OAuthConfig fields (AllowedScopes, AllowedClaims, DefaultScopeClaimsMapping, AllowedSubjectTypes) have no yaml/json struct tags

backend/internal/system/config/config.go
thiva-k: Seems ``yaml:"-" json:"-"`` will ignore parsing the specific field

ThaminduDilshan: Have to duplicate the model due to https://github.com/thunder-id/thunderid/pull/5099#discussion_r3819726366

backend/pkg/thunderidengine/config/config.go
ThaminduDilshan: Added

thiva-k (approved):

Issue #3089, opened 2026-06-02
[SCIM] Support SCIM v2.0 Server Protocol for Automated User Provisioning


ravindu439:
### What problem are we solving?

When an administrator wants to automatically and continuously synchronize user accounts and group memberships from external Identity Providers and HR platforms into ThunderID, they currently have to hand-craft custom synchronization scripts or resort to manual administrative configuration because ThunderID lack support for the industry-standard SCIM protocol, which causes significant security risks (under/over provisioning, delayed account deactivation), high maintenance costs, and delay in user onboarding.

### Who are we solving this for?

Primary: Enterprise Identity Administrators
- Goal: Streamline user lifecycle management (joiners, movers, leavers) by automating directory synchronization.
- Pain: High manual overhead, delayed deprovisioning, and lacks standard protocol endpoints (SCIM) compatible out-of-the-box with major external identity providers.

Secondary: Integration Engineers / Developers
- Goal: Connect software tools and user directories to ThunderID quickly and reliably.
- Pain: Forced to build and maintain proprietary synchronization glue-code using ThunderID's unique API rather than leverage standard protocol interfaces.

### Why should we solve this now?

Why now:
- Value to Persona: Secures organizational access directories. Eliminates account provisioning delays, automatically terminates permissions on employee departure, and guarantees seamless integration with leading directory portals.
- Value to Org: Eliminates administrative costs of Custom API integrations. Elevates our position to be a corporate-ready Identity Provider, as SCIM v2.0 is a mandatory checkbox requirement for major SaaS procurement.

### Proposed Solution

Implement a lightweight SCIM v2.0 (RFC 7643 and RFC 7644) server wrapper on top of existing entity, user, and group services. 

### Key Technical Aspects:
1. **Core Translation Wrapper:** Translate incoming SCIM query and mutation contracts on standard endpoints (`/scim/v2/Users` and `/scim/v2/Groups`) to existing backend calls.
2. **Two-Tier Attribute Mapping:**
   - **Tier 1 (Explicit & Configurable):** Map flat local attributes to hierarchical SCIM properties (e.g. `empID` ↔ `userName`, `firstname` ↔ `name.givenName`, `workEmail` ↔ `emails[type eq "work"].value`) using a configurable mapping stored inside the `SystemAttributes` schema JSON.
[...]

PR #5110, opened 2026-08-20
feat(scim): Add SCIM 2.0 provisioning support (Users, Groups, discovery)

ravindu439:
### Purpose
Adds SCIM 2.0 provisioning support to ThunderID, implementing RFC 7643 (Core Schema) and RFC 7644 (Protocol) so external identity providers and provisioning clients (Okta, Azure AD, Entra ID, etc.) can create, read, update, and delete Users and Groups against `/scim/v2` using the standard SCIM wire format.

scim connection drawio (1)

### Approach

**New package: `backend/internal/scim`**
Standalone SCIM layer sitting on top of the existing user, group, and entitype (user-type schema) services — it does not replace them, it translates between the SCIM wire protocol and ThunderID's native resource model. Organized as one subpackage per concern: `common` (shared types, errors, filter parsing, attribute-rule tables), `config`, `discovery`, `users`, `groups`.

**Discovery endpoints** (`discovery/handler.go`, `discovery/service.go`, `discovery/schema_builder.go`)
- `GET /ServiceProviderConfig` — advertises supported capabilities (patch, bulk, filter, sort, pagination), public, no auth.
- `GET /Schemas`, `GET /Schemas/{id}` — returns the static core `User`/`Group` schemas (RFC 7643 §4.1/§4.2) plus one dynamically generated extension schema per registered ThunderID user type, built from that type's attribute schema.
- `GET /ResourceTypes`, `GET /ResourceTypes/{id}` — advertises `User` and `Group` resource types; the `User` type's `schemaExtensions` array lists every registered user type's extension URN.

**Users** (`users/handler.go`, `users/service.go`, `users/resource.go`, `users/model.go`, `users/validator.go`, `users/attr_mapper.go`)
- `GET/POST /Users`, `GET/PUT/DELETE /Users/{id}`, `POST /Users/.search`, `GET/PUT /Me`.
- `PATCH /Users/{id}` returns `501 Not Implemented` per spec (not yet supported).
- Requests carry a SCIM Core User schema URN plus exactly one ThunderID extension URN (`urn:thunderid:params:scim:schemas:<userType>:2.0:User`) identifying the target user type — unless the server has exactly one configured user type, in which case the extension URN may be omitted.
- `users/attr_mapper.go` bidirectionally maps SCIM core fields (`userName`, `name.givenName/familyName`, `emails`, `phoneNumbers`, `addresses`, etc.) to/from ThunderID attribute names (`username`, `given_name`, `family_name`, `email`, ...), so a SCIM client's `userName` and a ThunderID admin's `username` attribute stay in sync.
- Core and Enterprise schema attributes are only accepted and returned for the designated core user type (`SCIMConfig.CoreUserTypeID`, or the sole configured user type when unset); sending either against a different target user type is rejected (`SCIM-1035` for Enterprise, `SCIM-1039` for core).
- Responses for the designated core user type merge the mapped core fields at the top level, with the Enterprise extension object (if applicable) alongside them; the custom extension-URN object carries only the remaining attributes that aren't already exposed via the core or Enterprise schema, so nothing is duplicated across schemas. Users of any other type return their custom extension schema only. Credential-typed attributes (passwords, etc.) are always stripped before the response leaves the server.
- Filtering supports `eq` comparisons optionally joined by `and` (`common/filter.go`); `or`, `not`, grouping, and any operator other than `eq` return `400 invalidFilter`. Sorting (`sortBy`/`sortOrder`) is not implemented and returns `400` if requested.
- `attributes`/`excludedAttributes` query params implement RFC 7644 §3.9 attribute projection on Users/Me; the two are mutually exclusive.

**Groups** (`groups/handler.go`, `groups/service.go`, `groups/resource.go`, `groups/model.go`)
- `GET/POST /Groups`, `GET/PUT/PATCH/DELETE /Groups/{id}`.
- `PATCH` (RFC 7644 §3.5.2) supports replacing `displayName`, adding/replacing/clearing the full `members` list, and removing a single member via `members[value eq "{id}"]`.
- Members may be of type `User` or `Group`.

**Cross-cutting**
- `common/response.go` — single `mapSCIMError` translator from internal `tidcommon.ServiceError` codes to SCIM HTTP status + `scimType` (RFC 7644 §3.12), so error-format decisions live in one place instead of being duplicated per handler. Internal ThunderID error codes are never sent to SCIM clients — only the standard `{schemas, status, scimType, detail}` shape.
- `common/error_constants.go` — one `SCIM-10xx` internal error per failure mode (missing schema URN, duplicate schemas, unknown user type, uniqueness conflict, mutability violation, invalid filter syntax, etc.), each carrying an i18n key + default message.
- `Content-Type: application/scim+json` is enforced on all write requests; wrong/missing Content-Type returns `400 invalidSyntax`.
- `/Bulk` and root `/.search` return `501 Not Implemented` rather than a generic 404, per spec.
- `config/config.go` — SCIM-specific server config (public URL for resource `location` fields, and `CoreUserTypeID`, which determines whether GET/POST/PUT responses include mapped core attributes for a given user type).

**Tests**
- Unit tests alongside every handler/service/resource/mapper file in `backend/internal/scim` (discovery, users, groups, filter parsing, core attribute mapping, error handling, schema building).
- Interface mocks for `SCIMDiscoveryServiceInterface`, `SCIMUsersServiceInterface`, `SCIMGroupsServiceInterface` under `backend/tests/mocks/discoverymock`, `backend/tests/mocks/usersmock`, and `backend/tests/mocks/groupsmock` respectively, generated via the project's mockery setup.
- Integration tests under `tests/integration/scim` covering discovery, Users, Groups, Me, search, `eq`/`and` filtering, and SCIM-specific authorization scoping (`scim_authz_test.go`).

### Related Issues
- #3089

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added SCIM v2.0 provisioning for users and groups, including CRUD, PATCH, search, and self-service `/Me` operations.
  * Added discovery endpoints for service capabilities, schemas, and resource types.
  * Added bearer authentication, pagination, filtering, attribute projection, and validation.
  * Added SCIM-standard errors, authorization controls, and unsupported-operation responses.
* **Documentation**
  * Added comprehensive OpenAPI documentation for the SCIM API.
* **Tests**
  * Added extensive coverage for users, groups, discovery, search, authentication, and authorization.


backend/.mockery.private.yml
```diff
@@ -415,6 +415,22 @@ packages:
...
+      pkgname: scim
+      filename: "{{.InterfaceName}}_mock_test.go"
+
+  github.com/thunder-id/thunderid/internal/scim:
+    config:
+      all: true
+      dir: internal/scim
+      structname: '{{.InterfaceName}}Mock'
+      pkgname: scim
+      filename: "{{.InterfaceName}}_mock_test.go"
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/config/config.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// Static SCIM protocol capability declarations.
+// These values reflect what this server implementation supports and are
+// not operator-configurable — they are facts about the codebase, not
+// deployment decisions.
+const (
+	// PatchSupported indicates that the SCIM PATCH operation is supported
+	// per RFC 7644 §3.5.2.
+	PatchSupported = false
```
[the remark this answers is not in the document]
ravindu439: since yet the users not supporting patch , lets keep this as false.

backend/internal/scim/scimusers/attr_mapper.go
```diff
@@ -0,0 +1,718 @@
...
+			}
+		case kindMultiComplex:
+			arr := normalizeToMultiComplex(val, rule.valueKey)
+			if rule.scimField == scimFieldAddresses {
+				arr = normalizeAndTranslateAddrOutbound(arr)
+			}
+			if len(arr) > 0 {
+				b, _ := json.Marshal(arr)
+				result[string(rule.scimField)] = b
+			}
```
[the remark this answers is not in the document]
ravindu439: instead of having mechanism to handle only addresses lets have , generalize function for all of the multi complex attributes.

backend/internal/scim/discovery_handler_test.go
```diff
@@ -0,0 +1,306 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/thunder-id/thunderid/internal/system/constants"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+const testBaseURL = "https://thunder.example.com"
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/groups_resource.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// thunderMemberTypeToSCIM maps Thunder member types to SCIM member types.
+// Thunder: "user"/"app"/"agent" → SCIM "User"; Thunder "group" → SCIM "Group".
+func thunderMemberTypeToSCIM(t group.MemberType) string {
+	if t == group.MemberTypeGroup {
+		return "Group"
+	}
+	return "User"
+}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/groups/resource.go
```diff
@@ -0,0 +1,82 @@
...
+// The ETag covers the group's DisplayName and Members list.
+func groupVersionState(g group.Group) any {
+	return struct {
+		DisplayName string
+		Members     []group.Member
+	}{
+		DisplayName: g.Name,
+		Members:     g.Members,
+	}
+}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,484 @@
...
+			return nil, mapGroupServiceErrorToSCIM(svcErr)
+		}
+	}
+	if len(thunderMembers) > 0 {
+		if _, svcErr = s.groupService.AddGroupMembers(ctx, groupID, thunderMembers); svcErr != nil {
+			logger.Error(ctx, "SCIM ReplaceGroup: failed to add new members",
+				log.String("groupID", groupID), log.Any("error", svcErr))
+			return nil, mapGroupServiceErrorToSCIM(svcErr)
+		}
+	}
```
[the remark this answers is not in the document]
ravindu439: To solve this we need to change the groups package, and need set atomic transaction for this .

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,484 @@
...
+			Name:        g.Name,
+			Description: g.Description,
+			OUID:        g.OUID,
+			OUHandle:    g.OUHandle,
+			IsReadOnly:  g.IsReadOnly,
+			Members:     members,
+		}
+		scimGroups = append(scimGroups, buildSCIMGroupResource(full, baseURL))
+	}
+	return buildSCIMGroupListResponse(scimGroups, listResp.TotalResults, startIndex, len(scimGroups)), nil
```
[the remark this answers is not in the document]
ravindu439: ListGroups currently calls fetchAllGroupMembers per group (N+1 calls to group.GroupServiceInterface.GetGroupMembers). To turn that into an actual batch lookup (not just moving the same loop behind a new name), we need a real multi-group fetch at the store layer — a single IN (...) query instead of one query per group.

That means:

- New method on groupStoreInterface (store.go) + SQL query with chunking, following the existing GetGroupsByIDs precedent (store.go:556, store_constants.go:327)
- Implementation in both backends: DB store (store.go) and file-based store (file_based_store.go), plus merge/dedup logic in composite_store.go (mirrors GetGroupsByIDs at composite_store.go:339)
- New method on GroupServiceInterface (service.go) wrapping the store call
- Mock regeneration (make mockery) for groupStoreInterface_mock_test.go and GroupServiceInterface_mock_test.go
- ListGroups in groups_service.go then calls this once with all group IDs from the page instead of looping

backend/internal/scim/init.go
```diff
@@ -0,0 +1,225 @@
...
+	mux.HandleFunc(middleware.WithCORS(
+		"OPTIONS "+SCIMBasePath+"/Groups",
+		func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusNoContent) },
+		optsCRUD,
+	))
+	mux.HandleFunc(middleware.WithCORS(
+		"OPTIONS "+SCIMBasePath+"/Groups/{id}",
+		func(w http.ResponseWriter, r *http.Request) { w.WriteHeader(http.StatusNoContent) },
+		optsCRUD,
+	))
```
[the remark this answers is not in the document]
ravindu439: lets keep this update for later.

backend/internal/scim/version.go
```diff
@@ -0,0 +1,61 @@
...
+	if ifMatch == "" || ifMatch == "*" {
+		return nil
+	}
+	for _, tag := range strings.Split(ifMatch, ",") {
+		if normalizeETag(tag) == normalizeETag(currentVersion) {
+			return nil
+		}
+	}
+	return &ErrorPreconditionFailed
+}
```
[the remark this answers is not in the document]
ravindu439: for this fix we need to touch other packages , outside of scim

backend/internal/scim/users/handler.go
```diff
@@ -0,0 +1,393 @@
...
+
+	if svcErr := validateSCIMContentType(r); svcErr != nil {
+		handleSCIMError(w, r, svcErr)
+		return
+	}
+	body, err := io.ReadAll(r.Body)
+	if err != nil || len(body) == 0 {
+		handleSCIMError(w, r, &ErrorInvalidRequestBody)
+		return
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

tests/integration/scim/scim_filter_test.go
```diff
@@ -0,0 +1,232 @@
...
+	urn, _, err := discoverExtensionSchema(ts.entityTypeName)
+	ts.Require().NoError(err, "failed to discover extension schema via GET /Schemas")
+	ts.extensionURN = urn
+
+	ts.engineeringUserID = ts.createFilterFixtureUser(
+		"scim.it.filter.jane@example.com", "Jane", "filter-Engineering", "Colombo")
+	ts.marketingUserID = ts.createFilterFixtureUser(
+		"scim.it.filter.marketing@example.com", "Alex", "filter-Sales and Marketing", "Galle")
+	ts.salesUserID = ts.createFilterFixtureUser(
+		"scim.it.filter.john@example.com", "John", "filter-Sales", "Kandy")
```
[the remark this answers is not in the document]
ravindu439: fixed

tests/integration/scim/users_test.go
```diff
@@ -0,0 +1,537 @@
...
+	var list scimUserListResponse
+	ts.Require().NoError(json.Unmarshal(listBody, &list))
+
+	ids := make([]string, 0, len(list.Resources))
+	for _, u := range list.Resources {
+		ids = append(ids, u.ID)
+	}
+	ts.NotContains(ids, orphanUserID, "user with unresolvable type should be omitted, not returned half-built")
+	ts.Contains(ids, goodUserID, "other users must still be listed")
+}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,484 @@
...
+	) (SCIMGroupListResponse, *tidcommon.ServiceError)
+	CreateGroup(ctx context.Context, displayName string, members []SCIMGroupMember,
+		baseURL string) (*SCIMGroup, *tidcommon.ServiceError)
+	GetGroup(ctx context.Context, groupID, baseURL string,
+	) (*SCIMGroup, *tidcommon.ServiceError)
+	ReplaceGroup(ctx context.Context, groupID, displayName string,
+		members []SCIMGroupMember, ifMatch, baseURL string) (*SCIMGroup, *tidcommon.ServiceError)
+	PatchGroup(ctx context.Context, groupID string, actions []SCIMGroupPatchAction,
+		ifMatch, baseURL string) (*SCIMGroup, *tidcommon.ServiceError)
+	DeleteGroup(ctx context.Context, groupID string, ifMatch string) *tidcommon.ServiceError
```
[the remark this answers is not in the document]
ravindu439: lets keep doc updates for later.

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,484 @@
...
+		switch action.Target {
+		case scimGroupPatchTargetDisplayName:
+			applyErr = s.applyDisplayNamePatch(ctx, groupID, g.OUID, g.Description, action)
+		case scimGroupPatchTargetMembers:
+			applyErr = s.applyMembersPatch(ctx, groupID, action)
+		}
+		if applyErr != nil {
+			return nil, applyErr
+		}
+	}
```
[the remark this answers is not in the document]
ravindu439: for these atomic operations we need to touch the other packages aswell.

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,484 @@
...
+			return nil, mapGroupServiceErrorToSCIM(svcErr)
+		}
+		g.Members = existingMembers
+		if svcErr := checkIfMatch(trimmed, generateVersion(groupVersionState(*g))); svcErr != nil {
+			return nil, svcErr
+		}
+	}
+
+	req := group.UpdateGroupRequest{Name: displayName, Description: g.Description, OUID: g.OUID}
+	_, svcErr = s.groupService.UpdateGroup(ctx, groupID, req)
```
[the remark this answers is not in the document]
ravindu439: need to touch other packages.

backend/internal/scim/scim_filter.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/thunder-id/thunderid/internal/system/database/utils"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+)
+
+var (
+	scimFilterQuotedStringRe = regexp.MustCompile(`"([^"\\]|\\.)*"`)
+	scimFilterEqRe           = regexp.MustCompile(`(?i)^((?:[A-Za-z0-9][A-Za-z0-9.\-_]*:)*)` +
+		`([A-Za-z][A-Za-z0-9.\-_]*)\s+eq\s+(.+)$`)
+	scimFilterAndRe           = regexp.MustCompile(`(?i)\s+and\s+`)
+	scimFilterUnsupportedOpRe = regexp.MustCompile(`(?i)\b(?:ne|co|sw|ew|pr|gt|lt|ge|le)\b`)
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/groups_service.go
```diff
@@ -0,0 +1,484 @@
...
+	thunderMembers, svcErr := scimMembersToThunder(members)
+	if svcErr != nil {
+		logger.Debug(ctx, "SCIM CreateGroup: invalid member type", log.Any("error", svcErr))
+		return nil, svcErr
+	}
+
+	req := group.CreateGroupRequest{
+		Name:    displayName,
+		OUID:    security.GetOUID(ctx),
+		Members: thunderMembers,
```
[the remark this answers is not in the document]
ravindu439: addressed in early fix.

backend/internal/scim/groups/handler.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// HandleGroupsListRequest handles GET /scim/v2/Groups.
+func (h *scimGroupsHandler) HandleGroupsListRequest(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	logger := log.GetLogger().With(log.String(log.LoggerKeyComponentName, groupHandlerLoggerComponentName))
+	if r.URL.Query().Get("filter") != "" {
+		handleSCIMError(w, r, &ErrorFilterNotSupported)
+		return
+	}
```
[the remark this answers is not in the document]
ravindu439: for now let's advertise filter support as true ,since user support filter eq oparations.

backend/internal/scim/scimusers/attr_mapper.go
```diff
@@ -0,0 +1,760 @@
...
+	}
+	for parent, obj := range parentObjs {
+		b, _ := json.Marshal(obj)
+		result[string(parent)] = b
+	}
+	for field, obj := range multiPartObjs {
+		obj["primary"], _ = json.Marshal(true)
+		b, _ := json.Marshal([]map[string]json.RawMessage{obj})
+		result[string(field)] = b
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/users/handler_test.go
```diff
@@ -0,0 +1,1348 @@
...
+	t.Run("MeReplace", func(t *testing.T) {
+		req := httptest.NewRequest(http.MethodPut, "/scim/v2/Me", oversizedBody)
+		req.Header.Set("Content-Type", constants.SCIMContentType)
+		authCtx := security.NewSecurityContextForTest("user-123", "", "", nil, nil)
+		req = req.WithContext(security.WithSecurityContextTest(req.Context(), authCtx))
+		rr := httptest.NewRecorder()
+		h.HandleMeReplaceRequest(rr, req)
+		require.Equal(t, http.StatusBadRequest, rr.Code)
+	})
+}
```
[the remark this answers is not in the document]
ravindu439: fixed

tests/integration/scim/users_test.go
```diff
@@ -0,0 +1,548 @@
...
+		var list scimUserListResponse
+		ts.Require().NoError(json.Unmarshal(listBody, &list))
+		for _, u := range list.Resources {
+			ids = append(ids, u.ID)
+		}
+		if len(list.Resources) == 0 || startIndex+scimPaginationMaxPageSize > list.TotalResults {
+			break
+		}
+		startIndex += scimPaginationMaxPageSize
+	}
```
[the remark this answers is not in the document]
ravindu439: fixed

backend/internal/scim/common/error_constants.go
```diff
@@ -0,0 +1,628 @@
...
+	ErrorUndeclaredCustomSchemaObject = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "SCIM-1033",
+		Error: tidcommon.I18nMessage{
+			Key:          "error.scim.undeclared_custom_schema_object",
+			DefaultValue: "Undeclared custom schema object",
+		},
+		ErrorDescription: tidcommon.I18nMessage{
+			Key: "error.scim.undeclared_custom_schema_object_description",
+			DefaultValue: "The request body includes a ThunderID extension object whose schema URN " +
```
ThaminduDilshan: Shall we avoid hard coding product name? Maybe we can refine this error to not include thunderid term

ravindu439: will fix it .

backend/internal/system/security/permissions.go
```diff
@@ -45,6 +45,11 @@ var publicPaths = []string{
 	"/auth/**",
 	"/register/passkey/**",
 	"/access/**",
+	// SCIM discovery endpoints exposed without authentication
+	// (ServiceProviderConfig, ResourceTypes).
+	"/scim/v2/ServiceProviderConfig",
+	"/scim/v2/ResourceTypes",
```
ThaminduDilshan: `/ResourceTypes` lists every registered user type's extension URN. Do we want that unauthenticated? `/Schemas` right below needs a token for the same names

backend/internal/system/security/permissions.go
```diff
@@ -248,6 +253,24 @@ func InitSystemPermissions(handle string) {
 		{"PUT /users/**", p.User},
 		{"DELETE /users/**", p.User},
 
+		// SCIM API.
+		{"GET /scim/v2/Me", ""},
+		{"PUT /scim/v2/Me", ""},
+		{"GET /scim/v2/Schemas", ""},
```
ThaminduDilshan, line 287: `GET /user-types` needs `system:usertype:view`. This returns the same attribute schemas to any authenticated user. Shall we gate it the same way?

backend/internal/system/security/permissions.go
```diff
@@ -248,6 +253,24 @@ func InitSystemPermissions(handle string) {
...
+		{"POST /scim/v2/Users", p.User},
+		{"POST /scim/v2/Users/.search", p.UserView},
+		{"GET /scim/v2/Users/**", p.UserView},
+		{"PUT /scim/v2/Users/**", p.User},
+		{"DELETE /scim/v2/Users/**", p.User},
+		{"GET /scim/v2/Groups", p.GroupView},
+		{"POST /scim/v2/Groups", p.Group},
+		{"GET /scim/v2/Groups/**", p.GroupView},
+		{"PUT /scim/v2/Groups/**", p.Group},
+		{"PATCH /scim/v2/Groups/**", p.Group},
```
ThaminduDilshan, line 301: `PATCH /scim/v2/Users/**` is missing here. It falls back to the root permission, so a SCIM client gets 403 instead of the 501 we register in `init.go`

backend/internal/scim/common/filter.go
```diff
@@ -0,0 +1,406 @@
...
+	prefix, attr, valid := splitSCIMAttrPath(attrTok.text)
+	if !valid {
+		return nil, fmt.Errorf("invalid filter expression; expected format: 'attrPath eq value'")
+	}
+	if p.rules.IsUnsupported(attr) {
+		return nil, fmt.Errorf("filtering on %q is not supported", attr)
+	}
+	// Custom/extension attributes must be qualified with their schema URN per RFC 7643 §3.10;
+	// core schema attributes may be referenced with or without a URN prefix.
+	if prefix == "" && !p.rules.IsCore(attr) {
```
ThaminduDilshan, line 310: The prefix is only checked for presence, never against a real schema. `filter=urn:x:y:2.0:User:password eq ".."` skips `IsCore`, `Translate` passes it through, and it reaches the store as an attribute predicate. `attributes=` rejects the same URN with SCIM-1034

backend/internal/scim/users/attr_mapper.go
```diff
@@ -0,0 +1,829 @@
...
+			valueKey := rule.ValueKey
+			if valueKey == "" {
+				valueKey = scimValueKey
+			}
+			set[strings.ToLower(string(rule.SCIMField)+"."+valueKey)] = struct{}{}
+		case scim.KindSubAttr, scim.KindMultiComplexPart:
+			set[strings.ToLower(string(rule.ParentField)+"."+rule.SubAttr)] = struct{}{}
+		}
+	}
+	set["id"] = struct{}{}
```
ThaminduDilshan, line 90: Filters are applied as JSON conditions on the `ATTRIBUTES` column, never on `ID`. So `filter=id eq ".."` always returns 0 results

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+		// Self-service replace: type and OU can't change (enforced above), so this
+		// goes through the same attribute-only update path as native /users/me,
+		// skipping user.UpdateUser's OU/type validation. That validation requires
+		// system:usertype:view with no self-access bypass, so self-service callers
+		// would otherwise hit it and fail.
+		result, svcErr = s.userService.UpdateUserAttributes(ctx, userID, attrsJSON)
+	} else {
+		updatedUser := &providers.User{
+			ID:         userID,
+			OUID:       et.OUID,
```
ThaminduDilshan: Shall we use `existingUser.OUID` here? `et.OUID` moves the user to the user type's OU on every PUT

backend/internal/scim/users/validator.go
```diff
@@ -0,0 +1,295 @@
...
+		if strings.EqualFold(k, "schemas") {
+			continue
+		}
+		if extensionURN != "" && strings.EqualFold(k, extensionURN) {
+			continue
+		}
+		if strings.EqualFold(strings.TrimSpace(k), scim.SCIMEnterpriseUserSchemaURN) {
+			continue
+		}
+		coreAttrs[k] = v
```
ThaminduDilshan, line 202: `id`, `meta` and `externalId` land in `coreAttrs` too. GET a user of a non core type and PUT the same body back and `processInboundPayload` returns `ErrorCoreSchemaNotSupported`. Shall we skip the readonly common attributes?

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+			log.String("userID", userID), log.Any("error", svcErr))
+		return nil, mapUserServiceErrorToSCIM(svcErr)
+	}
+
+	extensionURN := scim.BuildSchemaURN(u.Type)
+	rawProps, typeID, svcErr := s.getSchemaProps(ctx, u.Type)
+	if svcErr != nil {
+		return nil, svcErr
+	}
+	includeCoreAttrs := s.cfg.CoreUserTypeID != "" && typeID == s.cfg.CoreUserTypeID
```
ThaminduDilshan, line 136: `processInboundPayload` uses `resolveIsCoreUserType`, which falls back to the sole user type. This and `ListUsers` don't. With `core_user_type_id` unset, POST returns the core attributes and the following GET doesn't

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+			continue
+		}
+		rawProps, ok := schemaPropsByType[u.Type]
+		if !ok {
+			var typeID string
+			var svcErr *tidcommon.ServiceError
+			rawProps, typeID, svcErr = s.getSchemaProps(ctx, u.Type)
+			if svcErr != nil {
+				s.logger.Warn(ctx, "SCIM ListUsers: omitting user with unresolvable user type",
+					log.String("userID", u.ID), log.String("userType", u.Type))
```
ThaminduDilshan, line 105: `totalResults` still counts these users but they're missing from `Resources`. A client paging through would silently skip them, right?

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+
+// getSchemaProps returns the parsed schema property definitions and entity type ID for the
+// given user type. The entity type ID lets callers decide whether this is the configured
+// CoreUserTypeID without a second lookup.
+func (s *scimUsersService) getSchemaProps(
+	ctx context.Context, resolvedUserTypeName string,
+) (map[string]scim.RawPropertyDef, string, *tidcommon.ServiceError) {
+	// Use elevated runtime context if necessary, but we are just reading schema info.
+	et, err := s.userTypeService.GetEntityTypeByName(
+		security.WithRuntimeContext(ctx), entitytype.TypeCategoryUser, resolvedUserTypeName)
```
ThaminduDilshan, line 395: Do we need runtime context on a user facing request path? Everywhere else `WithRuntimeContext` is for declarative loading and internal providers

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+		urn := attr[:idx]
+		field := attr[idx+1:]
+		if strings.Contains(field, ".") {
+			return scim.NewSubAttrProjectionError(attr)
+		}
+		userTypeName, ok := scim.ParseUserTypeFromSchemaURN(urn)
+		if !ok {
+			return scim.NewUnrecognizedSchemaURNError(attr)
+		}
+		resolved, svcErr := scim.ResolveUserTypeNameForSchemaURN(ctx, s.userTypeService, userTypeName)
```
ThaminduDilshan, line 368: Every other user type lookup in this file elevates. Without `system:usertype:view` a valid URN qualified `attributes=` returns 400 here, and `/Me` needs no permission at all

backend/internal/scim/users/service.go
```diff
@@ -0,0 +1,546 @@
...
+}
+
+// GetUser fetches a single user by ID and returns a SCIM User resource.
+func (s *scimUsersService) GetUser(
+	ctx context.Context, userID, baseURL string,
+) (*SCIMUser, *tidcommon.ServiceError) {
+	u, svcErr := s.userService.GetUser(ctx, userID, false)
+	if svcErr != nil {
+		s.logger.Debug(ctx, "SCIM GetUser: user service error",
+			log.String("userID", userID), log.Any("error", svcErr))
```
ThaminduDilshan, line 127: Shall we use `log.MaskedString(log.LoggerKeyUserID, ...)`? The user service masks the same ID on the same request. Check other places as well

backend/internal/scim/discovery/service.go
```diff
@@ -0,0 +1,481 @@
...
+		return et, page.TotalResults, nil
+	}
+
+	if page.TotalResults != 1 {
+		s.logger.Debug(ctx, "Core user type unavailable, omitting SCIM core User schema",
+			log.Any("error", &scim.ErrorMissingCustomSchema))
+		return nil, page.TotalResults, nil
+	}
+	et, svcErr := s.userTypeService.GetEntityTypeByName(
+		runtimeCtx, entitytype.TypeCategoryUser, page.Types[0].Name,
```
ThaminduDilshan, line 414: `ResolveDefaultUserTypeName` also checks `len(page.Types) != 1`. Count and list are separate queries, so this can panic. Shall we do the same here?

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,410 @@
...
+	members []SCIMGroupMember, baseURL string) (*SCIMGroup, *tidcommon.ServiceError) {
+	thunderIDMembers, svcErr := scimMembersToThunderID(members)
+	if svcErr != nil {
+		s.logger.Debug(ctx, "SCIM CreateGroup: invalid member type", log.Any("error", svcErr))
+		return nil, svcErr
+	}
+
+	req := group.CreateGroupRequest{
+		Name:    displayName,
+		OUID:    security.GetOUID(ctx),
```
ThaminduDilshan, line 132: SCIM Users takes the OU from the entity type. This takes it from the `ouId` claim, so a client credentials token without `ouId` can't create a group and gets `400 invalidSyntax`

backend/internal/scim/groups/resource.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/thunder-id/thunderid/internal/group"
+	scim "github.com/thunder-id/thunderid/internal/scim/common"
+)
+
+// thunderIDMemberTypeToSCIM maps ThunderID member types to SCIM member types.
+func thunderIDMemberTypeToSCIM(t group.MemberType) string {
+	if t == group.MemberTypeGroup {
+		return "Group"
+	}
+	return "User"
```
ThaminduDilshan, line 18: `app` and `agent` members come back as `type: "User"` with a `/Users/` ref. The GET then PUT round trip sends them back as users and `validateEntityMembers` rejects it

backend/internal/scim/groups/validator.go
```diff
@@ -0,0 +1,151 @@
...
+		return validateDisplayNamePatchOp(normalizedOp, op.Value)
+	case strings.EqualFold(path, "members"):
+		return validateMembersPatchOp(normalizedOp, op.Value, "")
+	case strings.HasPrefix(strings.ToLower(path), "members["):
+		filterValue, svcErr := parseMembersFilterPath(path)
+		if svcErr != nil {
+			return SCIMGroupPatchAction{}, svcErr
+		}
+		return validateMembersPatchOp(normalizedOp, op.Value, filterValue)
+	default:
```
ThaminduDilshan, line 81: `path` is optional for `add` and `replace` (RFC 7644 §3.5.2.1/§3.5.2.3). A pathless `{"op":"replace","value":{"displayName":"x"}}` is what Entra sends for a rename and it gets 400 invalidPath

backend/internal/scim/groups/validator.go
```diff
@@ -0,0 +1,151 @@
...
+		return SCIMGroupPatchAction{}, &scim.ErrorInvalidPatchPath
```
ThaminduDilshan, line 82: §3.5.2.2 wants `noTarget` for a pathless remove and for a valuePath that matches nothing. We only have `invalidPath`

backend/internal/scim/groups/handler.go
```diff
@@ -0,0 +1,184 @@
...
+		baseURL: cfg.PublicURL,
+		logger:  *log.GetLogger().With(log.String(log.LoggerKeyComponentName, groupHandlerLoggerComponentName)),
+	}
+}
+
+// HandleGroupsListRequest handles GET /scim/v2/Groups.
+func (h *Handler) HandleGroupsListRequest(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	if r.URL.Query().Get("filter") != "" {
+		scim.HandleSCIMError(w, r, &scim.ErrorFilterNotSupported, h.logger)
```
ThaminduDilshan, line 40: ServiceProviderConfig advertises `filter.supported: true` but every filter on `/Groups` is rejected. Okta and Entra both do `GET /Groups?filter=displayName eq ".."` before create. Shall we support `eq` here too?

backend/internal/scim/common/constants.go
```diff
@@ -0,0 +1,61 @@
...
+// SCIM Error types. Exported because tests in the users/groups/discovery
+// packages assert on the ScimType field of a returned SCIMErrorResponse.
+const (
+	ScimErrorTypeInvalidValue   ScimErrorType = "invalidValue"
+	ScimErrorTypeInvalidPath    ScimErrorType = "invalidPath"
+	ScimErrorTypeMutability     ScimErrorType = "mutability"
+	ScimErrorTypeInvalidFilter  ScimErrorType = "invalidFilter"
+	ScimErrorTypeInvalidSyntax  ScimErrorType = "invalidSyntax"
+	ScimErrorTypeUniqueness     ScimErrorType = "uniqueness"
+	ScimErrorTypeNotImplemented ScimErrorType = "notImplemented"
```
ThaminduDilshan, line 60: `notImplemented` isn't in the RFC 7644 §3.12 Table 9 list. Shall we drop `scimType` on the 501 and keep only the status?

backend/internal/scim/common/response.go
```diff
@@ -0,0 +1,173 @@
...
+		ErrorMultipleCustomSchemas.Code,
+		ErrorInvalidCustomSchemaURN.Code,
+		ErrorMissingCustomSchemaObject.Code,
+		ErrorUndeclaredCustomSchemaObject.Code,
+		ErrorEnterpriseSchemaNotSupported.Code,
+		ErrorCoreSchemaNotSupported.Code,
+		ErrorUndeclaredEnterpriseSchemaObject.Code,
+		ErrorUnknownUserType.Code,
+		ErrorSchemaValidationFailed.Code,
+		ErrorMutabilityViolation.Code,
```
ThaminduDilshan, line 42: Shall we move `ErrorMutabilityViolation` to the `mutability` case below? `ErrorImmutableUserType` already maps there, and `api/scim.yaml:834` documents `mutability` for this same PUT

backend/internal/scim/users/resource.go
```diff
@@ -0,0 +1,259 @@
...
+) SCIMUser {
+	location := fmt.Sprintf("%s%s/Users/%s", baseURL, scim.SCIMBasePath, u.ID)
+
+	scimUser := SCIMUser{
+		ID:           u.ID,
+		Schemas:      []string{scim.SCIMCoreUserSchemaURN, extensionURN},
+		ExtensionURN: extensionURN,
+		Meta: scim.SCIMMeta{
+			ResourceType: "User",
+			Location:     location,
```
ThaminduDilshan, line 68: `meta.created` and `meta.lastModified` are never set on Users or Groups, though discovery sets them. #3089 item 4 asked for these and the spec examples show them

backend/internal/scim/users/model.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	scim "github.com/thunder-id/thunderid/internal/scim/common"
+)
+
+// SCIMUser is the SCIM wire representation of a ThunderID user resource.
+type SCIMUser struct {
+	ID              string                     `json:"id"`
```
ThaminduDilshan, line 15: No `externalId`. That's the correlation key clients reconcile on, and `attr_mapper.go:120` also makes it unfilterable. Are we deferring it?

backend/internal/scim/common/attr_rules.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// CoreAttrRules is the pre-configured mapping table, one candidate per ThunderID library attribute.
+//
+// This table is global and static: it applies to whichever single ThunderID user type is
+// designated as SCIMConfig.CoreUserTypeID, the only user type SCIM maps to/from the core User
+// schema. Other user types are exposed through their custom extension schema only. When
+// per-user-type mapping configuration (frontend + database backed) is added, this table will
+// need to become keyed by user type instead of being one global rule set.
+var CoreAttrRules = []CoreAttrRule{
```
ThaminduDilshan, line 80: No `active` mapping anywhere. Deprovisioning is the main use case in #3089 and clients set `active: false` rather than sending DELETE. Separate PR?

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+                  maxPayloadSize: 0
+                filter:
+                  supported: true
+                  maxResults: 100
+                changePassword:
+                  supported: false
+                sort:
+                  supported: false
+                etag:
+                  supported: true
```
ThaminduDilshan, line 72: `etag.supported: true` here but `ETagSupported = false` in `scimconfig`. The live endpoint returns the opposite of the doc

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+        Supports `attributes`/`excludedAttributes` for attribute projection (RFC 7644 §3.9); the two are mutually exclusive.
+      parameters:
+        - name: id
+          in: path
+          required: true
+          description: The unique identifier of the user.
+          schema:
+            type: string
+          example: "9a475e1e-b0cb-4b29-8df5-2e5b24fb0ed3"
+        - name: If-Match
```
ThaminduDilshan, line 772: `If-Match` and the 412s are documented but no handler reads the header, and `SCIMMeta` has no `version` field. The write goes through unconditionally

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+
+  /Users:
+    get:
+      tags:
+        - SCIM Provisioning
+      summary: List Users
+      description: |
+        Returns a paginated list of users. Supports `startIndex` and `count` query parameters for pagination (RFC 7644 §3.4.2).
+
+        Supports `filter` with one or more `eq` comparisons joined by `and` (e.g. `userName eq "alice"` or `userName eq "alice" and active eq true`); `or`, `not`, grouping, and any operator other than `eq` are rejected with `400 invalidFilter`. Sorting via `sortBy`/`sortOrder` is not supported and returns `400 invalidValue` if requested.
```
ThaminduDilshan, line 405: `active` is in `scimUnsupportedFilterAttrs`, so this example returns 400

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+                    id: "9a475e1e-b0cb-4b29-8df5-2e5b24fb0ed3"
+                    userName: "alice.wu"
+                    name:
+                      givenName: "Alice"
+                      familyName: "Wu"
+                    emails:
+                      - value: "alice.wu@company.inc"
+                        primary: true
+                    urn:thunderid:params:scim:schemas:employee:2.0:User:
+                      username: "alice.wu"
```
ThaminduDilshan, line 474: `stripMappedCandidates` removes these from the extension object, so the response won't repeat the mapped core attributes

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+    All write requests (`POST`, `PUT`, `PATCH`) must send `Content-Type: application/scim+json`. Requests with a missing or different Content-Type are rejected with a `400 invalidSyntax` error.
+
+    ### Unimplemented Endpoints
+    Per the SCIM specification, endpoints that are not yet implemented by the server (such as `/Bulk` and the root `/.search`) will return a **501 Not Implemented** HTTP status code rather than a generic 404. This correctly signals to SCIM clients that the feature is unsupported.
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0.html
+
+servers:
+  - url: https://{host}:{port}/scim/v2
```
ThaminduDilshan, line 25: `merge-openapi-specs.mjs` discards each spec's `servers`, so every path renders without `/scim/v2`. No sibling spec carries a prefix here

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+        displayName:
+          type: string
+        members:
+          type: array
+          items:
+            $ref: '#/components/schemas/GroupMember'
+        meta:
+          $ref: '#/components/schemas/Meta'
+
+    GroupListResponse:
```
ThaminduDilshan, line 2223: `GroupListResponse` and `UserListResponse` already exist in `group.yaml` and `user.yaml`. The merger flattens `components.schemas` last wins, so these overwrite theirs in the published reference

api/scim.yaml
```diff
@@ -0,0 +1,2316 @@
...
+          description: Unauthorized
+          content:
+            application/scim+json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+              example:
+                schemas: ["urn:ietf:params:scim:api:messages:2.0:Error"]
+                status: "401"
+                detail: "Authentication failed"
+        '409':
```
ThaminduDilshan, line 648: No endpoint declares 403 anywhere in this file, though the middleware returns it and `scim_authz_test.go` asserts it on eight operations. And 409 is only here, not on `/Groups`, where a name conflict maps to `uniqueness`

backend/internal/scim/init.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/thunder-id/thunderid/internal/scim/users"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	"github.com/thunder-id/thunderid/internal/system/middleware"
+	"github.com/thunder-id/thunderid/internal/user"
+)
+
+var scimServerStartTime = time.Now().UTC().Format(time.RFC3339)
+
+// unsupportedRouteComponentName identifies the unimplemented-endpoint routes for log attribution.
+const unsupportedRouteComponentName = "SCIMUnsupported"
```
ThaminduDilshan, line 29: Shall we get rid of the logger name constants? We define the logger in the constructor now. Seven of these across the package

backend/internal/scim/init.go
```diff
@@ -0,0 +1,234 @@
...
+
+// Initialize sets up the SCIM module and registers all /scim/v2 routes.
+func Initialize(
+	mux *http.ServeMux,
+	userService user.UserServiceInterface,
+	userTypeService entitytype.EntityTypeServiceInterface,
+	groupService group.GroupServiceInterface,
+	cfg scimconfig.SCIMConfig,
+) {
+	dh := discovery.NewHandler(userTypeService, cfg, scimServerStartTime, cfg.PublicURL)
```
ThaminduDilshan, line 39: `cfg.PublicURL` is already in `cfg`

backend/internal/system/security/middleware.go
```diff
@@ -30,8 +33,26 @@ func middleware(service SecurityServiceInterface) (func(http.Handler) http.Handl
 			// Process the security checks
 			ctx, err := service.Process(r)
 			if err != nil {
-				// Write error response and stop request processing
-				writeSecurityError(ctx, w, err)
+				if strings.HasPrefix(r.URL.Path, "/scim/") {
```
ThaminduDilshan, line 36: Shall we keep the SCIM error shape out of the generic middleware? The error URN is duplicated from `scim/common/constants.go` too

backend/cmd/server/deployment.yaml
```diff
@@ -78,3 +78,6 @@ email:
     from_address: "noreply@thunderid.dev"
     enable_start_tls: false
     enable_authentication: true
+
+gate_client:
```
ThaminduDilshan: Is `gate_client` related to SCIM? It derives from the server config when omitted

backend/internal/system/config/config.go
```diff
@@ -398,6 +398,16 @@ type GroupConfig struct {
 	Store string `yaml:"store" json:"store"`
 }
 
+// SCIMConfig holds the SCIM service configuration.
+type SCIMConfig struct {
+	// CoreUserTypeID designates the ThunderID user type whose schema backs the SCIM
+	// core User schema (RFC 7643 §4.1): its attribute characteristics are reflected in
+	// /scim2/Schemas, and it is the default target for payloads carrying only the core
```
ThaminduDilshan, line 405: `/scim2/Schemas` should be `/scim/v2/Schemas`. Same in `scim/common/usertype_resolver.go:48`

backend/cmd/server/deployment.yaml
thiva-k: Let's remove this

backend/cmd/server/config/default.json
```diff
@@ -136,7 +136,7 @@
       "validity_period": 3600
     },
     "dcr": {
-      "enabled" : true,
+      "enabled": true,
```
thiva-k: Can we revert the unrelated whitespace reformatting here and keep only the new `scim` block?

backend/internal/scim/users/service.go
thiva-k: Shouldn't this be `existingUser.OUID`? This will move a user in a child OU to the user type's OU

backend/internal/scim/config/config.go
thiva-k, line 20: Is setting `PatchSupported` as false correct? Group PATCH is implemented right?

backend/internal/scim/groups/service.go
```diff
@@ -0,0 +1,410 @@
...
+		s.logger.Debug(ctx, "SCIM PatchGroup: group service error",
+			log.String("groupID", groupID), log.Any("error", svcErr))
+		return nil, mapGroupServiceErrorToSCIM(svcErr)
+	}
+	if g.IsReadOnly {
+		s.logger.Debug(ctx, "SCIM PatchGroup: group is read-only", log.String("groupID", groupID))
+		return nil, &scim.ErrorMutabilityViolation
+	}
+
+	for _, action := range actions {
```
thiva-k, line 210: Here patch fails on the first failed action and leaves the earlier ones applied right? Shouldn't this be done atomically. Let's check the spec

backend/cmd/server/config/default.json
```diff
@@ -191,6 +203,9 @@
   "group": {
     "store": "composite"
   },
+  "scim": {
+    "core_user_type_id": "01900000-0000-7000-8000-000000000010"
```
thiva-k, line 195: Can we use the user type name here instead of hardcoding the UUID?

backend/cmd/server/deployment.yaml
ravindu439: fixed it .

backend/cmd/server/config/default.json
ravindu439: fixed

backend/internal/scim/users/service.go
ravindu439: fixed it .

backend/internal/scim/config/config.go
ravindu439: I'd suggest keeping `PatchSupported` as `false` for now. 
`ServiceProviderConfig.patch.supported` is a single, service-wide flag per RFC 7644, it doesn't distinguish between resource types. Since User PATCH isn't implemented yet, advertising `true` would tell clients PATCH works everywhere, including on /Users, which isn't true and would break clients that rely on discovery to decide which operations to use.

Group PATCH is implemented, but I don't think that alone is enough to flip the flag. Concurrent member updates on the same group still need proper concurrency/transaction control (e.g. read-modify-write races on the members list), which we haven't added yet. Until that's in place, I'd rather not present Group PATCH as fully production-ready either.

My proposal: leave `PatchSupported` as `false` for this PR, and flip it to `true` once User PATCH is implemented and Group PATCH has concurrency control for member updates. That way the flag accurately reflects PATCH being supported and safe across both resource types, not just partially.

backend/internal/scim/groups/service.go
ravindu439: Agreed, this can't be fixed at the SCIM level. The proper transaction handling should come from the groups package itself. Until that's in place, may I restrict PATCH to the displayName target only, instead of also supporting members?

backend/cmd/server/config/default.json
ravindu439: We use the ID here on purpose. ThunderID allows creating user types whose names differ only by case (e.g. Employee vs employee), so a name-based reference could be ambiguous or resolve to the wrong type. The ID guarantees uniqueness with no risk of that kind of collision.

Name would admittedly be more readable in a config file, but the ID is safer for now. This is only a backend implementation detail at this point, the idea is to eventually let admins configure the core user type for SCIM through the UI, and this hardcoded value in default.json won't stick around once that frontend support exists.

backend/internal/scim/users/service.go
ravindu439: fixed already

backend/cmd/server/config/default.json
```diff
@@ -191,6 +191,9 @@
...
```
ThaminduDilshan, line 195: I'm planning to introduce a handle for the entity types. Once that is added, we should use handle here. Shall we remove id part from the config name and keep it agnostic from the type so that it will be easier to replace it with handle later?

On a separate thought, from a cloud product perspective I feel like this is not the correct place to have such config. 
Maybe server config API with a dedicated config resource for scim will be a better place for this.

backend/internal/scim/common/attr_rules.go
```diff
@@ -0,0 +1,349 @@
+// Copyright 2025-2026 The ThunderID Authors
```
ThaminduDilshan, line 1: We shouldn't have a license year range here right? Let's only keep 2026.
Check other files too

backend/internal/scim/common/attr_rules.go
```diff
@@ -0,0 +1,349 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// Package common provides shared SCIM 2.0 types, error constants, and helpers used across the
```
ThaminduDilshan, line 4: Shall we keep package comment only in the constants.go file?

backend/internal/scim/common/attr_rules_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestCandidatesForSubAttr_MappedSubAttr_ReturnsItsCandidate tests that a sub-attribute backed
+// by its own CoreAttrRule (e.g. name.givenName) returns exactly that candidate.
+func TestCandidatesForSubAttr_MappedSubAttr_ReturnsItsCandidate(t *testing.T) {
```
ThaminduDilshan, line 14: Shall we follow test suite pattern. Check in all the places

backend/internal/scim/config/config.go
ThaminduDilshan, line 20: Do we need to have public constants for all of these atomic fields? What if we move all of these inside `SCIMConfig` struct and only expose that?

backend/internal/scim/discovery/consts_test.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2025-2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package discovery
+
+// testServerStartTime is a fixed serverStartTime value fed into newSCIMDiscoveryService
+// for tests, standing in for the real value init.go computes from time.Now().
+const testServerStartTime = "2024-01-01T00:00:00Z"
```
ThaminduDilshan, line 8: Do we need to have a separate file just to store this?

backend/internal/system/security/permissions.go
ravindu439: I've made /ResourceTypes match /Schemas: it now needs a valid bearer token, and only ServiceProviderConfig stays public.

backend/internal/system/security/permissions.go
ravindu439: I did not gate them with system:usertype:view. A SCIM provisioning client's token is normally scoped for the CRUD operations, system:user and system:group. Requiring usertype:view for discovery would stop those clients from reading /Schemas and /ResourceTypes, which they need to learn the extension URNs before provisioning. Every SCIM client would then need an extra permission that has nothing to do with provisioning.

The gap with GET /user-types remains: any authenticated token can read the attribute schemas through SCIM discovery. Do you want us to close that by requiring usertype:view on discovery, and grant that permission to the provisioning client role? Or is authenticated-only acceptable for SCIM discovery?

backend/internal/scim/discovery/handler.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/thunder-id/thunderid/internal/entitytype"
+	scim "github.com/thunder-id/thunderid/internal/scim/common"
+	scimconfig "github.com/thunder-id/thunderid/internal/scim/config"
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+const discoveryHandlerLoggerComponentName = "SCIMDiscoveryHandler"
+
+// Handler handles SCIM discovery HTTP requests.
+type Handler struct {
```
ThaminduDilshan, line 18: Let's not expose public structs and constructor methods unnecessarily. Pattern is to keep structs and constructor methods package private and only expose instances through a Initialize() method in init.go.
If needed to expose outside, we can define a public interface, still we shouldn't have public structs.

Check this for all the applicable places in this PR

ThaminduDilshan, line 18: better to rename this to scimDiscoveryHandler or something

backend/internal/scim/groups/consts_test.go
```diff
@@ -0,0 +1,9 @@
+// Copyright 2025-2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package groups
+
+const (
```
ThaminduDilshan, line 6: Do we need a separate file just to store test constants? But don't move them to source files too.
Check other places too where we have the same pattern

backend/internal/scim/common/constants.go
```diff
@@ -0,0 +1,61 @@
...
+	SCIMEnterpriseUserSchemaURN = "urn:ietf:params:scim:schemas:extension:enterprise:2.0:User"
+
+	// SCIMErrorSchemaURN is the SCIM error schema URN.
+	SCIMErrorSchemaURN = "urn:ietf:params:scim:api:messages:2.0:Error"
+
+	// SCIMListResponseSchemaURN is the SCIM list response schema URN.
+	SCIMListResponseSchemaURN = "urn:ietf:params:scim:api:messages:2.0:ListResponse"
+
+	// ThunderIDURNPrefix is the custom URN prefix for ThunderID SCIM schemas.
+	ThunderIDURNPrefix = "urn:thunderid:params:scim:schemas:"
```
ThaminduDilshan, line 27: Do we need to have thunderid term here? Thinking whether customers have to expose internal deployment details with this. I.e. there're using thunderid.

Two options;
1. If it's not needed to be there, we can remove it
2. If needed to be there, maybe we should make it configurable through deployment.yaml

@sadilchamishka @AnuradhaSK WDYT?

backend/cmd/server/config/default.json
ThaminduDilshan, line 195: Btw if we're having any deployment configs, we should template them in deployment resources too

backend/internal/system/security/permissions.go
ravindu439: fixed

backend/internal/scim/common/filter.go
ravindu439: fixed

backend/internal/scim/users/attr_mapper.go
ravindu439: fixed by moving it in to unsupported section.

backend/internal/scim/users/validator.go
ravindu439: fixed

backend/internal/scim/users/service.go
ravindu439: fixed

backend/internal/scim/users/service.go
ravindu439: fixed it

backend/internal/scim/users/service.go
ravindu439: Intentional. SCIM clients are scoped for user/group CRUD and don't hold system:usertype:view, so reading the type schema on that path would fail with 401. The runtime context is used only for that schema read. User create, update, get and delete still run on the caller's own context, so OU scoping applies there. Same pattern as the OU-by-path lookups in the user and group services.

backend/internal/scim/users/service.go
ravindu439: fixed

backend/internal/scim/users/service.go
ravindu439: fixed

backend/internal/scim/discovery/service.go
ravindu439: fixed

backend/internal/scim/groups/service.go
ravindu439: This is intentional. A group belongs to an organization unit, and SCIM has no attribute for the OU, so we use the caller's OU from the token's ouId claim. Users can take their OU from the entity type. Groups have no type, and a user type's OU has nothing to do with where a group should live, so falling back to it would be wrong.

For client credentials, ouId is opt-in. It appears in the token only if the client lists it in ClientConfig.Attributes. A SCIM client that provisions groups needs ouId in its token attributes, which we can document.

You're right that the 400 invalidSyntax is misleading when the claim is missing. I'm happy to return a clearer error for that case if you'd like.

backend/internal/scim/groups/resource.go
ravindu439: The SCIM spec only defines User and Group, but the group store also allows app and agent, and we were reporting them as User, which broke GET then PUT. Members now come back as App or Agent, and both are accepted on input. There's no SCIM endpoint for them, so those members carry no $ref. /Schemas canonicalValues is updated to match

PR #5132, opened 2026-08-21
Reject unsupported request objects instead of ignoring them

### LINKED ISSUE #5111: Authorization requests carrying a `request` object are processed on the query string alone, silently dropping `state` and `nonce`
### Description

ThunderID does not implement JAR (RFC 9101), which is a legitimate position: OIDC Core §6 makes
request object support optional for an OP, and OIDC Discovery defines
`request_parameter_supported` so a server can advertise that it does not.

The problem is what happens instead. An authorization request that carries a `request`
parameter is neither honoured nor rejected. The parameter is ignored and the request proceeds
using only the query string, so any parameter that the client placed *inside* the request
object is lost. The client receives what looks like a successful authorization response that
silently violates its own request.

`state` and `nonce` are the two that matter. Both are security parameters: `state` is the
client's CSRF defence and `nonce` binds the ID token to the authorization request. A client
that sends them inside a request object gets a response with neither, and no error explaining
why.

Discovery makes this worse rather than better. `request_parameter_supported`,
`request_uri_parameter_supported` and `request_object_signing_alg_values_supported` are all
absent from `/.well-known/openid-configuration`, and none of them appear in
`backend/internal/oauth/oauth2/discovery/model.go`. The two booleans have opposite defaults in
OIDC Discovery §3:

| Field | Default when omitted |
|---|---|
| `request_parameter_supported` | `false` |
| `request_uri_parameter_supported` | **`true`** |

So omitting `request_parameter_supported` correctly says the `request` parameter is not
supported, but omitting `request_uri_parameter_supported` advertises that `request_uri` **is**
supported. A client reading discovery is therefore told it may pass a `request_uri`, does so,
and has it silently ignored.

`request_uri` is only recognised as a PAR handle
(`backend/internal/oauth/oauth2/par/`), so a `request_uri` pointing at a JWT is not resolved.

### Steps to Reproduce

Send an authorization request that carries its parameters in a `request` object rather than on
the query string:

```
GET /oauth2/authorize
  ?client_id=<id>
  &request=<unsigned JWT containing response_type, redirect_uri, scope, state, nonce>
```

The request succeeds and redirects to the client, but the authorization response carries no
`state`, and the ID token's `nonce` does not match the one inside the object. A `request_uri`
pointing at the same JWT is likewise not resolved, since `request_uri` is only recognised as a
PAR handle.

This was found running the OpenID Foundation conformance suite, Basic OP plan
(`oidcc-basic-certification-test-plan`). Two modules cover it.

`oidcc-unsigned-request-object-supported-correctly-or-rejected-as-unsupported` sends the
parameters in an unsigned JWT in `request`, with `state` and `nonce` inside the object, and
fails:

```
CheckStateInAuthorizationResponse: State was passed in request, but is missing from
response (or returned in the wrong place)
   expected: HXuVIfU2Qa
ValidateIdTokenNonce: Nonce values mismatch
   expected: u64iFelu1S
```

`oidcc-request-uri-unsigned-supported-correctly-or-rejected-as-unsupported` sends the same by
reference in `request_uri` and stalls after the redirect to the authorization endpoint.

The module names state the contract: *supported correctly **or rejected as unsupported***.
Either behaviour passes. Silently ignoring is the one outcome neither branch allows.

#### Expected

Two things, both small, and neither requires implementing JAR.

1. **Advertise the truth in discovery.** Add `request_uri_parameter_supported: false`, since
   the default when omitted is `true` and that is the opposite of what ThunderID does.
   `request_parameter_supported` can stay omitted, as its default is already `false`, though
   stating it explicitly costs nothing and removes the asymmetry for anyone reading the
   document.
2. **Reject rather than ignore.** An authorization request carrying a `request` parameter
   should be rejected with the error OIDC Core §6.1 defines for it, `request_not_supported`,
   and a `request_uri` that is not a PAR handle should be rejected with
   `request_uri_not_supported`. Both are returned to the client's `redirect_uri` in the
   ordinary way, so the client learns its request was not honoured.

The discovery fix stops well-behaved clients sending something that will be dropped. The
rejection covers clients that do not read discovery, and turns both conformance modules above
into passes.

Note that PAR issues its own `request_uri` values, so whatever wording is chosen should make
clear that `request_uri_parameter_supported: false` refers to client-supplied request objects by
reference, not to PAR handles.

### Version

v1.0.0 (`oidc-conformnace-suite` branch, distribution `thunderid-1.0.0-macos-arm64`)

macOS, SQLite (default deployment), server on `https://thunderid.conformance.test:8090`.

Thumimku:
### Purpose

An authorization request carrying a `request` parameter was neither honored nor rejected. The
parameter was ignored and the request proceeded on the query string alone, so any parameter the
client placed inside the request object was lost, including `state` and `nonce`. Both are security
parameters: `state` is the client's CSRF defence and `nonce` binds the ID token to the
authorization request. The client received an apparently successful response that silently
violated its own request, with no error explaining why.

Discovery compounded this. `request_parameter_supported` and `request_uri_parameter_supported` were
both absent from `/.well-known/openid-configuration`, and their defaults in OIDC Discovery 3 differ:
the former defaults to `false`, but the latter defaults to **`true`**. Omitting
`request_uri_parameter_supported` therefore advertised support that does not exist, so a client
reading discovery was told it may pass a `request_uri`, did so, and had it silently ignored.

Found running the OpenID Foundation conformance suite, Basic OP plan
(`oidcc-basic-certification-test-plan`). Two modules cover it, and their names state the contract:
*supported correctly **or rejected as unsupported***. Either branch passes. Silently ignoring is the
one outcome neither allows.

### Approach

JAR (RFC 9101) remains unimplemented, which OIDC Core 6 permits an OP to do. Rather than
implementing it, the server now tells the truth and rejects:

1. **Advertise the truth in discovery.** `request_parameter_supported` and
   `request_uri_parameter_supported` are both emitted as `false`. Both are stated explicitly rather
   than relying on defaults that point in opposite directions, which also removes the asymmetry for
   anyone reading the document.

2. **Reject rather than ignore.** A `request` parameter is rejected with `request_not_supported`
   (OIDC Core 6.1), and a `request_uri` that is not a PAR handle is rejected with
   `request_uri_not_supported`.

Verified manually against a local build:

| Request | Result |
|---|---|
| `request=<unsigned JWT>` | `302` to client: `error=request_not_supported`, `state` echoed |
| `request_uri=https://client.example.org/request.jwt` | `302` to client: `error=request_uri_not_supported`, `state` echoed |
| `request` at `/oauth2/par` | `{"error":"request_not_supported"}` |
| `request_uri=urn:ietf:params:oauth:request_uri:...` (PAR handle) | unchanged, resolves normally |
| no request object | unchanged, proceeds to login |

Before the fix, the first case redirected to the login page with no error at all, and the second
redirected to the server's own error page rather than the client, which is why the `request_uri`
conformance module stalled after the redirect.

### Related Issues
- Fixes thunder-id/thunderid#5111

### Related PRs
- N/A

### Checklist
    - `docs/content/guides/protocols/oauth-oidc/server-metadata.mdx`: added both fields to the
      sample document and documented the behavior, making clear that
      `request_uri_parameter_supported: false` refers to client-supplied request objects by
      reference, not to PAR handles.
    - [x] Unit Tests
        - `requestvalidator/validator_test.go`: `request` rejected, rejected before other
          parameters, empty value ignored.
        - `authz/service_test.go`: non-PAR `request_uri` rejected to the client; not redirected when
          the `redirect_uri` is unregistered; falls back to the single registered URI when omitted.
        - `par/service_test.go`: `IsPARRequestURI` prefix handling.
        - `discovery/discovery_test.go`: both booleans false and present in the serialized document.
    - [x] Integration Tests
        - `tests/integration/oauth/authz/request_object_test.go`: both rejections reach the client's
          `redirect_uri` with `state` echoed.
        - `tests/integration/oauth/discovery/discovery_test.go`: both booleans present and false in
          the raw JSON.

### Security checks

## Summary by CodeRabbit

* **OAuth/OIDC Authorization**
  * Unsupported request objects return `request_not_supported` or `request_uri_not_supported`.
  * Valid redirect URIs preserve client redirection and state; invalid or unsafe URIs are not used.
  * Missing redirect URIs may fall back to the client’s sole registered URI.
  * Pushed Authorization Request (PAR) request URIs remain supported.

* **Discovery**
  * Provider metadata now explicitly reports unsupported request objects and non-PAR request URIs.

* **Documentation**
  * Updated server metadata guidance for supported and unsupported request parameters.

### DIFF
--- a/backend/internal/oauth/oauth2/authz/requestvalidator/validator.go
+++ b/backend/internal/oauth/oauth2/authz/requestvalidator/validator.go
@@ -19,7 +19,7 @@ import (
 // ValidateAuthorizationRequestParams validates the common authorization request parameters
 // shared by both the standard authorize endpoint and the PAR endpoint.
 //
-// This validates: prompt, grant_type, response_type, PKCE, nonce, and dpop_jkt.
+// This validates: request, prompt, grant_type, response_type, PKCE, nonce, and dpop_jkt.
 // Callers are responsible for validating client_id and redirect_uri before calling this
 // function, since those validations have endpoint-specific error handling semantics
 // (e.g., the authorize endpoint must not redirect errors when the redirect_uri is invalid).
@@ -37,6 +37,13 @@ func ValidateAuthorizationRequestParams(
 	params := url.Values(rawParams)
 	responseType := params.Get(constants.RequestParamResponseType)
 	responseMode := params.Get(constants.RequestParamResponseMode)
+
+	// ThunderID does not implement JAR (RFC 9101). A request object must be rejected rather than
+	// ignored: honoring only the query string would silently drop security parameters the client
+	// placed inside the object, notably state and nonce (OIDC Core 6.1).
+	if params.Get(constants.RequestParamRequest) != "" {
+		return constants.ErrorRequestNotSupported, "The request parameter is not supported"
+	}
 
 	// Validate the prompt parameter if present.
 	if params.Has(constants.RequestParamPrompt) {
--- a/backend/internal/oauth/oauth2/authz/service.go
+++ b/backend/internal/oauth/oauth2/authz/service.go
@@ -188,8 +188,15 @@ func (as *authorizeService) HandleInitialAuthorizationRequest(ctx context.Contex
 		}
 	}
 
-	// If request_uri is present, resolve the pushed authorization request.
+	// If request_uri is present, resolve the pushed authorization request. A request_uri that is
+	// not a PAR handle is a client-supplied request object by reference (RFC 9101), which is not
+	// supported: reject it rather than ignoring it, so the client learns its request was not
+	// honored (OIDC Core 6.1).
 	if requestURI != "" {
+		if !par.IsPARRequestURI(requestURI) {
+			return nil, as.newRequestObjectError(ctx, msg, app,
+				oauth2const.ErrorRequestURINotSupported, "The request_uri parameter is not supported")
+		}
 		return as.handlePARAuthorizationRequest(ctx, requestURI, clientID, app)
 	}
 
@@ -207,6 +214,38 @@ func (as *authorizeService) HandleInitialAuthorizationRequest(ctx context.Contex
 	}
 
 	return as.handleStandardAuthorizationRequest(ctx, msg, app, initiatorReq)
+}
+
+// newRequestObjectError builds the rejection for an unsupported request object. The error is
+// returned to the client's redirect_uri only when that redirect_uri validates against the
+// client's registration, since an unvalidated redirect_uri must not be used as a redirect
+// target; otherwise it is shown on the server error page.
+func (as *authorizeService) newRequestObjectError(
+	ctx context.Context, msg *OAuthMessage, app *providers.OAuthClient, code string, message string,
+) *AuthorizationError {
+	queryParams := url.Values(msg.RequestQueryParams)
+	redirectURI := queryParams.Get(oauth2const.RequestParamRedirectURI)
+
+	authErr := &AuthorizationError{
+		Code:    code,
+		Message: message,
+		State:   queryParams.Get(oauth2const.RequestParamState),
+	}
+	if err := app.ValidateRedirectURI(ctx, redirectURI); err != nil {
+		as.logger.Debug(ctx, "Validation failed for redirect URI", log.Error(err))
+		return authErr
+	}
+	// ValidateRedirectURI accepts an omitted redirect_uri when the client has exactly one
+	// registered; fall back to that so the rejection still reaches the client.
+	if redirectURI == "" {
+		if len(app.RedirectURIs) == 0 {
+			return authErr
+		}
+		redirectURI = app.RedirectURIs[0]
+	}
+	authErr.SendErrorToClient = true
+	authErr.ClientRedirectURI = redirectURI
+	return authErr
 }
 
 // handlePARAuthorizationRequest resolves a request_uri from a PAR and continues the authorization flow.
--- a/backend/internal/oauth/oauth2/authz/service.go
+++ b/backend/internal/oauth/oauth2/authz/service.go
@@ -188,8 +188,15 @@ func (as *authorizeService) HandleInitialAuthorizationRequest(ctx context.Contex
 		}
 	}
 
-	// If request_uri is present, resolve the pushed authorization request.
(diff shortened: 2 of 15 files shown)


backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -209,6 +216,38 @@ func (as *authorizeService) HandleInitialAuthorizationRequest(ctx context.Contex
...
+		State:   queryParams.Get(oauth2const.RequestParamState),
+	}
+	if err := app.ValidateRedirectURI(ctx, redirectURI); err != nil {
+		as.logger.Debug(ctx, "Validation failed for redirect URI", log.Error(err))
+		return authErr
+	}
+	// ValidateRedirectURI accepts an omitted redirect_uri when the client has exactly one
+	// registered; fall back to that so the rejection still reaches the client.
+	if redirectURI == "" {
+		if len(app.RedirectURIs) == 0 {
```
thiva-k: Isn't this already checked at ValidateRedirectURI ?

backend/internal/oauth/oauth2/authz/requestvalidator/validator.go
```diff
@@ -38,6 +38,13 @@ func ValidateAuthorizationRequestParams(
 	responseType := params.Get(constants.RequestParamResponseType)
 	responseMode := params.Get(constants.RequestParamResponseMode)
 
+	// ThunderID does not implement JAR (RFC 9101). A request object must be rejected rather than
+	// ignored: honoring only the query string would silently drop security parameters the client
+	// placed inside the object, notably state and nonce (OIDC Core 6.1).
+	if params.Get(constants.RequestParamRequest) != "" {
+		return constants.ErrorRequestNotSupported, "The request parameter is not supported"
+	}
```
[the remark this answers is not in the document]
Thumimku: The finding assumes repeated request / request_uri values can reach the service layer, so that url.Values.Get returning only the first value would let ?request=&request=<jwt> bypass the check. That can't happen: repeated parameters are already rejected one layer up, before OAuthMessage is ever constructed.

backend/internal/oauth/oauth2/authz/service.go
Thumimku: Yes, removed

thiva-k (approved):

Issue #5134, opened 2026-08-21
JWKS endpoint publishes kid and alg that can mismatch the signing key


anushasunkada:
### Description

The JWKS endpoint (GET /oauth/jwks via internal/oauth/jwks) can publish a JWK whose kid and alg don't correspond to the key actually used to sign issued JWTs.

kid: GetPublicKeys() in the runtime_crypto_provider.go set PublicKeyInfo.KeyID to the internal cert/key store id, while the JWT signer (internal/system/jose/jwt/service.go) uses the key's thumbprint as kid. The JWKS service was building the published kid from Thumbprint.

alg: internal/oauth/jwks/service.go's per-key-type JWK builders (RSA, ECDSA, EdDSA, ML-DSA) each hardcoded or independently recomputed the alg value instead of using the algorithm already resolved by GetPublicKeys(), risking a published alg that doesn't match the key's actual signing algorithm.

A relying party validating a token against the JWKS document could fail to find a matching key, or match the wrong key/algorithm.

### Steps to Reproduce

1. Configure the key manager with a signing key.
2. Issue a token via the OAuth token endpoint and inspect its JWT header kid/alg.
3. Fetch GET /oauth2/jwks and compare the kid/alg of the matching entry.
4. Observe the published kid (thumbprint-derived) does not correspond to PublicKeyInfo.KeyID used elsewhere, and alg is independently hardcoded per key type rather than sourced from the provider.

### Version

1.0.0

PR #5135, opened 2026-08-21
Fixed the JWKS kid and alg mismatch issue

anushasunkada:
### Purpose

Fixed the JWKS kid and alg mismatch issue.

### Approach

GetPublicKeys() now computes the thumbprint once and assigns it to both KeyID and Thumbprint.
jwks/service.go reads kid := keyInfo.KeyID and alg := keyInfo.Algorithm directly from the provider and passes alg through to every builder function, removing all the per-type hardcoded/recomputed algorithm logic.

Added a guard: if KeyID comes back empty, the service logs and returns InternalServerError rather than publishing a JWK with no kid.

### Related Issues
- #5134

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JWKS reliability by using consistent certificate thumbprints as key identifiers.
  * Preserved the runtime-provided signing algorithm across supported key types.
  * JWKS generation now reports an internal error when a key lacks a required identifier.

### DIFF
--- a/backend/internal/oauth/jwks/service.go
+++ b/backend/internal/oauth/jwks/service.go
@@ -61,7 +61,12 @@ func (s *jwksService) GetJWKS(ctx context.Context) (*JWKSResponse, *tidcommon.Se
 	var jwksKeys []JWKS
 
 	for _, keyInfo := range publicKeys {
-		kid := keyInfo.Thumbprint
+		kid := keyInfo.KeyID
+		if kid == "" {
+			s.logger.Error(ctx, "Public key is missing key ID")
+			return nil, &tidcommon.InternalServerError
+		}
+		alg := keyInfo.Algorithm
 
 		var x5c []string
 		var x5t, x5tS256 string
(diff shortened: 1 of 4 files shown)


backend/internal/system/kmprovider/defaultkm/runtime_crypto_provider.go
```diff
@@ -319,11 +319,13 @@ func (s *runtimeCryptoService) GetPublicKeys(
 			continue
 		}
 
+		thumbprint := s.pkiService.GetCertThumbprint(id)
+
 		keys = append(keys, providers.PublicKeyInfo{
-			KeyID:               id,
+			KeyID:               thumbprint,
```
thiva-k, line 325: This field is for the key's id from the deployment config (e.g "default-key") in defaultkm right? Do we need to have two fields for thumbprint?

backend/internal/oauth/jwks/service.go
```diff
@@ -61,7 +61,12 @@ func (s *jwksService) GetJWKS(ctx context.Context) (*JWKSResponse, *tidcommon.Se
 	var jwksKeys []JWKS
 
 	for _, keyInfo := range publicKeys {
-		kid := keyInfo.Thumbprint
+		kid := keyInfo.KeyID
```
thiva-k, line 64: In runtime we are still reading the Thumbprint to embed the kid in jwt header. Better to read from a single field for both jwks and jwt

ThaminduDilshan:
@coderabbitai review

backend/internal/system/kmprovider/defaultkm/runtime_crypto_provider.go
anushasunkada: This method should return the same keyId as used in the JWT signing in the header part.

This is a provider which will be implemented by the embedder's. KeyId and thumbprint can hold different or same value depends on the embedder's implementation. So it is better to keep it separate.

backend/internal/oauth/jwks/service.go
anushasunkada: This is a provider which will be implemented by the embedder's. KeyId and thumbprint can hold different or same value depends on the embedder's implementation.

backend/internal/system/kmprovider/defaultkm/runtime_crypto_provider.go
thiva-k: If an embedder passes different values for KeyId and Thumbprint, then with the current changes the JWKS will publish the KeyID while the JWT header will still carry the Thumbprint as kid(https://github.com/thunder-id/thunderid/blob/1.0.x/backend/internal/system/jose/jwt/service.go#L89)

Those two kids have to be the same value, so both should read from the same field.

thiva-k: @coderabbitai Can you verify?

thiva-k: @coderabbitai PublicKeyInfo.Thumbprint is used only for the kid field as of now right?

anushasunkada: @thiva-k you are right, only when JWT is signed with thunderID JWT service. But the problem arises when the thunderID engine is integrated with MOSIP ID system, which signs the userinfo JWT. 

so any such ID systems which is capable of signing the userinfo JWT, those ID system userinfo signing keys must also be hosted in the same oauth2/jwks endpoint. So that is the reason we should stick to the value given in the KeyID field. All the ID system userinfo signing keys will be shared with engine in the RuntimeCryptoProvider.GetPublicKeys method.

thiva-k: Then ThunderID issued JWTs should also be signed with the KeyID field as kid instead of the Thumbprint field, and the redundant Thumbprint field can be removed. Our defaultkm can then set only the KeyID field to the thumbprint. 

This would also stop any embedder whose JWTs ThunderID signs from setting different values for the two fields and breaking signature verification.

This can be refactored later

thiva-k: @coderabbitai verfiy

thiva-k (approved):

PR #5142, opened 2026-08-23
Fixes #5139

### LINKED ISSUE #5139: UserInfo endpoint (and self-issued token middleware) does not validate the `aud` claim on access tokens
### Description

The `/userinfo` endpoint accepts any signed, unexpired ThunderID-issued access token without checking that its `aud` claim matches the userinfo endpoint / this resource server. A token minted for a completely different resource server can be replayed against `/userinfo`.

### Root cause
`ValidateAccessToken` calls `VerifyJWT` with an empty `expectedAud`:

`backend/internal/oauth/oauth2/tokenservice/validator.go:82`
```go
if err := tv.jwtService.VerifyJWT(ctx, token, "", expectedIss); err != nil {
```

In `backend/internal/system/jose/jwt/service.go:465`, the audience check is gated by `if expectedAud != ""`, so passing `""` skips audience validation entirely.

Additionally, `extractAudiences` in `backend/internal/oauth/oauth2/tokenservice/utils.go:101-119` only verifies that `aud` is *present*, never compares its value against anything.

### Impact
Any access token that:
- is signed by this server,
- is unexpired,
- has `sub` / `iss` / `client_id` claims,
- carries the `openid` scope,
- was not issued via `client_credentials`,

will be accepted at `/userinfo` regardless of what resource server its `aud` claim actually names.

**Note:** This is not unique to `/userinfo`. The same pattern (empty `expectedAud` passed to `VerifyJWT`) exists in the general API auth middleware for self-issued tokens

### Steps to Reproduce

Step 1 — Authorization request, requesting openid scope and binding to that resource server:

GET /oauth2/authorize?response_type=code&client_id=<client_id>&redirect_uri=<redirect_uri>&scope=openid+profile&resource=https://api.example.com/payments&state=xyz
After login/consent this redirects back with ?code=<code>&state=xyz.

Step 2 — Token exchange, minting an access token whose aud is the resource server's identifier while still carrying openid:

POST /oauth2/token
Authorization: Basic base64(client_id:client_secret)
Content-Type: application/x-www-form-urlencoded

grant_type=authorization_code&code=<code>&redirect_uri=<redirect_uri>&resource=https://api.example.com/payments
Resulting access token claims: aud: "https://api.example.com/payments", scope: "openid profile", grant_type: "authorization_code".

Step 3 — Replay at /userinfo (this is where the bug fires):

GET /oauth2/userinfo
Authorization: Bearer <access_token_from_step_2>
Expected: rejected, because aud names a different resource server, not this auth server / userinfo endpoint.
Actual: 200 OK with the user's claims — buildResponseFromClaims 

### Version

1.0.0

_No response_

anushasunkada:
### Purpose

The /userinfo endpoint accepts any signed, unexpired ThunderID-issued access token without checking that its aud claim matches the userinfo endpoint / this resource server. A token minted for a completely different resource server can be replayed against /userinfo.

### Approach

validateAudience validates that the access token's audience is the client's own default audience (its configured DefaultAudience, or the client_id when unset). A token whose audience was bound to an external resource server via the 'resource' parameter is scoped to that resource server and must not be redeemable at the UserInfo endpoint.

Note: "client_id" is still considered to support backward compatibility.

### Related Issues
- #5139

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Enhancements**
  * OAuth tokens without an explicitly targeted resource now use the server issuer as their default audience.
  * Explicit resource targets and configured default audiences continue to take precedence.
  * UserInfo now verifies that access tokens target the client’s accepted audience.

* **Bug Fixes**
  * UserInfo rejects tokens intended for unrelated resource servers with a clear invalid-token error.
  * Refresh-token handling correctly distinguishes issuer-audience tokens from resource-bound tokens.

### DIFF
--- a/backend/internal/oauth/oauth2/userinfo/service.go
+++ b/backend/internal/oauth/oauth2/userinfo/service.go
@@ -165,6 +165,10 @@ func (s *userInfoService) buildResponseFromClaims(
 
 	oauthApp := s.getOAuthApp(ctx, tokenClaims)
 
+	if svcErr := s.validateAudience(ctx, accessTokenClaims, oauthApp); svcErr != nil {
+		return nil, svcErr
+	}
+
 	// Extract allowed user attributes
 	var allowedUserAttributes []string
 	if oauthApp != nil && oauthApp.UserInfo != nil {
@@ -404,6 +408,22 @@ func (s *userInfoService) validateGrantType(
 		return &errorClientCredentialsNotSupported
 	}
 
+	return nil
+}
+
+// validateAudience validates that the access token's audience is the client's own default
+// audience (its configured DefaultAudience, or the server issuer ID when unset). A token whose
+// audience was bound to an external resource server via the 'resource' parameter is scoped to
+// that resource server and must not be redeemable at the UserInfo endpoint.
+func (s *userInfoService) validateAudience(
+	ctx context.Context, accessTokenClaims *tokenservice.AccessTokenClaims, oauthApp *providers.OAuthClient,
+) *tidcommon.ServiceError {
+	expectedAud := oauthApp.ResolveDefaultAudience(s.cfg.JWT.Issuer)
+	if !slices.Contains(accessTokenClaims.Aud, expectedAud) {
+		s.logger.Debug(ctx, "UserInfo request token audience does not match the client's default audience",
+			log.String("client_id", accessTokenClaims.ClientID))
+		return &errorAudienceNotAccepted
+	}
 	return nil
 }
 
--- a/backend/internal/oauth/oauth2/granthandlers/authorization_code.go
+++ b/backend/internal/oauth/oauth2/granthandlers/authorization_code.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/thunder-id/thunderid/internal/attributecache"
+	oauthconfig "github.com/thunder-id/thunderid/internal/oauth/config"
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/authz"
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/dpop"
@@ -28,6 +29,7 @@ type authorizationCodeGrantHandler struct {
 	tokenBuilder    tokenservice.TokenBuilderInterface
 	attributeCache  attributecache.AttributeCacheServiceInterface
 	resourceService providers.ResourceServerProvider
+	cfg             oauthconfig.Config
 }
 
 // newAuthorizationCodeGrantHandler creates a new instance of AuthorizationCodeGrantHandler.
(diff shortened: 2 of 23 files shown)


anushasunkada:
@ThaminduDilshan @senthalan - Infact we shouldnt use "client_id" as "aud". Pls check the PR, should we retain backward compatibility?
Instead of client_id fallback should be issuer identifier. What do you think?

backend/internal/oauth/oauth2/userinfo/service.go
```diff
@@ -407,6 +411,22 @@ func (s *userInfoService) validateGrantType(
 	return nil
 }
 
+// validateAudience validates that the access token's audience is the client's own default
+// audience (its configured DefaultAudience, or the client_id when unset). A token whose audience
+// was bound to an external resource server via the 'resource' parameter is scoped to that
+// resource server and must not be redeemable at the UserInfo endpoint.
+func (s *userInfoService) validateAudience(
```
thiva-k, line 418: Tokens issued for external resource servers being able to call userinfo when openid is present is a valid and common usecase right?

We may need to add userinfo/issuer url in the aud array for access tokens if this aud validation at userinfo is needed.

thiva-k: Also this would be a breaking change for userinfo right?

anushasunkada: By design does thunderID support "aud" as array? IMO, we should keep the change easily manageable as this may be a breaking change. Setting the "aud" equal to issuer identifier.

ThaminduDilshan:
@coderabbitai review

backend/internal/oauth/oauth2/userinfo/service.go
thiva-k: On second thought, what additional improvement are we going to gain with aud validation at userinfo endpoint?

If openid scope is present in the access token issued by the authorization server, then that client is eligible to access the userinfo endpoint. Even if we include the issuer/userinfo url in the aud array, it would be redundant since we are always going to add it if openid scope is present and that scope validation is already done at userinfo.

If token theft is an issue, that is already solved by DPoP.

This aud validation at userinfo will make sense only if we need an additional validation(apart from the openid scope validation) on whether the userinfo needs to be restricted to a client on a certain scenario.

Please let know if there is any missing concern here?

anushasunkada: IMO, I think there's still a meaningful gap even with openid scope validation in place.

Scope and audience answer two different questions: `openid` scope says *what* the token is authorized to do, `aud` says *where* it's valid. A token minted with `openid` scope for Resource Server A is, today, equally valid at Resource Server B's /userinfo as long as B trusts the same AS signing key — scope alone doesn't stop that cross-RS replay. That's essentially the confused-deputy scenario RFC 8707 (Resource Indicators) and the OAuth 2.0 Security BCP call out audience restriction as the mitigation for, independent of scope.

DPoP solves a different threat, here the token isn't stolen — it's a legitimately-issued, unexpired, correctly-signed token being presented to the wrong resource server by the legitimate client or a cooperating malicious RS. DPoP doesn't stop that.

That said, I think your underlying question is fair: if in practice all resource servers in a given ThunderID deployment implicitly trust each other (no real trust boundary between RSs), then aud enforcement at /userinfo may seem redundant overhead. I'd lean toward keeping the aud check as the safer default, given it's the same protection RFC 9700 recommends generally.

thiva-k: /userinfo isn't a third-party resource server here. It's always one, it belongs to the AS, and it shares the AS's key pair. Resource Server B doesn't have a /userinfo endpoint to replay against, so I don't think this example justifies the validation.

WDYT @senthalan @ThaminduDilshan @Thumimku ?

PR #5144, opened 2026-08-24
Support --version flag for the thunderid CLI

### LINKED ISSUE #4217: Support --version flag for npx thunderid
### Current Limitation

`npx thunderid` has no way to report its version. There is no `--version` / `-V` flag, and the flag is not listed in `--help`.

Currently `tools/cli/cmd/thunderid/main.go` only handles `upgrade`, `try`, `integrate`, `--help`/`-h`, `--verbose`/`-v`, and `--setup`. Any unrecognized flag (including `--version`) silently falls through to `cli.Run`, which starts the full install/setup/REPL flow instead of printing a version and exiting.

There is also no version embedded in the CLI binary to report: the build (`tools/cli/scripts/build.sh`) compiles with `-ldflags="-s -w"` only, so no version string is injected. This makes it hard for users and bug reports to state exactly which build they are running.

### Suggested Improvement

Add a `--version` (and `-V`) flag to `npx thunderid` that prints the version and exits immediately, without triggering install/setup. For example:

```
$ npx thunderid --version
thunderid 0.2.0
```

Suggested scope:

1. Add `--version` / `-V` handling in `main.go`, short-circuiting before `cli.Run` (same pattern as `--help`).
2. Introduce a build-time version variable (e.g. `var version = "dev"`) injected via `-ldflags "-X main.version=<v>"` in `build.sh` / `build.ps1`, sourced from the npx `package.json` version (currently `0.2.0`) or a release tag. Fall back to `dev` for local `go run` builds.
3. List `--version` in `printUsage()`.
4. Consider reporting both the CLI/npx package version and the active installed ThunderID product version when an install exists.

This is a small, self-contained CLI enhancement and improves supportability (version is easy to capture in bug reports).

pasindubalasooriya:
### Purpose
`npx thunderid` had no way to report its version. There was no `--version` / `-V` flag, the flag was not listed in `--help`, and any unrecognized flag fell through to `cli.Run`, so `npx thunderid --version` started the full install and setup flow instead of printing a version and exiting.

There was also no version embedded in the binary to report, since both build scripts compiled with `-ldflags="-s -w"` only. This makes it hard for users and bug reports to state exactly which build they are running.

```
$ npx thunderid --version
thunderid 1.0.2
```

### Approach
**Flag handling.** `main.go` gets a `--version` / `-V` branch placed next to the existing `--help` handling, short-circuiting before `cli.Run`. It follows the same shape as `--help`: matched on `args[0]`, prints and returns. `--version` is also listed in `printUsage()`. `-V` is used because `-v` is already taken by `--verbose`.

**Version injection.** A package-level `var version = "dev"` in `main.go` is populated at build time via `-ldflags "-X main.version=<v>"`, and stays `dev` for local `go run` and `go build`.

The value is sourced from `tools/npx/package.json`, which is what npm actually publishes. Rather than parsing JSON in both bash and PowerShell, `tools/npx/scripts/build.js` reads it (trivial in Node, and it already orchestrates both scripts) and passes it through the `VERSION` environment variable. `build.sh` and `build.ps1` each default to `dev` when `VERSION` is unset, so running either script directly still works.

No change is needed in `.github/workflows/release-tools.yml` or the `release-npm` action: the action runs `pnpm version` to bump `package.json` before building and publishing, so the published binary picks up the new version automatically.

**Scope.** This reports the CLI version only. The CLI also tracks the active ThunderID server install (`config.ReadActiveVersion`), but that version is already surfaced through the install and upgrade output, so `--version` is kept unambiguous and about the tool itself.

**Verification.** Beyond the unit tests, the full chain was exercised on each build path:

| Command | Output |
|---|---|
| `go build -ldflags "-X main.version=9.9.9"` | `thunderid 9.9.9` |
| `go build` (no ldflags) | `thunderid dev` |
| `VERSION=7.7.7 bash scripts/build.sh` | `thunderid 7.7.7` |
| `build.ps1` with `$env:VERSION=8.8.8` | `thunderid 8.8.8` |
| `node scripts/build.js` (shipped path) | `thunderid 1.0.2`, matching `package.json` |

Note: the issue mentions the npx version as `0.2.0`; it is currently `1.0.2`. Nothing in the approach depends on the specific value.

### Related Issues
- Fixes #4217

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

Unit tests in `tools/cli/cmd/thunderid/main_test.go`: the injected version renders as `thunderid <v>`, `version` defaults to `dev` when nothing is injected, and `printUsage()` lists the new flag.

The flag is self-documenting through `--help`, so no `docs/` page describes the CLI flag set today. Happy to add documentation if maintainers would like it covered.

### Security checks

---

Note for reviewers: this touches `tools/cli/scripts/build.sh`, `build.ps1`, and `tools/npx/scripts/build.js`, which the issue anticipated in its suggested scope. Happy to adjust the injection point if you would prefer the version come from the release tag instead of `package.json`.

## Summary by CodeRabbit

* **New Features**
  * Added `--version` and `-V` options to display the CLI product version and exit.
  * Version flags are recognized regardless of their position in the command.
  * Built CLI binaries now include the package version.

* **Documentation**
  * Updated CLI usage help and documentation with the new version options.

* **Bug Fixes**
  * Standardized version embedding across build environments, with a consistent fallback version when unavailable.

### DIFF
--- a/tools/cli/cmd/thunderid/main.go
+++ b/tools/cli/cmd/thunderid/main.go
@@ -16,8 +16,19 @@ import (
 	"github.com/thunder-id/thunderid/tools/cli/internal/ui"
 )
 
+// version is the CLI version. It is injected at build time via
+// -ldflags "-X main.version=<v>" and keeps the placeholder below for local builds.
+var version = "0.0.0-semantically-released"
+
 func main() {
 	args := os.Args[1:]
+
+	// --version is resolved before command dispatch so it always prints and exits,
+	// rather than falling through to a command or to install and setup.
+	if hasVersionFlag(args) {
+		fmt.Println(versionLine())
+		return
+	}
 
 	// upgrade — stop the running version, install the latest, restart on the same port.
 	if len(args) > 0 && args[0] == "upgrade" {
@@ -66,6 +77,21 @@ func main() {
 	cli.Run(verbose, forceSetup)
 }
 
+// hasVersionFlag reports whether args request the CLI version, in any position.
+func hasVersionFlag(args []string) bool {
+	for _, a := range args {
+		if a == "--version" || a == "-V" {
+			return true
+		}
+	}
+	return false
+}
+
+// versionLine renders the version string printed by --version.
+func versionLine() string {
+	return fmt.Sprintf("%s %s", product.Slug, version)
+}
+
 func printUsage() {
 	fmt.Printf(`Usage: %s [command] [flags]
 
@@ -77,6 +103,7 @@ Commands:
 Flags:
   --verbose, -v        Show detailed output
   --setup              Force re-run setup
+  --version, -V        Show the CLI version
   --help, -h           Show this help message
 `, product.Slug, product.Name)
 }
--- a/tools/cli/cmd/thunderid/main_test.go
+++ b/tools/cli/cmd/thunderid/main_test.go
@@ -1,0 +1,91 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+)
+
+// captureStdout runs fn with os.Stdout redirected and returns what it printed.
+func captureStdout(t *testing.T, fn func()) string {
+	t.Helper()
+
+	r, w, err := os.Pipe()
+	if err != nil {
+		t.Fatalf("failed to create pipe: %v", err)
+	}
+
+	original := os.Stdout
+	os.Stdout = w
+	defer func() { os.Stdout = original }()
+
+	fn()
+
+	if err := w.Close(); err != nil {
+		t.Fatalf("failed to close pipe writer: %v", err)
+	}
+
+	out, err := io.ReadAll(r)
+	if err != nil {
+		t.Fatalf("failed to read captured output: %v", err)
+	}
+
+	return string(out)
+}
+
+func TestVersionLineUsesInjectedVersion(t *testing.T) {
+	original := version
+	t.Cleanup(func() { version = original })
+
+	version = "1.2.3"
+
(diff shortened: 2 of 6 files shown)


tools/cli/cmd/thunderid/main.go
```diff
@@ -77,6 +103,7 @@ Commands:
 Flags:
   --verbose, -v        Show detailed output
   --setup              Force re-run setup
+  --version, -V        Show the CLI version
```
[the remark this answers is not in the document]
pasindubalasooriya: I did not create docs/content/guides/cli-version.mdx, because there is no CLI documentation under docs/ to extend. Added to tools/npx/README.md instead.

tools/cli/cmd/thunderid/main.go
```diff
@@ -16,9 +16,20 @@ import (
 	"github.com/thunder-id/thunderid/tools/cli/internal/ui"
 )
 
+// version is the CLI version. It is injected at build time via
+// -ldflags "-X main.version=<v>" and stays "dev" for local builds.
+var version = "dev"
```
brionmario: @coderabbitai Would it be better to use an actual semantic version as a placeholder here?

Like:

```go
var version = "0.0.0"
```

brionmario: @pasindubalasooriya Shall we use `0.0.0-semantically-released` as the placeholder instead of `dev`?

pasindubalasooriya: @brionmario I have fixed this

brionmario (approved):

brionmario:
I've merged your PR @pasindubalasooriya.
This will be released under `npx thunderid@next` in the future once we do a release from the `main branch`.

Thanks a lot for the contribution. Keep it up 🙌

PR #5179, opened 2026-08-25
Revoke application artifacts through administration flows

indeewari:
### Purpose

Deleting an application, or rotating its client secret, is a security event: the artifacts already issued to that application's OAuth client must stop being accepted. Neither action revoked anything before this change, so a deleted application's signed bearer tokens stayed valid until they expired, and a rotated secret left every token minted under the old one working.

Both actions now run through an administration flow, so the revocation, the session detachment and the action itself happen as one orchestrated sequence. Two flows ship by default:

| Flow | Handle |
|---|---|
| Application deletion | `default-application-deletion-flow` |
| Client secret regeneration | `default-client-secret-regeneration-flow` |

The console drives both. Its delete action and its regenerate-secret action resolve the configured handle, execute the flow, and fall back to the native endpoint only when no such flow is configured, matching how user deletion already works.

### Approach

**Four new executors, two flows.** The preparatory node (`PreApplicationDeleteExecutor` / `PreSecretRegenerationExecutor`) validates the target and publishes a trusted revocation plan into shared runtime data. The criteria and session nodes act on that plan. The final node (`ApplicationDeleteExecutor` / `ClientSecretRegenerationExecutor`) performs the delete or the rotation. Every refusal lands in the preparatory node, before anything has been revoked, rather than surfacing mid-flow with a deny-list row already written.

One type backs both preparatory names. They differ only in which validation runs, which reason is recorded, and whether a cutoff is stamped, so those are parameterized at construction. They stay two registered executors so each pins a single supported mode, which flow creation validates, and so the designer palette names the action rather than hiding it in a property.

**Terminal vs bounded revocation.** Deletion records `application_deleted` in `ModeAll`: the client id is retired with the application, so no future artifact can legitimately carry it and there is nothing to unfreeze. Rotation records `application_secret_regenerated` in `ModeBeforeAction` with a cutoff, so only artifacts established at or before that instant are rejected and tokens minted with the new secret carry a later `iat` and pass.

**Deny-list row lifetime.** A row is sized from the target application's own token validity plus the authorization-code window, not from the deployment default. Token validity is configured per application and is not capped by that default, so a row sized from it would go inert while the client's tokens were still valid, and both enforcement points would stop matching it. The authorization-code window is included because a code issued before the revocation can still be exchanged for a token after it.

**The native endpoints are unchanged.** `DELETE /applications/{id}` and `PUT /applications/{id}` behave exactly as before, so revocation has one home rather than two. That also means a caller who bypasses the flow gets no revocation, which is the deliberate trade: the flow is the orchestrator, and duplicating the revocation in the service would write the same deny-list row twice on the flow path.

**Server config.** `applicationDeletionFlow` and `clientSecretRegenerationFlow` join `userDeletionFlow` in the `flow` server-config section, with handle validation and layer merging. A deployment that unsets either handle keeps the native behaviour, so this is opt-out without a separate mode switch.

**Secret entropy.** The rotation flow's response is the only moment the new secret is readable, and the server generates it. The console's fallback path still generates one in the browser, because the update endpoint has no way to ask the server for one.

**Bug fixed along the way.** A refused flow step reports its executor error in the response's `error` envelope (`{code, message, description}`). The console was reading a `failureReason` field that `flowexec.FlowResponse` has never sent, so a refusal's reason and code were discarded and every failure showed a generic message. Both the new application paths and the existing user-deletion path now read the real envelope and carry the code on the thrown error, and the flow-executor error codes have console catalog entries.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests (suite added; not yet executed against a rebuilt distribution)

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added administration flows for deleting applications and regenerating client secrets.
  * Application deletion now revokes issued tokens and detaches the application from SSO sessions.
  * Client-secret regeneration revokes artifacts associated with the previous secret and returns the new secret once.
  * Console actions use configured flows, with native API fallback only when no flow is configured.
* **Bug Fixes**
  * Improved flow error handling, localized messages, validation, and fail-closed behavior.
  * Enhanced token revocation to recognize application-specific credentials and lifetimes.


backend/internal/flow/executor/application_action_validator.go
rajithacharith, line 1: Can we rename these to ValidateApplicationDeletion something like.. without using PreApplication...

backend/cmd/server/servicemanager.go
```diff
@@ -368,6 +372,9 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 			ResourceService:       resourceServerProvider,
 			UserService:           userService,
 			CriteriaRevoker:       revocationSvc,
+			ApplicationProvider: func() providers.ApplicationAdminProvider {
```
rajithacharith: Why we need to deviate from the existing pattern for ApplicationProvider?

backend/internal/application/error_constants.go
```diff
@@ -575,4 +575,18 @@ var (
 				"and string-typed in an allowed user type",
 		},
 	}
+	// ErrorApplicationHasNoClientSecret is the error returned when a client secret regeneration targets an
+	// application that authenticates without one: a public client, or one using private_key_jwt.
+	ErrorApplicationHasNoClientSecret = tidcommon.ServiceError{
```
rajithacharith, line 607: What is the current behaviour? 
Is the validation missing in current implementation?

backend/internal/application/init.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 The ThunderID Authors
+// Copyright 2025-2026 The ThunderID Authors
```
rajithacharith: This doesn't need to update? 
Are we following this @senthalan ?

backend/internal/application/service.go
```diff
@@ -82,6 +90,71 @@ func newApplicationService(
...
+//
+// It exists because token validity is configured per application and is not capped by the deployment
+// default, while the revocation service sizes rows from that deployment-wide default alone. An
+// application whose validity was raised in the console would otherwise outlive its own revocation: the
+// row goes inert on the default while artifacts issued to the client are still valid, and both
+// enforcement points stop matching it.
+//
+// Zero means "no opinion", which leaves the revocation service on its configured default. That is the
+// outcome when the client cannot be read or the runtime config is not loaded, so neither blocks the
+// operation that is revoking. Mirrors how the authorize service sizes the user-attribute cache.
```
rajithacharith: Shall we make this short? Feel like no one will ever read this

backend/internal/application/service.go
```diff
@@ -82,6 +90,71 @@ func newApplicationService(
...
+	// sub-config, while client_credentials reads the client one. Only the longest of the two bounds
+	// how long an access token issued to this client can live, so sizing from either alone leaves the
+	// row inert while tokens minted under the other are still valid.
+	seconds := tokenservice.ResolveTokenConfig(cfg, client, tokenservice.TokenTypeAccess,
+		client.UserAccessTokenConfig().ValidityPeriodOrZero()).ValidityPeriod
+	clientAccess := tokenservice.ResolveTokenConfig(cfg, client, tokenservice.TokenTypeAccess,
+		client.ClientAccessTokenConfig().ValidityPeriodOrZero()).ValidityPeriod
+	if clientAccess > seconds {
+		seconds = clientAccess
+	}
```
rajithacharith: Shall we use variable names which represents the true meaning of this 
ex: accessTokenValidity, clientAccessValidity <- not exactly the same

Then we can remove the above comment as well, WDYT?

backend/internal/application/service.go
```diff
@@ -82,6 +90,71 @@ func newApplicationService(
...
+// is still available while a delete is in progress.
+func applicationClientID(e *providers.Entity) string {
+	if e == nil || len(e.SystemAttributes) == 0 {
+		return ""
+	}
+	var sysAttrs map[string]interface{}
+	if err := json.Unmarshal(e.SystemAttributes, &sysAttrs); err != nil {
+		return ""
+	}
+	clientID, _ := sysAttrs[fieldClientID].(string)
```
rajithacharith: Don't we need to handle this err?

backend/internal/application/service.go
```diff
@@ -597,6 +671,130 @@ func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
...
+		return nil, svcErr
+	}
+	if as.inboundClientService.IsDeclarative(ctx, appID) {
+		return nil, &ErrorCannotModifyDeclarativeResource
+	}
+	if as.dependencyRegistry == nil {
+		as.logger.Error(ctx, "Dependency registry not set; refusing to delete application",
+			log.String("appID", appID))
+		return nil, &tidcommon.InternalServerError
+	}
```
rajithacharith, line 695: Don't we need to validate with dependencyRegistry whether there are dependencies? 
Does the application has blocking dependent relationship with anyone?

backend/internal/application/service.go
```diff
@@ -597,6 +671,130 @@ func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
...
+	if err != nil || client == nil {
+		as.logger.Error(ctx, "Failed to load OAuth client before secret regeneration",
+			log.String("appID", appID))
+		return nil, &tidcommon.InternalServerError
+	}
+	if client.PublicClient ||
+		(client.TokenEndpointAuthMethod != providers.TokenEndpointAuthMethodClientSecretBasic &&
+			client.TokenEndpointAuthMethod != providers.TokenEndpointAuthMethodClientSecretPost) {
+		return nil, &ErrorApplicationHasNoClientSecret
+	}
```
rajithacharith: Can we use application type for this? Will having `TokenEndpointAuthMethodClientSecretBasic` guarantee the existance of a client secret?
Don't we have existing validation for these?

backend/internal/flow/executor/application_action_validator.go
indeewari: Fixed

backend/internal/application/init.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

.github/backend-coverage-thresholds.yml
```diff
@@ -37,8 +37,15 @@ default: 70
 # Applies regardless of labels, but only when the applicable tags say nothing
 # about the package. A tag that sets a `default` therefore overrides these, so an
 # exemption recorded here does not survive such a label.
-packages: {}
-  # backend/internal/system/i18n: 0
+packages:
```
ThaminduDilshan, line 40: This change will affect other PRs which can actually cover the missing lines. Shall we revert this?

backend/cmd/server/bootstrap/01-default-resources.yaml
```diff
@@ -1991,6 +1991,168 @@ nodes:
...
+      width: 206
+      height: 113
+    position:
+      x: 300
+      y: 175
+- id: validate_secret_regeneration
+  type: TASK_EXECUTION
+  executor:
+    name: ValidateSecretRegenerationExecutor
+    mode: revoke_before_action
```
ThaminduDilshan, line 2113: Are we planning to introduce more modes in future or is this the sole mode that we're going to have?

backend/cmd/server/bootstrap/01-default-resources.yaml
```diff
@@ -1991,6 +1991,168 @@ nodes:
...
+    size:
+      width: 206
+      height: 113
+    position:
+      x: 920
+      y: 175
+- id: regenerate_secret
+  type: TASK_EXECUTION
+  executor:
+    name: ClientSecretRegenerationExecutor
```
ThaminduDilshan: We have some other future use cases related to client secret generation.
- Generate additional client secrets
- Rotate a given/ all client secret/s
- Regenerate client secret

Rotation can be a flavour of regeneration where we keep the existing secret valid for a certain period of time. Shall we keep this executor generic (rename/ modify files names, etc) so that it can be extended to support these other use cases in future?

ThaminduDilshan: Also so far we're keeping the flow level abstract of spec specific details. With this, we're onboarding oauth specific executor to the flow. I think we don't have a option for a component like this. Or else we can make it a generic app secret generator

backend/cmd/server/servicemanager.go
```diff
@@ -405,10 +412,12 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	consentEnforcer.SetConsentService(initConsentService(ctx, logger, inboundClientService))
 
 	// TODO: Remove entityService dependency after finalizing declarative resource loading pattern
-	applicationService, applicationExporter, err := application.Initialize(
+	appService, applicationExporter, err := application.Initialize(
 		mux, mcpServer, entityProvider, entityService, inboundClientService, ouService, i18nService,
 		runtimeCryptoSvc, serverConfigService)
 	fatalOnError(ctx, logger, err, "Failed to initialize ApplicationService")
+	// Publish the service to the resolver the administration-flow executors already hold.
+	applicationService = appService
```
ThaminduDilshan: We can directly assign this to the variable rather than declaring another right?

backend/cmd/server/servicemanager.go
ThaminduDilshan: This diverges from the two-phase injection pattern already used in the same wiring path.

`FlowConfigHandler.SetHandleValidator` (server_config.go:38) injects a function post-construction via a setter, and ouService.SetOUFlowResolver (servicemanager.go:389) / consentEnforcer.SetConsentService (servicemanager.go:412) do the same for services. Can we expose a SetApplicationProvider on the executor registry and call it after appService is initialized, instead of embedding a closure over a later-mutated applicationService variable in ExecutorDependencies?

backend/internal/application/init_test.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 The ThunderID Authors
+// Copyright 2025-2026 The ThunderID Authors
```
ThaminduDilshan: Do we need this? None of the code files were modified in this PR

backend/internal/application/service.go
```diff
@@ -45,6 +48,11 @@ type ApplicationServiceInterface interface {
 		ctx context.Context, appID string, app *model.ApplicationDTO) (
 		*model.ApplicationDTO, *tidcommon.ServiceError)
 	DeleteApplication(ctx context.Context, appID string) *tidcommon.ServiceError
+	ValidateDeleteApplication(ctx context.Context, appID string) (
+		*providers.ApplicationRevocationTarget, *tidcommon.ServiceError)
+	ValidateRegenerateClientSecret(ctx context.Context, appID string) (
```
ThaminduDilshan: Client secret is specific to oauth, however thunderid applications can be of different types which doesn't hold a client secret. 
We can either try to introduce a generic credential validation method (1) or keep the pattern consistent with existing `GetOAuthApplication()` method (2). I prefer the first option, but if not possible, we can goahead with second.
In our current pattern, application has to be of one of the type. Hence it cannot hold multiple types of secrets together. Since we get the id, inside service we are able to identify the app type. With that we should be able to determine the applicable secret type and do the validations. This way IMO we should be able to have a generic credential validation method.

Same applies to `RegenerateClientSecret` method as well.

backend/internal/application/service.go
```diff
@@ -45,6 +48,11 @@ type ApplicationServiceInterface interface {
 		ctx context.Context, appID string, app *model.ApplicationDTO) (
 		*model.ApplicationDTO, *tidcommon.ServiceError)
 	DeleteApplication(ctx context.Context, appID string) *tidcommon.ServiceError
+	ValidateDeleteApplication(ctx context.Context, appID string) (
```
ThaminduDilshan, line 49: I don't think app validation result (either deletion/ secret regeneration) should be a revocation target. Revocation is a specific usecase that uses validation, but validation is not specific to revocation right?
Shall we try to generalize these two methods?

backend/internal/application/service.go
ThaminduDilshan: Also IMO better if we can keep the method signature generic so that it can be used for other listed secret generation/rotation requirements in future without introducing dedicated methods for these use cases

backend/internal/application/service.go
```diff
@@ -82,6 +90,73 @@ func newApplicationService(
 	}
 }
 
+// resolveArtifactLifetime returns how long a deny-list row against this application's OAuth client
+// must survive: the longest validity any artifact issued to that client can carry, plus the
+// authorization-code window. Sizing from the deployment default alone would let an application whose
+// validity was raised outlive its own revocation.
+//
+// Zero means "no opinion", which leaves the revocation service on its configured default.
+func (as *applicationService) resolveArtifactLifetime(ctx context.Context, clientID string) time.Duration {
```
ThaminduDilshan: Shall we order the files to have public service methods on top, then private service methods and helper methods as per their usages to increase the readability?

backend/internal/application/service.go
```diff
@@ -82,6 +90,73 @@ func newApplicationService(
...
+		as.logger.Warn(ctx, "Failed to resolve application token validity for revocation; "+
+			"falling back to the deployment default", log.MaskedString("clientID", clientID))
+		return 0
+	}
+
+	if !config.IsServerRuntimeInitialized() {
+		return 0
+	}
+
+	cfg := oauthconfig.FromServerRuntime()
```
ThaminduDilshan: This method isn't supposed to be called at service level and only from service manager. Let's remove this

backend/internal/application/service.go
```diff
@@ -82,6 +90,73 @@ func newApplicationService(
...
+// Zero means "no opinion", which leaves the revocation service on its configured default.
+func (as *applicationService) resolveArtifactLifetime(ctx context.Context, clientID string) time.Duration {
+	client, err := as.inboundClientService.GetOAuthClientByClientID(ctx, clientID)
+	if err != nil || client == nil {
+		as.logger.Warn(ctx, "Failed to resolve application token validity for revocation; "+
+			"falling back to the deployment default", log.MaskedString("clientID", clientID))
+		return 0
+	}
+
+	if !config.IsServerRuntimeInitialized() {
```
ThaminduDilshan: This feels like a unnecessary method call. Not sure why this was introduced in the first place. Comment explain it is added only for tests

backend/internal/application/service.go
```diff
@@ -82,6 +90,73 @@ func newApplicationService(
...
+		return 0
+	}
+
+	cfg := oauthconfig.FromServerRuntime()
+	// User-subject tokens and client_credentials tokens read separate validity sub-configs, so only
+	// the longest of the two bounds how long an access token issued to this client can live.
+	userAccessValidity := tokenservice.ResolveTokenConfig(cfg, client, tokenservice.TokenTypeAccess,
```
ThaminduDilshan: Is this a correct logic to determine how long the deny list entry should be kept? This determines the lifetime of the revocation entry based on the app's configs right? But if the app configs are updated after generating some tokens (have higher validity period initially and reduce later), this provides a incorrect result. Isn't it?

IMO we can't correctly determine the longest validity period for a issued token without looking at actually issued tokens which we cannot do without storing the token details.

backend/internal/application/service.go
```diff
@@ -82,6 +90,73 @@ func newApplicationService(
...
+}
+
+// applicationClientID returns the OAuth client id recorded on the application entity, or an empty
+// string when the application genuinely has none. It reads the entity rather than the inbound client so
+// the value is still available while a delete is in progress.
+//
+// An unreadable attribute blob is an error rather than an empty id: callers size a revocation from this
+// value, and an empty one means "nothing to revoke". Returning it for a blob that merely failed to parse
+// would delete the application while its issued artifacts stay valid.
+func applicationClientID(e *providers.Entity) (string, error) {
```
ThaminduDilshan: Shall we rename this method to be something meaningful?

backend/cmd/server/bootstrap/01-default-resources.yaml
indeewari: There are two modes for the moment, there can be modes upcoming 
https://github.com/thunder-id/thunderid/blob/49b77967d1976dd4c87f2ed99115a7cada5890c8/backend/internal/revocation/model.go#L58

backend/cmd/server/bootstrap/01-default-resources.yaml
indeewari: Generalizing the name - Accepted.

There are  OAuthExecutor, OIDCAuthExecutor, GithubOAuthExecutor, GoogleOIDCAuthExecutor already. Lets clear out the abstraction we need to preserve.

.github/backend-coverage-thresholds.yml
indeewari: Fixed

backend/internal/application/service.go
```diff
@@ -582,21 +658,165 @@ func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
...
 		return false
 	}
-	if cfg.PublicClient {
+	return authMethodRequiresClientSecret(cfg.PublicClient, cfg.TokenEndpointAuthMethod)
+}
+
+// authMethodRequiresClientSecret reports whether a client authenticating this way holds a client secret.
+// Shared by the write paths that decide whether to store one and by the regeneration path that decides
+// whether there is one to rotate, so the two cannot disagree on the same client.
+func authMethodRequiresClientSecret(publicClient bool, method providers.TokenEndpointAuthMethod) bool {
```
ThaminduDilshan, line 599: Shall we try to merge appRequiresClientSecret and authMethodRequiresClientSecret functions without introducing new methods unnecessarily? Both serves the same functionality, just accepts different parameters.

backend/cmd/server/bootstrap/01-default-resources.yaml
ThaminduDilshan: > There are OAuthExecutor, OIDCAuthExecutor, GithubOAuthExecutor, GoogleOIDCAuthExecutor already. Lets clear out the abstraction we need to preserve.

These are executors related to external IDPs, not protocol specific client implementations right? These are of two different categories IMO

backend/internal/application/service.go
ThaminduDilshan: I feel like methods are being added to random places. Personally I prefer if we can order them aligning with how we have ordered these methods in the interface

backend/internal/application/service.go
```diff
@@ -582,21 +658,165 @@ func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
...
+		as.logger.Error(ctx, "Failed to persist regenerated client secret", log.Error(entErr),
+			log.String("appID", appID))
+		return "", &tidcommon.InternalServerError
+	}
+	return secret, nil
+}
+
+// loadApplicationEntity returns the application's entity, mapping a missing or non-application record to
+// the not-found error the API surfaces.
+func (as *applicationService) loadApplicationEntity(ctx context.Context, appID string) (
```
ThaminduDilshan, line 809: Can we reuse this function in the `getApplication()` method too to avoid duplicate code?

backend/internal/application/service.go
ThaminduDilshan: I also feel like we should validate dependencies

backend/internal/application/service.go
```diff
@@ -582,21 +658,165 @@ func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
...
+	clientID, err := clientIDFromEntity(existing)
+	if err != nil {
+		as.logger.Error(ctx, "Failed to read the application's OAuth client id",
+			log.String("appID", appID), log.Error(err))
+		return nil, &tidcommon.InternalServerError
+	}
+	if clientID == "" {
+		return nil, &ErrorApplicationHasNoClientSecret
+	}
+	client, err := as.inboundClientService.GetOAuthClientByClientID(ctx, clientID)
```
ThaminduDilshan, line 765: We're first loading the entity, validation and then loading the oauth profile. Can't we simply use the existing app/ oauth client retrieval methods?

.github/backend-coverage-thresholds.yml
ThaminduDilshan: Is this fixed?

backend/cmd/server/bootstrap/01-default-resources.yaml
```diff
@@ -1991,6 +1991,168 @@ nodes:
...
+    size:
+      width: 206
+      height: 113
+    position:
+      x: 300
+      y: 175
+- id: validate_application_deletion
+  type: TASK_EXECUTION
+  executor:
+    name: ValidateApplicationDeletionExecutor
```
ThaminduDilshan: Alternative suggestion: `ApplicationDeletionValidator` or `AppDeletionValidator`. We don't necessarily need to have executor part in executor names

backend/cmd/server/bootstrap/01-default-resources.yaml
```diff
@@ -1991,6 +1991,168 @@ nodes:
...
+    size:
+      width: 85
+      height: 34
+    position:
+      x: 1760
+      y: 200
+---
+resource_type: flow
+id: 01900000-0000-7000-8000-00000000007a
+name: Default Client Secret Regeneration Flow
```
ThaminduDilshan: I'm thinking whether we should keep these terms generic to accommodate future use cases like regenerating flow secret. Or otherwise we'll ended up writing multiple flows for each secret type.

backend/cmd/server/bootstrap/01-default-resources.yaml
```diff
@@ -1991,6 +1991,168 @@ nodes:
...
+    size:
+      width: 206
+      height: 113
+    position:
+      x: 300
+      y: 175
+- id: validate_secret_regeneration
+  type: TASK_EXECUTION
+  executor:
+    name: ValidateSecretRegenerationExecutor
```
ThaminduDilshan: Suggestion: `SecretRegenerationValidator` or more generally `SecretGenerationValidator`

backend/cmd/server/servicemanager.go
indeewari: Fixed

backend/internal/application/init_test.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/cmd/server/servicemanager.go
indeewari: Fixed

backend/internal/application/error_constants.go
```diff
@@ -588,4 +588,18 @@ var (
...
+		Type: tidcommon.ClientErrorType,
+		Code: "APP-1047",
+		Error: tidcommon.I18nMessage{
+			Key:          "error.applicationservice.application_has_no_client_secret",
+			DefaultValue: "Application has no client secret",
+		},
+		ErrorDescription: tidcommon.I18nMessage{
+			Key:          "error.applicationservice.application_has_no_client_secret_description",
+			DefaultValue: "The application authenticates without a client secret, so there is none to regenerate",
```
ThaminduDilshan: Better to refine the error description saying app doesn't have a client secret/ client secret is not applicable/ or required for this application

backend/internal/flow/config/config.go
```diff
@@ -33,13 +33,19 @@ type FlowTypeConfig struct {
 // UserDeletionFlow names the administration flow that carries out a user deletion. There is no
 // separate mode switch: a console deletes through the flow when one is configured and present, and
 // through DELETE /users/{id} otherwise, mirroring how user onboarding falls back to manual creation.
+//
+// ApplicationDeletionFlow and ClientSecretRegenerationFlow name the administration flows for the two
```
ThaminduDilshan: Shall we avoid adding unnecessary comments? Existing comment is also too bulky

backend/internal/flow/executor/application_delete_executor.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// setApplicationProvider injects the application service. See ExecutorRegistryInterface.
+func (e *applicationDeleteExecutor) setApplicationProvider(provider providers.ApplicationAdminProvider) {
+	e.provider = provider
+}
+
+// applicationProvider returns the injected application service, or an error when the second-phase
+// injection never ran.
+func (e *applicationDeleteExecutor) applicationProvider() (providers.ApplicationAdminProvider, error) {
```
ThaminduDilshan: Do we need this additional method? Id on't see a value of it

backend/internal/flow/executor/application_workflow_executors_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/thunder-id/thunderid/internal/system/cache"
+	"github.com/thunder-id/thunderid/internal/system/config"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+	"github.com/thunder-id/thunderid/tests/mocks/applicationadminprovidermock"
+	"github.com/thunder-id/thunderid/tests/mocks/flow/sessionmock"
+	"github.com/thunder-id/thunderid/tests/mocks/oauth/oauth2/revocationmock"
+)
+
+type ApplicationWorkflowExecutorsTestSuite struct {
```
ThaminduDilshan: Can we try to have separate test files for each source file? Go's convention is foo.go → foo_test.go, and here `application_workflow_executors_test.go` covers two independent source files.

I don't prefer to have this combined test pattern. This is raised as a followup comment in the previous PR for such existing tests as well

backend/internal/flow/executor/client_secret_executor.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/thunder-id/thunderid/internal/revocation"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// clientSecretExecutor acts on the client secret of the application named in the trusted revocation
+// plan. It regenerates the secret, and runs after the artifacts issued under the old secret have been
+// denied.
+//
+// The name is action-neutral while the behavior is not. An executor name is persisted into stored
```
ThaminduDilshan: Isn't this is a implementation time instruction? Do we need to keep this in the code

backend/internal/flow/executor/client_secret_executor.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// setApplicationProvider injects the application service. See ExecutorRegistryInterface.
+func (e *clientSecretExecutor) setApplicationProvider(provider providers.ApplicationAdminProvider) {
+	e.provider = provider
+}
+
+// applicationProvider returns the injected application service, or an error when the second-phase
+// injection never ran.
+func (e *clientSecretExecutor) applicationProvider() (providers.ApplicationAdminProvider, error) {
```
ThaminduDilshan: Same as previous comment. Do we need this method?

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

.github/backend-coverage-thresholds.yml
indeewari: Had to revert due to the coverage, will increase the coverage and fix this

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: This is already a known gap and will be done via a followup issue, as performance is also a concern. Currently we are modelling this concern altogether with role/group case as well.

- https://github.com/thunder-id/thunderid/issues/5292

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: These two look similar but handle a missing entity in opposite ways on purpose. `getApplication()` keeps going when the entity is gone and builds the response from the inbound client, while `loadApplicationEntity` treats that as application-not-found. Reusing it in `getApplication()` would make the GET API start returning 404 for an application whose entity is missing, so I have left them separate.

.github/backend-coverage-thresholds.yml
indeewari: Fixed

backend/internal/flow/executor/constants.go
```diff
@@ -72,6 +76,14 @@ const (
 	userInputConsentDecisions = "consent_decisions"
 	userInputLoginHint        = "login_hint"
 	revocationInputSubject    = "subject"
+	// revocationInputApplication names the application an administrative flow acts on. It is deliberately
+	// not "applicationId": the flow execution request already carries a top-level applicationId meaning
+	// the application a flow runs for, and reusing that name one level down would invite supplying the
+	// value in the wrong place, where the failure is a flow that pauses asking for input.
+	revocationInputApplication = "targetApplicationId"
```
ThaminduDilshan, line 78: Shall we avoid having bulky comments in all the places? Let's keep it consistent with the existing code.
It's okay to have a smaller comment if absolutely needed

backend/internal/flow/executor/constants.go
```diff
@@ -72,6 +76,14 @@ const (
...
+	// dataKeyClientSecret carries a regenerated client secret back to the caller. AdditionalData is the
+	// only executor output the engine serializes, and this is the single moment the value is readable.
+	dataKeyClientSecret = "clientSecret" // #nosec G101 -- response field name, not a secret
```
ThaminduDilshan: Shall we move this to backend/internal/flow/common/constants.go where we have defined all other data keys?
Also avoid having bulky comments

backend/internal/flow/executor/criteria_revocation_executor.go
```diff
@@ -30,6 +31,13 @@ func newCriteriaRevocationExecutor(factory core.FlowFactoryInterface,
 }
 
 // Execute records every criterion from the trusted revocation plan.
+//
+// A plan that declares NothingToRevoke writes nothing and completes: the preparatory node established
```
ThaminduDilshan: Do we need this much of a comment?

backend/pkg/thunderidengine/providers/interface.go
```diff
@@ -130,6 +130,26 @@ type FlowProvider interface {
 	GetFlow(ctx context.Context, flowID string) (*CompleteFlowDefinition, *common.ServiceError)
 }
 
+// ApplicationAdminProvider exposes the application operations an administration flow performs. It is a
+// seam rather than a direct dependency because the application service is built after the flow
+// executors and sits behind them in the import graph.
+type ApplicationAdminProvider interface {
```
ThaminduDilshan: Do we need this public interface? Having this in pkg will make this exported from thunderid lib. We should avoid exposing unnecessary stuff

backend/internal/flow/executor/register.go
```diff
@@ -58,6 +59,26 @@ func newExecutorRegistry() ExecutorRegistryInterface {
 	}
 }
 
+// applicationProviderConsumer is implemented by the executors that act on applications. They are
+// built before the application service exists, so it reaches them in a second phase.
+type applicationProviderConsumer interface {
```
ThaminduDilshan: Shall we keep the interface name consistent with it's functionality? Maybe appProviderConsumerInterface or something like that?

Also IMO all the executors and functions that require such functionality can refer this interface rather than a interface in pkg or some other place. If this is the consumer interface, all the consumers can reuse it right?
With this we should be able to get rid of the pkg interface

backend/internal/flow/executor/session_revocation_executor.go
```diff
@@ -31,7 +31,16 @@ func newSessionRevocationExecutor(factory core.FlowFactoryInterface,
...
+//
+// A subject plan ends that subject's sessions outright, since all of them belong to the user being acted
+// on. An application plan instead detaches only that application's participation, because an SSO session
+// routinely spans several applications and ending it would sign the user out of ones that still exist.
+//
+// The application branch keys off the plan's target rather than its criteria, and therefore runs even when
+// there is nothing to revoke: participation is recorded per application whether or not it ever held a
+// token, so an application with no OAuth component can still be a participant. It applies to a deletion
+// only, never to a secret rotation.
 func (e *sessionRevocationExecutor) Execute(ctx *providers.NodeContext) (*providers.ExecutorResponse, error) {
```
ThaminduDilshan, line 38: Do we need such bulky comment?

backend/internal/flow/executor/utils.go
```diff
@@ -49,12 +64,32 @@ func decodeRevocationPlan(data map[string]string) (revocationPlan, error) {
 	if err := json.Unmarshal([]byte(encoded), &plan); err != nil {
 		return revocationPlan{}, fmt.Errorf("failed to decode trusted revocation plan: %w", err)
 	}
-	if len(plan.Criteria) == 0 {
+	if len(plan.Criteria) == 0 && !plan.NothingToRevoke {
```
ThaminduDilshan, line 63: Is it the correct behavior to fail the execution if there's nothing to revoke? I feel like this is a no-op and should return a successful response

backend/internal/flow/executor/validate_application_action_executor.go
```diff
@@ -0,0 +1,164 @@
...
+	*providers.ApplicationArtifactProfile, *tidcommon.ServiceError)
+
+// validateApplicationActionExecutor is the preparatory node of an application administration flow. It
+// validates the target and publishes the trusted revocation plan the nodes after it consume, the same
+// contract PreDeleteExecutor holds for user deletion.
+//
+// One type backs both registered names. The nodes differ only in which validation runs, which reason is
+// recorded, and whether a cutoff is stamped, so they are parameterized at construction rather than
+// duplicated. They stay two registered executors so each pins a single supported mode, which flow
+// creation validates, and so the designer palette names the action rather than hiding it in a property.
```
ThaminduDilshan: Okay it looks like two executor related codes are added to the same shared code base, but only in the implementation. Executor `mode` has been introduced to support this use case; where the same executor can behave in different ways at runtime.

Can't we follow the same pattern here rather than exposing these as separate executors to the runtime?

backend/internal/flow/executor/validate_application_action_executor.go
```diff
@@ -0,0 +1,164 @@
...
+
+// setApplicationProvider injects the application service. See ExecutorRegistryInterface.
+func (e *validateApplicationActionExecutor) setApplicationProvider(
+	provider providers.ApplicationAdminProvider) {
+	e.provider = provider
+}
+
+// Execute validates the target application and publishes the trusted plan the later nodes act on.
+//
+// The mode is not read from the node: each registered name supports exactly one, so a node carrying
```
ThaminduDilshan: This is not a expected use case for mode. Related to previous comment.

backend/internal/flow/session/service.go
```diff
@@ -62,6 +62,12 @@ type Service interface {
 	// grants first so no session is deleted while its tokens remain live. It is idempotent, returning
 	// nil when the subject holds no sessions.
 	TerminateBySubject(ctx context.Context, subjectID string) error
+
+	// RemoveApplication detaches an application from every SSO session it participates in: it revokes
+	// the token family of that participation and drops the participant row, deleting the session itself
+	// only when the application was its last participant. It is idempotent, returning nil when the
+	// application participates in none.
+	RemoveApplication(ctx context.Context, appID string) error
```
ThaminduDilshan: Maybe we should rename this method. It doesn't delete the application, rather detaches the session entries right?
Method name could be misleading

backend/internal/flow/session/service.go
```diff
@@ -371,6 +377,67 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
 
 // revokeSessionFamilies revokes the token family of every application participating in the session,
 // so signing out of a login drops all of that login's grants. It is a no-op when no family revoker is
+// RemoveApplication detaches an application from every SSO session it participates in, for use when the
```
ThaminduDilshan: This method and method comment is added in middle of revokeSessionFamilies() method's comment

backend/internal/flow/session/service.go
```diff
@@ -371,6 +377,67 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
...
+// is left to participate: at that point it can no longer back SSO for anything.
+//
+// Each participation's token family is revoked before its row is dropped, because the row is the only
+// record linking the session to that grant, so losing it first would leave the grant unrevocable. All
+// writes share one transaction, so a partial failure leaves every session intact rather than some subset
+// detached.
+//
+// Applications that hold no tokens still participate: TFID is empty for a participant that joined
+// without a grant, and revoking an empty family is a no-op, so such a participation is simply dropped.
+func (s *service) RemoveApplication(ctx context.Context, appID string) error {
```
ThaminduDilshan: Comment is too bulky. We don't have to explain the entire function in the method comment right?
Check this for all the places. This can be seen in many places

backend/internal/flow/session/service.go
```diff
@@ -371,6 +377,67 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
...
+			if delErr := s.store.DeleteSession(txCtx, participation.SessionID); delErr != nil {
+				return delErr
+			}
+			if delErr := s.store.Delete(txCtx, participation.SessionID); delErr != nil {
+				return delErr
+			}
+		}
+		return nil
+	}); txErr != nil {
+		return fmt.Errorf("failed to remove application from sessions: %w", txErr)
```
ThaminduDilshan: Suggestion: "failed to remove session for the application"

backend/internal/flow/session/store_constants.go
```diff
@@ -108,6 +108,22 @@ var (
 		Query: `DELETE FROM "SSO_SESSION" WHERE SESSION_ID = $1 AND DEPLOYMENT_ID = $2`,
 	}
 
+	// queryListParticipantsByAppID returns every session the application participates in, oldest first.
```
ThaminduDilshan, line 132: Shall we move these two constants to the end of constant block so that they're in ascending order?

backend/internal/flow/session/service_test.go
```diff
@@ -672,3 +672,122 @@ func (suite *ServiceTestSuite) TestTerminate_DeleteError() {
...
+
+	err := svc.RemoveApplication(context.Background(), "app-embedded")
+
+	suite.Require().NoError(err)
+}
+
+// The revocation precedes the row delete, since the participant row is the only record linking the
+// session to that grant: a failed revocation must leave the row in place rather than orphaning a live
+// family. The whole detachment shares one transaction, so nothing is dropped.
+func (suite *ServiceTestSuite) TestRemoveApplication_RevocationFailureAbortsTheDetachment() {
```
ThaminduDilshan: Not a hard requirement, but adding for info. We don't need to add method comments for the tests

backend/internal/oauth/oauth2/revocation/service_test.go
```diff
@@ -285,3 +285,68 @@ func TestRevokeTokenFamily_NonPositiveTTLFallsBack(t *testing.T) {
 	assert.NoError(t, err)
 	assert.WithinDuration(t, captured.RevokedAt.Add(defaultTokenFamilyRevocationTTL), captured.ExpiryTime, time.Second)
 }
+
+// A caller that knows its artifacts outlive the configured lifetime can ask for a longer row, which is
+// what an application-scoped revocation needs: token validity is per application and uncapped, so the
+// deployment-wide default can expire the row while matching tokens are still valid.
+func TestRevokeByCriteria_RequestedTTLExtendsTheRow(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

backend/internal/revocation/model.go
```diff
@@ -81,6 +81,15 @@ type CriteriaRevocation struct {
...
 	Reason    Reason
+	// TTL is how long the deny-list row must survive, for a producer that knows the lifetime of the
+	// artifacts its criterion matches. A criterion scoped to one application is the case that needs it:
+	// token validity is configurable per application and uncapped, so the deployment-wide default the
+	// revocation service falls back to can expire the row while matching artifacts are still valid.
+	//
+	// It raises the row's lifetime and never lowers it: the service writes the longer of this and its own
+	// default, so a zero value keeps the default and no producer can shorten a row by supplying a small
+	// value.
+	TTL time.Duration
```
ThaminduDilshan, line 87: Too bulky comment

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -1046,6 +1046,29 @@ type Application struct {
 	Metadata           map[string]interface{}        `yaml:"metadata,omitempty" json:"metadata,omitempty" jsonschema:"Generic metadata key-value pairs."`
 }
 
+// ApplicationArtifactProfile describes the artifacts an application issues: the OAuth client they
+// carry, and the longest any of them can live. It is a fact about the application rather than a plan
+// for one operation on it. A revocation is one consumer, which sizes a deny-list row from the lifetime.
+type ApplicationArtifactProfile struct {
```
ThaminduDilshan: Do we need all of these changes in the pkg? We should be cautious about what we're adding to the code in pkg. Adding unnecessary stuff would get exposed from the thunderid lib and we won't be able to later modify/remove them without a major release version

backend/internal/system/revocationcache/query_constants.go
```diff
@@ -20,9 +20,10 @@ var querySnapshotRevokedTokenFamilies = dbmodel.DBQuery{
 		`WHERE CRITERION_TYPE = $1 AND EXPIRY_TIME > $2 AND DEPLOYMENT_ID = $3`,
 }
 
-// querySnapshotRevokedSubjects reads subject criteria together with the reason and action boundary
-// needed to distinguish permanent revocations from time-bounded revocations.
-var querySnapshotRevokedSubjects = dbmodel.DBQuery{
+// querySnapshotBoundedCriteria reads the criteria of one dimension together with the reason and action
+// boundary needed to distinguish permanent revocations from time-bounded ones. The criterion type is a
+// bind parameter, so a new dimension reuses this query rather than adding another.
+var querySnapshotBoundedCriteria = dbmodel.DBQuery{
```
ThaminduDilshan, line 25: This is one disadvantage of having bulky comment. We have to update them every time when a related code gets updated even none of the implementation line near that code is changed.
And these can get outdated very frequently too if some developer missed updating them

backend/internal/application/init.go
```diff
@@ -31,9 +32,10 @@ func Initialize(
 	i18nService i18nmgt.I18nServiceInterface,
 	cryptoSvc providers.RuntimeCryptoProvider,
 	serverConfigSvc serverconfig.ServerConfigService,
+	oauthCfg oauthconfig.Config,
```
ThaminduDilshan: I prefer to plug a separate config defined in application package (check some other packages. This pattern should already be there) rather than passing the oauth config here. Same comment applies to the service too.
App is not oauth specific and is common for many other types of apps. It doesn't feel right to pass a oauth specific config here

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/flow/executor/constants.go
indeewari: Fixed

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/flow/session/store_constants.go
indeewari: Fixed

backend/internal/flow/session/service_test.go
indeewari: Fixed

backend/internal/oauth/oauth2/revocation/service_test.go
indeewari: Fixed

backend/internal/application/init.go
indeewari: Fixed

backend/internal/flow/executor/constants.go
indeewari: Fixed

backend/internal/flow/executor/criteria_revocation_executor.go
indeewari: Fixed

backend/internal/flow/executor/session_revocation_executor.go
indeewari: Fixed

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/revocation/model.go
indeewari: Fixed

backend/internal/system/revocationcache/query_constants.go
indeewari: Fixed

backend/cmd/server/bootstrap/01-default-resources.yaml
indeewari: Fixed

backend/cmd/server/bootstrap/01-default-resources.yaml
indeewari: Fixed

backend/internal/application/error_constants.go
indeewari: Fixed

backend/internal/flow/executor/client_secret_executor.go
indeewari: Fixed

backend/internal/flow/executor/register.go
indeewari: Fixed

backend/pkg/thunderidengine/providers/interface.go
indeewari: Fixed

backend/pkg/thunderidengine/providers/model.go
indeewari: Fixed

backend/internal/flow/executor/constants.go
indeewari: Fixed

backend/internal/flow/executor/criteria_revocation_executor.go
indeewari: Fixed

backend/internal/flow/executor/session_revocation_executor.go
indeewari: Fixed

backend/internal/flow/session/service.go
indeewari: Fixed

backend/internal/revocation/model.go
indeewari: Fixed

backend/internal/system/revocationcache/query_constants.go
indeewari: Fixed

backend/internal/flow/config/config.go
indeewari: Fixed by having a generic comment for admin flow vs native API call

backend/internal/flow/executor/application_delete_executor.go
indeewari: Fixed

backend/cmd/server/bootstrap/01-default-resources.yaml
indeewari: Fixed

backend/internal/flow/executor/utils.go
indeewari: It does not fail when there is nothing to revoke. A plan that establishes there is nothing to deny sets NothingToRevoke, and the criteria node then writes nothing and completes successfully.
This check only rejects a plan that has no criteria and does not set that flag, which means the criteria were lost on the way rather than never existing.

backend/internal/flow/executor/validate_application_action_executor.go
indeewari: Fixed

backend/internal/flow/executor/validate_application_action_executor.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Corrected. My earlier reply was premature: the two calls had only moved from service.go into config.go in the same package, so neither was actually removed.

The application package now takes a Config from the composition root and no longer reads the server runtime at all. The lifetime computation moved to tokenservice.ArtifactLifetime, where the OAuth configuration belongs, and servicemanager supplies the resolver.

backend/internal/application/service.go
indeewari: Corrected. The guard had only moved into config.go in the same package, so my earlier reply was wrong.

It is gone now: the application package takes a Config from the composition root instead of reading the server runtime, so there is nothing to guard against being uninitialised.

backend/internal/application/init.go
indeewari: Fixed

backend/internal/application/error_constants.go
```diff
@@ -588,4 +588,47 @@ var (
...
+			Key: "error.applicationservice.application_has_blocking_dependencies_description",
+			DefaultValue: "The application cannot be deleted because other resources depend on it. " +
+				"Remove or reassign them first",
+		},
+	}
+	// ErrorApplicationHasNoClientSecret is the error returned when a client secret regeneration targets an
+	// application that authenticates without one: a public client, or one using private_key_jwt.
+	ErrorApplicationHasNoClientSecret = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "APP-1047",
```
ThaminduDilshan: Shall we please keep error constants in the correct order? All APP-1047, APP-1048 and APP-1049 are random placed

backend/internal/application/config.go
```diff
@@ -51,3 +53,12 @@ func isDeclarativeModeEnabled() bool {
 func getAppIndexedAttributes() []string {
 	return []string{"name", "clientId"}
 }
+
+// Config holds the deployment settings this package depends on. The composition root supplies them, so
+// the package neither reads global runtime state nor takes another domain's configuration type.
+type Config struct {
```
ThaminduDilshan: 1. Shall we place this in a config sub package (application/config/config.go)?

2. Then define a FromServerRuntime() function that will get invoked from the servicemanager.go file.

    ```go
    // FromServerRuntime builds flow configuration from the global server runtime.
    func FromServerRuntime() Config {
    ```

Refer backend/internal/flow/config/config.go

ThaminduDilshan: Wait. But this is determining the artifact lifetime at runtime based on client config calculations right?
In that case this is not the correct pattern. This config model should only be used to pass static server configs/ like any deployment.yaml configs.

In that case discard my previous comment and let's get this removed. This should be retrieved from runtime injecting required services

backend/internal/application/service.go
```diff
@@ -577,23 +589,23 @@ func seedClientSubTypeAttribute(processedDTO *model.ApplicationProcessedDTO) {
...
 }
 
-// appRequiresClientSecret reports whether the OAuth config implies a confidential client requiring a secret.
-func appRequiresClientSecret(cfg *providers.OAuthConfigWithSecret) bool {
-	if cfg == nil {
-		return false
-	}
-	if cfg.PublicClient {
+// authMethodRequiresClientSecret reports whether a client authenticating this way holds a client secret.
+// Shared by the write paths that decide whether to store one and by the regeneration path that decides
```
ThaminduDilshan: Do we need this in the comment?

backend/internal/flow/common/constants.go
```diff
@@ -145,6 +145,8 @@ const (
 	// pre-processing node produces for the executors that follow. It travels on the engine context's
 	// cross-frame store, so it survives a CALL into another flow.
 	RuntimeKeyRevocationPlan = "revocationPlan"
+	// AdditionalDataKeyClientSecret carries a regenerated client secret back to the caller.
+	AdditionalDataKeyClientSecret = "clientSecret" // #nosec G101 -- response field name, not a secret
```
ThaminduDilshan: Why is this placed in middle of runtime keys? We already have a place where we define all additional data keys

backend/internal/flow/config/config.go
```diff
@@ -30,16 +30,17 @@ type FlowTypeConfig struct {
 // FlowSectionConfig is the value of the server-config "flow" section. It carries per-type default
 // handles and context TTLs. A zero ExpirySeconds falls back to the built-in default for that type.
 //
-// UserDeletionFlow names the administration flow that carries out a user deletion. There is no
-// separate mode switch: a console deletes through the flow when one is configured and present, and
-// through DELETE /users/{id} otherwise, mirroring how user onboarding falls back to manual creation.
+// An administration flow runs when its handle is configured and the flow exists; otherwise the native
```
ThaminduDilshan: Is this comment needed to be here?

backend/internal/flow/executor/register.go
```diff
@@ -367,3 +402,15 @@ func registerBuiltInExecutor(
 	}
 	return nil
 }
+
+// ApplicationAdminProvider is the application seam the administration executors consume. It is declared
+// here rather than in pkg so it stays internal: the application service satisfies it structurally.
+type ApplicationAdminProvider interface {
```
ThaminduDilshan: Shall we make this private? We shouldn't expose unnecessary interfaces to other packages

backend/internal/flow/executor/validate_application_action_executor.go
```diff
@@ -0,0 +1,165 @@
...
+		},
+		reason:   revocation.ReasonApplicationSecretRegenerated,
+		notFound: ErrSecretRegenerationNotAllowed,
+	},
+}
+
+// validateApplicationActionExecutor is the preparatory node of an application administration flow: it
+// validates the target and publishes the trusted revocation plan the later nodes consume. The node's
+// mode selects the action, so one executor serves every application administration flow.
+type validateApplicationActionExecutor struct {
```
ThaminduDilshan: Shall we rename;
1. Executor struct to `applicationActionValidator`
2. Constructor method to `newApplicationActionValidator`
3. File name to `application_action_validator.go`

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -698,3 +700,30 @@ func resolveClientGroupRoleClaims(
 	}
 	return claims, nil
 }
+
+// ArtifactLifetime returns the longest an artifact issued to this client can remain valid: the longest
+// access-token validity across both token subjects, the refresh-token validity when the client may use
+// it, plus the authorization-code window.
+func ArtifactLifetime(cfg oauthconfig.Config, client *providers.OAuthClient) time.Duration {
```
ThaminduDilshan: I still have a question whether this is the correct way to determine artifact validity time. If the client or configs get updated, it will honor a invalid validity time right?

backend/internal/flow/executor/validate_application_action_executor.go
indeewari: x

backend/internal/flow/executor/client_secret_executor.go
indeewari: Fixed

backend/internal/application/error_constants.go
indeewari: Fixed

backend/internal/application/config.go
indeewari: Fixed

backend/internal/application/service.go
indeewari: Fixed

backend/internal/flow/common/constants.go
indeewari: Fixed

backend/internal/flow/config/config.go
indeewari: Fixed

backend/internal/flow/executor/validate_application_action_executor.go
indeewari: Fixed

backend/internal/flow/executor/application_workflow_executors_test.go
indeewari: Fixed

backend/internal/flow/executor/register.go
indeewari: Fixed

backend/internal/application/config.go
ThaminduDilshan: Second concern is not addressed right? Seems first comment is addressed, but that conflicts with the concern raised in second comment

backend/internal/flow/common/constants.go
```diff
@@ -324,6 +324,8 @@ const (
 
 // ForwardedData key constants define keys used in the ForwardedData map.
 const (
+	// AdditionalDataKeyClientSecret carries a regenerated client secret back to the caller.
+	AdditionalDataKeyClientSecret = "clientSecret" // #nosec G101 -- response field name, not a secret
```
ThaminduDilshan: Now this is in the forwarded data key constant block. If this is a additional data key, let's move it to line 95.

https://github.com/thunder-id/thunderid/blob/b3f4c434e4c2d15913f9f6daa4529c32f98169d2/backend/internal/flow/common/constants.go#L64-L95

backend/internal/flow/executor/application_action_validator.go
```diff
@@ -0,0 +1,165 @@
...
+		return nil, err
+	}
+	execResp.SharedRuntimeData = map[string]string{common.RuntimeKeyRevocationPlan: plan}
+	return execResp, nil
+}
+
+// buildApplicationPlan records the trusted intent for this action. An application with no OAuth client
+// publishes no criteria and says so explicitly, which keeps it distinguishable from a plan that lost its
+// criteria.
+func buildApplicationPlan(mode revocation.Mode, reason revocation.Reason, appID string,
```
ThaminduDilshan: Prefer to keep this as a struct method, otherwise this will be available for the all executors in the package

backend/internal/flow/common/constants.go
indeewari: Fixed

backend/internal/flow/executor/application_action_validator.go
indeewari: Fixed

backend/internal/application/config.go
indeewari: You are right, and the two asks do pull against each other. Putting FromServerRuntime() in application/config means that package imports oauthconfig and tokenservice, so the OAuth dependency moved one directory deeper rather than going away.

Only one of the two can hold:

A. application/config declares the Config type only, and servicemanager builds the resolver from tokenservice.ArtifactLifetime. The application domain then has no OAuth import at all, but there is no FromServerRuntime() in application/config.

B. Keep FromServerRuntime() as it is now, and the application config sub-package keeps the OAuth import.

I would pick A, since the composition root is the one place that legitimately sees both domains. Happy to switch to it, or leave B if you prefer the FromServerRuntime() shape. Let me know which you want.

indeewari: Went with option A.

application/config now declares the Config type only, and servicemanager builds the resolver from tokenservice.ArtifactLifetime. The application packages carry no OAuth configuration dependency and read no global runtime state; the composition root supplies both.

There is no FromServerRuntime() in application/config as a result, which is the part of your earlier suggestion this trades away.

indeewari: Correcting my previous reply. Wrapping the resolver in a Config type was wrong: a function supplied at wiring time is a collaborator, not configuration, so it did not belong in a config file or a config package at all.

The application/config sub-package is removed. The application service now declares an artifactLifetimeResolver and takes it as a constructor dependency alongside the other services, and servicemanager passes a closure over tokenservice.ArtifactLifetime. No configuration type, no config package, no OAuth configuration dependency in the application package, and no global runtime read.

ThaminduDilshan (approved):

PR #5183, opened 2026-08-25
Add End-to-End tests for the CLI

### LINKED ISSUE #3954: [E2E] End-to-End tests for the CLI
### Current Limitation

The CLI currently lacks end-to-end (E2E) test coverage to validate complete user workflows. While individual components and utilities may be tested, there are no automated tests that verify the CLI behaves correctly from a user's perspective, including command execution, interactive prompts, project scaffolding, and generated outputs.

### Suggested Improvement

Introduce a comprehensive E2E test suite for the CLI that validates common user journeys across supported platforms. The tests should cover scenarios such as:

* CLI installation and execution.
* Interactive command flows and prompt handling.
* Project scaffolding with different templates and configurations.
* Validation of generated project structure and files.
* Error handling for invalid inputs and failed operations.
* Regression tests for previously reported CLI issues.

This will improve confidence in releases, prevent regressions, and ensure the CLI provides a consistent experience across environments.

brionmario:
### Purpose

The CLI under `tools/cli` had no end to end coverage, and nothing under `tools/` was built, linted or tested by any PR workflow. This adds an e2e suite that drives the CLI the way users actually invoke it, plus a workflow to run it.

It also unbreaks two pre existing failures on `main` that the new workflow gates on, so the gate is green from the start rather than red on arrival.

### Approach

**The suite drives `npx thunderid`, not the Go binary.** Specs spawn `node tools/npx/bin/thunderid.js` under a real pseudo terminal, so the wrapper, the download and extract path, `setup.sh`, the port negotiation and the bubbletea REPL are exercised as one chain. Nothing is stubbed: a run reaches the release manifest, downloads a real distribution, boots the server, and finally launches the Wayfinder sample and signs into it in a browser. 18 specs, about two minutes warm.

**Playwright Test plus node-pty, in `tests/e2e-cli`.** A pseudo terminal is required rather than a pipe. The CLI branches on `ui.Interactive()` to decide whether to prompt at all, so a piped run would assert the scripted fallbacks instead of the paths users see, and would cover neither the port conflict prompt, the credential prompt, the onboarding picker, nor the REPL.

**Why a separate Playwright project.** `tests/e2e` starts a backend on port 8090 through its `webServer` before the first test runs. Here the CLI is the thing that installs and starts ThunderID, and it needs that port free, so sharing the configuration is not possible. The new project defines no browser projects and runs no `playwright install`, since it drives a terminal.

**Isolation.** Each spec gets a temp `HOME`, which relocates `~/.thunderid/state.json`, and a temp working directory, which relocates the `./thunderid/vX` install. Every `THUNDERID_*` variable is stripped from the child environment so local shell configuration cannot change what is under test.

**Structure.** Specs are grouped by the part of the CLI they exercise, one Playwright project per folder, ordered by the project dependency graph (the same mechanism `tests/e2e` uses for `wayfinder-setup`):

| Project | Depends on | Covers |
|---|---|---|
| `commands` | nothing | `--help`, `-h`, `integrate`, `try` with no install |
| `install` | nothing | Manifest, download, extract, `setup.sh`, start, `state.json`, files on disk |
| `repl` | `install` | Onboarding picker and command overlay, `/status`, `/logs`, `/open-console`, `/integrate-react`, `/stop` |
| `restart` | `install` | Warm start, `--setup` |
| `port-conflict` | `install` | Alternate port, abort, reclaim |
| `sample` | all of the above | Unknown sample name, and `try wayfinder` through a browser login |

**Cost.** An extracted distribution is roughly 85MB, so `install` performs it once and everything downstream reuses it, warm starting in seconds. Only two specs download: `install`, and the first-run port conflict, which needs an install setup has not yet pinned to a port. The suite is single worker by necessity: every spec binds the default port or deliberately holds it.

Deliberately not covered, with reasons recorded in `tests/e2e-cli/README.md`: `/try-consumer` from inside the REPL, a real `upgrade`, and the offline fallback. The last one is unreachable until the releases URL becomes injectable; it is a constant in `internal/product` today.

**Three findings from writing the sample spec**, all worth a look in review:

- **`try consumer` is not a valid command.** The only sample name the CLI accepts is `wayfinder` (`sample.go:56-64`); `consumer` exists only as the REPL slash command `/try-consumer`. So the name a user is most likely to type from the REPL fails on the command line. The error does name what is available, and `unknown-usecase.spec.ts` now pins that message down, but the mismatch may be worth reconsidering.
- **`try` leaves everything running.** It installs no signal handler and never calls `StopServices`, so it exits with npm and all five sample services still up, plus the ThunderID server it restarted. The `Press Ctrl+C to stop.` line it prints is misleading, since the command has already returned. The spec cleans up after itself.
- **The sample spec runs the CLI without a pty**, unlike every other spec. `try` exits as soon as the services are up, and closing a pty at that point SIGHUPs the ThunderID server it just backgrounded, which broke the browser login with a discovery failure until I traced it. A real terminal is owned by the shell, so the pipe is the honest model here.

**node-pty needs two pieces of workspace wiring**, both worth a look in review:

- A `node-pty: true` entry in `allowBuilds`. Without it pnpm skips the install script and no prebuilt binary is fetched.
- A `postinstall` that restores the executable bit on node-pty's `spawn-helper`. The prebuild's mode bits are lost on extraction into the pnpm store, after which the addon loads fine but every spawn fails with the opaque `posix_spawnp failed`. This bit me during development and would bite anyone on a fresh install.

**Pre existing fixes included here.** The new `unit` job runs `go test ./...` and golangci-lint over `tools/cli`, and both failed on `main`:

- `TestExit_UndeliverableSignalLeavesThePortAlone` failed deterministically. Commit 2da93ee30 moved `if !explicit` into an `else if` so an undeliverable signal always falls through to the port stop, which is what stops an exit from stranding the server it started. The test still encoded the previous behavior and was never updated. It is now renamed and asserts the intended behavior.
- 23 golangci-lint findings, all in committed code. Long lines are wrapped, plus `gofmt`, `gocritic`, `ineffassign` and two misspells. `cancelling` joins `cancelled` in `ignore-words`, since the codebase consistently uses the British spelling in user facing strings.

The line wrapping touches user facing walkthrough copy, so every change was verified to preserve string contents exactly: the SHA-256 of all string literal contents in source order is unchanged in each of the three files touched.

**One behavior worth flagging for review.** The onboarding picker reappears on every run until a use case is actually selected, because `onboardingDone` is only recorded by `ui.selectOnboarding`. The specs document this rather than assert it, in case it is not intended.

### Related Issues
- Fixes https://github.com/thunder-id/thunderid/issues/3954

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive cross-platform end-to-end coverage for CLI installation, startup, REPL commands, port conflicts, restarts, and sample applications.
  * Added automated validation for CLI help, integration flows, readiness, authentication, and console launching.
  * Added continuous integration checks for CLI linting, unit tests, and end-to-end tests on Ubuntu and macOS.

* **Bug Fixes**
  * Improved cleanup behavior so servers started by the CLI are stopped when exiting after an undeliverable signal.

* **Documentation**
  * Added instructions for running and extending the CLI end-to-end test suite.

### DIFF
--- a/tools/i18n-extractor/generator.go
+++ b/tools/i18n-extractor/generator.go
@@ -68,12 +68,7 @@ func (g *Generator) Generate(messages []ExtractedMessage, outputPath string) err
 	sortedMessages := make([]MessageEntry, 0, len(keys))
 	for _, k := range keys {
 		msg := messageMap[k][0] // Take the first one since we are sure there are no duplicates now
-		sortedMessages = append(sortedMessages, MessageEntry{
-			Key:          msg.Key,
-			DefaultValue: msg.DefaultValue,
-			SourceFile:   msg.SourceFile,
-			Line:         msg.Line,
-		})
+		sortedMessages = append(sortedMessages, MessageEntry(msg))
 	}
 
 	data := struct {
--- a/.github/actions/setup-cli-e2e/action.yml
+++ b/.github/actions/setup-cli-e2e/action.yml
@@ -1,0 +1,22 @@
+name: Setup CLI E2E Suite
+description: >-
+  Installs the tests/e2e-cli pnpm workspace and Chromium, the two setup steps every
+  job that runs the CLI e2e suite needs. Assumes Node.js and pnpm are already set up.
+
+runs:
+  using: composite
+  steps:
+    # tests/e2e-cli is its own pnpm workspace root, deliberately outside the repository's, so
+    # its node-pty dependency cannot force a native build on every other job. See the comment
+    # in tests/e2e-cli/pnpm-workspace.yaml.
+    - name: 📦 Install E2E Dependencies
+      shell: bash
+      working-directory: tests/e2e-cli
+      run: pnpm install --frozen-lockfile
+
+    # Only chromium, and only because the sample spec signs in through a browser. Every other
+    # project in this suite drives a terminal and needs no engine at all.
+    - name: 🌐 Install Chromium
+      shell: bash
+      working-directory: tests/e2e-cli
+      run: pnpm exec playwright install --with-deps chromium
--- a/.github/actions/upload-cli-e2e-report/action.yml
+++ b/.github/actions/upload-cli-e2e-report/action.yml
@@ -1,0 +1,24 @@
+name: Upload CLI E2E Report
+description: Uploads the Playwright report and test results from tests/e2e-cli. Meant to run on failure.
+
+inputs:
+  name:
+    description: Artifact name
+    required: true
+  retention-days:
+    description: How long to keep the artifact
+    required: false
+    default: '7'
+
+runs:
+  using: composite
+  steps:
+    - name: 📄 Upload Playwright Report
+      uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
+      with:
+        name: ${{ inputs.name }}
+        path: |
+          tests/e2e-cli/playwright-report/
+          tests/e2e-cli/test-results/
+        if-no-files-found: ignore
+        retention-days: ${{ inputs.retention-days }}
--- a/.github/actions/verify-cli-e2e-ports/action.yml
+++ b/.github/actions/verify-cli-e2e-ports/action.yml
@@ -1,0 +1,29 @@
+name: Verify CLI E2E Ports Are Free
+description: >-
+  Fails the job if any port the CLI e2e suite needs is already bound on the runner,
+  rather than letting the suite fail later with a confusing port-conflict assertion.
+
+inputs:
+  extra-ports:
+    description: >-
+      Additional space-separated ports to check beyond the suite's fixed set (8090 for
+      ThunderID, 5173/8787/8788/2525/8795 for the Wayfinder sample). Used for a local
+      release server, for example.
+    required: false
+    default: ''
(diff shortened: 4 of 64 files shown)


tools/i18n-extractor/generator.go
```diff
@@ -68,12 +68,7 @@ func (g *Generator) Generate(messages []ExtractedMessage, outputPath string) err
 	sortedMessages := make([]MessageEntry, 0, len(keys))
 	for _, k := range keys {
 		msg := messageMap[k][0] // Take the first one since we are sure there are no duplicates now
-		sortedMessages = append(sortedMessages, MessageEntry{
-			Key:          msg.Key,
-			DefaultValue: msg.DefaultValue,
-			SourceFile:   msg.SourceFile,
-			Line:         msg.Line,
-		})
+		sortedMessages = append(sortedMessages, MessageEntry(msg))
```
brionmario, line 71: The linter was failing here. So had to do this fix to solve.
cc: / @ThumulaPerera

DonOmalVindula (approved):

Issue #5109, opened 2026-08-20
Let a client select its ID token signing algorithm via `id_token_signed_response_alg`


Thumimku:
### Current Limitation

The server signs ID tokens with a single algorithm chosen server-side. A client cannot request
one, even though the server supports more than one and says so in discovery:

```
id_token_signing_alg_values_supported: ["RS256", "ES256"]
```

`id_token_signed_response_alg` is defined by OIDC Registration §2 for exactly this purpose, and
dynamic client registration accepts the parameter, but the value is discarded: it does not
appear in the registration response and has no effect on the tokens the client subsequently
receives. Silently accepting it is tracked separately as a bug; this issue is the missing
capability behind it.

The signing machinery itself is not the gap. Multiple algorithms are already supported and
advertised, ID token construction runs through
`backend/internal/oauth/oauth2/tokenservice/builder.go`, and `jose/jwt` takes an `alg` override
(`backend/internal/system/jose/jwt/service.go:97`). What is missing is per-client selection:
storing the client's choice at registration and honouring it at signing time.

Two Dynamic OP modules test this, `oidcc-idtoken-rs256` and `oidcc-userinfo-rs256`. Neither has
been run: that profile is blocked earlier by the absence of RFC 7592 management. So this
limitation is established from the discovery document and the DCR response rather than from a
conformance failure.

The equivalent for UserInfo, `userinfo_signed_response_alg`, has the same gap.
`userinfo_signing_alg_values_supported` is advertised in
`backend/internal/oauth/oauth2/discovery/service.go`, and a client cannot select from it either.

### Suggested Improvement

Accept `id_token_signed_response_alg` at registration, persist it with the client, and use it
when building that client's ID tokens, falling back to the server default when the client did
not ask for one. Validate the requested value against
`id_token_signing_alg_values_supported` at registration and reject an unsupported one with
`invalid_client_metadata`, rather than accepting and ignoring it.

Handle `userinfo_signed_response_alg` the same way, since it is the same shape of change against
an already-advertised capability list.

Both values should appear in the registration response once stored, so a client can confirm what
the server kept, matching how `logo_uri` and the other retained metadata already behave.

Thumimku:
this was fixed by https://github.com/thunder-id/thunderid/pull/5198

PR #5198, opened 2026-08-28
Support client-selected ID token and UserInfo signing algorithms

### LINKED ISSUE #5195: UserInfo signing fails for every algorithm except the server's preferred key's
### Description

userinfo_signed_response_alg is accepted at registration, echoed back in the DCR response, and persisted, but it only works when the requested value happens to equal the algorithm of the server's preferred_key_id. On a default install that is RS256. Any other value produces HTTP 500 on every subsequent UserInfo call.

The parameter is validated nowhere in DCR. ES256 and bogus-alg are indistinguishable to the client, though they are entirely different situations. ES256 is a legitimate request: the server advertises it in discovery and publishes a P-256 key for it in JWKS. So the server advertises ES256, publishes the key to verify it, accepts the client's request for it, then returns 500 when asked to use it.

Cause. Three defects compound.

No validation at registration. buildUserInfoConfig (backend/internal/oauth/oauth2/dcr/service.go:270) copies request.UserInfoSignedResponseAlg into UserInfoConfig.SigningAlg unchecked. Nothing compares it against supported algorithms, and nothing rejects strings that are not algorithms at all. ErrorInvalidClientMetadata already exists (backend/internal/oauth/oauth2/dcr/error_constants.go:58) and is used for other metadata problems; it is not applied here.

jwtService is bound to one key for the process lifetime. newJWTService (backend/internal/system/jose/jwt/service.go:62) resolves preferred_key_id once at startup and stores a single jwsAlg and kid. The alg parameter on GenerateJWT is therefore not an override but a guard (service.go:105):

if alg != "" {
    if alg != js.jwsAlg {
        return "", 0, &ErrorUnsupportedJWSAlgorithm
    }
}
The layers beneath are not the constraint. Sign(ctx, keyRef, alg, content) (backend/pkg/thunderidengine/providers/interface.go:275) already takes key and algorithm per call, SignAlgorithmFor maps ES256 to ECDSA-SHA256, and signing with the P-256 key produces a correct 64 byte JOSE R||S signature that verifies. Only service.go:197 prevents it, by passing the two frozen fields:

signature, err := js.cryptoProvider.Sign(ctx, js.keyRef, js.jwsAlg, []byte(signingInput))
A client error is reported as a server error. ErrorUnsupportedJWSAlgorithm is declared ClientErrorType (JWT-1003). UserInfo recognises this exact case, has a dedicated branch for it, then discards the classification (backend/internal/oauth/oauth2/userinfo/service.go:340):

if err.Code == jwt.ErrorUnsupportedJWSAlgorithm.Code {
    s.logger.Error(ctx, "UserInfo signing algorithm is not supported by the server key", ...)
} else { ... }
return nil, &tidcommon.InternalServerError   // both branches
The branch changes only the log line. The handler then correctly maps ServerErrorType to 500 (userinfo/handler.go:147) and scrubs the detail into a generic server_error body, omitting the WWW-Authenticate header that client errors receive.

Expected behaviour. A client registering an advertised algorithm receives UserInfo responses signed with it, carrying the matching kid. A client registering an unsupported or malformed algorithm is rejected at registration with 400 invalid_client_metadata, not at first use. No configuration reachable through DCR results in a 500.

Suggested fix.

Validate userinfo_signed_response_alg in buildUserInfoConfig against the discovery derived supported list, returning ErrorInvalidClientMetadata. Validate against the deployment's actual keys, not the ten algorithms GetSupportedSigningAlgorithms reports the provider can do, since a deployment configuring one key advertises one algorithm.
Make jwtService resolve key and algorithm per call: when alg is non-empty select the matching key and use its id and thumbprint for that call, falling back to preferred_key_id when alg is empty. The kid header must move with the key. Signing with the EC key while advertising the RSA kid yields a token no client can verify, which fails more quietly than the current 500.
Propagate ClientErrorType from GenerateJWT through UserInfo so any remaining mismatch surfaces as a 4xx with a usable reason.
Fix 2 is shared with #5109 (client selected ID token signing algorithm), since the same single key binding blocks both. This issue can be fixed independently and is arguably more urgent: ID tokens degrade silently to RS256 and still work, whereas UserInfo confirms the client's choice and then breaks the endpoint. Fix 1 is worth landing on its own regardless, since it converts a runtime 500 into a correct registration time rejection.

Out of scope, noticed while tracing: pkiAlgorithmToJWSAlgorithms (backend/internal/system/kmprovider/defaultkm/pki/service.go:188) maps RSA to RS256 only, so RS512 and PS256 are never advertised despite being signable with the same RSA key.

### Steps to Reproduce

Start ThunderID with the default two key configuration (default-key RSA, ecdsa-key P-256) and preferred_key_id: "default-key".

Confirm the server advertises both algorithms:

curl -sk https://localhost:8090/.well-known/openid-configuration | jq .userinfo_signing_alg_values_supported
["RS256","ES256"]

curl -sk https://localhost:8090/oauth2/jwks | jq '.keys[] | {kty,alg,crv}'
{"kty":"RSA","alg":"RS256","crv":null}
{"kty":"EC","alg":"ES256","crv":"P-256"}
Register three clients at POST /oauth2/dcr/register (requires a system permission token), identical apart from userinfo_signed_response_alg, set to ES256, RS256, and bogus-alg.

Observe all three return 201 Created with the value echoed back verbatim, including bogus-alg, which is not an algorithm.

Obtain an access token for each client and call GET /oauth2/userinfo.

Actual:

Requested	DCR response	UserInfo result
ES256	201, "userinfo_signed_response_alg": "ES256"	500 {"error":"server_error"}
RS256	201, "userinfo_signed_response_alg": "RS256"	200, signed JWT
bogus-alg	201, "userinfo_signed_response_alg": "bogus-alg"	500 {"error":"server_error"}
Expected: ES256 returns 200 with an ES256 signed JWT and the EC kid; bogus-alg is rejected at step 3 with 400 invalid_client_metadata.

Setting preferred_key_id: "ecdsa-key" and repeating inverts the result: ES256 succeeds and RS256 returns 500. This confirms the outcome depends solely on which key the server was started with, not on the client's request.

### Version

1.0.0

_No response_

Thumimku:
### Purpose

A client could not choose the algorithm signing its ID tokens, and choosing one
for UserInfo broke the endpoint.

The JWT service bound a single signing key at startup, resolved from
`jwt.preferred_key_id`, and signed every token with it. The `alg` parameter on
`GenerateJWT` was therefore a guard rather than a selector: it rejected anything
other than that one algorithm.

That left two defects, both reachable with algorithms the server advertises in
discovery and publishes keys for in JWKS:

| Client registers | Before | After |
|---|---|---|
| `id_token_signed_response_alg: ES256` | 201, value absent from response, ID token signed RS256 | 201, echoed, ID token signed **ES256** |
| `userinfo_signed_response_alg: ES256` | 201, echoed, then **500** on every UserInfo call | 201, echoed, UserInfo signed **ES256** |
| either, with an unsupported value | 201, stored, fails later | **400 `invalid_client_metadata`** |

`id_token_signed_response_alg` was absent from the DCR request model entirely,
so the value was discarded at JSON decode. `userinfo_signed_response_alg` was
accepted, stored and echoed, but any value other than the preferred key's
algorithm failed at signing, and UserInfo reported that client error as
`InternalServerError`, so the client received an opaque 500 for a configuration
the server had confirmed at registration. Registration validated neither field,
so even a value that is not an algorithm at all was accepted.

### Approach

**Resolve the signing key per call instead of at startup.** A requested
algorithm selects the configured key for it, and the `kid` header carries that
key's thumbprint so clients can verify against the published JWKS. Signing with
one key while advertising another's `kid` would fail more quietly than the
original 500, so `kid` moves with the key. An empty algorithm keeps using the
preferred key, which is what access tokens and unconfigured clients continue to
get.

**Validate at registration, against the deployment.** `SupportedSigningAlgorithms`
was added to `JWTServiceInterface` so DCR validates through the service rather
than an ad-hoc crypto call or a hardcoded list, matching the existing convention
for JWE algorithms. The list derives from the configured keys, so a deployment
running one key advertises and accepts one algorithm. This turns a runtime
failure into a registration-time rejection with the correct RFC 7591 error.

**Two details worth review:**

`buildIDTokenConfig` returned `nil` unless an encryption field was set and
hardcoded `ResponseType: JWE`. A client requesting only a signing algorithm
would still have been dropped, so the response type is now derived from the
algorithm fields (`JWT` / `JWE` / `NESTED_JWT`), mirroring `buildUserInfoConfig`.

`resolveOAuthTokens` rebuilds `IDTokenConfig` field by field and omitted the new
signing algorithm, silently dropping it between the application API and the
token builder. This is the same class of bug as the original issue and was
invisible to code reading, since every other layer passes the token config
wholesale. The end-to-end test below is what caught it.

### Related Issues
- Related #5109
- Fixes #5195 - The UserInfo 500 and the missing registration validation are fixed here too;
  happy to split them into their own issue if preferred.

### Related PRs
- N/A

### Checklist
    - `docs/content/guides/protocols/oauth-oidc/dynamic-client-registration.mdx`
    - `docs/content/guides/protocols/oauth-oidc/token-formats.mdx`
    - `docs/content/guides/protocols/oauth-oidc/openid-connect.mdx`
    - `docs/content/guides/protocols/oauth-oidc/userinfo.mdx`

      Four existing statements said tokens are signed with *the* deployment
      signing key and that the algorithm follows that key's type. That is no
      longer true and has been corrected.
    - [x] Unit Tests

      `TestGenerateJWTSelectsKeyForAlg` decodes the header and asserts the alg
      and the EC key's `kid`, then verifies the signature against that key.
      `TestBuildIDTokenConfig` covers all four sign/encrypt combinations.
      `TestBuildIDToken_UsesConfiguredSigningAlg` covers the configured and
      fallback paths. `TestResolveOAuthTokens_PreservesIDTokenAlgFields` guards
      the field-by-field copy.
    - [x] Integration Tests

      `tests/integration/oauth/authz/id_token_signing_alg_test.go` runs a full
      authorization_code flow per advertised algorithm and asserts `alg`, `typ`
      and `kid` on the issued ID token, with `kid` matched against JWKS. DCR
      coverage in `dcr_encryption_test.go` asserts every advertised algorithm is
      accepted and that unsupported values are rejected.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting the signing algorithm for ID tokens and signed UserInfo responses.
  * ID tokens now support signing-only, encrypted, and signed-and-encrypted formats.
  * Dynamic client registration now exposes the selected ID token signing algorithm.
  * Tokens include the appropriate signing algorithm and key identifier.

* **Bug Fixes**
  * Unsupported signing algorithms are rejected with clear client errors instead of server errors.

* **Documentation**
  * Updated OAuth/OIDC guidance for signing algorithm selection, defaults, validation, and key discovery.


backend/internal/oauth/oauth2/dcr/service.go
```diff
@@ -266,6 +275,28 @@ func (ds *dcrService) convertDCRToApplication(request *DCRRegistrationRequest) (
 	return appDTO, nil
 }
 
+// validateSignedResponseAlgs checks that requested signing algorithms are ones the deployment can
+// actually sign with, so an unsupported value is rejected at registration rather than failing on
```
thiva-k: Why are we doing this validation at the DCR layer? DCR should be a wrapper and these validations should happen at the inbound client layer so applications, agents and DCR paths can get it

Thumimku: This is to give DC specifc errors, duplicate validations are no harm when we needed api task specific messages or quick turn arounds, for example, if application name is empty then we can validate it from front end rather than passing empty value and validate in backend.

I see this as a similar case

docs/content/guides/protocols/oauth-oidc/token-formats.mdx
```diff
@@ -30,9 +30,11 @@ For **UserInfo**, JSON is the default. Switch to JWS when integrity matters; to
...
+<ProductName /> signs every token (access token, ID Token, and signed UserInfo response) with a deployment signing key. A deployment advertises the signing algorithms its configured keys support, and the requested algorithm selects a matching key, for example `RS256` selects an RSA key and `EdDSA` selects an Ed25519 key.
 
-To see which algorithm a deployment signs with, read `id_token_signing_alg_values_supported` (and `userinfo_signing_alg_values_supported`) in the [Server Metadata](../server-metadata) document, or inspect the published keys at the [JWKS](../jwks) endpoint. Clients verify signatures against those keys.
+A deployment can configure more than one signing key. A client can select which algorithm signs its ID Tokens and signed UserInfo responses, using `id_token_signed_response_alg` and `userinfo_signed_response_alg` at registration. The value must be one of the algorithms the deployment advertises, otherwise registration is rejected. When a client selects nothing, and for access tokens always, the deployment's signing key is used.
```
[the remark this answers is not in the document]
Thumimku: Technically, "preferred" isn't more accurate. preferred_key_id is the config key's name, but the docs describe behaviour, not config field names — and the current text already explains the selection contract fully: clients pick an algorithm, the server matches a key, and absent a choice the deployment's signing key is used.

backend/internal/oauth/oauth2/dcr/service.go
thiva-k: If we are already doing this validation in inboundclient, we can map it in DCR layer to dcr specific errors via existing `mapApplicationErrorToDCRError`. If not we need to add this to inbound client. With the current approach we will still be allowing to register apps with invalid algorithms via the mgt API right?

backend/internal/system/jose/jwt/service.go
```diff
@@ -206,6 +210,48 @@ func (js *jwtService) GenerateJWT(
 	return signingInput + "." + signatureBase64, iat.Unix(), nil
 }
 
+// SupportedSigningAlgorithms returns the JWS "alg" values usable for signing, derived from the
+// signing keys configured for this deployment. It matches the algorithms advertised in discovery.
+func (js *jwtService) SupportedSigningAlgorithms(ctx context.Context) []string {
```
thiva-k: Are we using this function now?

thiva-k:
Don't we need to update console's alg selection logic reflect these changes?

tests/integration/oauth/dcr/dcr_encryption_test.go
```diff
@@ -405,11 +427,101 @@ func (ts *DCREncryptionTestSuite) TestDCRIDTokenEncryptedResponseWithJWKS() {
...
+			name:   "IDTokenNotAnAlgorithm",
+			mutate: func(r *DCRRegistrationRequest) { r.IDTokenSignedResponseAlg = "bogus-alg" },
+		},
+		{
+			name:   "UserInfoUnsupportedAlg",
+			mutate: func(r *DCRRegistrationRequest) { r.UserInfoSignedResponseAlg = "PS512" },
+		},
+		{
+			name:   "UserInfoNotAnAlgorithm",
+			mutate: func(r *DCRRegistrationRequest) { r.UserInfoSignedResponseAlg = "bogus-alg" },
```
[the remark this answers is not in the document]
thiva-k: @Thumimku Shall we fix this if valid?

backend/internal/oauth/oauth2/dcr/service.go
Thumimku: Done fixed it.

tests/integration/oauth/dcr/dcr_encryption_test.go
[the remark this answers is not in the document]
Thumimku: @coderabbitai @thiva-k ,

PS512 is already guaranteed to be unavailable. GetPublicKeys maps each key type to exactly one alg (RSA→RS256, P-256→ES256, P-384→ES384, P-521→ES512, Ed25519→EdDSA, ML-DSA→its own), so no key type or deployment config can ever produce PS512 — it doesn't appear anywhere in the crypto provider.

The two cases are deliberately different: PS512 is a valid alg name with no configured key, bogus-alg isn't an algorithm at all. If both used bogus-alg we'd lose coverage for a validator that checks "is this a known algorithm" instead of "do we have a key for it" — which is the bug this PR fixed on the userinfo side.

thiva-k (approved):

PR #5200, opened 2026-08-28
Add agent and user provider for runtime user and agent creation

Dilusha-Madushan:
### Purpose

Runtime capabilities have no supported way to work with agent or user identities. The runtime only understands generic entities, so anything needing to provision an identity must either skip the operation or re-implement domain rules over the entity boundary. That duplicates security-sensitive logic which can drift from the management services that own it.

Other domains consumed by the runtime (organization units, identity providers, consent, design, resource servers, authorization) already have a provider boundary. Agents and users were the missing pieces.

This PR establishes one consistent rule:

> **Management services own their domain's rules. The runtime reaches them through a `*MgtProvider`. The generic entity boundary is for reads and identification, not creation.**

Five changes deliver that, and the last one is only possible once the others land:

1. `providers.AgentMgtProvider` and `internal/agentmgtprovider`, delegating to the agent management service.
2. `providers.UserMgtProvider` and `internal/usermgtprovider`, delegating to the user management service.
3. The provisioning executor creates users through `UserMgtProvider` instead of `EntityProvider.CreateEntity`.
4. The application service talks to `entityService` directly instead of routing through the runtime's `entityProvider`.
5. `CreateEntity` is removed from `EntityProviderInterface`, because 3 and 4 eliminate its only two callers.

Neither provider adds domain behaviour. Both delegate to the existing management service, which remains the single owner of the rules.

---

### Approach

Both contracts are published in `pkg/thunderidengine/providers`, with implementations in `internal/agentmgtprovider` and `internal/usermgtprovider` following the shape of `internal/entityprovider` (config-driven `Initialize`, default plus disabled implementation).

Key design decisions:

**Contracts live in `pkg`, not `internal`.** The full server and the embedded engine are separate composition roots (`servicemanager.go` and `thunderidengine.New`). Publishing the interfaces in `providers` lets both carry them.

**`Agent` and `User` move to `providers`, and the domain types become aliases.** `internal/agent/model` and `internal/user` already import `providers`, so `providers` cannot import them back. This is the same pattern `internal/inboundclient/model` uses for `InboundClient` and `OAuthClient`. An alias is the identical type, so all existing references compile unchanged, including field promotion off the embedded profile. `go build ./...` across the agent, user, importer, and executor packages is the proof.

**Errors pass through as `*common.ServiceError`, unchanged.** This is the main correctness constraint. The entity path lost the cause at three consecutive boundaries: `entityprovider.mapEntityError` collapsed sentinels into nine coarse codes, `application.mapEntityProviderError` returned `nil` for everything except not-found, and `handleCreateUserError` flattened everything except attribute conflicts into "An error occurred while provisioning the user". `entityprovider` has to classify because `internal/entity` returns bare sentinel errors. The management services do not: they already return fully translated errors with specific `AGT-*` and `USR-*` codes and registered i18n keys. Both providers pass those through with no `default:` branch, so a name conflict, an invalid user type, and a database outage stay distinguishable. This is also the type the flow engine serializes, since `ExecutorResponse.Error` is `*common.ServiceError`.

**Duplicate validation is removed from the executor, not from the service.** `createUserInStore` previously guarded on empty OU ID and user type, returning untyped `fmt.Errorf` values that were then flattened. Those guards are gone because `validateOrganizationUnitForUserType` already performs them and returns typed errors. The user service is untouched; the executor simply stops re-doing its work.

**`Owner` is required on the agent provider input.** The Agent API falls back to `security.GetSubject(ctx)`, dependable there because the subject comes from a verified token. On the runtime path it is not: `/flow/execute/**` is a public path, so the subject is only conditionally present. An unresolved owner is silently omitted from the agent's system attributes rather than rejected, so the provider requires it explicitly. The Agent API path is unchanged. `UserMgtProvider` has no equivalent requirement, since users have no owner.

**Both providers take a `context.Context`.** `entityprovider` discards the caller context; every other provider propagates it. Both call `security.WithRuntimeContext(ctx)`, which sets the runtime marker exactly as before (no authorization behaviour changes) while preserving the request trace ID for log correlation. The application service gains the same benefit: it previously held a `ctx` and dropped it at the entity boundary.

**The application service moves to `entityService`.** It is a management service, the peer of `agent` and `user`, but it reached the entity layer through the runtime's provider. The switch is mechanical because the `entityprovider` converters are `providers.Entity` to `providers.Entity`, the same type, so no result handling changed. It fixes two latent defects beyond the status codes: `toServiceEntity` dropped `OUHandle` and `IsReadOnly` on the way in, and `toProviderEntity` dropped `IsReadOnly` on the way out.

**Absent providers degrade, they do not fail.** User provisioning is optional for embedders, so the engine defaults to a disabled implementation instead of adding a required nil check to `validateEngineContext`. The provisioning executor also nil-guards the provider. An embedder that wires nothing gets `ErrorUserProvisioningDisabled` rather than a startup failure or a nil panic. Deployments can opt out via `user_mgt_provider.type: disabled`.

**`AgentMgtProvider` ships without runtime wiring.** The provisioning executor handles users only in this PR; agent provisioning follows in a later one. Rather than leave plumbing that nothing reads, the executor dependency, engine option, and engine default for the agent provider are deliberately omitted. The contract, implementation, config, and mock are complete and tested; the wiring lands with its consumer. This also keeps `internal/agent` out of the embedded engine's dependency graph.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5176

## Summary by CodeRabbit

* **New Features**
  * Added configurable runtime provisioning for users and agents.
  * Added options to disable user or agent provisioning with clear error messages.
  * Added ID token signing algorithm configuration.
  * Added user and agent management configuration settings.

* **Bug Fixes**
  * Improved propagation of user-service errors during provisioning.
  * Application entity operations now support contextual service errors and handle missing entities more gracefully.
  * Added validation and clearer error handling for invalid user and agent provisioning requests.


backend/cmd/server/servicemanager.go
```diff
@@ -404,9 +406,8 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	// flow services (which themselves depend on the enforcer) are initialized.
 	consentEnforcer.SetConsentService(initConsentService(ctx, logger, inboundClientService))
 
-	// TODO: Remove entityService dependency after finalizing declarative resource loading pattern
```
sahandilshan, line 407: Have we finalise the declarative approach with this?

backend/cmd/server/servicemanager.go
```diff
@@ -356,6 +357,7 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 			RoleService:           roleService,
 			RoleAssignmentService: roleAssignmentService,
 			EntityProvider:        entityProvider,
+			UserMgtProvider:       usermgtprovider.InitializeUserMgtProvider(userService),
```
sahandilshan: Two questions:
1. Entity provider is being used in AuthZen, assume since we only introduce management operation of user resources with this PR we don't need to pass it to the authzen service
2. I can see all other services are initialize and passed as a variable here. So all of those had additional usages other than initializing here. Assume there are no any other usage of usermgt provider. If no usage initializing here would be fine

backend/internal/agent/model/agent.go
```diff
@@ -33,23 +33,10 @@ type AgentRequestWithID struct {
...
-	Attributes  json.RawMessage `json:"attributes,omitempty"`
-
-	// The service-level model carries the full internal profile, including SubjectAttribute, which
-	// the API-facing request and response types deliberately exclude.
-	providers.InboundAuthProfile
-	InboundAuthConfig []providers.InboundAuthConfigWithSecret `json:"inboundAuthConfig,omitempty"`
-}
+type (
+	// Agent is the service-level model for agent create operations.
+	Agent = providers.Agent
```
sahandilshan: Is it the pattern to get the model from the provider level?

backend/internal/agentmgtprovider/default_agent_mgt_provider_test.go
```diff
@@ -0,0 +1,292 @@
...
+	suite.Equal(full.RecoveryFlowHandle, resp.RecoveryFlowHandle)
+	suite.Equal(full.IsRecoveryFlowEnabled, resp.IsRecoveryFlowEnabled)
+	suite.Equal(full.SignOutFlowID, resp.SignOutFlowID)
+	suite.Equal(full.SignOutFlowHandle, resp.SignOutFlowHandle)
+	suite.Equal(full.ThemeID, resp.ThemeID)
+	suite.Equal(full.LayoutID, resp.LayoutID)
+	suite.Equal(assertion, resp.Assertion)
+	suite.Equal(loginConsent, resp.LoginConsent)
+	suite.Equal(full.AllowedUserTypes, resp.AllowedUserTypes)
+	suite.Equal(full.SubjectAttribute, resp.SubjectAttribute)
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check this with latest changes

backend/internal/user/model.go
```diff
@@ -11,16 +11,10 @@ import (
...
-	OUHandle   string          `json:"ouHandle,omitempty"`
-	Type       string          `json:"type,omitempty"`
-	Attributes json.RawMessage `json:"attributes,omitempty"`
-	Display    string          `json:"display,omitempty"`
-	IsReadOnly bool            `json:"isReadOnly"`
-}
+type (
+	// User represents a user in the system.
+	User = providers.User
+)
```
sahandilshan: Let's remove this as well. Same as agent type

Dilusha-Madushan: removed this in user and agent model files in service side

backend/internal/agent/model/agent.go
Dilusha-Madushan: Yes, the models are defined on the provider side under `pkg/thunderidengine/providers/model.go`, and those defined models are used across the provider and service interfaces.

backend/cmd/server/servicemanager.go
Dilusha-Madushan: Yes, currently `userMgtProvider` is only used by the flow executors, with a single usage in `serviceManager`. I’ve now moved the provider definition to a separate code line and use that variable here instead.

backend/cmd/server/servicemanager.go
Dilusha-Madushan: The original TODO read "Remove entityService dependency after finalizing declarative resource loading pattern". It was there because `application.Initialize` took **both** `entityProvider` (the primary entity access, 13 service call sites) and `entityService` (used only in `init.go` for `LoadIndexedAttributes` and `LoadDeclarativeResources`). The note tracked that second one as temporary.

This PR removes `entityProvider` from the application service instead. `applicationService` is a management service. That makes `entityService` the primary dependency, not a temporary extra, so the TODO's premise no longer holds.

The declarative-loading concern is still open, so I moved a TODO to `internal/application/init.go`, on the loading block itself:

    // TODO: Revisit once the declarative resource loading pattern is finalized.

backend/internal/agent/model/agent.go
sahandilshan: We have to remove this, right?

sahandilshan (dismissed):

backend/internal/agent/model/agent.go
Dilusha-Madushan: Yes, this is removed and directly calling the providers.Agent in agent service.

backend/pkg/thunderidengine/providers/interface.go
```diff
@@ -70,6 +70,26 @@ type ActorProvider interface {
 	GetActorRoles(actorID string, groupIDs []string) ([]string, *common.ServiceError)
 }
 
+// AgentMgtProvider provisions agents on behalf of runtime capabilities. The rules and semantics of an
+// agent remain owned by the agent management service; this exposes only what the runtime needs.
+type AgentMgtProvider interface {
+	// CreateAgent provisions an agent. The returned agent carries the generated identifier and, on
```
thiva-k: Shall we remove these verbose comments if not needed?

backend/cmd/server/servicemanager.go
```diff
@@ -187,6 +188,9 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	fatalOnError(ctx, logger, err, "Failed to initialize UserService")
 	exporters = append(exporters, userExporter)
 
+	// Initialize user management provider
+	userMgtProvider := usermgtprovider.InitializeUserMgtProvider(userService)
```
thiva-k:
```suggestion
	userMgtProvider := usermgtprovider.Initialize(userService)
```

backend/internal/system/config/config.go
```diff
@@ -266,8 +266,8 @@ type AuthnProviderConfig struct {
 	Rest RestConfig `yaml:"rest" json:"rest"`
 }
 
-// UserProviderConfig holds the user provider configuration details.
-type UserProviderConfig struct {
+// UserMgtProviderConfig holds the user management provider configuration details.
+type UserMgtProviderConfig struct {
```
thiva-k, line 270: Are there any usages for this existing UserProviderConfig? Let's check if this rename won't be a breaking change

backend/cmd/server/servicemanager.go
Dilusha-Madushan: resolved

backend/internal/system/config/config.go
Dilusha-Madushan: I checked in source repo and confirm that there is no usage from 'UserProviderConfig'.

thiva-k (dismissed):

backend/internal/system/config/config.go
thiva-k, line 270: Shall we add this and other missing provider configs to the helm templates too?

Dilusha-Madushan: it's broader than the new configs: the chart templates no provider section at all today, including entity_provider and authn_provider on main. Unset is safe (each Initialize falls through default: on the zero value), so nothing is broken, but there's no way to set type: disabled from Helm since the chart has no passthrough for untemplated keys.

so better if we handle it through separate issue.

sahandilshan: @Dilusha-Madushan,
Let's create an issue. and put it in here. And then let's merge

thiva-k: @coderabbitai Can you create an issue for the missing provider templates in helm

cc: @rajithacharith

sahandilshan (dismissed):

sahandilshan (approved):

PR #5224, opened 2026-08-31
Integrate external AuthZEN PDP authorization

Yathusiga27:
### Purpose
Add backend support for integrating an external AuthZEN PDP with ThunderID. Resource servers can reference an external PDP connection, and authorization requests are evaluated through the configured PDP.

This PR includes:

- External AuthZEN PDP connection management.
- Resource-server association with an external PDP.
- AuthZEN authorization evaluation and routing.
- Outbound authentication support for PDP requests.
- Authorization integration for client-credentials and refresh-token flows.
- Dependency checks when deleting an external PDP connection.
- Unit and integration test coverage.

### Approach
- Added an external AuthZEN PDP authorization engine and router.
- Added external PDP connection models, stores, handlers, and declarative configuration support.
- Routed authorization decisions using the resource server identifier.
- Added integration coverage using a local AuthZEN PDP test server.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5220

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring external AuthZEN Policy Decision Point (PDP) connections, including endpoints, timeouts, retries, subject properties, and attribute mappings.
  * Added connection management APIs for creating, viewing, updating, deleting, listing, and checking connection usage.
  * Resource servers can now use an external AuthZEN PDP for authorization decisions, including batch evaluations.
  * Added declarative export/import support for AuthZEN PDP connections.
  * Added validation for endpoint URLs and protection against deleting connections still in use.


backend/internal/connection/authzen_pdp.go
```diff
@@ -0,0 +1,334 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package connection
```
thiva-k: Let's move authzen pdp files to connection/authzenpdp for now and have a service.go

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -191,16 +191,18 @@ type Resource struct {
...
-	Resources   []Resource         `yaml:"resources,omitempty"   json:"resources,omitempty"`
+	ID                      string             `yaml:"id"                    json:"-"`
+	Name                    string             `yaml:"name"                  json:"name"`
+	Description             string             `yaml:"description,omitempty" json:"description,omitempty"`
+	Identifier              string             `yaml:"identifier"            json:"identifier"`
+	Type                    ResourceServerType `yaml:"type,omitempty"        json:"type,omitempty"`
+	OUID                    string             `yaml:"ouId,omitempty"        json:"ouId"`
+	OUHandle                string             `yaml:"ouHandle,omitempty"    json:"-"`
+	Delimiter               string             `yaml:"delimiter,omitempty"   json:"delimiter,omitempty"   yamlfmt:"quoted"`
+	AuthorizationEngine     string             `yaml:"authorizationEngine,omitempty"     json:"authorizationEngine,omitempty"`
```
thiva-k: Lets have this like

```
authorizationEngine: {
  type: RBAC/authzenpdp,
  properties: { externalPDPConnectionId: id, etc. 
  }
}
```

backend/internal/connection/authzen_pdp.go
Yathusiga27: Done. Moved the AuthZEN PDP connection files into backend/internal/connection/authzenpdp and added service.go.

backend/pkg/thunderidengine/providers/model.go
Yathusiga27: Done. Restructured authorizationEngine into a nested object with type and properties.

backend/internal/oauth/oauth2/granthandlers/client_credentials.go
```diff
@@ -159,16 +159,23 @@ func buildAccessEvaluationsRequest(
 	groupIDs []string,
 	permissions []string,
 	resourceServerID string,
+	resourceServerIdentifier string,
 ) providers.AccessEvaluationsRequest {
 	evaluations := make([]providers.AccessEvaluationRequest, 0, len(permissions))
 	for _, permission := range permissions {
 		evaluations = append(evaluations, providers.AccessEvaluationRequest{
 			Subject: providers.Subject{
+				Type:     "application",
```
thiva-k: We shouldn't hardcode since this can be agents as well. Let's resolve and pass the entity category if needed

Yathusiga27: removed the hardcoded "application" value and now derive the subject type from `entityCategory` using `authZENSubjectType`

thiva-k:
@coderabbitai please review

backend/internal/authz/external_router.go
```diff
@@ -0,0 +1,234 @@
...
+		return &engine.AccessEvaluationResponse{}, nil
+	}
+	return &response.Evaluations[0], nil
+}
+
+func (r *dynamicExternalAuthZENRouter) EvaluateAccessBatch(
+	ctx context.Context,
+	request engine.AccessEvaluationsRequest,
+) (*engine.AccessEvaluationsResponse, error) {
+	routes, err := buildExternalAuthZENRoutes(ctx, r.logger, r.externalAuthZEN, r.resourceService, false)
```
thiva-k: Can we remove the separate router? `resolveEngine` could just be a method on `authorizationService`. Look up the resource server's engine type, build the PDP client if needed, done. Can remove `dynamicExternalAuthZENRouter` and the `buildExternalAuthZENRoutes` call happening on every `EvaluateAccessBatch`.

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,571 @@
...
+			_ = resp.Body.Close()
+			return fmt.Errorf("failed to decode AuthZEN response: %w", err)
+		}
+		_ = resp.Body.Close()
+		return nil
+	}
+	return fmt.Errorf("AuthZEN PDP request failed after retries")
+}
+
+func (p *authZENPDP) resolveEndpoints(ctx context.Context) (string, string, error) {
```
thiva-k: Do we need discovery here? We don't discover endpoints for OAuth/OIDC IdPs, we just ask for `authorizationEndpoint` and `tokenEndpoint` directly. Could do the same for the PDP. Ask for the evaluation and batch endpoints up front instead of hitting `.well-known` on every call.

backend/internal/system/config/config.go
```diff
@@ -617,6 +617,31 @@ func (c OAuthConfig) ToEngineConfig() engineconfig.OAuthConfig {
 	}
 }
 
+// AuthorizationConfig configures authorization evaluation providers.
+type AuthorizationConfig struct {
+	ExternalAuthZEN ExternalAuthZENConfig `yaml:"external_authzen" json:"external_authzen"`
+}
+
+// ExternalAuthZENConfig configures an external AuthZEN policy decision point.
+type ExternalAuthZENConfig struct {
```
thiva-k: Let's remove this in server config. PDP connections should be configured through the `connection` module.

Yathusiga27: Removed

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: Removed the .well-known discovery flow. The PDP connection now requires the single evaluation endpoint and batch evaluation endpoint directly, and runtime evaluation uses those configured endpoints without discovery calls.

backend/internal/authz/external_router.go
Yathusiga27: Removed the separate external AuthZEN router. `authorizationService.resolveEngine` now looks up the resource server, builds or reuses the configured PDP engine, and send requests directly.

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -191,16 +191,28 @@ type Resource struct {
...
+	OUHandle            string                    `yaml:"ouHandle,omitempty"    json:"-"`
+	Delimiter           string                    `yaml:"delimiter,omitempty"   json:"delimiter,omitempty"   yamlfmt:"quoted"`
+	AuthorizationEngine AuthorizationEngineConfig `yaml:"authorizationEngine,omitempty" json:"authorizationEngine,omitempty"`
+	IsReadOnly          bool                      `yaml:"-"                     json:"-"`
+	Resources           []Resource                `yaml:"resources,omitempty"   json:"resources,omitempty"`
+}
+
+// AuthorizationEngineConfig selects the authorization engine for a resource server.
+type AuthorizationEngineConfig struct {
+	Type       string                        `yaml:"type,omitempty"       json:"type,omitempty"`
```
Thumimku, line 215: what is meant by type can u give me some example

backend/internal/resource/store.go
```diff
@@ -950,6 +954,9 @@ func resolveProperties(row map[string]interface{}, rs *providers.ResourceServer)
 // buildPropertiesJSON builds the PROPERTIES JSON for a providers.ResourceServer.
 func buildPropertiesJSON(rs providers.ResourceServer) interface{} {
 	properties := resourceServerProperties{Delimiter: rs.Delimiter}
+	if rs.AuthorizationEngine.Type != "" || rs.AuthorizationEngine.Properties.ExternalPDPConnectionID != "" {
```
Thumimku: why we need special logic for AuthorizationEngine?

backend/internal/resource/service.go
```diff
@@ -444,6 +444,12 @@ func (rs *resourceService) UpdateResourceServer(
 	// Type is immutable and always preserved from the existing record
 	resourceServer.Type = existingResServer.Type
 
+	if resourceServer.AuthorizationEngine.Type == "" {
+		resourceServer.AuthorizationEngine = existingResServer.AuthorizationEngine
+	} else if resourceServer.AuthorizationEngine.Type != "external_authzen_pdp" {
```
Thumimku: can we make this magic string as constnat and whats the purpose of this check

backend/internal/resource/declarative_resource_test.go
```diff
@@ -516,7 +516,7 @@ type: "MCP"
 ouId: "ou1"
 `)
 
-	parser := parseAndValidateResourceServerWrapper(nil)
+	parser := parseAndValidateResourceServerWrapper()
```
Thumimku, line 553: why we need not to pass the nill

backend/pkg/thunderidengine/providers/model.go
Yathusiga27: `type` selects which authorization engine a resource server should use.

Examples:
- `rbac` means evaluate with ThunderID’s built-in authorization engine.
- `authzen-pdp` means send the AuthZEN evaluation request to the configured external PDP connection.

backend/internal/resource/store.go
Yathusiga27: This is only to avoid persisting an empty `authorizationEngine` object for normal resource servers.

`authorizationEngine` is optional and stored inside the existing `PROPERTIES` JSON column. We only add it to that JSON when either the engine type or external PDP connection id is configured. Otherwise existing resource servers keep the same properties shape as before.

backend/internal/resource/service.go
Yathusiga27: Yes, we can move `external_authzen_pdp` to a constant.

The purpose of this check is to keep update behavior safe. If `authorizationEngine.type` is omitted, we preserve the existing resource-server engine config. If it is set to the supported external PDP engine type, we keep it. If any other value is provided, we clear the authorization engine config so an unsupported engine type is not persisted.

backend/internal/resource/declarative_resource_test.go
Yathusiga27: Earlier the function accepted one argument, so we passed `nil`. But that argument was not used anywhere inside the function. Since the wrapper does not need any input now, I removed the unused parameter.

backend/internal/authz/init.go
```diff
@@ -5,12 +5,28 @@ package authz
...
+// Initialize creates and initializes the authorization service with the selected engine.
+func Initialize(
+	authorizationEngine engine.AuthorizationEngine,
+	userServices ...userpkg.UserServiceInterface,
+) providers.AuthorizationProvider {
+	return newAuthorizationService(authorizationEngine, userServices...)
+}
+
+// InitializeWithResourceService initializes authorization with the default engine and resource service.
+func InitializeWithResourceService(
```
thiva-k: Why do we need this separate Initialize function? Shall we update the exiting Initialize to take the resource service as dependency

backend/internal/authz/external_engine.go
```diff
@@ -0,0 +1,82 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/thunder-id/thunderid/internal/authz/engine"
+	"github.com/thunder-id/thunderid/internal/connection"
+)
+
+func externalAuthZENRuntimeConfigEngine(
```
thiva-k: Shall we refactor and move external authzen pdp related logic into auth/engine/external_authzen_pdp and remove this file. All engine specific logic should be there. The authorization service should just pick the engine and delegate

thiva-k:
Shall we add integration tests covering all the scenarios we captured in the usecases?

backend/internal/authz/init.go
Yathusiga27: removed `InitializeWithResourceService` and updated the existing Initialize function to accept the resource service dependency.

backend/internal/authz/external_engine.go
Yathusiga27: Implemented as suggested. The external AuthZEN PDP engine construction and defaults now live in `authz/engine/external_authzen_pdp.go`, and `external_engine.go` has been removed. The authorization service now resolves the configured connection, selects or caches the engine, and delegates the evaluation.

Yathusiga27:
> Shall we add integration tests covering all the scenarios we captured in the usecases?

Added integration tests for the applicable end-to-end use cases: permitted, denied, batch, partially permitted, all denied, attribute-based evaluation, PDP unavailability, external PEP delegation, unauthenticated PDP access, and fallback to the default engine.

backend/internal/authz/service.go
```diff
@@ -6,25 +6,74 @@ package authz
...
+	signature string
+	engine    engine.AuthorizationEngine
+	failOpen  bool
+}
+
+var getAuthZENPDPRuntimeConfig = connection.GetAuthZENPDPRuntimeConfig
+
 // newAuthorizationService creates a new instance of authorizationService.
-func newAuthorizationService(engine engine.AuthorizationEngine) providers.AuthorizationProvider {
+func newAuthorizationService(
```
thiva-k, line 34: Do we need both constructors? The only real caller (Initialize) always goes through the one with resourceService. Could we drop this one and just make resourceService a required param everywhere?

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,636 @@
...
+
+	results := make([]AccessEvaluationResponse, 0, len(response.Evaluations))
+	for _, evaluation := range response.Evaluations {
+		results = append(results, AccessEvaluationResponse(evaluation))
+	}
+	return &AccessEvaluationsResponse{Evaluations: results}, nil
+}
+
+// evaluateSingle sends one ThunderID authorization request to the configured AuthZEN PDP.
+func (p *authZENPDP) evaluateSingle(
```
thiva-k, line 272: This duplicates the retry loop below almost line for line. Should just delegate to it instead of reimplementing it, so we're not keeping retry/timeout logic in sync in two places.

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,636 @@
...
+
+func validateAuthZENEndpoint(endpoint string) error {
+	parsedEndpoint, err := url.Parse(endpoint)
+	if err != nil || parsedEndpoint.Scheme == "" || parsedEndpoint.Host == "" {
+		return fmt.Errorf("endpoint must be an absolute URL")
+	}
+	return nil
+}
+
+func defaultAuthZENBatchEndpoint(evaluationEndpoint string) string {
```
thiva-k: Since we moved to explicit endpoints instead of discovery, the batch endpoint should be required too, not guessed from the single endpoint by string suffix.

backend/internal/resource/store.go
Thumimku: so for each resource servers, rather than having a dedicated column for authoirzation servers, we are having it as json property, wont it make hard to query what resource/resource servers used by this authorization type

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: Fixed. Both single and batch evaluation paths now delegate HTTP execution, retries, timeout handling, status validation, and response decoding to the shared post method.

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: Fixed. The batch evaluation endpoint is now required and validated explicitly.

backend/internal/authz/service.go
Yathusiga27: Fixed. I removed `newAuthorizationServiceWithResourceService` and kept a single `newAuthorizationService` constructor. resourceService, userService, and the shared HTTP client are now passed directly as required dependencies from Initialize.

backend/internal/authz/service.go
```diff
@@ -75,6 +99,237 @@ func (s *authorizationService) EvaluateAccessBatch(
...
+		}
+	}
+	if err := evaluateResolvedBatch(ctx, s.engine, defaultRequest, defaultIndexes, responses); err != nil {
+		return nil, err
+	}
+
+	return &engine.AccessEvaluationsResponse{Evaluations: responses}, nil
+}
+
+func evaluateResolvedBatch(
```
thiva-k, line 159: Add method-level comments. Check other places as well

thiva-k:
Address and resolve valid Coderabbit comments

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,380 @@
...
+
+const (
+	// DefaultTimeoutMS is the default external PDP request timeout in milliseconds.
+	DefaultTimeoutMS = 500
+	// DefaultRetryCount is the default number of retries for external PDP requests.
+	DefaultRetryCount = 1
+)
+
+// ConnectionRequest is the API representation of an external AuthZEN PDP connection request.
+type ConnectionRequest struct {
```
thiva-k, line 37: Move structs to model.go. Check other places

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,41 @@
...
+	store Store
+}
+
+// NewService creates an AuthZEN PDP connection service.
+func NewService(store Store) *Service {
+	return &Service{store: store}
+}
+
+// Create stores an external AuthZEN PDP connection.
+func (s *Service) Create(ctx context.Context, connection AuthZENPDPConnection) error {
```
thiva-k, line 75: Where are we handling validations, declarative checks, etc.? Check service.go from other packages and align with that pattern

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,585 @@
...
+)
+
+type authZENEvaluationRequest struct {
+	Subject  authZENSubject         `json:"subject"`
+	Resource authZENResource        `json:"resource"`
+	Action   authZENAction          `json:"action"`
+	Context  map[string]interface{} `json:"context,omitempty"`
+}
+
+type authZENSubject struct {
```
thiva-k: Can't we reuse the structs from the existing authzen implementation rather than redefining here?

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,585 @@
...
+			Properties: authZENSubjectProperties(
+				evaluation.Subject,
+				allowedSubjectProperties,
+				subjectPropertyMappings,
+				subjectAttributeMappings,
+			),
+		},
+		Resource: authZENResource{
+			Type:       resourceType,
+			ID:         resourceID,
```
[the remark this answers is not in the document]
Yathusiga27: > Addressed in `toAuthZENEvaluationRequest`. When `ResourceServer.Type` or `ResourceServer.ResourceID` is empty, both values now fall back to `ResourceServer.ID`, ensuring standard AuthZEN requests include the required resource identity.

backend/internal/authz/service.go
```diff
@@ -75,6 +99,237 @@ func (s *authorizationService) EvaluateAccessBatch(
...
+			enriched.Evaluations = append(enriched.Evaluations, evaluation)
+			continue
+		}
+		user, ok := users[evaluation.Subject.ID]
+		if !ok {
+			var svcErr *tidcommon.ServiceError
+			user, svcErr = s.userService.GetUser(ctx, evaluation.Subject.ID, false)
+			if svcErr != nil {
+				return providers.AccessEvaluationsRequest{}, &tidcommon.InternalServerError
+			}
```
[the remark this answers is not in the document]
Yathusiga27: Addressed. `enrichRequest` now continues without user enrichment when `GetUser` returns `userpkg.ErrorUserNotFound`, allowing the rest of the batch to be evaluated. Unexpected user lookup errors still return `tidcommon.InternalServerError`. A unit test covers this behavior.

backend/internal/resource/service.go
```diff
@@ -444,6 +444,12 @@ func (rs *resourceService) UpdateResourceServer(
 	// Type is immutable and always preserved from the existing record
 	resourceServer.Type = existingResServer.Type
 
+	if resourceServer.AuthorizationEngine.Type == "" {
+		resourceServer.AuthorizationEngine = existingResServer.AuthorizationEngine
+	} else if resourceServer.AuthorizationEngine.Type != providers.AuthorizationEngineTypeExternalAuthZENPDP {
+		resourceServer.AuthorizationEngine = providers.AuthorizationEngineConfig{}
+	}
+
```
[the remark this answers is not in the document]
Yathusiga27: Addressed. `validateResourceServerUpdate` now rejects any non-empty `AuthorizationEngine.Type` other than `providers.AuthorizationEngineTypeExternalAuthZENPDP` before normalization. Empty types still inherit the existing configuration, and supported external AuthZEN types remain accepted.

tests/integration/authzen/external_authzen_pdp_test.go
```diff
@@ -0,0 +1,825 @@
...
+		http.MethodDelete,
+		testutils.TestServerURL+"/connections/external-authzen-pdp/"+s.connectionID,
+		nil,
+	)
+	s.Require().NoError(err)
+	response, err = testutils.GetHTTPClient().Do(request)
+	body, readErr := io.ReadAll(response.Body)
+	_ = response.Body.Close()
+	s.Require().NoError(err)
+	s.Require().NoError(readErr)
```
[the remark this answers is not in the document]
Yathusiga27: Addressed.

backend/internal/authz/service.go
Yathusiga27: Added.

backend/internal/connection/authzenpdp/model.go
Yathusiga27: Addressed.

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: Addressed. The external PDP engine now reuses the existing AuthZEN models from backend/internal/authzen/model.go, and the duplicate struct definitions were removed.

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Endpoint validation is centralized in authzenpdp/model.go  `authzenpdp.ValidateConnection` and enforced before creating or updating connections, including declarative imports.

backend/internal/connection/authzenpdp/store_constants.go
```diff
@@ -0,0 +1,57 @@
...
+
+import dbmodel "github.com/thunder-id/thunderid/internal/system/database/model"
+
+var (
+	queryCreateAuthZENPDPConnection = dbmodel.DBQuery{
+		ID: "CON-AUTHZEN-PDP-01",
+		Query: `INSERT INTO "AUTHZEN_PDP_CONNECTION"
+			(ID, DEPLOYMENT_ID, NAME, DESCRIPTION, ENDPOINT,
+			SUBJECT_PROPERTIES, SUBJECT_PROPERTY_MAPPINGS)
+			VALUES ($1, $2, $3, $4, $5, $6, $7)`,
```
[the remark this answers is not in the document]
Yathusiga27: Addressed through the existing JSON storage format. BatchEndpoint, TimeoutMS, RetryCount, and SubjectAttributeMappings are encoded in SUBJECT_PROPERTIES during create/update and decoded during get/list. The integration test verifies that all settings persist correctly.

backend/internal/connection/connection_declarative_model.go
```diff
@@ -35,6 +35,20 @@ type connectionExportModel struct {
...
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	TimeoutMS int `yaml:"timeoutMs,omitempty"                json:"timeoutMs,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	RetryCount int `yaml:"retryCount,omitempty"               json:"retryCount,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	SubjectProperties []string `yaml:"subjectProperties,omitempty"        json:"subjectProperties,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	SubjectPropertyMappings string `yaml:"subjectPropertyMappings,omitempty"  json:"subjectPropertyMappings,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	SubjectAttributeMappings []authZENPDPSubjectAttributeMapping `yaml:"subjectAttributeMappings,omitempty" json:"subjectAttributeMappings,omitempty"`
```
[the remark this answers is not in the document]
Yathusiga27: `FailOpen` is now included in the declarative export model, mapped during export and import, and covered by a true-value round-trip test.

backend/internal/connection/authzenpdp/service.go
thiva-k: Validations and logic should be in service.go. model.go should have only structs and related helpers. 

Also, how we are handling declarative resources validations? e.g. blocking create/update for declarative connections

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,617 @@
...
+	SubjectAttributeMappings []SubjectAttributeMapping
+}
+
+// AuthZENPDPRuntimeConfigResolver loads an external AuthZEN PDP connection by ID.
+type AuthZENPDPRuntimeConfigResolver func(
+	context.Context, string,
+) (*connection.AuthZENPDPRuntimeConfig, error)
+
+// NewAuthZENPDPFromConnection creates an external AuthZEN PDP engine from a saved connection.
+func NewAuthZENPDPFromConnection(
```
thiva-k: Why do we a new separate `NewAuthZENPDPFromConnection` ? Lets add this parameter to `NewAuthZENPDP` and remove this

backend/internal/authz/service.go
```diff
@@ -6,25 +6,45 @@ package authz
...
 // authorizationService is the default implementation of providers.AuthorizationProvider.
 type authorizationService struct {
-	engine engine.AuthorizationEngine
+	engine          engine.AuthorizationEngine
+	userService     userpkg.UserServiceInterface
+	resourceService resource.ResourceServiceInterface
+	httpClient      httpservice.HTTPClientInterface
 }
 
+var getAuthZENPDPRuntimeConfig = connection.GetAuthZENPDPRuntimeConfig
```
thiva-k: Let's get this config in external_authzen_pdp.go instead of passing from here

backend/internal/authz/engine/external_authzen_pdp.go
```diff
@@ -0,0 +1,617 @@
...
+	ctx context.Context,
+	connectionID string,
+	resolve AuthZENPDPRuntimeConfigResolver,
```
thiva-k: Why do we need to get this as an arguement?

backend/internal/authz/service.go
```diff
@@ -75,6 +98,209 @@ func (s *authorizationService) EvaluateAccessBatch(
...
+) (*engine.AccessEvaluationsResponse, error) {
+	if s.resourceService == nil {
+		return s.engine.EvaluateAccessBatch(ctx, request)
+	}
+	responses := make([]engine.AccessEvaluationResponse, len(request.Evaluations))
+	defaultRequest := engine.AccessEvaluationsRequest{}
+	defaultIndexes := make([]int, 0, len(request.Evaluations))
+	externalRequests := map[engine.AuthorizationEngine]engine.AccessEvaluationsRequest{}
+	externalIndexes := map[engine.AuthorizationEngine][]int{}
+	externalFailOpen := map[engine.AuthorizationEngine]bool{}
```
thiva-k: What is the use of having these separate maps for external and resolved engine evaluation?

Yathusiga27: These maps separate evaluations by resolved engine so external PDP requests can be batched while preserving the original evaluation order. The index map is needed to merge PDP results back into the complete response, and the fail-open map keeps the connection-specific fallback behavior

backend/internal/connection/authzenpdp/service.go
Yathusiga27: declarative checks are handled in connection/service.go using CheckDeclarativeCreate, CheckDeclarativeUpdate, and CheckDeclarativeDelete guards.

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: Removed the separate `NewAuthZENPDPFromConnection` constructor and moved the connection ID parameter into `NewAuthZENPDP`.

backend/internal/authz/service.go
Yathusiga27: Addressed. The AuthZEN PDP connection lookup and configuration handling now live in external_authzen_pdp.go.

backend/internal/authz/engine/external_authzen_pdp.go
Yathusiga27: The resolver is no longer passed as an argument. `NewAuthZENPDP` now receives the connection ID and resolves the saved PDP configuration internally.

backend/internal/authz/service.go
```diff
@@ -87,6 +309,8 @@ func toEngineAccessEvaluationsRequest(request providers.AccessEvaluationsRequest
 			},
 			ResourceServer: engine.ResourceServer{
 				ID:         evaluation.ResourceServer.ID,
+				Type:       evaluation.ResourceServer.Type,
+				ResourceID: evaluation.ResourceServer.ResourceID,
```
thiva-k, line 296: 1. Why do we need ID and ResourceID both? And what is the use of ResourceServer.Type here? 
2. If both are needed can't we resolve it from the ID?

backend/internal/connection/authzen_pdp.go
```diff
@@ -0,0 +1,115 @@
...
```
thiva-k: Why do we have this outside of connection/authzenpdp ?

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,369 @@
...
+			SubjectPropertyMappings:  cloneStringMap(connection.SubjectPropertyMappings),
+			SubjectAttributeMappings: cloneSubjectAttributeMappings(connection.SubjectAttributeMappings),
+			FailOpen:                 connection.FailOpen,
+		})
+	}
+	return configs, nil
+}
+
+// GetAuthZENPDPRuntimeConfig returns a saved AuthZEN PDP connection for token-issuance routing.
+func GetAuthZENPDPRuntimeConfig(ctx context.Context, id string) (*AuthZENPDPRuntimeConfig, error) {
```
thiva-k: Why is this in model.go?

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,369 @@
...
+	TimeoutMS                int
+	RetryCount               int
+	SubjectProperties        []string
+	SubjectPropertyMappings  map[string]string
+	SubjectAttributeMappings []SubjectAttributeMapping
+	FailOpen                 bool
+}
+
+// ListAuthZENPDPRuntimeConfigs returns saved AuthZEN PDP connections for token-issuance routing.
+func ListAuthZENPDPRuntimeConfigs(ctx context.Context) ([]AuthZENPDPRuntimeConfig, error) {
```
thiva-k: Why is this in model.go?

backend/internal/connection/init.go
```diff
@@ -18,19 +18,23 @@ import (
 // services, registers the /connections routes, loads declarative connection resources, and
 // returns the connection exporter for the export API.
 func Initialize(mux *http.ServeMux, idpService idp.IDPServiceInterface,
-	notificationService notification.NotificationSenderMgtSvcInterface) (
+	notificationService notification.NotificationSenderMgtSvcInterface,
+	resourceService resourceServerLister) (
 	declarativeresource.ResourceExporter, error) {
-	svc := newService(idpService, notificationService)
+	authZENPDPStore := newAuthZENPDPStoreForService()
+	svc := newService(idpService, notificationService, resourceService, authZENPDPStore)
```
thiva-k: We should pass the authzenPDPService here

backend/internal/connection/init.go
```diff
@@ -18,19 +18,23 @@ import (
 // services, registers the /connections routes, loads declarative connection resources, and
 // returns the connection exporter for the export API.
 func Initialize(mux *http.ServeMux, idpService idp.IDPServiceInterface,
-	notificationService notification.NotificationSenderMgtSvcInterface) (
+	notificationService notification.NotificationSenderMgtSvcInterface,
+	resourceService resourceServerLister) (
 	declarativeresource.ResourceExporter, error) {
-	svc := newService(idpService, notificationService)
+	authZENPDPStore := newAuthZENPDPStoreForService()
```
thiva-k: Let's remove the ForService part

backend/internal/connection/init.go
```diff
@@ -18,19 +18,23 @@ import (
...
-	if err := loadDeclarativeResources(idpService); err != nil {
+	if err := loadDeclarativeResources(idpService, authZENPDPStore); err != nil {
 		return nil, err
 	}
 
-	return newConnectionExporter(idpService, notificationService), nil
+	return newConnectionExporter(idpService, notificationService, authZENPDPStore), nil
 }
 
+var newAuthZENPDPStoreForService = newAuthZENPDPStore
```
thiva-k: Why do we need this variable decalration?

backend/internal/connection/init.go
Yathusiga27: removed

backend/internal/connection/init.go
Yathusiga27: addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: moved to service.go

backend/internal/connection/authzen_pdp.go
Yathusiga27: Removed and moved under connection/authzenpdp

backend/internal/connection/init.go
Yathusiga27: removed

backend/internal/authz/service.go
Yathusiga27: Addressed. ResourceServer.ID is the internal resource-server UUID used to load its configuration and select the authorization engine. ResourceID identifies the actual protected resource and is sent as AuthZEN resource.id, so it cannot be derived from the resource-server UUID.  
The previous ResourceServer.Type field was removed. We now resolve the resource server from ID and use its Identifier as AuthZEN resource.type.

api/connections.yaml
```diff
@@ -78,6 +78,94 @@ paths:
...
+      - { $ref: '#/components/parameters/ConnectionID' }
+    get:
+      tags: [Connections]
+      summary: Get an external AuthZEN PDP connection
+      responses:
+        "200":
+          description: Connection details
+          content:
+            application/json:
+              schema: { $ref: '#/components/schemas/ExternalAuthZENPDPConnectionResponse' }
```
Thumimku: can we remove the external part

Thumimku: check other places

backend/cmd/server/servicemanager.go
```diff
@@ -224,7 +224,7 @@ func registerServices(mux *http.ServeMux, cacheManager cache.CacheManagerInterfa
 	ouAuthzService.SetPermissionResolver(
 		role.NewEffectivePermissionResolver(roleService, groupService, entityService))
 
-	authZService := authz.Initialize(roleService)
+	authZService := authz.Initialize(roleService, resourceService, userService)
```
Thumimku: why we dont include the AI service?

api/connections.yaml
Yathusiga27: Removed

backend/cmd/server/servicemanager.go
Yathusiga27: updated this to use `entityService` instead of `userService`. With entityService, the authorization service can resolve shared entity attributes for supported subject types. Currently it enriches user and agent subjects.

backend/cmd/server/config/default.json
```diff
@@ -320,6 +320,10 @@
       }
     }
   },
+  "authzen_pdp": {
+    "timeout_ms": 500,
```
Thumimku: let's make this as 1 sec

backend/internal/authz/engine/engine.go
```diff
@@ -33,6 +33,8 @@ type Subject struct {
 // ResourceServer identifies the resource server for an access evaluation.
 type ResourceServer struct {
 	ID         string
+	Identifier string
```
Thumimku: whats the purpose of ID and Identifier? why we need both?

backend/internal/connection/init.go
```diff
@@ -98,6 +109,8 @@ func registerRoutes(mux *http.ServeMux, h *handler) {
 		getSMSHandler(h, ncommon.NotificationProviderTypeCustom, smsGatewayFromSenderDTO),
 		updateSMSHandler(h, ncommon.NotificationProviderTypeCustom, smsGatewayToSenderDTO, smsGatewayFromSenderDTO),
 		collectionOpts, itemOpts)
+
+	registerAuthZENPDPVendorRoutes(mux, h, "/connections/"+authzenpdp.VendorName, collectionOpts, itemOpts)
```
Thumimku, line 106: why `"/connections/"+` is hard coded?

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+		TimeoutMS:                connection.TimeoutMS,
+		RetryCount:               connection.RetryCount,
+		SubjectProperties:        append([]string(nil), connection.SubjectProperties...),
+		SubjectPropertyMappings:  cloneStringMap(connection.SubjectPropertyMappings),
+		SubjectAttributeMappings: cloneSubjectAttributeMappings(connection.SubjectAttributeMappings),
+	}, nil
+}
+
+// ValidateEndpoint reports whether endpoint is an absolute HTTP or HTTPS URL.
+func ValidateEndpoint(endpoint string) error {
```
Thumimku: can we check for other places how they generalise this

backend/internal/connection/init.go
Yathusiga27: This follows the existing connection route pattern

backend/internal/authz/engine/engine.go
Yathusiga27: ID- internal uuid, Identifier- unique value for identifying the rs
removed identifier as it can be resolved from id

backend/cmd/server/config/default.json
Yathusiga27: updated

backend/internal/connection/connection_declarative_model.go
```diff
@@ -35,6 +38,23 @@ type connectionExportModel struct {
...
+	Endpoint string `yaml:"endpoint,omitempty"                 json:"endpoint,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	BatchEndpoint string `yaml:"batchEndpoint,omitempty"            json:"batchEndpoint,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	TimeoutMS int `yaml:"timeoutMs,omitempty"                json:"timeoutMs,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	RetryCount *int `yaml:"retryCount,omitempty"               json:"retryCount,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
+	SubjectProperties []string `yaml:"subjectProperties,omitempty"        json:"subjectProperties,omitempty"`
+	//nolint:lll // long struct tag: both yaml and json keys are required
```
thiva-k, line 50: Remove this unwanted comments and nolint if not needed

backend/internal/connection/service.go
```diff
@@ -270,6 +304,115 @@ func (s *service) deleteSMSByProvider(ctx context.Context, provider ncommon.Noti
 	return s.notificationService.DeleteSender(ctx, id)
 }
 
+// createAuthZENPDP validates and stores an AuthZEN PDP connection.
+func (s *service) createAuthZENPDP(
+	ctx context.Context,
+	connection authzenpdp.AuthZENPDPConnection,
+) (*authzenpdp.AuthZENPDPConnection, *tidcommon.ServiceError) {
+	if svcErr := declarativeresource.CheckDeclarativeCreate(); svcErr != nil {
```
thiva-k: This logic like declarative check, normalize, existing check, uuid generation should be in authzen/service.go. From connection you should just call the authzen/service.go function's. Check for other authzen CRUD operations too and match with the existing pattern of other CRUD functions in connection/service.go calling domain specific service.go.

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,161 @@
...
+
+// VendorName is the connection vendor identifier for AuthZEN PDPs.
+const VendorName = "authzen-pdp"
+
+// DefaultTimeoutMS returns the configured server timeout default.
+func DefaultTimeoutMS() int {
+	if !config.IsServerRuntimeInitialized() {
+		return 0
+	}
+	return config.GetServerRuntime().Config.AuthZENPDP.TimeoutMS
```
thiva-k: config.GetServerRuntime().Config.AuthZENPDP. should be passed from servicemanager.go. authzenpdp package should use only the passed config. Check how other pacakges get this config from servicemanager.go

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+// NewService creates an AuthZEN PDP connection service.
+func NewService(store Store) *Service {
+	return &Service{store: store}
+}
+
+// GetAuthZENPDPRuntimeConfig returns a saved AuthZEN PDP connection for token-issuance routing.
+var GetAuthZENPDPRuntimeConfig = getAuthZENPDPRuntimeConfig
+
+func getAuthZENPDPRuntimeConfig(ctx context.Context, id string) (*AuthZENPDPRuntimeConfig, error) {
+	connection, err := NewStore().Get(ctx, id)
```
thiva-k: Shall we use the get from service.go?

backend/internal/connection/init.go
```diff
@@ -18,17 +19,19 @@ import (
 // services, registers the /connections routes, loads declarative connection resources, and
 // returns the connection exporter for the export API.
 func Initialize(mux *http.ServeMux, idpService idp.IDPServiceInterface,
-	notificationService notification.NotificationSenderMgtSvcInterface) (
+	notificationService notification.NotificationSenderMgtSvcInterface,
+	resourceService resourceServerLister,
+	authZENPDPService *authzenpdp.Service) (
```
thiva-k: We should get the authzenpdp.authzenpdpserviceinterface here like other arguements

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// Service provides storage operations for AuthZEN PDP connections.
+type Service struct {
+	store Store
+}
+
```
thiva-k, line 36: Lets define an authzenpdpserviceinterface here, similar to how other packages have done

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+// Create stores an AuthZEN PDP connection.
+func (s *Service) Create(ctx context.Context, connection AuthZENPDPConnection) error {
+	if err := validateConnection(connection); err != nil {
+		return err
+	}
+	return s.store.Create(ctx, connection)
+}
+
+// List returns all AuthZEN PDP connections.
+func (s *Service) List(ctx context.Context) ([]AuthZENPDPConnection, error) {
```
thiva-k: Use meaningful full names. e.g. ListAuthZenPDPs . Check other functions too. And check store.go functions too

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+	"strings"
+)
+
+// Service provides storage operations for AuthZEN PDP connections.
+type Service struct {
+	store Store
+}
+
+// NewService creates an AuthZEN PDP connection service.
+func NewService(store Store) *Service {
```
thiva-k: `NewAuthzenPDPService`

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// Service provides storage operations for AuthZEN PDP connections.
+type Service struct {
```
thiva-k:
```suggestion
type AuthZENPDPService struct {
```

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("invalid access evaluation endpoint: %w", err)
+	}
+	if err := ValidateEndpoint(connection.BatchEndpoint); err != nil {
+		return fmt.Errorf("invalid access evaluations endpoint: %w", err)
+	}
+	return nil
+}
+
+// NormalizeEndpoints trims and validates the configured AuthZEN PDP endpoints.
+func NormalizeEndpoints(connection *AuthZENPDPConnection) error {
```
thiva-k: Let's move the helper functions to bottom of file and make it private

backend/internal/connection/service.go
```diff
@@ -24,13 +27,28 @@ import (
 type service struct {
 	idpService          idp.IDPServiceInterface
 	notificationService notification.NotificationSenderMgtSvcInterface
+	resourceService     resourceServerLister
+	authZENPDPService   *authzenpdp.Service
+}
+
+type resourceServerLister interface {
```
thiva-k: Why do we need this new resourceServerLister interface? Cant we use the existing resourceServerInterface from resource server package?

backend/internal/connection/service.go
```diff
@@ -270,6 +304,115 @@ func (s *service) deleteSMSByProvider(ctx context.Context, provider ncommon.Noti
...
+// updateAuthZENPDP validates and updates an AuthZEN PDP connection by ID.
+func (s *service) updateAuthZENPDP(
+	ctx context.Context,
+	id string,
+	connection authzenpdp.AuthZENPDPConnection,
+) (*authzenpdp.AuthZENPDPConnection, *tidcommon.ServiceError) {
+	if svcErr := declarativeresource.CheckDeclarativeUpdate(); svcErr != nil {
+		return nil, svcErr
+	}
+	if _, svcErr := s.getAuthZENPDP(ctx, id); svcErr != nil {
```
thiva-k: This logics should happen in authzenpdp/service.go. The connection/service.go should just delegate to it

backend/internal/connection/service.go
```diff
@@ -270,6 +304,115 @@ func (s *service) deleteSMSByProvider(ctx context.Context, provider ncommon.Noti
...
+	}
+	existing, err := s.authZENPDPService.GetByName(ctx, connection.Name)
+	if err != nil {
+		return nil, &tidcommon.InternalServerError
+	}
+	if existing != nil && existing.ID != id {
+		return nil, &ErrorAuthZENPDPAlreadyExists
+	}
+	if err := s.authZENPDPService.Update(ctx, id, connection); err != nil {
+		return nil, &tidcommon.InternalServerError
```
thiva-k: The authzenPDPService should send the appropriate error and from connection/service.go we should just return that error. We shouldn't always return internal server error from connection/service.go. Check how the errors are returned from idp/notification functions in connection/service.go

backend/internal/authz/engine/authzen_pdp.go
```diff
@@ -0,0 +1,575 @@
...
+}
+
+// AuthZENPDP evaluates requests using configuration supplied separately for each call.
+type AuthZENPDP struct {
+	client httpservice.HTTPClientInterface
+	logger *log.Logger
+}
+
+// NewAuthZENPDP creates a reusable evaluator with a shared HTTP client.
+func NewAuthZENPDP(client httpservice.HTTPClientInterface) *AuthZENPDP {
```
thiva-k: Can we make this symmetric with RBAC? `NewRBACEngine` takes the role service and fetches what it needs during evaluation. Let's do the same: take the AuthZEN PDP connection service here, drop the `config` param from `EvaluateAccess`/`EvaluateAccessBatch`, and have the engine resolve the connection from the id on the evaluation. Then it satisfies `AuthorizationEngine` like RBAC does and the service just picks an engine.

It'll need to group by connection id internally, similar to `groupEvaluations`, so one batch can span connections and settings get built once per group instead of on every call. An empty connection id or a connection that no longer exists should be an error from the engine, not a skipped or empty decision. Note `Get` returns `(nil, nil)` for not found today, so that needs handling explicitly.

This also lets us drop `AuthZENPDPRuntimeConfig` and the `GetAuthZENPDPRuntimeConfig` var, which builds a new store on each call. Wire the service through `authz.Initialize`. `authzenpdp.Initialize()` is constructed after it in servicemanager today, but it takes no deps so it can move up.

backend/internal/authz/service.go
```diff
@@ -75,6 +97,271 @@ func (s *authorizationService) EvaluateAccessBatch(
...
+	}
+	return enriched, nil
+}
+
+func supportsSubjectEnrichment(subjectType string) bool {
+	return subjectType == providers.EntityCategoryUser.String() ||
+		subjectType == providers.EntityCategoryAgent.String()
+}
+
+func toEngineResourceServer(resourceServer providers.AccessEvaluationResourceServer) engine.ResourceServer {
```
thiva-k, line 289: For the above, `engine.ResourceServer` needs to carry what the engine routes on. Can we add `Identifier` and an `Engine` field holding the type and properties (connection id), mirroring `AuthorizationEngineConfig`? Keep them mirrored in the engine package, it doesn't import `providers` and that's worth preserving.

RBAC ignores both. The PDP needs the connection id to resolve, and `Identifier` for the AuthZEN `Resource.Type`, since external policies key on the resource server identifier and not our UUID.

They can't be filled here though, since they come from the stored record rather than the API request. Let's fill them in the routing loop after `GetResourceServer`.

backend/internal/authz/service.go
```diff
@@ -75,6 +97,271 @@ func (s *authorizationService) EvaluateAccessBatch(
 	return fromEngineAccessEvaluationsResponse(evaluationResp), nil
 }
 
+// evaluateWithResolvedEngines routes evaluations to connection-backed or default engines and preserves order.
+func (s *authorizationService) evaluateWithResolvedEngines(
```
thiva-k, line 99: With the above, this becomes just picking an engine and bucketing by it. Switch on the resource server's engine type, with `rbac`/empty and `authzen_pdp` as explicit cases and an error otherwise. `resolvePDPConfig`, `resolveConnectionConfig`, `loadConnectionConfig`, `toEngineSubjectAttributeMappings` and `connectionBatches` can all go. Keying that map by config pointer is a bug too, two resource servers sharing a connection become two entries and two separate calls.

Let's keep `evaluateResolvedBatch` and the per request memo (keyed on resource server id, for the `GetResourceServer` call) so ordering is preserved and we don't refetch per evaluation. `AuthorizationProvider`, `enrichRequest` and the translation helpers stay as they are.

Worth noting in the PR description that this changes behavior: today a missing or unresolvable PDP connection silently falls through to RBAC. The description currently mentions fail-open configuration, and there's no such field in the code.

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ConnectionRequest is the API representation of an AuthZEN PDP connection request.
+type ConnectionRequest struct {
+	Name                     string                    `json:"name"`
+	Description              string                    `json:"description,omitempty"`
+	Endpoint                 string                    `json:"-"`
+	BatchEndpoint            string                    `json:"batchEndpoint,omitempty"`
+	TimeoutMS                int                       `json:"timeoutMs,omitempty"`
+	RetryCount               int                       `json:"retryCount,omitempty"`
+	SubjectProperties        string                    `json:"subjectProperties,omitempty"`
```
thiva-k: Can these be native JSON, `SubjectProperties []string` and `SubjectPropertyMappings map[string]string`? `SubjectAttributeMappings` just below is already a typed array, and the stored shape in `authZENPDPProperties` is already `[]string`/`map[string]string`, so the string form exists for one hop and gets parsed straight back. It also removes a silent failure: a mapping without a colon is dropped today with no error.

Then `splitSubjectProperties`, the `Split`/`Join` helpers and both custom marshalers can go, and the timeout/retry defaults can be applied once in the service instead of in `UnmarshalJSON`, `FromRequest` and `buildAuthZENPDPConnection`. `api/connections.yaml` needs updating to match as well.

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,161 @@
...
+	BatchEndpoint            string
+	TimeoutMS                int
+	RetryCount               int
+	SubjectProperties        []string
+	SubjectPropertyMappings  map[string]string
+	SubjectAttributeMappings []SubjectAttributeMapping
+}
+
+// SubjectAttributeMapping maps ThunderID user-type attributes to PDP subject attributes.
+type SubjectAttributeMapping struct {
```
thiva-k, line 60: Following up on the comment above, I think we should drop `SubjectProperties` and `SubjectPropertyMappings` instead of converting them. This struct already covers both levels: being in the list means released, `PDPAttribute` is an optional rename. No entry for a type means nothing is released.

```json
"subjectAttributeMappings": [
  {
    "userType": "employee",
    "attributes": [
      { "attribute": "email", "pdpAttribute": "mail" },
      { "attribute": "groups" }
    ]
  }
]
```

It needs to key on the real user type though, not the category. `subjectMappingForSubject` matches `UserType` against `Subject.Type` (`user`/`agent`), so `"employee"` never matches. `enrichRequest` already loads the entity and `providers.Entity.Type` is right there, we just discard it. Then we can validate the type and attribute against the schema at create/update like `inboundclient.validateSubjectAttributeMapping` does, with `groups` kept as a documented reserved name since it comes from `GroupIDs`.

backend/internal/connection/declarative_resource.go
```diff
@@ -468,6 +569,18 @@ func (s *connectionDeclarativeStore) Create(id string, data interface{}) error {
 		return s.idpStore.Create(id, dto)
 	case *ncommon.NotificationSenderDTO:
 		return s.senderStore.Create(id, dto)
+	case *authzenpdp.AuthZENPDPConnection:
```
thiva-k, line 555: This writes declarative connections into the config DB, while the IdP and sender cases above go to an in-memory `GenericFileBasedStore`. The declarative id is the same every boot and this is a plain INSERT against `PRIMARY KEY (DEPLOYMENT_ID, ID)`, so the second startup hits a PK violation and `fatalOnError` kills the server.

Let's add `file_based_store.go` and `composite_store.go` to `authzenpdp`, with a `store` key on `AuthZENPDPConfig` and `Initialize` resolving the mode like `idp.initializeStore` does. This case then writes to a `GenericFileBasedStore`, which needs a new `entity.KeyType` so the package's file store can read them back. No cache wrapper needed, notification manages without one. Also should use the caller's context here, not `context.Background()`.

backend/internal/connection/connection_declarative_model.go
Yathusiga27: removed

backend/internal/connection/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed. AuthZEN now uses the initialized `AuthZENPDPService` and its `GetAuthZENPDP`method.

backend/internal/connection/init.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Renamed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: renamed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/service.go
Yathusiga27: Addressed

backend/internal/connection/service.go
Yathusiga27: Addressed.

backend/internal/connection/service.go
Yathusiga27: updated

backend/internal/authz/engine/authzen_pdp.go
Yathusiga27: Addressed. The AuthZEN PDP engine now receives the connection service during initialization, resolves connection settings by connection ID during evaluation, and groups batch evaluations by connection.

backend/internal/authz/service.go
Yathusiga27: Addressed

backend/internal/authz/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: Addressed. `SubjectProperties` and `SubjectPropertyMappings` were removed, leaving `SubjectAttributeMappings` as the single mapping configuration. The engine now uses the concrete `providers.Entity.Type` through` Subject.EntityType` when matching UserType, while the public Subject.Type remains the entity category.

backend/internal/connection/declarative_resource.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+	) (*AuthZENPDPConnection, *tidcommon.ServiceError)
+	DeleteAuthZENPDPConnection(ctx context.Context, id string) *tidcommon.ServiceError
+	CreateAuthZENPDP(ctx context.Context, connection AuthZENPDPConnection) error
+	ListAuthZENPDPs(ctx context.Context) ([]AuthZENPDPConnection, error)
+	GetAuthZENPDP(ctx context.Context, id string) (*AuthZENPDPConnection, error)
+	UpdateAuthZENPDP(ctx context.Context, id string, connection AuthZENPDPConnection) error
+	DeleteAuthZENPDP(ctx context.Context, id string) error
+}
+
+type entityTypeAttributeProvider interface {
```
thiva-k: Let's remove this and use entitytype.EntityTypeServiceInterface directly

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+	declarativeresource "github.com/thunder-id/thunderid/internal/system/declarative_resource"
+	"github.com/thunder-id/thunderid/internal/system/log"
+	sysutils "github.com/thunder-id/thunderid/internal/system/utils"
+	tidcommon "github.com/thunder-id/thunderid/pkg/thunderidengine/common"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+// AuthZENPDPServiceInterface defines AuthZEN PDP connection management operations.
+type AuthZENPDPServiceInterface interface {
+	FromRequest(request ConnectionRequest) AuthZENPDPConnection
```
thiva-k: Let's remove this from the interface and make it private and move it to bottom of file

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+		ctx context.Context,
+		connection AuthZENPDPConnection,
+	) (*AuthZENPDPConnection, *tidcommon.ServiceError)
+	UpdateAuthZENPDPConnection(
+		ctx context.Context,
+		id string,
+		connection AuthZENPDPConnection,
+	) (*AuthZENPDPConnection, *tidcommon.ServiceError)
+	DeleteAuthZENPDPConnection(ctx context.Context, id string) *tidcommon.ServiceError
+	CreateAuthZENPDP(ctx context.Context, connection AuthZENPDPConnection) error
```
thiva-k: Let's use CreateAuthZENPDPConnection/UpdateAuthZENPDPConnection and remove the plain delete,create and update functions altogether and update the usages

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+	transactioner providers.Transactioner
+	entityTypes   entityTypeAttributeProvider
+	logger        *log.Logger
+}
+
+var _ AuthZENPDPServiceInterface = (*AuthZENPDPService)(nil)
+
+const reservedSubjectGroupsAttribute = "groups"
+
+const loggerComponentName = "AuthZENPDPService"
```
thiva-k: Remove this constant and use it directly while intializing the logger

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+}
+
+// FromRequest converts an API request into an AuthZEN PDP connection.
+func (s *AuthZENPDPService) FromRequest(request ConnectionRequest) AuthZENPDPConnection {
+	return fromRequest(request)
+}
+
+// ApplyDefaults fills unset connection-level timeout and retry settings.
+func (s *AuthZENPDPService) ApplyDefaults(connection *AuthZENPDPConnection) {
```
thiva-k: Make this private and move it to bottom of file

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+			}
+			if _, ok := allowed[attribute.Attribute]; !ok {
+				return &ErrorInvalidSubjectAttributeMapping
+			}
+		}
+	}
+	return nil
+}
+
+func (s *AuthZENPDPService) getSubjectTypeAttributes(
```
thiva-k, line 288: Add method level comments. Check all the new functions added in all the files

backend/internal/connection/authzenpdp/store.go
```diff
@@ -0,0 +1,212 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/thunder-id/thunderid/internal/system/database/provider"
+	"github.com/thunder-id/thunderid/internal/system/deployment"
+)
+
+// Store persists AuthZEN PDP connections.
+type Store interface {
```
thiva-k: Rename to newAuthZENPDPStore and have it as private. Check other functions/ struct/interface names too whether they are meaningful

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+var _ AuthZENPDPServiceInterface = (*AuthZENPDPService)(nil)
+
+const reservedSubjectGroupsAttribute = "groups"
+
+const loggerComponentName = "AuthZENPDPService"
+
+// NewAuthZENPDPService creates an AuthZEN PDP connection service.
+func NewAuthZENPDPService(
```
thiva-k: CAn we remove this and use only the `newAuthZENPDPService` in callsites?

backend/internal/connection/authzenpdp/model.go
```diff
@@ -0,0 +1,56 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package authzenpdp
+
+// VendorName is the connection vendor identifier for AuthZEN PDPs.
+const VendorName = "authzen-pdp"
```
thiva-k: Define this directly in the place it is used

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Added

backend/internal/connection/authzenpdp/service.go
Yathusiga27: removed

backend/internal/connection/authzenpdp/store.go
Yathusiga27: Addressed

backend/internal/connection/authzenpdp/model.go
Yathusiga27: Addressed

backend/internal/authz/engine/authzen_pdp_test.go
```diff
@@ -0,0 +1,613 @@
...
+	if err := json.Unmarshal(data, &cfg); err != nil {
+		panic(err)
+	}
+	if err := config.InitializeServerRuntime("", &cfg); err != nil {
+		panic(err)
+	}
+	os.Exit(m.Run())
+}
+
+func TestAuthZENPDPConcurrentConfigurations(t *testing.T) {
```
ThaminduDilshan: Shall we follow test suite pattern?

Yathusiga27: updated

backend/internal/authz/engine/engine.go
Yathusiga27: - ID is internal resource-server UUID. We use it to resolve the persisted resource-server configuration and its selected authorization engine. 
- Identifier is the configured ThunderID resource-server identifier, which the AuthZEN PDP engine sends as resource.type.  

Although Identifier can be resolved from ID, authz/service.go resolves it once and passes both values to the engine. This keeps the engine independent of resourceService and avoids a second lookup during evaluation.

backend/internal/connection/authzenpdp/service.go
```diff
@@ -0,0 +1,392 @@
...
+		ctx context.Context,
+		request ConnectionRequest,
+	) (*AuthZENPDPConnection, *tidcommon.ServiceError)
+	UpdateAuthZENPDPConnection(
+		ctx context.Context,
+		id string,
+		request ConnectionRequest,
+	) (*AuthZENPDPConnection, *tidcommon.ServiceError)
+	DeleteAuthZENPDPConnection(ctx context.Context, id string) *tidcommon.ServiceError
+	ListAuthZENPDPs(ctx context.Context) ([]AuthZENPDPConnection, error)
```
thiva-k: Let's return service errors here

backend/internal/authz/engine/init.go
```diff
@@ -0,0 +1,29 @@
...
+package engine
+
+import (
+	"github.com/thunder-id/thunderid/internal/role"
+	httpservice "github.com/thunder-id/thunderid/internal/system/http"
+	"github.com/thunder-id/thunderid/internal/system/log"
+)
+
+// NewRBACEngine creates an RBAC authorization engine.
+func NewRBACEngine(roleService role.RoleServiceInterface) AuthorizationEngine {
```
thiva-k: Let's have this as private function inside the rbacengine.go. In engine/init.go, we can have a Initialize method that calls newRBACEngine and newAuthZENPDPEngine and returns those two to servicemanager

backend/internal/connection/authzenpdp/service.go
Yathusiga27: Updated

backend/internal/authz/engine/init.go
Yathusiga27: updated

backend/pkg/thunderidengine/providers/model.go
```diff
@@ -1236,6 +1259,7 @@ func getDuration(startTime int64, endTime int64) int64 {
 // Subject identifies the principal for an access evaluation.
 type Subject struct {
 	Type       string                 `json:"type,omitempty"`
+	EntityType string                 `json:"-"`
```
thiva-k: Why do we need EntityType in the access subject model? Is it used?  And one should be renamed to category.

Yathusiga27: Addressed. Renamed `Type` to `Category`. `EntityType` is used to select the correct subject attribute mapping.

thiva-k: Let's rename EntityType to Type to be consistent with the engine model and add json tags

Yathusiga27: updated

thiva-k (approved):

PR #5253, opened 2026-09-02
Handle incompatible yq implementations

### LINKED ISSUE #5216: setup.sh fails with Debian yq and does not generate certificates
### Description

I believe this issue could affect quite a few users, especially those installing ThunderID on Debian- or Ubuntu-based systems. Since these distributions may provide a different yq implementation than the one expected by setup.sh, users can run into this problem even when they have installed all seemingly required dependencies.

I found an issue with setup.sh in ThunderID 1.0.1 on Debian.

The setup script detects yq using:

if command -v yq >/dev/null 2>&1; then
and then executes commands such as:

yq eval '.server.hostname // "localhost"' deployment.yaml
The problem is that Debian can provide a different yq implementation which does not support the yq eval syntax expected by the script.

Because setup.sh also uses:

set -e
the script exits immediately when that yq command fails.

As a result, it never reaches:

configure_direct_auth_secret
configure_certificates
and files such as the following are not generated:

config/certs/server.cert
config/certs/server.key
config/certs/signing.cert
config/certs/signing.key
config/certs/ecdsa-signing.cert
config/certs/ecdsa-signing.key
config/certs/crypto.key
config/secrets/direct_auth_secret
The startup then fails with errors such as:

Failed to initialize certificate service
certificate file not found at .../config/certs/signing.cert
I confirmed the failure with bash -x ./setup.sh --verbose. The trace stops around:

+ command -v yq
++ yq eval '.server.hostname // "localhost"' ./deployment.yaml
+ HOSTNAME=
A possible fix would be to check whether the installed yq actually supports the required syntax before using it, for example:

if command -v yq >/dev/null 2>&1 &&
   yq eval '.' /dev/null >/dev/null 2>&1; then
and otherwise use the existing grep/awk fallback.

Alternatively, the setup script could explicitly check for Mike Farah's yq v4 and provide a clear error message if an incompatible implementation is installed.

It may also be useful to avoid having an optional YAML parser failure terminate the entire setup process under set -e.

### Steps to Reproduce

Run `setup.sh` in ThunderID 1.0.1 on Debian.

### Version

1.0.1 Linux x64

OS: Debian
Shell: Bash 5.2.x

ZiyamSanthosh:
### Purpose
Prevent `setup.sh` and `build.sh` from terminating when an incompatible `yq` implementation is present in `PATH`.

Fixes #5216

### Approach
- Verify that the detected `yq` supports the required `eval` syntax before using it.
- Use the existing grep/awk configuration fallback when the compatibility probe fails.
- Apply the same check to both setup and source-build configuration parsing.

### Related Issues
- Fixes #5216

### Related PRs
- N/A

### Checklist
- [ ] Documentation provided. Not required for this change.
- [ ] Tests provided. No new automated test framework was introduced for this shell capability check.
    - [ ] Unit Tests
    - [ ] Integration Tests
- [ ] Breaking changes. None.

### Validation
- `bash -n setup.sh build.sh`
- `git diff --check`
- Verified the compatibility probe against Python/jq `yq` 3.1.0, which is rejected and therefore selects the existing fallback.

### Security checks
- [x] Confirmed that this PR does not commit keys, passwords, tokens, usernames, or other secrets.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration parsing when the installed `yq` tool is unavailable or not functioning correctly.
  * Automatically falls back to an alternate parsing method instead of producing empty configuration values.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -142,7 +142,7 @@ read_config() {
         PUBLIC_HOSTNAME=""
     else
         # Try yq first (YAML parser)
-        if command -v yq >/dev/null 2>&1; then
+        if command -v yq >/dev/null 2>&1 && yq eval '.' /dev/null >/dev/null 2>&1; then
             HOSTNAME=$(yq eval '.server.hostname // "localhost"' "$config_file" 2>/dev/null)
             PORT=$(yq eval '.server.port // 8090' "$config_file" 2>/dev/null)
             HTTP_ONLY=$(yq eval '.server.http_only // false' "$config_file" 2>/dev/null)
--- a/setup.sh
+++ b/setup.sh
@@ -254,7 +254,7 @@ read_config() {
     log_debug "Reading configuration from: $config_file"
 
     # Try yq first (YAML parser)
-    if command -v yq >/dev/null 2>&1; then
+    if command -v yq >/dev/null 2>&1 && yq eval '.' /dev/null >/dev/null 2>&1; then
         HOSTNAME=$(yq eval '.server.hostname // "localhost"' "$config_file" 2>/dev/null)
         PORT=$(yq eval '.server.port // 8090' "$config_file" 2>/dev/null)
         HTTP_ONLY=$(yq eval '.server.http_only // false' "$config_file" 2>/dev/null)


rajithacharith (approved):

ZiyamSanthosh:
Closing in favor of #5304. The 1.0.x pull request will create the corresponding automated pull request for the main branch.

PR #5255, opened 2026-09-02
Answer OIDC prompt, max_age, and id_token_hint from the SSO session

### LINKED ISSUE #5105: auth_time in the ID token is taken from authorization code creation rather than the session's authentication time
### Description

OIDC Core §2 defines `auth_time` as "Time when the End-User authentication occurred". The
authorization code grant instead derives it from the moment the authorization code was created,
at `backend/internal/oauth/oauth2/granthandlers/authorization_code.go:196`:

```go
AuthTime: authCode.TimeCreated.Unix(),
```

With no session those two moments coincide, so the claim looks correct. Once an existing SSO
session is reused, they diverge: the user authenticated earlier, but a new authorization code
is created for each authorization request, so `auth_time` advances every time even though no
new authentication took place.

The session subsystem already records what the claim needs. `backend/internal/flow/session/`
holds the session store, resolver, state and timeouts, `SSOCheckExecutor`
(`backend/internal/flow/executor/sso_check_executor.go`) resolves an existing session, and the
Console App authentication flow uses it at
`backend/cmd/server/bootstrap/01-default-resources.yaml:274` (`start` -> `sso_check` ->
`session`, falling back to `prompt_credentials`). The plumbing to carry the value is also
present: `AuthTime` is a field on `IDTokenBuildContext`
(`backend/internal/oauth/oauth2/tokenservice/model.go:105`) and is emitted at
`backend/internal/oauth/oauth2/tokenservice/builder.go:437`. The code grant simply does not ask
the session for it.

This makes the claim wrong in exactly the case it exists for. `max_age` handling depends on it:
a client uses `auth_time` to decide whether the authentication is recent enough, and a value
that tracks code creation always looks fresh.

Note the CIBA grant handler reads a stored `record.AuthTime`
(`backend/internal/oauth/oauth2/granthandlers/ciba.go:260`) rather than a creation timestamp,
so the two grant types disagree about what the claim means.

### Steps to Reproduce

Observed against the OpenID Foundation conformance suite, Basic OP plan. 

1. Run the Conformance suite.
1. `oidcc-max-age-10000` authorizes once, then authorizes again with `max_age=10000`. The
   existing authentication is well inside that window, so §3.1.2.1 requires it to be reused and
   `auth_time` to stay the same across both ID tokens. It fails:

   ```
   CheckIdTokenAuthTimeClaimsSameIfPresent: The id_tokens contain different auth_time
   claims, but must contain the same auth_time.
   ```

Two related modules, `oidcc-max-age-1` and `oidcc-prompt-login`, pass their `auth_time`
assertions (`CheckSecondIdTokenAuthTimeIsLaterIfPresent: auth_time is later in the second
id_token`) and then stop on an unrelated harness limitation, so they neither confirm nor
contradict this.

#### Expected

`auth_time` should carry the time the user actually authenticated, taken from the session that
authorization resolved, so that it stays stable while a session is reused and only advances on
a genuine re-authentication.

Fixing this on its own does not make `oidcc-max-age-10000` pass, because the authorization
endpoint does not consult an existing session for OAuth clients yet; that part is tracked as a
separate improvement. This issue is the claim being derived from the wrong source, which is
worth correcting independently.

### Version

v1.0.0 (`oidc-conformnace-suite` branch, distribution `thunderid-1.0.0-macos-arm64`)

macOS, SQLite (default deployment), server on `https://thunderid.conformance.test:8090`.

Thumimku:
### Purpose

The authorization endpoint and the authentication flow did not consult the SSO session when
answering the OIDC authentication-context parameters, so four related behaviours were wrong:

- `auth_time` was derived from authorization-code creation rather than the session's
  authentication time, so it advanced on every authorization even when no new authentication
  took place (#5105).
- `prompt=login` was silently ignored: an existing session was reused despite the request
  asking for a fresh authentication.
- A `max_age` smaller than the session's age failed the request at the end of the flow
  (`FET-1082`) instead of re-authenticating, leaving the client unable to proceed even with the
  End-User present.
- `prompt=none` always returned `login_required`, and `id_token_hint` was not consulted at all,
  so silent re-authentication was unavailable.

### Approach

Four changes, all resolving to the same idea: the SSO session is the authority on when and as
whom the subject authenticated.

**`auth_time` from the session.** `AuthAssertExecutor` now stamps the flow assertion with the
resolved session's authentication time, and the assertion decoder prefers it over the `iat`
fallback. `AuthorizationCode` gained an explicit `AuthTime` field so the claim no longer shares
`TimeCreated`, whose other job is computing the code's expiry. Without that split, a code minted
over a reused session inherited the session's age and was rejected at insertion as already
expired.

**Forced re-authentication.** `prompt=login` sets a runtime flag that `SSOCheckExecutor` honors by
declining to reuse the session, and an exceeded `max_age` does the same. Both route down the
existing `onFailure: prompt_credentials` edge rather than adding a new path, and the session
handle is still shared so the re-authentication attaches to the same session.

**Session authentication time refresh.** A re-authentication attaches to the existing session, so
`SaveCheckpoint` now moves its `AuthenticatedAt` forward. Without this the session kept claiming
the original login, which under-reported `auth_time` and made a later `max_age` check reject a
request it actually satisfied.

**`prompt=none` and `id_token_hint` at the authorize endpoint.** The endpoint resolves the
client's flow, reads the per-flow SSO cookie, and answers from the session: `login_required` when
there is none, when `id_token_hint` names a different subject, or when the authentication is
older than `max_age`. The unconditional `login_required` moved out of the request validator so
PAR is unaffected. Both new dependencies are optional and nil in the embedded engine, which has
no session store; `prompt=none` keeps returning `login_required` there.

Two deliberate decisions worth flagging for review:

- The `id_token_hint` signature and issuer are verified but its expiry is not. OIDC Core requires
  an expired hint to still be accepted: it identifies who was authenticated, not whether they
  still are.
- A malformed or negative `max_age` is treated as no constraint rather than as a zero-second
  window, matching the existing assurance check. All three sites that read `max_age` share these
  semantics.

**Known limitation.** `checkPromptNone` verifies that a live session exists, not that this flow's
specific checkpoint is present in it. A flow with more than one SSO-Check node (step-up
authentication) could therefore start under `prompt=none` and still reach a prompt. No shipped
flow has that shape, and the behaviour is not a regression, so it is left as follow-up: the
intended fix is a passive-authentication mode on the flow engine, which needs its own discussion
and a change to how flow errors map to OAuth error codes.

Enabling SSO on the Default Authentication Flow is deliberately **not** included: it changes the
behaviour every new application inherits and deserves a separate discussion.

### Related Issues
- Refs https://github.com/thunder-id/thunderid/issues/5113
- Fixes https://github.com/thunder-id/thunderid/issues/5105

### Related PRs
- N/A

### Checklist
    - `docs/content/guides/protocols/oauth-oidc/openid-connect.mdx`: rewrote the `prompt` values
      table, added the missing `max_age` and `id_token_hint` parameter rows, and added a
      `max_age` / `auth_time` section.
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added OpenID Connect `prompt=none` support for silent authorization with a valid SSO session.
  * Added `prompt=login` and `max_age` handling to require fresh authentication.
  * Added `id_token_hint` validation during authorization.
  * Preserved authentication time in ID tokens and authorization flows.
  * PAR requests using `prompt=none` are now accepted and evaluated during authorization.

* **Bug Fixes**
  * Reauthentication now refreshes session authentication timestamps for subsequent requests.

* **Documentation**
  * Updated OpenID Connect guidance for these authentication parameters and behaviors.

### DIFF
--- a/backend/internal/flow/executor/auth_assert_executor.go
+++ b/backend/internal/flow/executor/auth_assert_executor.go
@@ -206,6 +206,11 @@ func (a *authAssertExecutor) generateAuthAssertion(
 		jwtClaims[oauth2const.ClaimCompletedAuthClass] = completedACR
 	}
 
+	// Carry the time the subject actually authenticated. On the SSO path that is the reused
+	// session's authentication time, which is older than this assertion's iat; without the claim
+	// the OAuth layer falls back to iat and auth_time advances on every authorization.
+	jwtClaims[oauth2const.ClaimAuthTime] = a.resolveAuthTime(ctx)
+
 	// Bind the assertion to the originating auth request so the corresponding callback can verify this assertion
 	// authorizes the specific request it accompanies.
 	if authReqID, exists := ctx.RuntimeData[common.RuntimeKeyAuthorizationRequestID]; exists && authReqID != "" {
--- a/backend/internal/flow/executor/sso_check_executor.go
+++ b/backend/internal/flow/executor/sso_check_executor.go
@@ -4,6 +4,7 @@
 package executor
 
 import (
+	"strconv"
 	"time"
 
 	"github.com/thunder-id/thunderid/internal/flow/common"
@@ -74,6 +75,14 @@ func (e *ssoCheckExecutor) Execute(ctx *providers.NodeContext) (*providers.Execu
 		execResp.RuntimeData[common.RuntimeKeySSOSessionHandle] = resolved.HandleID
 	}
 
+	// A request may demand a fresh authentication even when a live session exists: prompt=login
+	// asks for one outright, and a max_age smaller than the session's age makes the existing
+	// authentication too old to satisfy. Both are answered by re-authenticating here, at the
+	// branch point, rather than by failing at the end of the flow once the assertion is due.
+	if resolved != nil && reauthRequired(ctx, resolved.AuthenticatedAt, logger) {
+		resolved = nil
+	}
+
 	var snapshot *session.SessionContext
 	if resolved != nil && checkpoint != "" {
 		if snapshot, err = e.sso.FindCheckpoint(ctx.Context, resolved.SessionID, checkpoint); err != nil {
@@ -104,6 +113,46 @@ func (e *ssoCheckExecutor) Execute(ctx *providers.NodeContext) (*providers.Execu
 	}
 
 	return execResp, nil
+}
+
+// reauthRequired reports whether the request rules out reusing a session that authenticated at
+// authenticatedAt: prompt=login always does, and max_age does when more than that many seconds
+// have passed since. A malformed max_age is treated as no constraint, matching the assurance
+// check that also reads it.
+func reauthRequired(ctx *providers.NodeContext, authenticatedAt time.Time, logger *log.Logger) bool {
+	if ctx.RuntimeData[common.RuntimeKeyForceReauth] == dataValueTrue {
+		logger.Debug(ctx.Context, "prompt=login requires a fresh authentication; not reusing the session")
+		return true
+	}
+
+	// A silent request cannot answer a credential prompt, and the authorize endpoint already
+	// confirmed this session satisfies it. Re-deciding here would compare max_age against a later
+	// clock than that check used, so a session on the boundary would be declined and the flow would
+	// prompt a request that forbids prompting. prompt=none cannot be combined with prompt=login, so
+	// the forced-reauth case above still takes precedence.
+	if ctx.RuntimeData[common.RuntimeKeySilentAuthOnly] == dataValueTrue {
+		logger.Debug(ctx.Context,
+			"Silent request already validated at the authorize endpoint; reusing the session")
+		return false
+	}
+
+	rawMaxAge := ctx.RuntimeData[common.RuntimeKeyMaxAge]
+	if rawMaxAge == "" || authenticatedAt.IsZero() {
+		return false
+	}
+	maxAge, err := strconv.ParseInt(rawMaxAge, 10, 64)
+	if err != nil || maxAge < 0 {
+		logger.Debug(ctx.Context, "Ignoring malformed max_age", log.String("maxAge", rawMaxAge))
+		return false
+	}
+	// max_age=0 admits no elapsed time at all, so it demands re-authentication even within the same
+	// second as the authentication (OIDC Core 3.1.2.1 treats it as equivalent to prompt=login)
+	if maxAge == 0 || time.Now().UTC().Unix()-authenticatedAt.Unix() > maxAge {
+		logger.Debug(ctx.Context, "Session authentication is older than max_age; re-authenticating",
+			log.String("maxAge", rawMaxAge))
+		return true
+	}
+	return false
 }
 
 // checkpointRef returns the Session (join) node id this SSO-Check node guards, read from
--- a/backend/internal/oauth/oauth2/authz/service.go
+++ b/backend/internal/oauth/oauth2/authz/service.go
@@ -11,11 +11,13 @@ import (
 	"maps"
 	"net/url"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
 	flowcm "github.com/thunder-id/thunderid/internal/flow/common"
 	"github.com/thunder-id/thunderid/internal/flow/flowexec"
+	flowsession "github.com/thunder-id/thunderid/internal/flow/session"
 	oauthconfig "github.com/thunder-id/thunderid/internal/oauth/config"
 	"github.com/thunder-id/thunderid/internal/oauth/oauth2/authz/requestvalidator"
 	oauth2const "github.com/thunder-id/thunderid/internal/oauth/oauth2/constants"
@@ -30,6 +32,10 @@ import (
 	"github.com/thunder-id/thunderid/internal/system/utils"
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
+
+// runtimeDataTrue is the value the flow layer reads as a set boolean flag on RuntimeData, which
+// carries string values only.
+const runtimeDataTrue = "true"
 
 // AuthorizeServiceInterface defines the interface for authorization services.
 type AuthorizeServiceInterface interface {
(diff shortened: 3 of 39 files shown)


backend/internal/flow/executor/sso_check_executor.go
```diff
@@ -74,6 +75,14 @@ func (e *ssoCheckExecutor) Execute(ctx *providers.NodeContext) (*providers.Execu
 		execResp.RuntimeData[common.RuntimeKeySSOSessionHandle] = resolved.HandleID
 	}
 
+	// A request may demand a fresh authentication even when a live session exists: prompt=login
+	// asks for one outright, and a max_age smaller than the session's age makes the existing
+	// authentication too old to satisfy. Both are answered by re-authenticating here, at the
+	// branch point, rather than by failing at the end of the flow once the assertion is due.
+	if resolved != nil && reauthRequired(ctx, resolved.AuthenticatedAt, logger) {
+		resolved = nil
```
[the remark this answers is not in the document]
Thumimku: the state described isn't reachable. checkPromptNone already returns login_required for a max_age-stale session at the authorize endpoint (service.go:136-138), and it runs before InitiateFlow, so such a request never starts a flow. Both sites resolve the same session with identical max_age semantics, and prompt=none can't be combined with prompt=login (rejected as invalid_request), so reauthRequired firing under a silent request can't happen — covered by TestCheckPromptNone_MaxAgeExceeded. There is a real case where prompt=none reaches a prompt, but it's a missing per-checkpoint snapshot in a multi-stage flow, already noted as a known limitation in the PR description.

tests/integration/oauth/sso/max_age_test.go
```diff
@@ -47,10 +47,43 @@ func (ts *SSOLogoutTestSuite) TestMaxAge_ExpiredSessionRequiresInteraction() {
...
+
+// TestMaxAge_ExpiredSessionReauthenticationSucceeds verifies the other half: after the prompt, the
+// subject can authenticate again and the flow completes, so max_age costs a login rather than
+// failing the request outright.
+func (ts *SSOLogoutTestSuite) TestMaxAge_ExpiredSessionReauthenticationSucceeds() {
+	client := ts.newSessionClient()
+	ts.login(client, ssoMaxAgeUsername, "max_age_reauth_1")
+	ts.Require().NotEmpty(ts.ssoCookieNames(client), "an SSO cookie should be set after first login")
+
+	time.Sleep(3 * time.Second)
```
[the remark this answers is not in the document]
Thumimku: Skipping this one because the mechanism it asks for doesn't exist. There's no helper to backdate a session's AuthenticatedAt and no clock seam in the session service — time.Now() is called directly, and the only runtime knob (/server-config/session) controls session timeouts, not the authentication time that max_age compares against. Building one means adding a clock abstraction to a security-relevant service plus a test-only override reachable from a running server: production change for test convenience, and a design decision rather than a cleanup. The sleeps are also the behaviour under test rather than incidental waiting — these tests verify max_age=1 against a session that must genuinely be older than a second, and the suite already uses the same idiom elsewhere (logout_negative_test.go sleeps out a real ID-token lifetime). Cost is 9 seconds in a suite that already runs 51s with 19s and 15s timeout tests, and there's been no flakiness. Worth doing properly as its own change, converting the time-dependent tests in both max_age_test.go and session_timeout_test.go together.

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -146,7 +146,8 @@ func (a *authAssertExecutor) checkAssurance(ctx *providers.NodeContext,
 			logger.Debug(ctx.Context, "Ignoring malformed max_age", log.String("maxAge", rawMaxAge))
 			return nil
 		}
-		if time.Now().UTC().Unix()-a.resolveAuthTime(ctx) > maxAge {
+		// max_age=0 admits no elapsed time, so it is never satisfied by a reused authentication.
+		if maxAge == 0 || time.Now().UTC().Unix()-a.resolveAuthTime(ctx) > maxAge {
```
thiva-k: Shouldn't the max age check be skipped for silent auth? We are already doing it at sso check executor. So better be consistent

backend/internal/oauth/oauth2/authz/service.go
```diff
@@ -81,10 +95,113 @@ func newAuthorizeService(
...
+	if !slices.Contains(strings.Fields(oauthParams.Prompt), oauth2const.PromptNone) {
+		return "", ""
+	}
+
+	sess := as.resolveSSOSession(ctx, app)
+	if sess == nil {
+		return oauth2const.ErrorLoginRequired, "User authentication is required"
+	}
+
+	if hint := oauthParams.IDTokenHint; hint != "" {
```
thiva-k, line 124: Can't we reuse the existing id token hint verification we already have at ciba, etc.?

backend/internal/flow/executor/auth_assert_executor.go
Thumimku: [max_age=0 means prompt=login](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest), so we have to check that assertion directly. 
in SSO check, onFailure means prompt_credentials and method reauthRequired returns true without any error.
SSO check added for this [comment](https://github.com/thunder-id/thunderid/pull/5255#discussion_r3914652350), Treat max_age=0 as mandatory reauthentication.
but in this check (auth_assertion check), we already had same second part of this check ErrInteractionRequired so I need to apply the fix to be consistent in this file also, cause sso authenticator may wont be there for the flows that doesn't have SSO.

backend/internal/oauth/oauth2/authz/service.go
Thumimku: CIBA has similar function but the diff is here we dont validate exp cause 
```
The OP SHOULD accept ID Tokens when the RP identified by the ID Token has a current session or had a recent session at the OP, even when the exp time has passed.
```
and returning error in the ciba method is ciba specific.

backend/internal/flow/executor/auth_assert_executor.go
Thumimku: @thiva-k 
You are correct. The max_age=0 check in this code is not only redundant but also logically incorrect.

If there is an existing session, the session executor will handle max_age=0 and require re-authentication as needed. If there is no existing session and max_age=0 is provided without prompt=none, the user is already going through a fresh login.

Therefore, checking max_age=0 again at this point—after authentication has already completed—would unnecessarily trigger another login.

thiva-k (approved):

Issue #4981, opened 2026-08-15
Review and rework the release builder pipeline


rajithacharith:
## Summary

The release builder pipeline needs an end to end review. A release run recently produced integration test failures that had nothing to do with the product, and the first two attempts to fix it did not take effect because of how the workflow is wired. The pipeline should be reworked so that a release can be built and verified from any branch, with the tests always running against the artifact that is actually being released.

## Background

Release run [31869966705](https://github.com/thunder-id/thunderid/actions/runs/31869966705) failed on both SQLite and PostgreSQL with five export test failures:

- `TestExportAPITestSuite/TestCredentialConfigurationExport`
- `TestExportAPITestSuite/TestCredentialConfigurationExportWithWildcard`
- `TestExportAPITestSuite/TestPresentationDefinitionExport`
- `TestExportAPITestSuite/TestPresentationDefinitionExportWithWildcard`
- `TestExportAPITestSuite/TestVCResourcesExportYAML`

All five returned `EXP-1002 No resources found`. The cause was a ref split, not a product defect. The `build` job checked out `1.0.0-release` and produced the distribution, but the `run-integration-tests` composite action performed its own `actions/checkout` with no `ref`, which resolved to the dispatch ref (`main`) and replaced the workspace. The OpenID4VCI and OpenID4VP export support was added in `4b0cbb54f`, which is on `main` but not on `1.0.0-release`, so tests for a feature that did not exist in the binary ran against it.

A follow up fix (#4980) added a `ref` input to the composite action, but run [31872087120](https://github.com/rajithacharith/thunder/actions/runs/31872087120) still failed with:

```
##[warning]Unexpected input(s) 'ref', valid inputs are ['database-type', 'coverage-enabled']
```

A local composite action referenced as `uses: ./...` is read from `$GITHUB_WORKSPACE`, so the runner loaded the copy of `action.yml` from the release branch, which predates the new input. The workflow file comes from the dispatch ref while the action file comes from whatever was checked out, and the two can silently disagree. This class of failure will recur on every release branch older than a workflow change.

## Requirements

1. **A release must be buildable and verifiable from any branch.** The `branch` input should select the source of truth for every job in the pipeline, with no job silently falling back to the dispatch ref.
2. **Tests must not run against a different ref than the build.** The distribution under test and the test sources must come from one commit, and that commit should be recorded in the run summary so a reviewer can confirm it at a glance.
[...]

PR #5273, opened 2026-09-03
Prepare branch for the pinned-commit release builder

Osara-B:
### Purpose

The release builder is being reworked so a release can be built and verified from any branch (#4981). It pins the source branch to a single immutable commit and checks that same commit out in every job, which means the files it reads at release time come from the branch being released rather than from the branch the workflow was dispatched from.

The files in this PR are read that way, so they have to be correct here before a release can be cut from this branch. This PR brings them in ahead of the workflow change on `main`.

The integration test action also fixes a live bug on this branch. It checks out a ref of its own, which replaces `$GITHUB_WORKSPACE` in the middle of the job and discards whatever the caller had checked out. On this branch the release builder checks out `RELEASE_BRANCH` and the action then reverts the workspace to the dispatched ref, so the integration suite can run against different sources than the ones the distribution under test was built from.

### Approach

Removed the `actions/checkout` step from `.github/actions/run-integration-tests`. Composite actions resolve out of `$GITHUB_WORKSPACE`, so a checkout inside one can silently swap the tree the rest of the job runs against. All three callers on this branch (`pr-builder.yml`, `postgres-tests.yml` and `release-builder.yml`) already check out the sources themselves before invoking the action, so nothing changes on the calling side.

Added a guard so the action skips its artifact download when the caller has already staged a distribution in `target/dist`. 

Added `.github/docker/Dockerfile.release`, which assembles the container image by unzipping the already published Linux archives instead of recompiling the product inside `docker buildx`. Extraction happens on `$BUILDPLATFORM` to avoid QEMU emulation, and the two `deployment.yaml` edits are applied to the packaged file rather than to the source tree.

The root `Dockerfile` is functionally unchanged and still builds from source for `make docker-build*`. It gains a comment naming the release Dockerfile, because the two share a runtime stage that has to be kept in sync.

### Related Issues
-  #4981 

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

### DIFF
--- a/.github/docker/Dockerfile.release
+++ b/.github/docker/Dockerfile.release
@@ -1,0 +1,63 @@
+# Copyright 2026 The ThunderID Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Built by .github/workflows/release-builder.yml from the published distribution
+# instead of from source. The build context is a directory of Linux archives.
+# Keep the runtime stage in sync with ./Dockerfile.
+
+# Unpack the target architecture's archive, on the build platform to avoid emulation.
+FROM --platform=$BUILDPLATFORM alpine:3.19 AS dist
+
+ARG TARGETARCH
+
+RUN apk add --no-cache unzip
+
+WORKDIR /dist
+
+COPY thunderid-*-linux-*.zip ./
+
+RUN set -eux; \
+    case "$TARGETARCH" in \
+      amd64) archive_arch=x64 ;; \
+      arm64) archive_arch=arm64 ;; \
+      *) echo "unsupported target architecture: $TARGETARCH" >&2; exit 1 ;; \
+    esac; \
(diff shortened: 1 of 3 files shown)


.github/docker/Dockerfile.release
```diff
@@ -0,0 +1,63 @@
+# Copyright 2025 The ThunderID Authors
```
rajithacharith:
```suggestion
# Copyright 2026 The ThunderID Authors
```

Osara-B: Resolved in [Prepare branch for the pinned-commit release builder](https://github.com/thunder-id/thunderid/pull/5273/commits/34a1360ee6297b83b2d05b61dfa41e8663c8163b)

rajithacharith (approved):

Issue #5192, opened 2026-08-27
Authorization support for federated entities


thiva-k:
## Problem

In many deployments, users are managed in an external identity provider such as Entra ID or Okta, while ThunderID manages the agents, applications, and the permission model. Those entities federate in carrying attributes issued by the provider, and none of it influences the access ThunderID grants.

### External attribute values carry no authorization meaning

ThunderID normalizes external claim *names* into local attribute names, but nothing interprets an attribute *value*. An entity arriving with `groups: ["engineering", "platform-admins"]` means nothing to the authorization engine.

The only value resolution that exists today, used to select a user type, reads a single scalar value. Attributes that carry authorization are usually collections, such as a list of groups or a space-delimited scope string, so that mechanism cannot consume the shapes providers actually emit.

Two consequences follow:

- An administrator assigns roles to every federated entity by hand, and repeats the work whenever group membership changes at the provider.
- A federated identity with no local record cannot be authorized at all, because permission resolution has always started from a local entity.

### Applications and agents acting on an entity's behalf

An application or agent that exchanges an external token to act for an entity receives nothing derived from that entity's attributes, so delegated access does not reflect what the entity actually holds. There is also no way to bound the result by the application or agent's own authorization. Agent identities managed in an external provider have the same gap when acting under their own identity.

### Organizations that already own their authorization policy

Some organizations manage authorization policy in an external policy decision point, over attributes ThunderID's role model does not represent. Without a way to delegate the decision, they must duplicate that policy inside ThunderID, accept tokens broader than their policy intends, or build a bespoke integration.

### Entry points

External identities reach ThunderID through three entry points, and none of them can use external attribute values today:

- Federated login, where ThunderID matches an existing local entity or provisions one just in time.
- Token exchange (RFC 8693), where an application or agent exchanges an external token to act on the entity's behalf.
- An identity assertion obtained from the provider for a specific resource server.

## Related

- #5126 - External AuthZEN PDP evaluation
- #4596 - Federated attribute-based permission assignment for externally managed users
- #3157 - Mapping external IdP claims to local claims

sahandilshan:
Discussion related to this issue can be found https://github.com/thunder-id/thunderid/discussions/5197

blev8824-ai:
[...]

PR #5275, opened 2026-09-03
Add authorization mapping for federation, token exchange, and ID-JAG grants

thiva-k:
### Purpose
Map external IDP claim values on a connection to local roles, groups, and permissions, resolved consistently across federated sign-in, RFC 8693 token exchange, and the ID-JAG (jwt-bearer) consumption leg. 

### Approach
- Resolution (`GetMappedAuthorizationTargets`, `SplitAuthorizationTargets`, `UnionMappedPermissionTargets`) lives in `internal/idp`, beside the existing attribute-mapping resolver; types live in `pkg/thunderidengine/providers`.
- Federated sign-in resolves the mapping in the OAuth/OIDC executors and carries results via flow runtime data to the authorization and provisioning executors.
- Token exchange and the ID-JAG jwt-bearer grant resolve the mapping while validating the presented token and apply it via a shared `tokenservice.ApplyMappedAuthorization`.
- Role/group targets are evaluated by the RBAC engine as subject context (no local entity id required); permission targets are granted directly.
- Connection create/update validation rejects unknown roles/groups/permissions and operator/value-type mismatches.

### Related Issues
- Refs #5192 

## Summary by CodeRabbit

* **New Features**
  * Added claim-based authorization mappings for federated sign-in, token exchange, and ID-JAG flows.
  * Map claims to roles, groups, or resource-server permissions using direct lookups or configurable rules.
  * Support scalar, list, delimited, typed, and membership-based claim values.
  * Mapped permissions combine with existing authorization and remain resource-server scoped.
  * Optionally seed mapped roles and groups during provisioning.
* **Validation**
  * Added validation for mapping targets, operators, value types, and resource servers.
* **Bug Fixes**
  * Prevented ambiguous, invalid, unmatched, or incorrectly scoped mappings from granting access.


backend/pkg/thunderidengine/providers/model.go
```diff
@@ -792,7 +792,168 @@ type AccountLinking struct {
...
+	}
+	return m.ValueType
+}
+
+// IsMultiValued reports whether the mapping declares a set-membership claim (an array, or a
+// delimited string) rather than a single value to compare.
+func (m AuthorizationRuleMapping) IsMultiValued() bool {
+	valueType := m.EffectiveValueType()
+	return valueType == AuthorizationValueTypeArray ||
+		(valueType == AuthorizationValueTypeString && m.Delimiter != "")
```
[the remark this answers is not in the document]
thiva-k: Docs will be added separately

backend/internal/role/store_constants.go
```diff
@@ -173,27 +173,59 @@ var (
...
 	requestedPermissions []string,
 	deploymentID string,
 ) (dbmodel.DBQuery, []interface{}) {
-	// Base query structure
+	// Base query structure. LEFT JOIN, not INNER JOIN: a role named directly in roleIDs (an
+	// externally derived role, holding no stored assignment) must still surface its permissions.
 	baseQuery := `SELECT DISTINCT rp.PERMISSION
 		FROM "ROLE_PERMISSION" rp
-		INNER JOIN "ROLE_ASSIGNMENT" ra ON rp.ROLE_ID = ra.ROLE_ID AND rp.DEPLOYMENT_ID = $1 AND ra.DEPLOYMENT_ID = $1
+		LEFT JOIN "ROLE_ASSIGNMENT" ra ON rp.ROLE_ID = ra.ROLE_ID AND rp.DEPLOYMENT_ID = $1 AND ra.DEPLOYMENT_ID = $1
```
sahandilshan, line 221: @coderabbitai

Can you check if this leads to privilage escalations or any other issues?

backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -358,6 +365,11 @@ func (tv *tokenValidator) ValidateIDJAGAssertion(
 		return nil, fmt.Errorf("invalid assertion signature: %v", svcErr.Error)
 	}
 
+	authorization, err := tv.resolveMappedAuthorization(ctx, &issuerInfo.IDPDTO, claims)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve authorization mapping for issuer %q: %w", iss, err)
+	}
+
```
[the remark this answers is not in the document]
thiva-k: Fixed

backend/internal/oauth/oauth2/granthandlers/token_exchange.go
```diff
@@ -456,23 +477,34 @@ func (h *tokenExchangeGrantHandler) validateAccessTokenType(
...
 
 	requestedScopes := tokenservice.ParseScopes(tokenRequest.Scope)
-
 	if len(requestedScopes) == 0 {
 		return []string{}, nil
 	}
 
-	// If subject token has no scopes, reject requests asking for scopes
+	if authorityIsMapping {
+		return requestedScopes, nil
```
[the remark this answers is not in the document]
thiva-k: Documentation will be done separately

thiva-k:
Related follow-up - #5399 (connection authorization mapping needs the same grant-ceiling check as role/group assignment when fine-grained access control is introduced for connections).

backend/internal/authz/engine/rbacengine.go
```diff
@@ -107,6 +108,7 @@ func findEvaluationGroup(groups []evaluationGroup, subject Subject, resourceServ
 		if group.subject.Type == subject.Type &&
 			group.subject.ID == subject.ID &&
 			slices.Equal(group.subject.GroupIDs, subject.GroupIDs) &&
+			slices.Equal(group.subject.RoleIDs, subject.RoleIDs) &&
```
sahandilshan, line 111: Why this was not used before and introduce with this PR? This should be a check that should be there from the begining, right?

thiva-k: Roles used to always come from assignments of the local entity and was derived inside the rbac engine. Now mapping can grant it with no local assignment at all, so RoleIDs is added to carry it.

backend/internal/flow/executor/utils.go
```diff
@@ -236,6 +238,53 @@ func setFederatedEntityState(ctx context.Context, execResp *providers.ExecutorRe
...
+	}
+	setMappedAuthorizationTargets(execResp, targets)
+}
+
+// setMappedAuthorizationTargets splits resolved authorization mapping targets (rule-based or direct)
+// by kind and stores them as runtime data for later executors. No-op when there is nothing to store.
+func setMappedAuthorizationTargets(execResp *providers.ExecutorResponse, targets []providers.AuthorizationTarget) {
+	if len(targets) == 0 {
+		return
+	}
```
sahandilshan: @coderabbitai verify this

This is very highly unlikely to happen scenario. But when we return if mapping is not empty, in this scenarion if a role or group has a role mapping like below
```
{
  "mapped_role_ids": "019abc-internal-role-uuid"
}
```
the map still preserve it and pass to the authz executor. And from authz executor it has the ability to resolve the role with id and it will internally resolve the user to a role with provided id, right?

sahandilshan: Verified mean this is an issue, right

sahandilshan: @coderabbitai Please re-analyze from the claim-ingestion path, not only RBAC. oauth_executor.go copies external claims into RuntimeData before mapping, so an IdP-supplied mapped_role_ids can survive empty or partial resolution and later be trusted as Subject.RoleIDs. Provisioning seeding can also persist it. Temporary authorization is still a security impact. Please add this trust-boundary pattern to your learnings: external claims must not share reserved internal authorization keys.

backend/internal/flow/executor/authz_executor.go
```diff
@@ -124,13 +125,21 @@ func (a *authorizationExecutor) Execute(ctx *providers.NodeContext) (*providers.
 		return nil, errors.Join(errors.New("Failed to extract group IDs"), err)
 	}
 
+	// Fold in any roles, groups, and permissions the federated login mapped from IDP claims (via the
+	// authorization mapping, rule-based or direct).
+	mappedRoleIDs := utils.ParseStringArray(ctx.RuntimeData[common.RuntimeKeyMappedRoleIDs], " ")
+	mappedGroupIDs := utils.ParseStringArray(ctx.RuntimeData[common.RuntimeKeyMappedGroupIDs], " ")
+	mappedPermissions := extractMappedPermissions(ctx, a.logger)
+	groupIDs = utils.MergeUniqueStrings(groupIDs, mappedGroupIDs)
```
sahandilshan, line 134: Here have we consider the cases of parent groups. In thunderID we can assign group to another group. In that case if the given map group has a parent group and when authorization mapping happens do we consider that case?

thiva-k: Updated to consider the transitive groups too for the mapped path

backend/internal/flow/executor/utils.go
thiva-k: This is fixed by adding a reserved key list

backend/internal/oauth/oauth2/granthandlers/jwt_bearer.go
```diff
@@ -106,12 +117,18 @@ func (h *jwtBearerGrantHandler) HandleGrant(ctx context.Context, tokenRequest *m
...
+	// when configured, ApplyMappedAuthorization (below).
+	requested := tokenservice.ParseScopes(tokenRequest.Scope)
+	var grantedScopes []string
+	if assertionClaims.Authorization.Configured {
+		grantedScopes = intersectScopes(assertionClaims.Scopes, requested)
+	} else {
+		grantedScopes = assertionClaims.Scopes
+		if tokenRequest.Scope != "" {
+			grantedScopes = intersectScopes(grantedScopes, requested)
+		}
```
sahandilshan, line 131: If requested scope is empty, does that mean even if the scope in the ID-Jag has a valid scope, it will be ignored. Is the implementation follow this
```
If request scope is provided:
    ID-JAG scopes ∩ request scopes ∩ mapping
If request scope is omitted:
    ID-JAG scopes ∩ mapping
```

@coderabbitai can you verify this

thiva-k: @coderabbitai can you check if this concern is fixed now and resolve if so

sahandilshan (approved):

PR #5290, opened 2026-09-04
Add current password validation to self-service credential updates

janithjay:
### Purpose

`POST /users/me/update-credentials` wrote a new password on the strength of the access token alone. Anyone holding a stolen or leaked token for an account could set a password on it, and that password keeps working long after the token expires, so temporary token possession turns into permanent account access. The user now has to prove they know the current password before the write goes through.

The endpoint also accepted any schema-declared credential, not just the password. That is narrowed here too, so proving your password cannot be used to write a different credential type.

---
Two changes to `POST /users/me/update-credentials`:

1. A `currentPassword` field is now required when the account already has a password stored. Requests that omit it, or supply the wrong value, are rejected with `403` and code `USR-1029`.
2. Only `password` may be written. Any other credential type in `attributes` is rejected with `400` and code `USR-1024`.

The request body schema changed from `UpdateSelfUserRequest` to a dedicated `UpdateSelfCredentialsRequest`, which carries `currentPassword` as a sibling of `attributes`.

#### 💥 Impact
Existing clients calling this endpoint without `currentPassword` will start receiving `403` for accounts that have a password. The endpoint has existed since the user self service API was introduced, so this affects any integration already built against it. Both bundled sample apps were affected and are updated in this PR.

Clients writing a credential type other than `password` through this endpoint will start receiving `400`. Use the admin credential update endpoint for those.

The admin reset path `POST /users/{userId}/update-credentials` is unchanged. It still needs no current password and still accepts any schema-declared credential, since an admin cannot know the target user's password.

#### 🔄 Migration Guide
Add the user's existing password to the request body:

```json
{
  "currentPassword": "0ldP@ssword!",
  "attributes": {
    "password": "n3wP@ssword!"
  }
}
```

Clients should surface a `403` carrying `USR-1029` against the current password field rather than as a generic failure, since it means only that the supplied password was wrong.

---

#### Backend
- `backend/internal/user/service.go` - `UpdateSelfUserCredentials` reads the stored password credential first, then requires and verifies `currentPassword` before delegating to `UpdateUserCredentials`. Verification reuses `AuthenticateEntityByID` instead of adding a second password comparison path, so there stays exactly one place in the codebase where a password is checked against storage.
- Kept as a separate method from `UpdateUserCredentials` (the admin path) rather than adding a flag to it. The two have different trust models, and a boolean parameter deciding whether to verify is the kind of thing that eventually gets passed wrong.
- `validateSelfCredentialPayload` rejects any credential key other than `password`. 
- An account with no password stored yet proceeds without verification. There is nothing to verify against, and demanding proof there would leave those users unable to ever set a password.
- `backend/internal/user/model.go` - adds `UpdateSelfCredentialsRequest`, with `CurrentPassword` as a sibling of `Attributes` rather than a member of it. The entity layer only accepts schema-declared credential keys inside `attributes`, so a current password nested there would be rejected as an unknown credential.
- `backend/internal/user/constants.go` - names `CredentialTypePassword`, since password now acts as the account level proof of ownership and not just one more credential type.
- `backend/internal/user/error_constants.go` and `backend/internal/user/handler.go` - add `ErrorInvalidCurrentPassword` (`USR-1029`), mapped to `403`. `403` rather than `401` because the caller is authenticated and it is this specific action that is refused; a `401` would tell the client its session is invalid and should be re-established, which is not the case. `USR-1024` for the rejected credential type maps to `400`.
- `backend/internal/system/i18n/core/defaults.go` - regenerated with `make generate_i18n` for the two new message keys, not hand edited.

#### Sample apps
Both bundled samples called this endpoint without a current password and broke against the new contract. Those are addressed here,

- `samples/apps/wayfinder-sample/frontend/src/api/userApi.js` and `src/pages/ProfilePage.jsx` - `updateMyCredentials` takes a current password and sends it when present, so a first-time set still works. The profile form gains a "Current password" field, wired into validation and the submit gate.
- `samples/apps/vanilla-sample/src/services/userProfileService.ts` and `src/views/ProfilePage.tsx` - the same change. The Next.js route at `src/app/api/profile/password/route.ts` proxies the body as-is and needed nothing.

This PR is related to [[Design Discussion] Require current credential verification on self-service credential update](https://github.com/thunder-id/thunderid/discussions/5227)

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5291

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Self-service password changes now require the current password when one is already set.
  * First-time password setup remains available without current-password verification.
  * Self-service credential updates now accept password credentials only.

* **Bug Fixes**
  * Incorrect or missing current passwords return HTTP 403 Forbidden.
  * Unsupported credential types return a documented validation error.
  * Admin credential resets continue without requiring the user’s current password.

* **Documentation**
  * Updated guidance to explain current-password requirements for password changes.

### DIFF
--- a/api/user.yaml
+++ b/api/user.yaml
@@ -1347,8 +1347,9 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UpdateSelfUserRequest'
+              $ref: '#/components/schemas/UpdateSelfCredentialsRequest'
             example:
+              currentPassword: "0ldP@ssword!"
               attributes:
                 password: "n3wP@ssword!"
       responses:
@@ -1371,6 +1372,16 @@ paths:
                     description:
                       key: "error.userservice.missing_credentials_description"
                       defaultValue: "At least one credential field must be provided"
+                unsupported-credential-type:
+                  summary: A credential type other than password was supplied
+                  value:
+                    code: "USR-1024"
+                    message:
+                      key: "error.userservice.invalid_credential"
+                      defaultValue: "Invalid request format"
+                    description:
+                      key: "error.userservice.invalid_credential_description"
+                      defaultValue: "Invalid credential fields in request"
         "401":
           description: Unauthorized - missing or invalid authentication token
           content:
@@ -1385,6 +1396,20 @@ paths:
                 description:
                   key: "error.unauthorized_description"
                   defaultValue: "Authentication is required to access this resource"
+        "403":
+          description: Forbidden - the supplied current password is incorrect
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+              example:
+                code: "USR-1029"
+                message:
+                  key: "error.userservice.invalid_current_password"
+                  defaultValue: "Invalid current password"
+                description:
+                  key: "error.userservice.invalid_current_password_description"
+                  defaultValue: "The provided current password is incorrect"
         "404":
           description: Authenticated user not found
           content:
@@ -2190,6 +2215,27 @@ components:
           type: object
           description: "User attributes"
           additionalProperties: true
+
+    UpdateSelfCredentialsRequest:
+      type: object
+      required: [attributes]
+      properties:
+        currentPassword:
+          type: string
+          format: password
+          description: "The user's existing password, verified before the write. Required once the account has a password. Omit it for a first-time set, on an account with no password yet."
+          example: "0ldP@ssword!"
+        attributes:
+          type: object
+          description: "Credential attributes to write. Only `password` is accepted here; other types are rejected with 400 USR-1024."
+          required: [password]
+          properties:
+            password:
+              type: string
+              format: password
+              description: "The new password to set."
+              example: "n3wP@ssword!"
+          additionalProperties: false
 
     UpdateUserCredentialsRequest:
       type: object
--- a/backend/internal/user/constants.go
+++ b/backend/internal/user/constants.go
@@ -14,6 +14,11 @@ const (
 	CredentialTypePasskey CredentialType = "passkey"
 )
 
+// CredentialTypePassword is the schema-defined password credential. It is not system-managed, but
+// it is named here because it acts as the account-level proof of ownership when a user changes
+// their own credentials.
+const CredentialTypePassword CredentialType = "password"
+
 // systemManagedCredentialTypes defines credential types that are managed by the system,
 // not through user types. These may support multiple values per user.
 var systemManagedCredentialTypes = []CredentialType{
--- a/backend/internal/user/error_constants.go
+++ b/backend/internal/user/error_constants.go
@@ -141,6 +141,20 @@ var (
 			DefaultValue: "At least one identifying attribute must be provided",
 		},
 	}
+	// ErrorInvalidCurrentPassword is returned when the authenticated user's supplied current
+	// password does not match the stored credential.
+	ErrorInvalidCurrentPassword = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "USR-1029",
+		Error: tidcommon.I18nMessage{
+			Key:          "error.userservice.invalid_current_password",
+			DefaultValue: "Invalid current password",
+		},
+		ErrorDescription: tidcommon.I18nMessage{
+			Key:          "error.userservice.invalid_current_password_description",
+			DefaultValue: "The provided current password is incorrect",
+		},
+	}
 	// ErrorMissingCredentials is the error returned when credentials are missing.
 	ErrorMissingCredentials = tidcommon.ServiceError{
 		Type: tidcommon.ClientErrorType,
--- a/backend/internal/user/handler_test.go
+++ b/backend/internal/user/handler_test.go
@@ -145,7 +145,7 @@ func TestHandleSelfUserCredentialUpdateRequest_Success(t *testing.T) {
 
 	mockSvc := NewUserServiceInterfaceMock(t)
 	credentialsJSON := json.RawMessage(`{"password":[{"value":"Secret123!"}]}`)
-	mockSvc.On("UpdateUserCredentials", mock.Anything, userID, credentialsJSON).Return(nil)
+	mockSvc.On("UpdateSelfUserPassword", mock.Anything, userID, "", credentialsJSON).Return(nil)
 
 	handler := newUserHandler(mockSvc)
 	req := httptest.NewRequest(http.MethodPost, "/users/me/update-credentials",
@@ -165,7 +165,7 @@ func TestHandleSelfUserCredentialUpdateRequest_StringValue(t *testing.T) {
 
 	mockSvc := NewUserServiceInterfaceMock(t)
 	credentialsJSON := json.RawMessage(`{"password":"plaintext-password"}`)
-	mockSvc.On("UpdateUserCredentials", mock.Anything, userID, credentialsJSON).Return(nil)
+	mockSvc.On("UpdateSelfUserPassword", mock.Anything, userID, "", credentialsJSON).Return(nil)
 
 	handler := newUserHandler(mockSvc)
 	req := httptest.NewRequest(http.MethodPost, "/users/me/update-credentials",
@@ -241,7 +241,7 @@ func TestHandleSelfUserCredentialUpdateRequest_ErrorCases(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			mockSvc := NewUserServiceInterfaceMock(t)
-			mockSvc.On("UpdateUserCredentials", mock.Anything, userID, tc.mockJSON).Return(tc.mockError)
+			mockSvc.On("UpdateSelfUserPassword", mock.Anything, userID, "", tc.mockJSON).Return(tc.mockError)
 
 			handler := newUserHandler(mockSvc)
 			req := httptest.NewRequest(http.MethodPost, "/users/me/update-credentials",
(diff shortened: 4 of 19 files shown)


api/user.yaml
```diff
@@ -1347,8 +1347,9 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UpdateSelfUserRequest'
+              $ref: '#/components/schemas/UpdateSelfCredentialsRequest'
             example:
+              currentPassword: "0ldP@ssword!"
               attributes:
                 password: "n3wP@ssword!"
```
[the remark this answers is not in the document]
janithjay: not valid anymore

api/user.yaml
```diff
@@ -1347,8 +1347,9 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/UpdateSelfUserRequest'
+              $ref: '#/components/schemas/UpdateSelfCredentialsRequest'
             example:
+              currentPassword: "0ldP@ssword!"
```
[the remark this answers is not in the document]
janithjay: Partly valid, took the doc fix and skipped the rest.

Skipping the SDK part: those will fix in `thunder-id/javascript-sdks` side.

docs/content/use-cases/b2c/try-it-out/profile-section.mdx
```diff
@@ -24,7 +24,7 @@ Complete [Set Up Your Environment](../../build-environment) before starting this
...
 
-In the redirect-based pattern, the consumer application renders the profile screen itself. Profile attributes are read from `/users/me` and ID token. Updates go through <ProductName />'s self-service endpoints: `/users/me` for attribute changes, `/users/me/meta` for schema metadata retrieve, and `/users/me/update-credentials` for password changes. These endpoints act on the signed-in user's own record, so no extra permissions are needed; the access token alone is enough.
+In the redirect-based pattern, the consumer application renders the profile screen itself. Profile attributes are read from `/users/me` and ID token. Updates go through <ProductName />'s self-service endpoints: `/users/me` for attribute changes, `/users/me/meta` for schema metadata retrieve, and `/users/me/update-credentials` for password changes. These endpoints act on the signed-in user's own record, so no extra permissions are needed and the access token alone is enough to read attributes and schema metadata. Changing a password additionally requires the user's current password in the request body, and `/users/me/update-credentials` returns `403` with code `USR-1029` when it is missing or incorrect.
```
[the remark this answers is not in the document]
janithjay: fixed

backend/internal/user/service.go
```diff
@@ -687,6 +689,57 @@ func (us *userService) UpdateUserAttributes(
...
+			case errors.Is(authErr, entity.ErrEntityNotFound):
+				return &ErrorUserNotFound
+			default:
+				return logErrorAndReturnServerError(ctx, logger, "Failed to verify current password", authErr,
+					log.MaskedString(log.LoggerKeyUserID, userID))
+			}
+		}
+	}
+
+	return us.UpdateUserCredentials(ctx, userID, credentials)
```
thiva-k, line 743: With this if the payload sends another credential (e.g. `secret`), then that will be updated for the user if the `password` is not in attribtues right? We should restrict the update to only password field by checking if the schema has it and only `password` is sent.

janithjay: fixed. `UpdateSelfUserCredentials` now rejects any payload key other than `password` with `400` and `USR-1024`, before it reads the stored credential. So proving your password no longer lets you write a different credential type.

backend/internal/user/service.go
```diff
@@ -687,6 +689,86 @@ func (us *userService) UpdateUserAttributes(
...
```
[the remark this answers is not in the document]
janithjay: The gap is real. `AuthenticateEntityByID` reads outside any transaction and `UpdateUserCredentials` opens its own, so there is a check-then-act window.

Low severity though. Both racing requests have to pass verification, so in security perspective if an attacker already knows the current password and gains nothing by winning the race.

Not fixing it here for two reasons. A conditional update means a new method on `EntityServiceInterface`, which is what the discussion https://github.com/thunder-id/thunderid/discussions/5227#discussioncomment-18244033 mentions of reusing `AuthenticateEntityByID`. 
And the locking it needs does not exist: `dbTransactioner.Transact` calls `BeginTx(ctx, nil)` with default isolation, and there is no `SELECT ... FOR UPDATE` or version column in `internal/entity`, so wrapping both calls
in one transaction would not actually close the window.

That is an entity-layer concurrency change affecting every caller, so it belongs in its own issue rather than in the change that adds the verification step.

backend/internal/user/service.go
```diff
@@ -687,6 +689,86 @@ func (us *userService) UpdateUserAttributes(
 	return &existingUser, nil
 }
 
+// UpdateSelfUserCredentials updates the authenticated user's own password after verifying their
+// current one. Kept separate from UpdateUserCredentials (the admin reset path), since an admin
+// can't supply the target's current password.
+// A user with no password stored yet is treated as a first-time set.
+func (us *userService) UpdateSelfUserCredentials(
```
thiva-k:
```suggestion
func (us *userService) UpdateSelfUserPassword(
```

janithjay: updated

thiva-k (approved):

backend/internal/user/service.go
```diff
@@ -687,6 +689,86 @@ func (us *userService) UpdateUserAttributes(
...
+	stored, err := us.entityService.GetCredentialsByType(ctx, userID, string(CredentialTypePassword))
+	if err != nil {
+		if errors.Is(err, entity.ErrEntityNotFound) {
+			return &ErrorUserNotFound
+		}
+		return logErrorAndReturnServerError(ctx, logger, "Failed to read stored credentials", err,
+			log.MaskedString(log.LoggerKeyUserID, userID))
+	}
+
+	if len(stored) > 0 {
```
ThaminduDilshan, line 721: For an account with no stored password, the access token alone still writes one. Is that an expected behaviour?

backend/internal/user/service.go
```diff
@@ -687,6 +689,86 @@ func (us *userService) UpdateUserAttributes(
...
+	}
+
+	return us.UpdateUserCredentials(ctx, userID, credentials)
+}
+
+// validateSelfCredentialPayload restricts a self-service credential write to the password.
+// Rejecting anything else is checked before the stored password is read, so an invalid payload
+// costs no lookup and reveals nothing about whether the account has a password set.
+func validateSelfCredentialPayload(credentials json.RawMessage) *tidcommon.ServiceError {
+	if len(credentials) == 0 {
```
ThaminduDilshan: `UpdateUserCredentials` already does these same checks right below. Shall we keep only the password type check here?

backend/internal/user/error_constants.go
```diff
@@ -141,6 +141,20 @@ var (
 			DefaultValue: "At least one identifying attribute must be provided",
 		},
 	}
+	// ErrorInvalidCurrentPassword is returned when the authenticated user's supplied current
+	// password does not match the stored credential.
+	ErrorInvalidCurrentPassword = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "USR-1029",
```
ThaminduDilshan, line 148: Shall we move this below `USR-1028` to keep the error constants in ascending order? Having them out of order could lead to duplicate codes in future

backend/internal/user/handler_test.go
```diff
@@ -267,7 +267,7 @@ func TestHandleSelfUserCredentialUpdateRequest_MultipleCredentialTypes(t *testin
 	mockSvc := NewUserServiceInterfaceMock(t)
 	// Test that multiple credential types are updated in a single atomic call
 	credentialsJSON := json.RawMessage(`{"password":"new-password","pin":"1234"}`)
-	mockSvc.On("UpdateUserCredentials", mock.Anything, userID, credentialsJSON).Return(nil)
+	mockSvc.On("UpdateSelfUserPassword", mock.Anything, userID, "", credentialsJSON).Return(nil)
```
ThaminduDilshan, line 270: The service now rejects `pin` alongside `password`. Shall we update this test? The name and the comment no longer match what the endpoint does

tests/integration/user/user_self_api_test.go
```diff
@@ -321,3 +334,97 @@ func (s *SelfUserEndpointsSuite) TestSelfUserUpdateCredentialsMissingAttributesR
...
+		})
+	s.Require().NoError(err)
+	defer resp.Body.Close()
+
+	s.requireSelfError(resp, http.StatusForbidden, "USR-1029")
+	s.requireExistingPasswordStillWorks()
+}
+
+// TestSelfUserUpdateCredentialsMissingCurrentPasswordRejected verifies that omitting the current
+// password fails the same way as supplying a wrong one, so the response does not reveal whether
```
ThaminduDilshan, line 371: An account with no stored password gets a `204` for this exact request, so the response does reveal it. Shall we fix the comment?

api/user.yaml
```diff
@@ -2191,6 +2216,27 @@ components:
           description: "User attributes"
           additionalProperties: true
 
+    UpdateSelfCredentialsRequest:
+      type: object
+      required: [attributes]
+      properties:
+        currentPassword:
```
ThaminduDilshan, line 2223: Shall we keep this generic rather than hard coding it to the password? In thunderid, credentials can be defined as you wish and there's no mandatory field called password. IMO our API design should also allow that flexibility. Existing `attributes` field allows that, and this conflicts with that.

backend/internal/user/constants.go
```diff
@@ -14,6 +14,11 @@ const (
 	CredentialTypePasskey CredentialType = "passkey"
 )
 
+// CredentialTypePassword is the schema-defined password credential. It is not system-managed, but
+// it is named here because it acts as the account-level proof of ownership when a user changes
+// their own credentials.
+const CredentialTypePassword CredentialType = "password"
```
ThaminduDilshan, line 20: Do we expect credential type to be password even if the credential attribute name is defined as something else in the schema? Or the attribute name itself has to be password for this to work?

api/user.yaml
ThaminduDilshan: Also raised in: https://github.com/thunder-id/thunderid/discussions/5227#discussioncomment-18342087

backend/internal/user/service.go
janithjay: Yes, expected. This is for accounts that never had a password, OTP-only, invited users, who need a way to set a first one. There's no stored value to prove against, so the token itself is the proof.

PR #5297, opened 2026-09-06
Make request headers and query params available to the HTTP Request Executor

### LINKED ISSUE #1354: Make request params available for the Http Request Executor
### Current Limitation

`HTTPRequestExecutor` allows to use context attributes in the request payloads. However currently we only allow the usage of user inputs and runtime data only. It's a valid requirement to publish some http request params such as hostname, user agent and some headers for analytics purposes.

### Suggested Improvement

Allow using http request params in HTTPRequestExecutor payloads

### Version

_No response_

Thisara-Welmilla:
### Purpose

Allow the `HTTPRequestExecutor` to publish HTTP request headers and query parameters (host, user agent, tracking params, etc.) in its outbound payloads, as requested in #1354.

Previously the executor could only template runtime data and user inputs via `{{ctx(...)}}`. This PR adds a `{{request(...)}}` placeholder syntax that resolves headers and query parameters from the request associated with the flow.

### Approach

Added a `{{request([source.]header|query.NAME)}}` placeholder resolved in `core.ResolvePlaceholder`, so it works in the executor's existing `url`, `headers`, and `body` fields with no new node configuration.

- `source` is `init` (the flow-initiation request, e.g. the OAuth2 `/authorize` call for redirect-based flows) or `flow` (the current flow-step request). Defaults to `init` when omitted.
- `type` is `header` (case-insensitive lookup) or `query` (case-sensitive lookup). First value wins for multi-valued entries. Unresolvable placeholders are kept as-is.

Key design decisions:

- **`init`** reuses the existing `InitiatorRequest` already captured at the OAuth `/authorize`, PAR, CIBA, and logout entry points, where credential headers are stripped via `FilterSensitiveHeaders`.
- **`flow`** is a new transient current-step request captured in the flow execution handler and threaded through the request `context.Context` (mirroring the existing inbound-SSO pattern), so no service interface or mock signatures change. The same `FilterSensitiveHeaders` filter is applied at capture, so `Authorization`/`Cookie`/`Set-Cookie`/`Proxy-Authorization` are never exposed.
- `{{ctx(...)}}` is resolved before `{{request(...)}}` so request-supplied text cannot be re-interpreted as a context placeholder.

### Related Issues
- Fixes thunder-id/thunderid#1354

### Related PRs
- N/A

### Checklist
- [x] Documentation provided. (`docs/content/guides/flows/advanced-configurations.mdx`)
- [x] Tests provided.
    - [x] Unit Tests (`core/utils_test.go`, `flowexec/handler_test.go`)
    - [x] Integration Tests (executor-level test in `http_request_executor_test.go`)
- [ ] Breaking changes. (Not applicable)

### Security checks
- [x] Followed secure coding standards. Credential-bearing headers are stripped at capture and are never exposable through the new syntax.

## Summary by CodeRabbit

- **New Features**
  - HTTP Request configurations can use `{{request(...)}}` placeholders to access headers and query parameters from the initiating or current request.
  - Supports explicit request sources, case-insensitive header matching, case-sensitive query matching, and first-value selection for multi-valued headers.
  - Unresolved or malformed placeholders remain unchanged.
  - Sensitive credential headers are excluded from request-based substitution.

- **Documentation**
  - Added guidance and examples covering request parameter syntax, supported sources, matching behavior, and unavailable credential headers.

### DIFF
--- a/backend/internal/flow/core/utils.go
+++ b/backend/internal/flow/core/utils.go
@@ -16,7 +16,19 @@ import (
 // TODO: Extend to support {{user(key)}}, {{env(key)}}, etc.
 var placeholderPattern = regexp.MustCompile(`{{\s*ctx\(\s*(\w+)\s*\)\s*}}`)
 
-// ResolvePlaceholder resolves a single placeholder string using the "{{ctx(key)}}" syntax.
+// requestPlaceholderPattern matches {{request(<selector>)}} with optional whitespace, capturing the
+// raw selector so it can be parsed into an optional source, a type (header/query) and a name.
+var requestPlaceholderPattern = regexp.MustCompile(`{{\s*request\(\s*([^)]*?)\s*\)\s*}}`)
+
+// Request selector source tokens used in {{request(...)}} placeholders. requestSourceInit refers
+// to the flow-initiation request (the default) and requestSourceFlow to the current flow step request.
+const (
+	requestSourceInit = "init"
+	requestSourceFlow = "flow"
+)
+
+// ResolvePlaceholder resolves a single placeholder string using the "{{ctx(key)}}" and
+// "{{request(...)}}" syntaxes.
 // If no placeholder is found, the original value is returned.
 // If a placeholder is found but the key doesn't exist in any data source, the placeholder is kept as-is.
 func ResolvePlaceholder(ctx *providers.NodeContext, value string, execResp *providers.ExecutorResponse,
@@ -27,7 +39,16 @@ func ResolvePlaceholder(ctx *providers.NodeContext, value string, execResp *prov
 
 	var contextUserRef *providers.EntityReference
 
-	return placeholderPattern.ReplaceAllStringFunc(value, func(match string) string {
+	// Resolve {{ctx(...)}} first, then {{request(...)}}. Because ReplaceAllStringFunc does not
+	// re-scan its own output, resolving request placeholders last keeps any {{ctx(...)}}-looking
+	// text that a request header or query value happens to contain from being resolved as context.
+	//
+	// Note the request pass does run over the ctx pass's output, so a {{ctx(...)}} value that itself
+	// contains the literal text {{request(...)}} would be resolved. This is intentional (it keeps the
+	// two passes simple) and safe: credential-bearing request data is stripped at capture — sensitive
+	// headers via FilterSensitiveHeaders and client-credential query params via the OAuth capture
+	// deny-list — so no secret is reachable through a request placeholder regardless of its origin.
+	value = placeholderPattern.ReplaceAllStringFunc(value, func(match string) string {
 		submatches := placeholderPattern.FindStringSubmatch(match)
 		if len(submatches) < 2 {
 			return match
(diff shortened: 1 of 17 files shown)


backend/internal/flow/flowexec/handler.go
```diff
@@ -60,6 +61,13 @@ func (h *flowExecutionHandler) HandleFlowExecutionRequest(w http.ResponseWriter,
 	// them available to the flow service, which selects the handle once the flow is known.
 	ctx := session.WithInbound(r.Context(), h.ssoTransport.Read(r))
 
+	// Carry the current step's request headers and query params so flow executors can publish them
+	// via {{request(flow.*)}} placeholders. Credential-bearing headers are stripped at this boundary.
+	ctx = withCurrentRequest(ctx, &providers.InitiatorRequest{
+		Headers:     sysutils.FilterSensitiveHeaders(r.Header),
```
ThaminduDilshan, line 67: `FilterSensitiveHeaders` only strips `Authorization`, `Cookie`, `Set-Cookie` and `Proxy-Authorization`. This request also carries `Flow-Secret` and `Attestation-Token`, so `{{request(flow.header.Flow-Secret)}}` would publish the flow secret to an arbitrary endpoint. Shall we add both to `sensitiveHeaders`?

backend/internal/flow/flowexec/handler.go
```diff
@@ -60,6 +61,13 @@ func (h *flowExecutionHandler) HandleFlowExecutionRequest(w http.ResponseWriter,
...
+		QueryParams: r.URL.Query(),
```
ThaminduDilshan, line 68: Query params aren't filtered here while the init path now strips the credential ones. Shall we apply the same deny-list?

backend/internal/oauth/oauth2/utils/oauthutils.go
```diff
@@ -19,6 +19,35 @@ import (
...
+var SensitiveQueryParams = map[string]bool{
+	constants.RequestParamClientSecret:        true,
+	constants.RequestParamClientAssertion:     true,
+	constants.RequestParamClientAssertionType: true,
+}
+
+// FilterSensitiveQueryParams returns a copy of params with SensitiveQueryParams removed. It is the
+// query-parameter analogue of system utils.FilterSensitiveHeaders. A nil or empty map is returned
+// unchanged.
+func FilterSensitiveQueryParams(params map[string][]string) map[string][]string {
```
ThaminduDilshan, line 37: Shall we move this to `system/utils` next to `FilterSensitiveHeaders`? The flow exec handler needs the same filter and shouldn't import oauth2 utils for it.

backend/internal/oauth/oauth2/utils/oauthutils.go
```diff
@@ -19,6 +19,35 @@ import (
 	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
 )
 
+// SensitiveQueryParams is the deny-list of OAuth request query parameters that must not be exposed
+// through a captured InitiatorRequest.QueryParams. They carry client credentials, so — like the
+// sensitive request headers stripped by system utils.FilterSensitiveHeaders — they are removed at
+// capture. That keeps them out of the plaintext flow-context store and out of any {{request(query.*)}}
+// placeholder a flow executor might publish. This is the authorize/CIBA counterpart of the PAR body
+// deny-list, and both strip the same parameters for the same reason.
+var SensitiveQueryParams = map[string]bool{
```
ThaminduDilshan, line 28: Do we need to expose this map? PAR is the only caller. Shall we keep it private and utilize the `FilterSensitiveQueryParams` method? If it cannot be used, we can expose a `IsSensitiveQueryParam` helper as well

backend/internal/flow/flowexec/service.go
```diff
@@ -287,6 +290,36 @@ func validateAdministrationCaller(ctx context.Context, flowType providers.FlowTy
 }
 
 // applyInboundSSO selects the SSO handle carried for this flow from the request-scoped
```
ThaminduDilshan, line 292: This block landed inside the `applyInboundSSO` doc comment and split it. Shall we move it above the comment?

backend/internal/flow/core/utils.go
```diff
@@ -77,6 +98,97 @@ func ResolvePlaceholder(ctx *providers.NodeContext, value string, execResp *prov
...
+			req = ctx.GetCurrentRequest()
+		}
+		if req == nil {
+			return match
+		}
+
+		var resolved string
+		var found bool
+		switch kind {
+		case "header":
```
ThaminduDilshan, line 137: Shall we define constants for `header` and `query` as well? We have them for the sources.

docs/content/guides/flows/advanced-configurations.mdx
```diff
@@ -1950,6 +1950,20 @@ The following fields are always available when the corresponding data exists in
...
+- `type` is `header` or `query`.
+- `name` is the header or query parameter name. Header lookups are case-insensitive; query parameter lookups are case-sensitive. When a header or query parameter has multiple values, the first is used.
+
+| Placeholder | Description |
+|---|---|
+| `{{request(header.User-Agent)}}` | The `User-Agent` header from the flow-initiation request. |
+| `{{request(init.query.utm_source)}}` | The `utm_source` query parameter from the flow-initiation request. |
+| `{{request(flow.header.User-Agent)}}` | The `User-Agent` header from the current flow step request. |
+
+Credential-bearing data is stripped at capture and is never available through this syntax: the headers `Authorization`, `Cookie`, `Set-Cookie`, and `Proxy-Authorization`, and the client-credential query parameters `client_secret`, `client_assertion`, and `client_assertion_type`. If a request, header, or query parameter is not present — or the selector is malformed (for example `{{request(header)}}` with no name, or `{{request(notaselector)}}`) — the placeholder is kept as-is.
```
ThaminduDilshan, line 1965: Let's not use em dashes here. Check the new code comments as well

backend/internal/flow/flowexec/model.go
```diff
@@ -41,6 +41,9 @@ type frame struct {
 type EngineContext struct {
 	Context          context.Context
 	initiatorRequest *providers.InitiatorRequest
+	// currentRequest carries the HTTP request driving the current flow step. It is transient:
+	// supplied on each execution from the request context and never persisted with the flow context.
+	currentRequest *providers.InitiatorRequest
```
ThaminduDilshan, line 46: `engineCtx.Context` is already the request context and it flows into `NodeContext.Context`, so `ResolvePlaceholder` can read the context key off it directly. Do we need these fields at all? SSO inputs are done that way already. The key would have to move out of `flowexec` since it imports `core`, `flow/session` looks like the right home.

tests/integration/flow/authentication/http_request_executor_test.go
```diff
@@ -168,6 +168,74 @@ var (
...
+				"type": "TASK_EXECUTION",
+				"properties": map[string]interface{}{
+					"url":     "http://localhost:9091/api/notifications",
+					"method":  "POST",
+					"headers": "{\"Content-Type\": \"application/json\"}",
+					// The flow-initiation request always carries Accept and Content-Type headers, so
+					// the flow-source lookups resolve; the remaining selectors are unresolvable and
+					// must be left as-is.
+					"body": "{" +
+						"\"flowAccept\": \"{{request(flow.header.Accept)}}\", " +
```
ThaminduDilshan, line 204: Do we need `flow.` prefix? If this is to differentiate `init.`, maybe we can have something like `current.`. @senthalan @brionmario WDYT of this pattern?

PR #5306, opened 2026-09-07
Preserve all hash parameters when storing and verifying credentials

### LINKED ISSUE #5303: ARGON2ID password hashing is completely broken — Memory/Parallelism dropped, all logins fail
### Description

Setting `crypto.password_hashing.algorithm` to `ARGON2ID` breaks password authentication completely: users can be created/bootstrapped successfully, but every subsequent login attempt with the correct password fails with a generic "Invalid credentials" (`AUTH-CRED-1002`) error. This affects normal users, the console/CLI admin bootstrap account, and agents alike, since all of them route through `entity.entityService`.

**Root cause**: `cryptolib.CredParameters` (`backend/internal/system/cryptolib/hash.go:38`) is copied field-by-field at storage and verification boundaries in `internal/entity/service.go`, and two of those copy sites drop the `Memory` and `Parallelism` fields that Argon2id requires:

- Write path — `hashPlaintextCredentials` (`backend/internal/entity/service.go:1098-1108`) persists only `Salt`, `Iterations`, `KeySize` from the `Generate()` output, discarding `Memory`/`Parallelism`.
- Read path — `verifyCredentials` (`backend/internal/entity/service.go:566-570`) reconstructs the reference credential with the same three fields only.

Since `argon2idHashProvider.Verify` (`backend/internal/system/cryptolib/hash.go:278-297`) requires `Memory` and `Parallelism` to be positive, the reconstructed reference credential (`Memory=0`, `Parallelism=0`) is rejected internally — but the resulting error is never logged (`backend/internal/entity/service.go:572-577` discards `verifyErr`), so it surfaces only as a generic authentication failure, indistinguishable from a wrong password.

The default algorithm is `PBKDF2` (unaffected), which is likely why this hasn't been reported: `ARGON2ID` has to be deliberately opted into via config to hit it.

A third site, `parseCredentialObject`'s pre-hashed-import branch (`backend/internal/user/declarative_resource.go:440-452`), has the same gap when importing already-hashed declarative credentials.

### Steps to Reproduce

Using a built distribution (or `make build`):

1. Unzip the distribution and edit `deployment.yaml`, changing only:
   ```yaml
   crypto:
     password_hashing:
       algorithm: "ARGON2ID"
   ```
2. Run `./setup.sh` — bootstrap succeeds and prints an admin username/password.
3. Run `./start.sh` to start the server.
4. Attempt to log in with the exact printed password:
   ```bash
   SECRET=$(cat config/secrets/direct_auth_secret)
   curl -sk -X POST https://localhost:8090/auth/credentials/authenticate \
     -H "Content-Type: application/json" \
     -H "Direct-Auth-Secret: $SECRET" \
     -d '{"identifiers": {"username": "admin"}, "credentials": {"password": "<printed password>"}}'
   ```

**Actual result**: `401` — `{"code":"AUTH-CRED-1002","message":{"defaultValue":"Invalid credentials"}}`, even though the password is correct and unmodified from the setup output.

**Expected result**: `200` with a signed assertion, as happens under the default `PBKDF2` algorithm (verified as a control against the same endpoint/flow on a separate port, same steps, `algorithm: "PBKDF2"` unchanged — returns `200` with a valid JWT).

A minimal Go-level repro (in `internal/entity`, package-internal, not committed) confirms the same mechanism directly: `hashPlaintextCredentials` generates a credential with real `Memory`/`Parallelism`, but the persisted JSON shows `"Parallelism":0,"Memory":0`, and feeding that back through `verifyCredentials` with the correct plaintext returns `authentication failed`.

### Version

v1.0.0 (commit `d4544c47d`)

- macOS (arm64), SQLite storage backend
- Reproduced via the packaged zip distribution (`thunderid-1.0.0-macos-arm64.zip`) and confirmed against source in `backend/internal/entity/service.go` on the `main` branch at the commit above.

### Suggested fix direction

Round-trip `CredParameters` as a whole at the write/read/import boundaries instead of copying named fields individually (so adding future algorithms doesn't reintroduce this class of bug), and log `verifyErr` from `hashService.Verify` at least at debug level so a config/parameter mismatch doesn't look identical to a wrong password in the logs.

rajithacharith:
### Purpose
`crypto.password_hashing.algorithm: ARGON2ID` broke authentication completely: users could be created and bootstrapped, but every login afterward failed with a generic "Invalid credentials", even with the exact correct password. Root cause was in `internal/entity/service.go`: `hashPlaintextCredentials` (write) and `verifyCredentials` (read) both reconstructed `cryptolib.CredParameters` by naming only `Salt`, `Iterations`, and `KeySize`, silently dropping `Memory` and `Parallelism`. Since Argon2id's provider requires both to be positive, the reconstructed reference credential was always rejected internally, and the resulting error was swallowed, so it looked identical to a wrong password. `PBKDF2` (the shipped default) was unaffected, which is likely why this went unnoticed.

Filed as #5303 with a full repro (code-level test and an end-to-end run against a built distribution).

### Approach
- `verifyCredentials` and `hashPlaintextCredentials` now assign `CredParameters` wholesale instead of naming individual fields, since `StoredCredential.StorageAlgoParams` is already typed as `cryptolib.CredParameters`. This is a copy-boundary fix, not a schema change.
- `verifyCredentials` now takes `ctx` and logs the internal `Verify` error at debug level, so a parameter/config mismatch is distinguishable from a wrong password going forward.
- `parseCredentialObject`'s pre-hashed-credential-import branch (declarative user resources) had the same gap for `memory`/`parallelism`; fixed to match.
- Added `hash_roundtrip_test.go`, a regression test that hashes and verifies a credential end-to-end for `SHA256`, `PBKDF2`, and `ARGON2ID`, covering both correct- and wrong-password cases. This would have caught #5303 before it shipped.

### Related Issues
- Fixes #5303

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved credential authentication reliability when checking multiple stored credentials.
  - Preserved complete hashing parameters, including memory and parallelism settings, for pre-hashed credentials.
  - Ensured authentication continues correctly when an individual stored credential cannot be verified.
- **Tests**
  - Added coverage for SHA-256, PBKDF2, and Argon2id credentials across persistence and authentication flows.
  - Added validation that incorrect passwords are consistently rejected.

### DIFF
--- a/backend/internal/entity/service.go
+++ b/backend/internal/entity/service.go
@@ -498,7 +498,7 @@ func (s *entityService) AuthenticateEntityByID(
 		return nil, ErrEntityNotFound
 	}
 
-	if err := s.verifyCredentials(credentials, result.SchemaCredentials, result.SystemCredentials); err != nil {
+	if err := s.verifyCredentials(ctx, credentials, result.SchemaCredentials, result.SystemCredentials); err != nil {
 		return nil, err
 	}
 
@@ -511,7 +511,7 @@ func (s *entityService) AuthenticateEntityByID(
 }
 
 // verifyCredentials verifies provided credentials from both schema and system credentials.
-func (s *entityService) verifyCredentials(credentials map[string]interface{},
+func (s *entityService) verifyCredentials(ctx context.Context, credentials map[string]interface{},
 	schemaCredsJSON, systemCredsJSON json.RawMessage) error {
 	// Merge both credential columns for verification.
 	storedCreds := make(map[string][]StoredCredential)
@@ -561,16 +561,17 @@ func (s *entityService) verifyCredentials(credentials map[string]interface{},
 		verified := false
 		for _, stored := range credList {
 			ref := cryptolib.Credential{
-				Algorithm: stored.StorageAlgo,
-				Hash:      stored.Value,
-				Parameters: cryptolib.CredParameters{
-					Salt:       stored.StorageAlgoParams.Salt,
-					Iterations: stored.StorageAlgoParams.Iterations,
-					KeySize:    stored.StorageAlgoParams.KeySize,
-				},
+				Algorithm:  stored.StorageAlgo,
+				Hash:       stored.Value,
+				Parameters: stored.StorageAlgoParams,
 			}
 			ok, verifyErr := s.hashService.Verify([]byte(credValue), ref)
-			if verifyErr == nil && ok {
+			if verifyErr != nil {
+				s.logger.Debug(ctx, "Credential verification error", log.String("credentialType", credType),
+					log.Any("error", verifyErr))
+				continue
+			}
+			if ok {
 				verified = true
 				break
 			}
@@ -1097,13 +1098,9 @@ func (s *entityService) hashPlaintextCredentials(creds json.RawMessage) (json.Ra
 			}
 			result[credType] = []StoredCredential{
 				{
-					StorageAlgo: credHash.Algorithm,
-					StorageAlgoParams: cryptolib.CredParameters{
-						Salt:       credHash.Parameters.Salt,
-						Iterations: credHash.Parameters.Iterations,
-						KeySize:    credHash.Parameters.KeySize,
-					},
-					Value: credHash.Hash,
+					StorageAlgo:       credHash.Algorithm,
+					StorageAlgoParams: credHash.Parameters,
+					Value:             credHash.Hash,
 				},
 			}
 		default:
--- a/backend/internal/entity/hash_roundtrip_test.go
+++ b/backend/internal/entity/hash_roundtrip_test.go
@@ -1,0 +1,57 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package entity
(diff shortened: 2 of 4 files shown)


hwupathum (approved):

backend/internal/entity/service.go
```diff
@@ -561,16 +561,17 @@ func (s *entityService) verifyCredentials(credentials map[string]interface{},
...
-					KeySize:    stored.StorageAlgoParams.KeySize,
-				},
+				Algorithm:  stored.StorageAlgo,
+				Hash:       stored.Value,
+				Parameters: stored.StorageAlgoParams,
 			}
 			ok, verifyErr := s.hashService.Verify([]byte(credValue), ref)
-			if verifyErr == nil && ok {
+			if verifyErr != nil {
+				s.logger.Debug(ctx, "Credential verification error", log.String("credentialType", credType),
```
thiva-k, line 570: Shouldn't this be `logger.Error(...)`?

PR #5309, opened 2026-09-07
Register a data plane, declaratively or through the API, and apply to it

bhagyasakalanka:
## Purpose
A control plane has no way to record the data plane it administers, and no way to send its configuration there. Doing it by hand means knowing the address and credentials out of band, which cannot be automated and cannot be shown in a console.

## Goals
Register the data plane a control plane administers, through the API or by declaring it in a file, hold its credential safely, bound how many it may hold, and apply the configuration the control plane currently holds.

## Approach
A new `internal/gateway` package with the usual shape: model, store, service, handler, init, a declarative resource, and a small client for the data plane.

- `GATEWAY` table in the config database, keyed by deployment like every other configuration resource, with the name unique per deployment.
- `GET|POST /gateways`, `GET|DELETE /gateways/{id}`, `POST /gateways/{id}/apply`.
- A gateway can also be declared under `gateways/`, carrying the same fields the API takes.
- An apply exports every resource type, joins the files into the document the import API takes, and posts it to the data plane's `/import` with a token obtained by client credentials from the machine to machine client the registration recorded.

The two planes talk over ordinary APIs. Neither needs an inbound path to the other, and there is no websocket in this path.

Choices worth reviewing:

- **The bound is configuration, not a constant.** `server.max_gateways` defaults to one, which is what a standalone deployment pairs with. Zero means the default rather than none, so an unset value cannot leave a deployment unable to register the gateway it is expected to have. The declarative path is bounded by the same value.
- **A declared gateway is matched by name.** A file cannot know an id this server generated, and these files are read on every start, so re-reading updates the gateway it already described rather than registering another. An edited file moves the gateway it describes.
- **No version history.** What an apply sends is what the control plane holds now. Captured versions, promotion and rollback stay outside the product.
- **Unresolved placeholders are reported, not hidden.** The export fills a value for every placeholder it can resolve and leaves the rest empty, which is what happens to a credential it only holds a hash of. Those names come back in the apply response, so an operator learns which credentials the data plane has to supply rather than discovering it when a login fails.
- **The client secret is encrypted at rest** through the same `cmodels.Property` path every other stored credential uses, whether it arrived through the API or a file, and is never returned by any endpoint.
- **A gateway failure is reported without its body.** A token endpoint or import error can echo the request, so the detail is logged and the caller gets a generic error.

A declared secret sits in the file as written. Nothing here can make a readable file safe, and the comment on the type says so rather than implying otherwise.

## User stories
As an administrator I can register the data plane this control plane administers, either through the API or by declaring it alongside my other resources, and apply the current configuration to it.

## Release note
A control plane can register the data plane it administers, through the API or declaratively, bounded by `server.max_gateways`, and apply its current configuration to it.

## Documentation
`server.max_gateways` is documented in `docs/content/deployment/configuration.mdx`.

## Training
N/A

## Tests
Service tests cover registration, the bound, the credential stored encrypted and recoverable, the credential never returned, refusal of an unreachable or unnamed gateway, unknown ids, deletion and a store failure. Apply tests cover the configuration and decrypted credential reaching the data plane, every resource type being exported, unresolved placeholders reported in order, an unknown gateway, nothing to apply, and a gateway failure not carrying its body into the response. Declarative tests cover parsing, refusal of an incomplete declaration, the same file read three times leaving one gateway, an edited file updating it, the bound applying, and the declared secret being encrypted.

`make lint_backend` reports 0 issues, the full unit suite passes, and the full integration suite passes against a fresh build: 50 suites, 0 failures.

## Not in this PR
Any console screen.

## Summary by CodeRabbit

- **New Features**
  - Added gateway management for registering, viewing, and removing administered data planes.
  - Added configuration deployment to registered data planes with import results and unresolved-variable reporting.
  - Added support for declarative gateway configurations.
  - Gateway credentials are securely stored and excluded from API responses.
  - Added configurable limits for administered gateways.
  - Environment variables can now provide values for unresolved import templates.

- **Configuration**
  - Added the `server.max_gateways` setting, defaulting to one gateway.

- **Documentation**
  - Documented the new gateway limit configuration.

- **Bug Fixes**
  - Added validation and localized error messages for invalid gateway requests and conflicts.


bhagyasakalanka:
Superseded by #5446, which registers the data planes a control plane administers using the management token from #5445 rather than an M2M application.

PR #5312, opened 2026-09-07
Add agent-onbaording-flow design spec

Dilusha-Madushan:
### Purpose

This PR adds the design specification  for making agent onboarding a configurable administrator flow in ThunderID.

Today, agent onboarding is a fixed sequence embedded in the Admin Console. The console decides what information is collected, the order of the steps, and how the resulting credentials are presented. Because the sequence lives in the console, organizations cannot adapt onboarding to their own security, governance, or operational requirements, and any change to the sequence needs a product change and a console release.

The specification proposes executing agent onboarding through a flow definition resolved at runtime, matching the model already used for user onboarding and user deletion. A default Agent Onboarding flow ships so the feature works out of the box, and organizations customize it through the existing flow management capabilities.

### Approach

The spec is written against the standard design spec template and settles on these key decisions:

- Reuse the existing Administrator Flow model rather than introducing a new flow
  type. The console resolves a configurable flow handle, loads the matching flow
  definition from the flow store, and executes it through the flow engine. The
  handle is configurable so the default flow can be replaced without a console
  change.

- Ship a default Agent Onboarding flow (permission check, agent type selection,
  organization unit, owner resolution, attribute collection, provisioning,
  credential presentation) as a customizable starting point.

- Delegate provisioning. A generalized provisioning step resolves a provisioning
  mode and delegates creation to a provider. A new agent provider handles the agent
  path and calls the existing agent management service.

- Keep creation rules in the agent service. Validation, the agent record, credential
  generation, and inbound client configuration stay in the existing service. The
  flow orchestrates the experience and does not duplicate these rules.

- Present credentials as a flow step rather than a hardcoded console completion
  screen, carrying the generated credentials as flow runtime data.

- Add new executor, one to resolve the owner.

## Summary by CodeRabbit

* **Documentation**
  * Added a specification for configurable agent onboarding flows.
  * Documented runtime flow resolution and execution, including ownership resolution, provisioning, credential presentation, validation, and failure handling.
  * Defined the default Agent Onboarding flow, its configurable handle, default value, validation rules, and deployment-level scope.
  * Added requirements and acceptance criteria covering onboarding flow configuration, execution, validation outcomes, and credential presentation.


docs-internals/agent-onbaording-flow/spec.md
```diff
@@ -0,0 +1,207 @@
...
+
+This feature moves agent creation onto the flow engine. An administrator runs an administration flow that resolves the agent type and owner, collects the agent's details from its entity type schema, provisions it, and shows its credentials once. The console renders whatever the flow returns and holds no knowledge of what an agent needs.
+
+The governing decision is that **the engine becomes category-generic rather than gaining a parallel agent path**. Provisioning, uniqueness validation and type resolution already existed for users; each is taught the entity category instead of being duplicated. Where a rule genuinely differs, the difference is declared in one lookup table rather than scattered through conditionals.
+
+In scope: the engine's agent support, the flow builder resources needed to compose such a flow, the shipped flow, and the console page that runs it. Out of scope: agent self-registration, which is refused for now, and the agent edit experience, which is unchanged.
+
+## Architecture
```
sahandilshan, line 16: @Dilusha-Madushan 
Please add a detailed diagram with proper explanation about how the architecture got changed after this. We decided multiple things here right. We have devided the entity provider into two providers
https://github.com/thunder-id/thunderid/discussions/5229

Why any of these info have not mentioned here?

Dilusha-Madushan: Yes, that part is missing. I initially considered using the current codebase as a starting point and writing the spec for this specific feature only. However, I agree with your suggestion that provider architecture changes need to be tracked in this spec, as concerns were raised about that issue with this feature.

docs-internals/agent-onbaording-flow/spec.md
```diff
@@ -0,0 +1,463 @@
...
+| `AgentMgtProvider` | Expose agent creation to the runtime, translate flow-level delegation intent into an inbound configuration, and delegate to the agent management service. |
+| User management service | Apply user creation rules and coordinate entity creation. |
+| Agent management service | Apply agent creation rules and coordinate the entity, inbound client, and authentication profile. |
+| Inbound client service | Validate and manage inbound authentication settings and their associated resources. |
+| Entity service | Manage the common entity representation used internally by the management services. |
+
+For example, a flow that provisions an agent will call `AgentMgtProvider.CreateAgent`. The provider will call the agent service, which will use the entity and inbound services. The executor will not create an inbound client or generate credentials itself.
+
+Management services will remain the authority for business validation. Flow validation will serve a different purpose: resolving trusted selections and collecting enough information to submit a meaningful creation request. A successful flow precheck will never replace service validation.
+
```
sahandilshan, line 49: Let's add the diagram of the proposed solution in the https://github.com/thunder-id/thunderid/discussions/5229 discussion on top of the explanation. It will give the better understanding

sahandilshan (approved):

PR #5316, opened 2026-09-07
Add the authorization support for federated entities specification

thiva-k:
## Summary
- Adds `docs-internals/federated-entity-authorization/spec.md`, the design specification for mapping external federated claims to local roles, groups, and permissions.
- Covers both mapping mechanisms (value-based rules and direct name matching), how they combine, and how they're evaluated at federated login, RFC 8693 token exchange, and ID-JAG assertion issuance/consumption.
- Based on [Discussion #5197](https://github.com/thunder-id/thunderid/discussions/5197) and Issue #5192.

## Test plan
- [ ] Structural review against `docs-internals/templates/spec.md`
- [ ] Confirm every requirement's acceptance criteria is covered by the detailed design

## Summary by CodeRabbit

* **Documentation**
  * Added the finalized specification for federated entity authorization.
  * Documented mappings from external claims to local roles, groups, and permissions.
  * Defined claim processing, authorization evaluation, and provisioning behavior.
  * Described federated login, token exchange, ID-JAG, and token-renewal flows.
  * Added API, user interface, acceptance criteria, and external policy decision point requirements.
  * Documented token-renewal handling for carried role and group identifiers.


docs-internals/federated-entity-authorization/spec.md
```diff
@@ -0,0 +1,332 @@
...
+scopes, unfiltered by mapping at that point; mapping is what determines the granted access at
+consumption, when the assertion is presented on the jwt-bearer grant and its own claims are resolved
+fresh against the connection's mapping.
+
+A renewed token has no external claims to re-read: mapped role/group identifiers are carried on the
+renewal credential and re-resolved against the RBAC engine, so a change to a role's own permissions is
+picked up, while a change to the entity's memberships at the provider only applies at the next federated
+login.
+
+### API
```
sahandilshan, line 212: Shall we add a sample API request here

docs-internals/federated-entity-authorization/spec.md
```diff
@@ -0,0 +1,332 @@
...
+
+A connection's attribute configuration gains an optional authorization mapping section holding the
+rule-based and direct mapping lists described above, alongside the connection's existing attribute
+mappings. The full schema, field-level requirements, and validation error catalog are defined in the
+connections API spec. A rule-based mapping's role, group, and permission targets are validated to exist
+at save time; a direct mapping has no such targets to validate, since its role, group, and permission
+names only exist as claim values resolved at request time, but a permission-target direct mapping's
+resource server is validated to exist at save time. The connection is rejected otherwise.
+
+### UI
```
sahandilshan, line 249: Let's add the UI mockups as well

docs-internals/federated-entity-authorization/spec.md
thiva-k: Added a sample request body combining a direct mapping and a rule-based mapping.

docs-internals/federated-entity-authorization/spec.md
thiva-k: Added two screenshots of the current UI (direct/default mode and Advanced Rules mode), and rewrote the surrounding text to match the current per-row design (it had gone stale after a UI redesign).

sahandilshan:
Not a blocker for this spec, just recording a broader concern.

Mapping a claim value directly to a permission means whoever configures the mapping can grant
privileges they do not already hold themselves. This is not introduced by this spec: the same
path already exists through role and group assignment. It is currently masked because most
management operations run on the system scope, so anyone who can configure mappings is already
fully privileged.

It becomes a real problem once we introduce personas like admin user or application developer.
At that point, anyone who can manage roles, groups, permissions, or claim mappings effectively
has a route to admin privileges. That needs one common answer at the authorization model level
rather than per feature, so let's keep it in mind as we build out the persona model.

I don't see any issue with the spec other than the above concern. +1 from me

thiva-k:
> Not a blocker for this spec, just recording a broader concern.
> 
> Mapping a claim value directly to a permission means whoever configures the mapping can grant privileges they do not already hold themselves. This is not introduced by this spec: the same path already exists through role and group assignment. It is currently masked because most management operations run on the system scope, so anyone who can configure mappings is already fully privileged.
> 
> It becomes a real problem once we introduce personas like admin user or application developer. At that point, anyone who can manage roles, groups, permissions, or claim mappings effectively has a route to admin privileges. That needs one common answer at the authorization model level rather than per feature, so let's keep it in mind as we build out the persona model.
> 
> I don't see any issue with the spec other than the above concern. +1 from me

This is tracked with https://github.com/thunder-id/thunderid/issues/5399

docs-internals/federated-entity-authorization/spec.md
```diff
@@ -0,0 +1,439 @@
...
+- **Status:** Final
+- **Version:** 1.0
+- **Related documents:** [Discussion #5197: Authorization support for federated entities](https://github.com/thunder-id/thunderid/discussions/5197), Issue #5192
+
+## Summary
+
+Organizations manage their users in an external identity provider while ThunderID manages agents,
+applications, and the permission model. Without a way to translate what the provider asserts into local
+authorization, every federated entity either goes unauthorized or has to be assigned roles and groups by
+hand, duplicating policy the organization already maintains elsewhere.
```
senthalan:
```suggestion
Organizations manage their users in an external identity provider while ThunderID manages agents, applications, and the permission model. Without a way to translate what the provider asserts into local authorization, every federated entity either goes unauthorized or has to be assigned roles and groups by hand, duplicating policy the organization already maintains elsewhere.
```

can we make the content in one line without manual break. The editor or visual tools will visualize accordingly. Let's make sure it is followed in other places too

thiva-k: Fixed

sahandilshan (approved):

Issue #5150, opened 2026-08-24
Support OAuth 2.0 Dynamic Client Registration Management (DCRM)


Thumimku:
### What problem are we solving?

ThunderID supports Dynamic Client Registration (DCR), allowing OAuth 2.0 / OpenID Connect clients to register dynamically with the authorization server. However, once a client has been dynamically registered, there is currently no standards-based mechanism for that client to manage its registration.

The OAuth 2.0 Dynamic Client Registration Management Protocol ([RFC 7592](https://www.rfc-editor.org/rfc/rfc7592)) extends Dynamic Client Registration by defining a Client Configuration Endpoint through which a dynamically registered client can:

- Retrieve its current registration information.
- Update its registered client metadata.
- Delete/unregister the client.

Without DCRM support, dynamically registered clients need to rely on administrative APIs, manual configuration, or re-registration when their metadata changes. This breaks the self-service lifecycle expected from clients using Dynamic Client Registration.

DCRM also defines a `registration_access_token`, bound to the dynamically registered client, which allows the client to authenticate when managing its registration without requiring its normal OAuth client credentials.

### Who are we solving this for?

**Primary: Developers integrating applications using Dynamic Client Registration**

- **Goal:** Dynamically register and subsequently manage the complete lifecycle of their OAuth/OIDC client without requiring administrative intervention.
- **Pain:** After registering a client, metadata such as redirect URIs or other client configuration cannot be managed through the standardized DCR protocol.

**Secondary: Platforms and automated clients integrating with ThunderID**

- **Goal:** Provision and manage OAuth clients programmatically using interoperable OAuth standards.
- **Pain:** Client provisioning can be automated through DCR, but subsequent read, update, and deletion operations require a different management mechanism.

### Why should we solve this now?

Supporting DCRM completes the lifecycle started by Dynamic Client Registration. Applications that dynamically create their OAuth clients can also retrieve, modify, and remove those registrations using a standardized protocol.

This is particularly useful for automated clients and platforms where client configuration may change over time, for example when redirect URIs or other registered metadata need to be updated.

Supporting RFC 7592 improves ThunderID's OAuth/OIDC standards coverage and interoperability.

DCR and DCRM together provide a complete standards-based client provisioning lifecycle:

```text
Dynamic Client Registration (RFC 7591)
              |
              v
       Register Client
              |
              v
 registration_access_token
 registration_client_uri
              |
              v
Dynamic Client Registration Management (RFC 7592)
              |
[...]

PR #5321, opened 2026-09-08
Add design docs for dynamic client registration management

Thumimku:
## Purpose

Adds the specification and threat model for OAuth 2.0 Dynamic Client Registration Management
(RFC 7592), ahead of the implementation.

Per [docs-internals/readme.md](../blob/main/docs-internals/readme.md), design docs are reviewed on
their own, separate from the implementation. The implementation for #5150 follows as its own PR.

Refs #5150

## What's included

| File | Contents |
|---|---|
| `spec.md` | Summary, architecture, detailed design, requirements, acceptance criteria |
| `threat-model.md` | Trust boundaries, actors, interactions, threats across STRIDE-LM, mitigations |
| `assets/architecture.md` | Component diagram and ownership boundaries |
| `assets/lifecycle.md` | Registration lifecycle from registration through management to deletion |
| `assets/authorization-flow.md` | Authorization decision flow and why the checks run in that order |

## Approach

The feature extends the existing dynamic client registration component rather than introducing a new
one. The registration endpoint and the client configuration endpoint share the same service, the same
client metadata model, and the same error vocabulary, so a registration response and a client
information response are the same shape. The endpoint owns no storage: the application component
remains the single owner of client state.

The client configuration endpoint is addressed per client under the existing registration path:

```
/oauth2/dcr/register/{client_id}
```

| Operation | Method | Success | Body |
|---|---|---|---|
| Read registration | `GET` | `200` | Current client metadata |
| Update registration | `PUT` | `200` | Updated client metadata |
| Delete registration | `DELETE` | `204` | None |

## Design decisions recorded

- **Management is administrative.** Every operation is authorized by the system permission, the same
  credential that already governs the registration endpoint, so one operator credential governs every
  dynamically registered client.
- **No per-client management credential is issued.** Registration returns neither a
  `registration_access_token` nor a `registration_client_uri`, and a registered client cannot manage
  itself. This is the shape an application portal needs: it provisions many clients and manages all of
  them on behalf of their owners, and a per-client credential would oblige it to store and rotate one
  secret per client to perform what a single secret already authorizes, making its own store of those
  secrets the thing worth attacking.
- **Authorization precedes client resolution.** An unauthorized request fails identically whether or
  not the client exists, so probing the endpoint cannot enumerate client identifiers.
- **There is no `403` path.** A caller either holds the permission that governs every registration or
  holds nothing, so there is no state in which a caller is authenticated for one registration and
  refused another. Failure is `401` with an RFC 6750 `WWW-Authenticate` challenge.
- **Full-replace update semantics** per the RFC. Client identifier, client secret, owning organization
  unit and application type are carried forward explicitly, because losing them would break the
  registration rather than change it.
- **Client self-management is deferred, not abandoned.** A later phase may reintroduce a per-client
  credential alongside the administrative path rather than in place of it, since the two serve
  different consumers. The reasoning is recorded in the spec so the question is settled rather than
  re-derived.

## Deviations from RFC 7592

Stated in the spec with reasoning, and the main thing worth reviewer attention:

1. **Registration does not return the management fields.** RFC 7592 §3 marks
   `registration_access_token` and `registration_client_uri` as required members of a client
   information response. This phase issues no per-client token, so neither field is present and a
   registrant has no self-service management route. The configuration endpoint is nonetheless
   reachable at the URI `registration_client_uri` would have named, to an administrative caller.
   Closing this deviation is the deferred scope above.
2. **`client_secret` is omitted from read and update responses.** ThunderID stores the client secret
   write-only: it is set once at registration and there is no read path anywhere in the platform, so
   the secret is omitted rather than fabricated. Adding a credential read path was considered and
   rejected, because it would open credential reads across the whole platform to serve one endpoint.
3. **A submitted `client_secret` on update is ignored rather than verified.** §2.2 says a
   `client_secret` included in an update must match the currently issued secret. That cannot be
   compared without a read path, so the field is ignored instead. The same section forbids a client
   choosing its own secret, which ignoring the field satisfies by construction.

## Residual risk worth reviewer attention

Scoping management to administrative callers concentrates risk: one administrative access token
governs every dynamically registered client, with nothing scoping the damage to a single
registration. This is the defining trade-off of the phase and is accepted rather than eliminated,
recorded as threat 8 in the threat model with its bounds. The alternative, a durable per-client
credential, would have scoped a single leak to a single registration but would have put a long-lived
credential in the hands of every registrant, and the leak of any one of them could not be revoked
individually.

## Note for reviewers

An earlier revision of this specification justified a registration access token by asserting that the
OpenID Connect conformance suite requires one. **That claim was checked against the suite's source and
is false:**

- `UnregisterDynamicallyRegisteredClient` logs and returns cleanly when no registration access token
  is present. It does not throw.
- Every OIDC call site wraps it as
  `.skipIfObjectsMissing("client").onSkip(INFO).onFail(WARNING).dontStopOnFailure()`, inside
  `cleanup()`.
- No OIDC test issues a `PUT` to the client configuration endpoint at all.

So no OIDC plan, Basic or Dynamic, requires the token. The consequence of its absence is orphaned test
clients left behind after a run, which is a cleanup concern for the deployment being certified, not a
conformance failure. Anything that reopens this decision should reopen it on the strength of a
consumer that needs self-management, not on a conformance requirement that does not exist.

No code in this PR. Diagrams are Mermaid and render in the GitHub UI.

## Related Issues
- Refs #5150

## Related PRs
- N/A

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive documentation for dynamic client registration management.
  - Documented administrative management of client registrations, including reading, replacing metadata, and deleting registrations.
  - Clarified authorization flows, permission requirements, response outcomes, and protection against client enumeration.
  - Documented registration lifecycle behavior, including client usability after registration and 404 responses after deletion.
  - Added architecture and threat-model documentation covering credential handling, identity validation, and security considerations.


docs-internals/dynamic-client-registration-management/assets/architecture.md
```diff
@@ -0,0 +1,46 @@
+# Component architecture
+
+How the client configuration endpoint relates to the components it reuses. The dynamic client
+registration component owns the protocol surface only; client state and credentials stay with their
+existing owners.
```
senthalan:
```suggestion
How the client configuration endpoint relates to the components it reuses. The dynamic client registration component owns the protocol surface only; client state and credentials stay with their existing owners.
```

can we make the content in one line without manual break. The editor or visual tools will visualize accordingly. Let's make sure it is followed in other places too

docs-internals/dynamic-client-registration-management/spec.md
```diff
@@ -0,0 +1,193 @@
...
+
+| Property | Why it is carried forward |
+|---|---|
+| Client identifier | The registration must keep its identity. Allowing it to lapse would issue a new identifier, orphan the client's own configuration URI, and drop any registered key material. |
+| Client secret | A metadata change must not silently invalidate the client's credentials. The update deliberately supplies no secret, which is what preserves the stored one. |
+| Owning organization unit and application type | Neither is expressible through RFC 7591 metadata, and the application type is immutable, so both are inherited rather than reasserted. |
+
+The client name is also inherited when the request omits it, because a name is required on the underlying record and an omitted name would otherwise fail validation rather than clear the field. A request is treated as omitting the name only when it carries neither `client_name` nor a localized variant of it, so supplying only a localized name still names the client. The update response reports the name that was actually applied, which is the inherited one in that case.
```
senthalan: Any specific reason to not to allow updating the client name?

Here we talk about the localized variant - how this is going to work with our existing translation capability,

docs-internals/dynamic-client-registration-management/spec.md
```diff
@@ -0,0 +1,193 @@
...
+
+**The conformance claim was checked and is false.** The registration access token is sometimes justified by asserting that the OpenID Connect conformance suite requires it. Reading the suite's source shows otherwise:
+
+- `UnregisterDynamicallyRegisteredClient` logs and returns cleanly when no registration access token is present. It does not throw.
+- Every OIDC call site wraps it as `.skipIfObjectsMissing("client").onSkip(INFO).onFail(WARNING).dontStopOnFailure()`, inside `cleanup()`. `AbstractOIDCCServerTest` and `AbstractOIDCCDynamicRegistrationTest` do this identically.
+- No OIDC test issues a `PUT` to the client configuration endpoint at all.
+
+So no OIDC plan, Basic or Dynamic, requires the token. The consequence of its absence is orphaned test clients left behind after a run, which is a cleanup concern for the deployment being certified, not a conformance failure. Anything that reopens this decision should reopen it on the strength of a consumer that needs self-management, not on a conformance requirement that does not exist.
```
senthalan: Do we need this level of details about the conformance suite in the spec.md

docs-internals/dynamic-client-registration-management/spec.md
```diff
@@ -0,0 +1,193 @@
...
+
+**Registration does not return the management fields.** RFC 7592 section 3 marks `registration_access_token` and `registration_client_uri` as required members of a client information response. Phase 1 issues no per-client token, so neither field is present in a registration response or in a client information response, and a registrant has no self-service management route. The configuration endpoint is nonetheless reachable at the URI that `registration_client_uri` would have named, to an administrative caller. Closing this deviation is the deferred scope described above.
+
+**A submitted client secret is ignored rather than verified.** RFC 7592 section 2.2 says that a `client_secret` included in an update must match the currently issued secret. ThunderID cannot compare it, because the stored secret has no read path, so the field is ignored instead. The same section forbids a client choosing its own secret, which ignoring the field satisfies by construction: there is no path from the submitted value to storage. Failing this way is the safe direction, since it never rotates a credential and never rejects a caller that correctly echoed the client's secret.
```
senthalan: Could you explain why  `ThunderID cannot compare it` ?

docs-internals/dynamic-client-registration-management/spec.md
```diff
@@ -0,0 +1,193 @@
...
+
+**The client secret is not returned on read or update.** RFC 7592's example client information response includes `client_secret`. ThunderID stores the client secret write-only: it is set once at registration and there is no read path anywhere in the platform. The secret is therefore omitted rather than fabricated. A client that loses its secret has no recovery route through this endpoint and must be registered again. Adding a credential read path was considered and rejected, because it would open credential reads across the whole platform to serve one endpoint.
```
senthalan: `ThunderID stores the client secret write-only` - I think this statement is confusing a bit. We store the secret hashed, so we can read and compare. Having this line is also confusing `there is no read path anywhere`

docs-internals/dynamic-client-registration-management/spec.md
Thumimku: no need removed and polished the spec md

docs-internals/dynamic-client-registration-management/spec.md
Thumimku: It can compare it so i updated the spec, check acceptance criteria 5 and 6

```
- **AC2.5:** Given a registered client, when an update body carries a `client_secret` that does not match the currently issued secret, then the request is rejected with `400` and the registration is left unchanged.
- **AC2.6:** Given a registered client, when an update body carries a `client_secret` that matches the currently issued secret, then the update is applied and the stored secret is unchanged.
- ```

File: docs-internals/dynamic-client-registration-management/spec.md (hunk 5 above)
Comment (Thumimku) [reply docs-internals/dynamic-client-registration-management/spec.md]: true, we can compare when they send the client secret but rotation is not part of the feature.

File: docs-internals/dynamic-client-registration-management/spec.md (hunk 2 above)
Comment (Thumimku) [reply docs-internals/dynamic-client-registration-management/spec.md]: 1. Any specific reason to not to allow updating the client name?

client_name can be updated normally. If the request includes a new client_name, it replaces the existing one.
The only special case is when the request contains neither client_name nor a localized version of it. Since the application requires a name, the existing name is kept instead of causing validation to fail. 

2. Here we talk about the localized variant - how this is going to work with our existing translation capability?

Reuse the existing translation capability rather than introducing storage of their own. A variant is a top-level `<field>#<bcp47-tag>` key alongside the base field. Sample request given below.

```json
{
  "client_id": "abc123",
  "client_name": "My App",
  "client_name#fr": "Mon Application",
  "client_name#ja": "マイアプリ",
  "redirect_uris": ["https://app.example.com/callback"]
}
```
this will update client name and itf fr and ja variants.

### PUT behaviour
- A client holding fr, ja and de that receives an update sending only client_name#fr keeps fr and loses the other two. Its general practise that do GET and change the required once before perform PUT.
- Sending a translation does not change the field's own value. A client named "My App" that receives client_name#fr only keeps "My App". A read returns each field with its translations, so a caller can see what an update would replace and send back what it read.

PR #5328, opened 2026-09-08
Improve duplicate user detection and handling for unique attributes in provisioning flow

### LINKED ISSUE #5300: Unable to create the same user even none of the attributes required to be unique
### Description

Unable to create the same user even none of the attributes required to be unique.

https://github.com/user-attachments/assets/c7d37a53-4319-4adf-9036-2c984fc1779f

### Steps to Reproduce

1. Login to console
2. Create a user type with non of the attributes being unique
3. Create two sub OU attach to the same user type
4. Create a user in one of the OUs
5. Try to create the same user in other OU

### Version

v1.0.0

_No response_

NipuniBhagya:
### Purpose

This pull request refines the user provisioning logic to strictly follow schema-defined uniqueness constraints when detecting duplicate users. Now, only attributes explicitly marked as unique in the schema are checked for duplicates, ensuring that users can be provisioned even if their non-unique attributes match existing users. The test suite is expanded to cover these scenarios, including edge cases where no attributes are unique or where some unique attributes are not searchable.

### Approach

**Provisioning logic improvements:**

* Introduced a new `uniqueAttribute` type and updated the provisioning logic so that only schema attributes marked as unique are checked for duplicates during user provisioning. Non-unique attributes are ignored for duplicate detection, and the probe order is stabilized by sorting attribute names. [[1]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cR33-R39) [[2]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cL128-R136) [[3]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cL139-L163) [[4]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cR320-R368) [[5]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cL582-R643) [[6]](diffhunk://#diff-8a20de7dc4bb45cd5a0961e2f54a45a2f799275c5ef52dc0c44d1d311a80439cR659-R671)

**Test suite enhancements:**

* Added comprehensive tests to cover cases where:
  - No attributes are unique (provisioning always proceeds).
  - Only some attributes are unique (only those are probed for duplicates).
  - Multiple unique attributes exist (each is probed individually).
  - Unique attributes are not searchable (they are skipped in probing).
  - Non-unique attributes match existing users (provisioning is not blocked).

* Updated existing tests and helper functions to align with the new logic, ensuring that only unique attributes are considered in duplicate checks and that the test mocks reflect the schema's uniqueness constraints. [[1]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L94-R96) [[2]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88R173-L175) [[3]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L396-R564) [[4]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L420-R588) [[5]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L434-R602) [[6]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L456-R624) [[7]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L486-R654) [[8]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L516-R684) [[9]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L545-R713) [[10]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L577-R745) [[11]](diffhunk://#diff-d84c0e159faf5eec6aae87e46f1dc53d59ea89c0847571ef75b6d5e9b3f49d88L602-R770)

### Related Issues
- Resolves https://github.com/thunder-id/thunderid/issues/5300

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duplicate-user detection by checking searchable attributes marked as unique.
  * Supports provisioning users with identical attribute values across organizational units when no unique attribute identifies an existing user.
  * Checks multiple unique attributes and prioritizes matches in the target organizational unit.
  * Continues probing when unique values are unused or matches occur outside the target organizational unit.

* **Tests**
  * Expanded coverage for unique-attribute detection, cross-organizational-unit onboarding, registration flows, and attribute matching.


backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -272,11 +262,11 @@ func (p *provisioningExecutor) handleNonProvisionableUserInRegistration(ctx *pro
 	execResp.Error = existsErr
 }
 
-// handleExistingUser handles the case where a user with the given ID already exists.
-// Returns true if provisioning should proceed (cross-OU case), false if execution should stop.
-func (p *provisioningExecutor) handleExistingUser(ctx *providers.NodeContext, userID string,
+// handleExistingUser handles the case where users already hold one or more of the colliding unique
+// values. Returns true if provisioning should proceed (cross-OU case), false if execution should stop.
+func (p *provisioningExecutor) handleExistingUser(ctx *providers.NodeContext, userIDs []string,
```
ThaminduDilshan: Shall we rename this to `handleExistingUsers` now that it takes a list?

backend/internal/flow/executor/provisioning_executor_test.go
```diff
@@ -3414,6 +3583,91 @@ func (suite *ProvisioningExecutorTestSuite) TestExecute_CrossOU_AmbiguousUser_Ma
 }
 
 // Ambiguous user + cross-OU NOT allowed → fail immediately without searching.
```
ThaminduDilshan: This comment belongs to `TestExecute_CrossOU_AmbiguousUser_CrossOUNotAllowed_Fails`. Shall we move it back?

tests/integration/flow/execution/cross_ou_provisioning_test.go
```diff
@@ -0,0 +1,315 @@
...
+// When every colliding value belongs to a user outside the target OU, cross-OU provisioning
+// proceeds. The create then still fails, because attribute uniqueness is enforced across the whole
+// deployment rather than per OU, so this records the boundary of the cross-OU mode.
+func (ts *CrossOUProvisioningTestSuite) TestCrossOU_AllMatchesOutsideTargetOU_ProceedsThenConflicts() {
+	ts.seedUser(ts.rootOUID, "cross_ou_only_elsewhere", "cross_ou_only_elsewhere@example.com")
+
+	step := ts.onboardInto(ts.childOUID, "cross_ou_brand_new_user", "cross_ou_only_elsewhere@example.com")
+
+	ts.Require().NotNil(step.Error, "The create must report the deployment-wide uniqueness conflict")
+	ts.Equal(errCodeProvisioningAttrConflict, step.Error.Code,
```
ThaminduDilshan: Uniqueness is deployment wide, so this path always ends in FET-1080. The user gets a hard failure here but a recoverable prompt when the match is in the target OU. Shouldn't both be recoverable?

backend/internal/flow/executor/provisioning_executor_test.go
```diff
@@ -3414,6 +3583,91 @@ func (suite *ProvisioningExecutorTestSuite) TestExecute_CrossOU_AmbiguousUser_Ma
...
+		Return(&providers.Entity{ID: secondForeignID, OUID: "ou-b"}, nil)
+	suite.mockUserMgtProvider.On("CreateUser", mock.Anything, mock.MatchedBy(func(u *providers.User) bool {
+		return u.OUID == testOUID && u.Type == testUserType
+	})).Return(&providers.User{ID: testNewUserID, OUID: testOUID, Type: testUserType}, nil)
+
+	resp, err := suite.executor.Execute(ctx)
+
+	assert.NoError(suite.T(), err)
+	assert.NotNil(suite.T(), resp)
+	assert.Equal(suite.T(), providers.ExecComplete, resp.Status)
```
ThaminduDilshan: The integration test shows this create actually fails with FET-1080. Shall we mock `CreateUser` to match instead of asserting `ExecComplete`?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -29,6 +30,13 @@ type entityRef struct {
 	ouID       string
 }
 
+// uniqueAttribute pairs a schema attribute the user type marks unique with the value
+// collected for it during the flow.
+type uniqueAttribute struct {
```
Dilusha-Madushan, line 35: why we need to defined separate struct for the uniqueAttribute and storing all identify attributes in mapper : _map[string]interface{}_?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -612,10 +681,19 @@ func (p *provisioningExecutor) getAttributesForProvisioning(
 			} else if runtimeValue, exists := ctx.RuntimeData[a.Attribute]; exists && runtimeValue != "" {
 				identifyingAttrs[a.Attribute] = convertToSchemaType(runtimeValue, a.Type)
 			}
+			if a.Unique && !slices.Contains(nonSearchableInputs, a.Attribute) {
+				if value, collected := identifyingAttrs[a.Attribute]; collected {
```
Dilusha-Madushan: Since we return three list of attributes data, better if we divide those as unique attributes, non unique attributes and credential attributes like wise. That will reduce additional lookup here right?

Dilusha-Madushan (changes requested):

backend/internal/flow/executor/provisioning_executor.go
NipuniBhagya: Hi @Dilusha-Madushan, 

`identifyingAttrs` and `credentialAttrs` are used as whole maps and they get merged into the create payload and are never iterated entry by entry, so their order doesn't matter.

`uniqueAttrs` is different: each entry becomes its own single-key `IdentifyEntity` search, run in sequence with an early exit on the first match. That makes order observable, it decides which existing user we report and how many queries we run. Go randomizes map iteration, and the source order already comes from `GetAttributes` ranging over the schema's property map, so a map here would be non-deterministic. Three tests depend on `email` being probed before `username`.

So we sort once and keep it in an ordered type. The struct is just the name/value pair, a map plus a sorted `[]string` of keys would do the same job (that's what `AttributeUniquenessValidator` does). The only requirement is that it's a slice rather than a map.

backend/internal/flow/executor/provisioning_executor.go
NipuniBhagya: On splitting the returns into unique / non-unique / credential: `identifyingAttrs` is deliberately the full non-credential set because it's consumed whole, it gets merged straight into the create payload. Making the three disjoint means reassembling it at the call site, so the work moves downstream rather than disappearing.

Dilusha-Madushan:
An entity has two types of attributes: schema-defined attributes and system attributes. Currently, when checking for duplicates, we only consider the schema-defined attributes, which is valid for the user category. However, for agents, both system and schema-defined attributes need to be considered for duplicate detection.

Since the duplication logic differs based on the entity category, it would be better to keep these validation logics within the respective management services.

@ThaminduDilshan @sahandilshan

backend/internal/flow/executor/provisioning_executor_test.go
ThaminduDilshan, line 3585: Seems this comment line belongs to `TestExecute_CrossOU_AllMatchesOutsideTargetOU_Proceeds`

ThaminduDilshan:
@Dilusha-Madushan agree, but IMO the question is one level up. Why is the executor deriving unique attributes at all? Uniqueness should be owned by the entity service, not the executors. Regarding the difference between agent and user, I'd say storing this kind of attribute as a system attribute for the agent feels wrong. We need to see whether there's a reason to do that, but these are not system attributes. They are similar to user attributes.

---
If you check the current implementation, `CreateEntity` already validates the unique attributes. The executor has a second copy of the same rule with its own definition of a duplicate. So duplicating the same logic here feels wrong.

IMO the preference would be to remove the `IdentifyUser` check from the provisioning executor and directly call the create user method. It will perform the required validations and return the result or an error. Based on the error, we can do the existing assertions. But there's one problem with the cross OU provisioning case. Without the identify check, it cannot detect whether this is a user disambiguation case and whether to allow or fail the execution.

I think this is due to the way we have introduced the cross OU provisioning support. In our design, uniqueness is enforced deployment wide. To create the same person in two OUs, we have to drop `unique` from the attributes. That also allows creating the same person twice in the same OU. `allowCrossOUProvisioning` should have been added to cover that.

But the flag doesn't do what its name says. If `unique` is still set, create blocks deployment wide anyway. The flag only suppresses our own check in the executor. It is also a node property deciding a data rule. `POST /users` doesn't honour it.

I think the proper fix would be to correctly define the scope of uniqueness. I.e. decide whether marking an attribute unique should make it unique across the deployment, the OU, or the user type. If it's again a choice, IMO that choice should also go into the schema, not the flow. For example, maybe we can define a scope near the unique flag.

I think with this, we should also be able to drop the `allowCrossOUProvisioning` node property. I feel like it does nothing when it's fixed at the schema level.

---
### 1.0.x

But the above suggestion will be a change of behaviour and we cannot port that fix to the 1.0.x branch. So we need to think of a different solution that preserves the existing behaviour and fixes the bug.

We first need to agree on the expected behaviour.

V1: `allowCrossOUProvisioning` only allows creating the user in another OU. Same OU duplicates are allowed once `unique` is dropped from the schema.
V2: `allowCrossOUProvisioning` allows creating the user in another OU, but also blocks creating the user in the same OU.

If someone doesn't have a disambiguation or cross provisioning use case, but still removes the uniqueness constraint from the schema, the natural expectation is V1. If the schema doesn't declare anything as unique, multiple users can have the same set of attributes. But practically IMO, having the same user in the same OU is of no use.

For the reported issue, I think the problem is the default user onboarding flow. It doesn't set `allowCrossOUProvisioning`, so it stays false. Console add user goes through that flow. That's why dropping the unique attributes and adding the same user to another OU fails.

If it's V2, setting this to true on the default onboarding flow is the complete fix. No code change needed.

If it's V1, we need one more change. The same OU case should also pass when nothing is unique. In `handleExistingUser`, resolve the existing user's OU first. If it's the target OU, go to create and let the schema decide. But we also need to validate this fix with regard to federated login. This could break the federated login cases which has JIT provisioning and allow authentication when a user exists in a registration flow as well.

Either way, most of this PR reverts. We don't need the unique attribute derivation in the executor.

PR #5338, opened 2026-09-08
Require an auth assertion to enroll a passkey through the Direct API

ImalshaD:
### Purpose
  Fix a privilege escalation vulnerability in Direct API passkey registration. Previously, a caller holding the Direct-Auth-Secret could enroll a passkey for an
  arbitrary user and authenticate as that user.

  ### ⚠️ Breaking Changes
  POST /register/passkey/start now requires both userId and an authentication assertion. The assertion’s subject must match userId.

  #### 💥 Impact
  Direct API integrations that register passkeys without an assertion will receive HTTP 400. Users need an existing authentication method before enrolling their
  first passkey through the Direct API.

  Orchestrated registration flows are unaffected.

  #### 🔄 Migration Guide
  Authenticate the user first, then include the returned assertion in the registration-start request:

```text
  {
    "userId": "<authenticated-user-id>",
    "assertion": "<assertion-from-prior-authentication>",
    "relyingPartyId": "localhost",
    "relyingPartyName": "Example"
  }
```
  Use an orchestrated registration flow when a passkey must be the account’s first credential.

  ### Approach
  - Require userId and assertion through request validation.
  - Verify the assertion’s signature, expiration, issuer, and assurance claim, then confirm its subject matches userId before initiating enrollment.
  - Enforce ownership at the Direct API service layer to preserve orchestrated registration flows.
  - Add unit and integration tests for missing, invalid, and mismatched assertions, attempted account takeover, and successful enrollment with the user’s own
    assertion.
  - Update API documentation and integration test setup for the authentication requirement.

  ### Related Issues
  - #5424
  ### Checklist
  - [x] Documentation provided.
  - [x] Tests provided.
      - [x] Unit Tests
      - [x] Integration Tests
  - [x] Breaking changes.
      - [ ] breaking change label added.

  ### Security checks
  - [ ] Followed secure coding standards in WSO2 Secure Coding Guidelines

## Summary by CodeRabbit

- **New Features**
  - Passkey registration now requires a valid authentication assertion proving the target user’s identity.
  - Registration rejects missing, unverifiable, or mismatched assertions with clearer error responses.
  - Added structured validation details for missing or blank required fields.

- **Documentation**
  - Updated passkey and authentication guidance to explain assertion requirements and Direct Auth Secret limitations.
  - Added examples for new registration error scenarios.

- **Tests**
  - Added comprehensive coverage for successful enrollment and assertion validation failures.

### DIFF
--- a/api/registration.yaml
+++ b/api/registration.yaml
@@ -47,11 +47,19 @@ paths:
           description: >
             Bad Request: The request body is malformed or contains invalid data. Client-side
             failures (invalid input, user not found, etc.) surface as a single enrollment failure
-            code (AUTHN-PSK-1002).
+            code (AUTHN-PSK-1002). A request that does not prove the target user is rejected
+            before any challenge is issued: an `assertion` that cannot be verified returns
+            AUTHN-1009, and one whose subject is not `userId` returns AUTHN-1010. A missing,
+            empty or blank `assertion` (or `userId`) is a required-field violation and is
+            rejected by request validation before the handler runs, which returns the
+            `INVALID_INPUT_METADATA` shape with a per-field `errors` map instead of the codes
+            above.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Error'
+                oneOf:
+                  - $ref: '#/components/schemas/Error'
+                  - $ref: '#/components/schemas/ValidationError'
               examples:
                 enrollmentFailed:
                   value:
@@ -62,6 +70,31 @@ paths:
                     description:
                       key: "error.authnservice.passkey_enrollment_failed_description"
                       defaultValue: "The passkey enrollment attempt failed"
+                missingAssertion:
+                  value:
+                    code: "INVALID_INPUT_METADATA"
+                    message: "Validation Failed"
+                    description: "One or more inbound fields failed structural edge-boundary rules."
+                    errors:
+                      assertion: "The field 'assertion' is missing but is strictly required."
+                invalidAssertion:
+                  value:
+                    code: "AUTHN-1009"
+                    message:
+                      key: "error.authnservice.invalid_assertion"
+                      defaultValue: "Invalid assertion"
+                    description:
+                      key: "error.authnservice.invalid_assertion_description"
+                      defaultValue: "The provided assertion token is invalid"
+                assertionSubjectMismatch:
+                  value:
+                    code: "AUTHN-1010"
+                    message:
+                      key: "error.authnservice.assertion_subject_mismatch"
+                      defaultValue: "Assertion subject mismatch"
+                    description:
+                      key: "error.authnservice.assertion_subject_mismatch_description"
+                      defaultValue: "The subject in the assertion does not match the authenticated user"
         "500":
           description: 'Internal Server Error: An unexpected error occurred while processing the request'
           content:
@@ -177,9 +210,18 @@ components:
           description: Attestation conveyance preference
           enum: ["none", "indirect", "direct", "enterprise"]
           example: "none"
+        assertion:
+          type: string
+          description: >
+            Auth assertion obtained from a prior authentication, proving the caller holds the
+            account named by `userId`. The credential is enrolled only for the assertion subject,
+            so the assertion `sub` and `userId` must be the same user. Holding the Direct Auth
+            Secret is not sufficient on its own.
+          example: "eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiI1NTBlODQwMC1lMjliIn0.signature"
       required:
         - userId
         - relyingPartyId
+        - assertion
 
     PasskeyRegistrationStartResponse:
       type: object
@@ -432,6 +474,36 @@ components:
           $ref: '#/components/schemas/I18nMessage'
         description:
           $ref: '#/components/schemas/I18nMessage'
+
+    ValidationError:
+      type: object
+      description: >
+        Returned when a request fails required-field or bounds validation, which happens before the
+        handler runs. Unlike `Error`, its `message` and `description` are plain strings rather than
+        internationalized objects, and it carries an `errors` map naming each field that failed.
+      required:
+        - code
+        - message
+        - description
+        - errors
+      properties:
+        code:
+          type: string
+          description: Always `INVALID_INPUT_METADATA` for a validation failure.
+          example: "INVALID_INPUT_METADATA"
+        message:
+          type: string
+          example: "Validation Failed"
+        description:
+          type: string
+          example: "One or more inbound fields failed structural edge-boundary rules."
+        errors:
+          type: object
+          description: One entry per field that failed validation, keyed by the field's JSON name.
+          additionalProperties:
+            type: string
+          example:
+            assertion: "The field 'assertion' is missing but is strictly required."
 
     I18nMessage:
       type: object
--- a/backend/internal/authn/service.go
+++ b/backend/internal/authn/service.go
@@ -61,7 +61,7 @@ type AuthenticationServiceInterface interface {
 	) (*common.AuthenticationResponse, *tidcommon.ServiceError)
 	// Passkey methods
 	StartPasskeyRegistration(ctx context.Context, userID, relyingPartyID, relyingPartyName string,
-		authSelection *PasskeyAuthenticatorSelectionDTO, attestation string,
+		authSelection *PasskeyAuthenticatorSelectionDTO, attestation, assertion string,
 	) (interface{}, *tidcommon.ServiceError)
 	FinishPasskeyRegistration(ctx context.Context, credential PasskeyPublicKeyCredentialDTO,
 		sessionToken string, skipAssertion bool, existingAssertion string,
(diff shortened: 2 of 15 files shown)


api/registration.yaml
```diff
@@ -47,7 +47,13 @@ paths:
...
             Bad Request: The request body is malformed or contains invalid data. Client-side
             failures (invalid input, user not found, etc.) surface as a single enrollment failure
-            code (AUTHN-PSK-1002).
+            code (AUTHN-PSK-1002). A request that does not prove the target user is rejected
+            before any challenge is issued: an `assertion` that cannot be verified returns
+            AUTHN-1009, and one whose subject is not `userId` returns AUTHN-1010. A missing,
+            empty or blank `assertion` (or `userId`) is a required-field violation and is
+            rejected by request validation before the handler runs, which returns the
+            `INVALID_INPUT_METADATA` shape with a per-field `errors` map instead of the codes
+            above.
```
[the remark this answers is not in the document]
ImalshaD: Thanks, but I don't think this one holds — the current path is correct, and the suggested change would break the build.

This site sets trailingSlash: true (docs/docusaurus.config.ts:61), so a page URL ends in / and relative links resolve against it as a directory — one more .. than the filesystem path.

Verified with a full build, which has onBrokenLinks: 'throw' (docusaurus.config.ts:67):

Broken links: 0
SUCCESS] Generated static files in "build".

The link renders as href="/docs/key-concepts/authentication/passwordless/passkeys/", and the target exists at build/docs/key-concepts/authentication/passwordless/passkeys.

The suggested ./passwordless/passkeys would resolve to /docs/key-concepts/authentication/integration-models/passwordless/passkeys, which does not exist — so applying it would introduce the broken link the comment warns about.

 Two existing pages follow the same convention: passkeys.mdx:191 uses ../../integration-models (one level up on disk, two in the link), and magiclink.mdx:27 uses
 ../../../../guides/... (three on disk, four in the link).

@coderabbitai please verify this against a full docs build and resolve.

docs/content/key-concepts/authentication/integration-models.mdx
```diff
@@ -352,6 +352,8 @@ On a fresh setup, the `setup.sh` / `setup.ps1` script generates a secret, writes
 
 This suits the intended Direct API callers, such as CLI tools, automation scripts, and server-to-server integrations, which can hold a configured secret. Browser and mobile applications, which cannot safely store a static secret, should use the [Redirect-Based](#redirect-based) or [App-Native](#app-native) model instead.
 
+The secret identifies the calling integration, not the end user, and every caller shares the same value. So it authorizes use of the Direct API but never stands in for a user's own identity. Endpoints that act on a specific account require separate proof of that user: enrolling a passkey through `/register/passkey/start`, for example, also requires the `assertion` from a prior authentication, and enrolls the credential only for that assertion's subject. See [Passkeys](../passwordless/passkeys#registration-flow).
```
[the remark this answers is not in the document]
ImalshaD: Agreed that Error doesn't model the INVALID_INPUT_METADATA response, but I'd rather not address it only here.

 The gap is repo-wide rather than specific to this endpoint — it comes from the shared DecodeJSONBody / validateStructNatively helper, so any endpoint using it can return that shape. In the current specs:

- no ValidationError schema exists anywhere under api/
 - INVALID_INPUT_METADATA appears exactly once, in the prose this PR adds
 - 18 spec files return 400 with the plain Error schema

Adding a oneOf here would make this the only endpoint in the API that models it, which is more misleading for generated clients than the current uniform omission. It belongs in a dedicated spec change that introduces one shared ValidationError schema and applies it across the affected paths, rather than in a security fix.

 I'll raise a separate issue to track it.

@coderabbitai please verify and resolve the thread — the underlying point is valid, but it will be handled repo-wide in a follow-up rather than in this PR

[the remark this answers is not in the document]
ImalshaD: @coderabbitai resolve

api/registration.yaml
```diff
@@ -47,7 +47,13 @@ paths:
           description: >
             Bad Request: The request body is malformed or contains invalid data. Client-side
             failures (invalid input, user not found, etc.) surface as a single enrollment failure
-            code (AUTHN-PSK-1002).
+            code (AUTHN-PSK-1002). A request that does not prove the target user is rejected
+            before any challenge is issued: an `assertion` that cannot be verified returns
+            AUTHN-1009, and one whose subject is not `userId` returns AUTHN-1010. A missing,
+            empty or blank `assertion` (or `userId`) is a required-field violation and is
+            rejected by request validation before the handler runs, which returns the
+            `INVALID_INPUT_METADATA` shape with a per-field `errors` map instead of the codes
```
ThaminduDilshan, line 55: The 400 schema is `Error`, which has no `errors` field. Shall we model the `INVALID_INPUT_METADATA` shape in the spec as well?

backend/internal/authn/service.go
```diff
@@ -571,6 +571,32 @@ func (as *authenticationService) extractClaimsFromAssertion(ctx context.Context,
...
 }
 
+// verifyAssertionSubject rejects the request unless assertion is a valid auth assertion whose
+// subject is userID. Callers use it to gate operations that act on an account rather than merely
+// reporting on one, where a caller supplied user ID is not trustworthy on its own.
+func (as *authenticationService) verifyAssertionSubject(ctx context.Context, assertion, userID string,
+	logger *log.Logger) *tidcommon.ServiceError {
+	// Callers arriving over HTTP are already stopped by the required-field validation on the DTO,
+	// which treats a blank string as absent. This guards the service contract for any other caller.
+	if strings.TrimSpace(assertion) == "" {
```
ThaminduDilshan: The DTO rejects a blank assertion first, and `extractClaimsFromAssertion` already returns `ErrorInvalidAssertion` for one. Do we need this guard and the two tests for it?

backend/internal/authn/service.go
```diff
@@ -571,6 +571,32 @@ func (as *authenticationService) extractClaimsFromAssertion(ctx context.Context,
...
+		logger.Debug(ctx, "Assertion missing on a request that requires proof of the target user")
+		return &common.ErrorInvalidAssertion
+	}
+
+	_, assertionSub, svcErr := as.extractClaimsFromAssertion(ctx, assertion, logger)
+	if svcErr != nil {
+		return svcErr
+	}
+
+	if assertionSub != userID {
```
ThaminduDilshan, line 591: A flow issued assertion sets `sub` from the application's subject attribute mapping (`resolveSubject` in `auth_assert_executor.go`), not the user ID. Won't this reject a caller that authenticated through a flow?

@coderabbitai verify and provide short answer

tests/integration/authn/passkey_registration_binding_test.go
```diff
@@ -0,0 +1,291 @@
...
+		if err := testutils.DeleteOrganizationUnit(ts.ouID); err != nil {
+			ts.T().Logf("teardown: failed to delete organization unit: %v", err)
+		}
+	}
+}
+
+// TestStartRejectsRequestWithoutAssertion asserts a registration start that proves nothing about
+// the caller is rejected, rather than handing out a challenge bound to the named user. An absent
+// assertion is the zero value of a required field, so it is caught by DTO validation before the
+// service runs, which is why the code here is the structural one rather than AUTHN-1012.
```
ThaminduDilshan: We don't have an `AUTHN-1012`. Did you mean `AUTHN-1009`?

tests/integration/authn/passkey_registration_binding_test.go
```diff
@@ -0,0 +1,291 @@
...
+// TestAttackerCannotTakeOverVictimAccount is the end to end proof. It runs the full ceremony the
+// way an attacker holding the Direct Auth Secret, its own account, and the victim's user ID would,
+// then tries to authenticate as the victim with the authenticator it enrolled. Enrollment must be
+// refused; if it is not, authentication as the victim must at least fail.
+func (ts *PasskeyRegistrationBindingTestSuite) TestAttackerCannotTakeOverVictimAccount() {
+	attackerAssertion := ts.assertionFor("passkeyattacker", "AttackerPassword123!", ts.attackerID)
+
+	authenticator, userHandle, err := testutils.RegisterPasskeyCredential(
+		ts.victimID, testRelyingPartyID, testRelyingPartyName, testPasskeyOrigin, attackerAssertion)
+	if err != nil {
```
ThaminduDilshan: This passes on any error from `RegisterPasskeyCredential`, not just the rejection. Shall we assert the status and error code instead of returning early?

tests/integration/authn/passkey_registration_binding_test.go
```diff
@@ -0,0 +1,291 @@
...
+		"assertion":      victimAssertion,
+	})
+
+	ts.Equal(http.StatusOK, status,
+		"registration start should accept an assertion proving the target user; body: %s", body)
+}
+
+// assertionFor authenticates the user with a password and returns the auth assertion, which is the
+// proof of identity the registration API is expected to require.
+func (ts *PasskeyRegistrationBindingTestSuite) assertionFor(username, password, expectedID string) string {
```
ThaminduDilshan: This duplicates `testutils.ObtainAuthAssertion` added in this PR. Shall we reuse it?

docs/content/key-concepts/authentication/passwordless/passkeys.mdx
```diff
@@ -82,13 +82,25 @@ This approach requires **no custom UI development** or direct WebAuthn API calls
...
+This means a user needs an existing credential before they can enroll their first passkey through
+the Direct API. To register a passkey as the very first credential on an account, drive a
+registration [flow](../../../../guides/flows/what-are-flows) instead.
+
 1) **Start registration.** Create WebAuthn creation options and a session token.
 
 ```bash
 curl -k -X POST https://localhost:8090/register/passkey/start \
   -H "Content-Type: application/json" \
+  -H "Direct-Auth-Secret: <secret>" \
```
ThaminduDilshan, line 100: The finish and authentication samples below don't carry this header. Shall we add it there as well?

tests/integration/authn/passkey_auth_test.go
```diff
@@ -48,6 +48,12 @@ var (
 			"displayName": map[string]interface{}{
 				"type": "string",
 			},
+			// Registration start requires an assertion proving the target user, so every user in
```
ThaminduDilshan: This file carries both passkey atomic API authentication and registration related tests right? I'd prefer to split authentication and registration into two separate entities, because with this we're kind of bounding passkey authentication to a other credential. I.e. password. But in a pure passwordless setup, schema shouldn't have a password or similar credential.

This change is required for your case, because what we have now is a passkey enrollment API, not a passkey user registration API. Hence I'd prefer to keep this schema as it is and split registration use case into a another schema/ or even test suite

backend/internal/authn/service.go
ImalshaD: @ThaminduDilshan this is a edge case right? this will happen only if the subject mapping is configured, but most importantly if someone is invoking the flows i don't see a need for them to invoke the atomic apis, given that atomic apis intended audience is the client who do not need to invoke flows or standard OAuth flows. Given the constraints, shall we proceed with this approach?

tests/integration/authn/passkey_auth_test.go
ImalshaD: @ThaminduDilshan Agreed on separating enrollment and authentication. I’ve consolidated the enrollment and ownership-check cases into passkey_enrollment_test.go, keeping authentication tests separate.

The authentication suite uses one schema and two users: one without credentials for negative cases, and one with a passkey for successful authentication cases. The schema includes an optional password because setup provisions that passkey through the atomic enrollment API. Before this fix, setup could enroll using only the user ID and Direct-Auth-Secret. Now it needs an assertion, so the password is used only to obtain that assertion during setup. The authentication tests themselves use the passkey.

Keeping the schema entirely passwordless would require provisioning through an orchestrated registration flow or a suitable fixture. For this PR, shall we retain the existing provisioning approach with the setup-only password? If you prefer preserving the passwordless schema, I can revisit the provisioning method.

backend/internal/authn/service.go
ThaminduDilshan: Makes sense. `validateAndAppendAuthAssertion` already rejects the same mismatch, so this isn't new. Let's proceed

ThaminduDilshan (approved):

PR #5346, opened 2026-09-09
Fix roles claim missing declarative roles assigned via the database

### LINKED ISSUE #5345: Declarative roles do not honor the `composite` store configuration for `GetUserRoles` function
### Description

`GetUserRoles` in the composite authz store does not bridge DB-stored role assignments onto roles that are defined in the file-based (declarative) store. It queries each store independently and merges the results, so a role defined declaratively but assigned via the DB never appears in a user's role list (and therefore never appears in the `roles` claim).

Three sibling methods on the same composite store already handle exactly this case:

| Composite method | Bridges DB assignment onto a file-defined role? |
|---|---|
| `GetAuthorizedPermissionsByResourceServer` | yes, via `crossStoreAuthorizedPermissions` (`:454`) |
| `GetAllPermissionsForAssignees` | yes, via `crossStoreAllPermissions` (`:558`) |
| `GetRoleAssignments` | yes, via `getCompositeAssignments` (`:230`) |
| `GetUserRoles` | **no** (`:608`), independent merge only |

Suggested fix, reusing existing primitives the same way `crossStoreAuthorizedPermissions:464` does:
```go
roleIDs := dbStore.GetEntityRoleIDs(entityID, groupIDs)
for _, id := range roleIDs {
    if fileStore.IsRoleExist(id) {
        role := fileStore.GetRole(id)
        names = append(names, role.Name)
    }
}
```
Error handling should match the sibling helper: skip ErrRoleNotFound and ErrRoleDataCorrupted, propagate anything else.

### Steps to Reproduce

1. Configure the authz store in `composite` mode (file-based + DB).
2. Define a role declaratively in a file-based resource (e.g. an admin role on a resource server).
3. Assign a user to that role through the API, so the assignment lands in the DB `ROLE_ASSIGNMENT` table.
4. `GET /roles/{id}/assignments` for that role. The user IS listed (works as expected).
5. Get a token for that user, or call the user-roles path that hits `GetUserRoles`.

Expected: the declarative role appears in the user's roles / `roles` claim.
Actual: the declarative role is absent. Only DB-defined roles are returned.

### Version

1.0.0

Environment Details

- ThunderID 1.0.0
- Authz store mode: `composite` (file-based declarative resources + DB)
- Database: PostgreSQL
- Deployed on Kubernetes via OpenChoreo (dev environment)

rajithacharith:
### Purpose
`GetUserRoles` on the composite role store queried the file-based (declarative) store and the database store independently and merged the results. A role defined declaratively but assigned to a user via the database (API) fell into the gap between the two independent lookups, so it never appeared in the user's roles / `roles` token claim, even though `GET /roles/{id}/assignments` correctly showed the assignment.

### Approach
Added a `crossStoreUserRoles` bridge on `compositeRoleStore`, mirroring the same pattern already used by three sibling methods (`crossStoreAuthorizedPermissions`, `crossStoreAllPermissions`, `getCompositeAssignments`): fetch assigned role IDs from the DB store, resolve any that exist in the file store to their declarative name, skip `ErrRoleNotFound`/`ErrRoleDataCorrupted`, and propagate other errors. `GetUserRoles` now merges DB roles, file-store roles, and this cross-store bridge.

Verified end-to-end against a locally built distribution (`make build`): configured `role.store: composite`, defined a declarative role with no YAML assignments, assigned it to a test user via the API (DB-only assignment), and confirmed the role now appears in the user's token `roles` claim. Reverting the fix and rebuilding reproduces the reported bug exactly (claim silently missing); reapplying the fix restores it.

### Related Issues
- Fixes #5345

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

* **Bug Fixes**
  * User role listings now include names for roles assigned at runtime when those roles are defined in configuration files.
  * Database-only, missing, or corrupted role definitions are excluded from results without disrupting valid role listings.
  * Unexpected role lookup errors are now surfaced appropriately.
  * Duplicate roles are removed from the returned results.

### DIFF
--- a/backend/internal/role/all_permissions_test.go
+++ b/backend/internal/role/all_permissions_test.go
@@ -497,6 +497,109 @@ func TestCrossStoreAllPermissions_RemovedRoleIsStillSkipped(t *testing.T) {
 	assert.Empty(t, result)
 }
 
+// A role defined declaratively in the file store but assigned to the user via the DB must still
+// appear in the user's role names, not just in GetRoleAssignments/GetAllPermissionsForAssignees.
+func TestGetUserRoles_BridgesDBAssignmentOntoFileDefinedRole(t *testing.T) {
+	ctx := context.Background()
+	dbStore := newRoleStoreInterfaceMock(t)
+	fileStore := newRoleStoreInterfaceMock(t)
+	store := newCompositeRoleStore(fileStore, dbStore)
+
+	dbStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{"dbRole"}, nil).Once()
+	fileStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{}, nil).Once()
+	dbStore.EXPECT().GetEntityRoleIDs(ctx, "user1", []string{}).Return([]string{"declarativeRole"}, nil).Once()
+	fileStore.EXPECT().IsRoleExist(ctx, "declarativeRole").Return(true, nil).Once()
+	fileStore.EXPECT().GetRole(ctx, "declarativeRole").
+		Return(RoleWithPermissions{Name: "Declarative Admin"}, nil).Once()
+
+	result, err := store.GetUserRoles(ctx, "user1", []string{})
+
+	require.NoError(t, err)
+	assert.ElementsMatch(t, []string{"dbRole", "Declarative Admin"}, result)
+}
+
+// A role ID from a DB assignment that is not declarative is already covered by dbStore.GetUserRoles,
+// so it must not be resolved (or double-counted) via the file store.
+func TestCrossStoreUserRoles_DBOnlyRoleIsSkipped(t *testing.T) {
+	ctx := context.Background()
+	dbStore := newRoleStoreInterfaceMock(t)
+	fileStore := newRoleStoreInterfaceMock(t)
+	store := newCompositeRoleStore(fileStore, dbStore)
+
+	dbStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{"dbRole"}, nil).Once()
+	fileStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{}, nil).Once()
+	dbStore.EXPECT().GetEntityRoleIDs(ctx, "user1", []string{}).Return([]string{"dbOnlyRole"}, nil).Once()
+	fileStore.EXPECT().IsRoleExist(ctx, "dbOnlyRole").Return(false, nil).Once()
+
+	result, err := store.GetUserRoles(ctx, "user1", []string{})
+
+	require.NoError(t, err)
+	assert.Equal(t, []string{"dbRole"}, result)
+}
+
+// A role whose declarative definition was removed after the assignment was made genuinely confers
+// no name, so skipping it remains correct.
+func TestCrossStoreUserRoles_RemovedRoleIsSkipped(t *testing.T) {
+	ctx := context.Background()
+	dbStore := newRoleStoreInterfaceMock(t)
+	fileStore := newRoleStoreInterfaceMock(t)
+	store := newCompositeRoleStore(fileStore, dbStore)
+
+	dbStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{}, nil).Once()
+	fileStore.EXPECT().GetUserRoles(ctx, "user1", []string{}).Return([]string{}, nil).Once()
+	dbStore.EXPECT().GetEntityRoleIDs(ctx, "user1", []string{}).Return([]string{"gone"}, nil).Once()
+	fileStore.EXPECT().IsRoleExist(ctx, "gone").Return(true, nil).Once()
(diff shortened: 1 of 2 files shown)


rajithacharith:
@coderabbitai full review

backend/internal/role/composite_store.go
```diff
@@ -618,7 +620,55 @@ func (c *compositeRoleStore) GetUserRoles(
 		return nil, err
 	}
 
-	return mergePermissions(dbRoleNames, fileRoleNames), nil
+	crossStoreRoleNames, err := c.crossStoreUserRoles(ctx, entityID, groupIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	return mergePermissions(mergePermissions(dbRoleNames, fileRoleNames), crossStoreRoleNames), nil
```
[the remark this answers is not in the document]
rajithacharith: @coderabbitai Does the docs mention that these assignments are excluded from the token? If so we need to change it. Otherwise we don't need to specify

backend/internal/role/composite_store.go
```diff
@@ -618,7 +620,55 @@ func (c *compositeRoleStore) GetUserRoles(
...
+}
+
+// crossStoreUserRoles resolves role names for the (declarative role definition in file store) +
+// (runtime assignment row in DB) case, mirroring crossStoreAuthorizedPermissions.
+func (c *compositeRoleStore) crossStoreUserRoles(
```
ThaminduDilshan: This is the third copy of the same roleIDs -> file store resolution loop. Shall we extract a shared helper that returns the resolved declarative roles?

backend/internal/role/composite_store.go
```diff
@@ -618,7 +620,55 @@ func (c *compositeRoleStore) GetUserRoles(
...
+	if err != nil {
+		return nil, err
+	}
+	if len(roleIDs) == 0 {
+		return []string{}, nil
+	}
+
+	names := make([]string, 0, len(roleIDs))
+	for _, id := range roleIDs {
+		exists, err := c.fileStore.IsRoleExist(ctx, id)
```
ThaminduDilshan, line 650: `IsRoleExist` and `GetRole` both do a `Get` on the same ID, and both the `!exists` and `ErrRoleNotFound` branches just `continue`. Do we need this check?

@coderabbitai validate and provide a short reply.

backend/internal/role/composite_store.go
```diff
@@ -618,7 +620,55 @@ func (c *compositeRoleStore) GetUserRoles(
...
+			return nil, err
+		}
+		if !exists {
+			// Role is DB-only; already covered by dbStore.GetUserRoles.
+			continue
+		}
+		role, err := c.fileStore.GetRole(ctx, id)
+		if err != nil {
+			// See crossStoreAuthorizedPermissions: benign cases are skipped, anything else propagates.
+			if errors.Is(err, ErrRoleNotFound) || errors.Is(err, ErrRoleDataCorrupted) {
```
ThaminduDilshan, line 661: `crossStoreAllPermissions` propagates `ErrRoleDataCorrupted` but this skips it, so a corrupted declarative role silently drops out of the roles claim. Any reason for the difference?

backend/internal/role/all_permissions_test.go
```diff
@@ -497,6 +497,109 @@ func TestCrossStoreAllPermissions_RemovedRoleIsStillSkipped(t *testing.T) {
 	assert.Empty(t, result)
 }
 
+// A role defined declaratively in the file store but assigned to the user via the DB must still
+// appear in the user's role names, not just in GetRoleAssignments/GetAllPermissionsForAssignees.
+func TestGetUserRoles_BridgesDBAssignmentOntoFileDefinedRole(t *testing.T) {
```
ThaminduDilshan, line 502: These are `GetUserRoles` tests, not permission tests. Shall we move them to `composite_store_test.go` and follow the `CompositeRoleStoreTestSuite` pattern?

backend/internal/role/all_permissions_test.go
```diff
@@ -497,6 +497,109 @@ func TestCrossStoreAllPermissions_RemovedRoleIsStillSkipped(t *testing.T) {
...
+
+	result, err := store.GetUserRoles(ctx, "user1", []string{})
+
+	require.NoError(t, err)
+	assert.Equal(t, []string{"dbRole"}, result)
+}
+
+// A role whose declarative definition was removed after the assignment was made genuinely confers
+// no name, so skipping it remains correct.
+func TestCrossStoreUserRoles_RemovedRoleIsSkipped(t *testing.T) {
```
ThaminduDilshan: `make lint_backend` fails with 3 `dupl` issues on these tests. Can't we use a table driven pattern here?

backend/internal/role/composite_store.go
```diff
@@ -618,7 +620,55 @@ func (c *compositeRoleStore) GetUserRoles(
...
+				continue
+			}
+			log.GetLogger().Error(ctx,
```
ThaminduDilshan, line 664: Are we both logging and returning a server error here? If so we should only return right? Service level will log it

PR #5385, opened 2026-09-13
[Feature] Add MOSIP eSignet as a login provider connection

Akila94:
# Add MOSIP eSignet as a login provider connection

## Purpose

Adds MOSIP eSignet as a login provider connection, so a ThunderID flow can federate
authentication to a government identity system that uses eSignet as the identity management platform.

An operator configures an eSignet connection once under Connections, then drops the eSignet
executor into any authentication or registration flow. The flow node references the connection
by `idpId` and carries nothing eSignet-specific, so endpoints and credentials are managed in one
place rather than restated per flow.

eSignet differs from the existing connections in one important way: it authenticates its
clients with `private_key_jwt` rather than a client secret. The connection therefore stores a
PEM-encoded RSA signing key and the `kid` of the matching JWK registered with eSignet, in place
of a client secret.

### Connection

Screenshot 2026-09-03 at 8 25 09 PM

Screenshot 2026-09-03 at 8 25 22 PM

A new `esignet` vendor in the connections API with full CRUD. The signing key is
treated exactly as a client secret is: stored encrypted, write-only on update (omit it to keep the
stored value), masked as `******` on read, and never decrypted into an API response. On
configuration export the parameterizer externalizes it to the generated `.env` file, so exported
YAML carries `signingKey: {{.EXPORT_E_SIGNET_SIGNING_KEY}}` and never the key itself.

### Executor 

Screenshot 2026-09-03 at 8 26 26 PM

Two-pass authorize and callback, following the existing OAuth and OIDC executors:

- The authorize leg mints a nonce, a state, and a PKCE S256 verifier, stashes all three in runtime
  data, and redirects. They are single-use and are cleared as soon as the callback is handled.
- The callback validates state when the client returns it (matching the OAuth executor, since a
  client handling CSRF itself may omit it), exchanges the code with a signed client assertion,
  then verifies both the ID token and the JWS userinfo response against eSignet's JWKS.

Key decisions worth calling out:

- **Signature algorithms are pinned** to `RS256` and `PS256` rather than read from the token
  header. This is what stops a token selecting `none` or an HMAC scheme and being verified against
  a public key. Both are accepted because eSignet advertises both and picks per deployment;
  accepting only RS256 failed every login against a PS256 deployment.
- **Audience handles both claim shapes** RFC 7519 permits. eSignet emits the array form, so
  asserting a plain string yielded `""` and failed every login.
- **The userinfo subject is cross checked** against the ID token subject before any claim is
  trusted, since the two responses are fetched independently.
- **The signing key is decrypted only where the client assertion is signed.** Configuration
  resolution keeps it encrypted and validates presence only. The authorize leg signs nothing, so
  the PEM never enters memory there. Go strings cannot be zeroed, so narrowing the key's reach is
  the available mitigation.

Verified claims are mapped onto the target user type's provisioning attributes and handed to the
authn provider for local user resolution and JIT provisioning, keyed on the eSignet subject.

**Console.** eSignet appears as its own connector card under a new **Login Provider** category.
The configuration form takes the endpoints, the signing key and `kid`, ACR values and an optional username prefix. The key is
entered through a multiline masked field, because a single line input strips the line breaks on
paste and corrupts the value. The executor and its widget are registered with the flow builder.

### Related Issues
- N/A 

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

    Not applicable. This is additive: a new vendor, a new executor and a new presentation
    category. No existing connection, flow or API shape changes.

**Tests added**

Backend unit tests:
- `backend/internal/flow/executor/esignet_oidc_executor_test.go` covers configuration
  resolution, the authorize redirect, the state gate, client assertion signing, JWKS
  verification, claim mapping, and the callback end to end against a mock eSignet serving the
  token, userinfo and JWKS endpoints, including the audience, nonce, missing subject, userinfo
  failure and subject mismatch branches.
- `backend/internal/connection/esignet_test.go` covers CRUD mapping and secret handling.
- `backend/internal/connection/declarative_resource_test.go` and
  `backend/internal/system/export/service_test.go` cover declarative import and export,
  asserting the exported YAML contains the placeholder and not the key.
- `backend/internal/idp/utils_test.go` covers the property helpers.

Integration tests:
- `tests/integration/connection/connection_api_test.go` and `vendor_list_test.go`.

Frontend tests: connection vendor metadata, form fields, the listing categories, and the flow
builder resource registration.

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added MOSIP eSignet as a supported identity provider and connection type.
  * Added eSignet connection management with signing-key configuration, secure masking, and multiline input.
  * Added “Continue with eSignet” flow widgets and login support.
  * Added eSignet authentication with PKCE, token verification, and user provisioning.
  * Added eSignet branding and configuration options in the console.

* **Documentation**
  * Added eSignet setup guidance and updated identity-provider and flow documentation.


ThaminduDilshan:
Hi @Akila94,

Thanks for your interest in proposing this feature. Shall we follow the feature proposal process [1] and create a GitHub issue first? Based on the triage and acceptance, we can proceed with the next steps.

[1] https://thunderid.dev/community/contributing/propose-a-feature/

PR #5391, opened 2026-09-14
Add Multi Value Support to ENTITY_IDENTIFIER table

Sadeesha-Sath:
### Purpose

`ENTITY_IDENTIFIER` stores indexed entity identifiers (e.g. email, phone, username) used for fast entity lookup during authentication and identification. Its primary key was `(ENTITY_ID, DEPLOYMENT_ID, NAME)`, which limited each entity to a single value per identifier name. An entity couldn't have, for example, two indexed email addresses at once: a second value for the same name silently replaced the first.

This PR widens the primary key to include `VALUE`, so an entity can hold multiple distinct values per identifier name.

---
The `ENTITY_IDENTIFIER` table's primary key changes from `(ENTITY_ID, DEPLOYMENT_ID, NAME)` to `(ENTITY_ID, DEPLOYMENT_ID, NAME, VALUE)`.

#### 💥 Impact
Any existing deployment that already ran the old schema script has an `ENTITY_IDENTIFIER` table with the old primary key. The table must be recreated for the new key to take effect. Fresh installs are unaffected.

#### 🔄 Migration Guide
For an existing deployment, before upgrading:
1. Back up the `ENTITY_IDENTIFIER` table's data if it needs to be retained.
2. Drop and recreate `ENTITY_IDENTIFIER` using the updated `postgres.sql`/`sqlite.sql` definition (or apply an equivalent `ALTER TABLE ... DROP CONSTRAINT` / add-PK migration by hand).
3. Re-populate the table by re-syncing indexed attributes for existing entities

---
### Approach

- **Schema** (`backend/dbscripts/entitydb/postgres.sql`, `sqlite.sql`): changed the `ENTITY_IDENTIFIER` primary key to `(ENTITY_ID, DEPLOYMENT_ID, NAME, VALUE)`.
- **Write path** (`backend/internal/entity/store.go`): the PK change alone wouldn't enable anything, since indexed attribute values were always scalars and the code deduplicated to one row per name before insert. To make the feature usable end-to-end:
  - `attrValueToStrings` is added alongside the existing `attrValueToString`: a scalar attribute value still produces a single identifier value, while an array attribute value now produces one identifier value per element (non-scalar elements are skipped, same as before).
  - `prepareIdentifierQuery` collects schema- and system-attribute entries separately and keeps the existing precedence rule (system attribute values fully override schema attribute values for the same name), now applied per name across however many values each side has.
  - Deduplication changed from "one row per name" to "one row per exact `(name, value)` pair", so distinct values for the same name are no longer collapsed, while exact repeats (e.g. duplicate array entries) still insert once.
- No changes were needed to the identifier lookup/identify queries (`buildIdentifyQueryFromIdentifiers`, `buildIdentifyQueryHybrid`) or the delete queries, since they already key off `ENTITY_ID`/`(NAME, VALUE)` and don't assume one row per name.

### Related Issues
- N/A

### Related PRs
- N/A

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Entities can store multiple identifier values under the same name, such as multiple email addresses.
  - Array-valued attributes are supported for indexed identifiers, creating an entry for each valid value.
  - Duplicate identifier values are removed while preserving their original order.
  - System-provided attributes take precedence over schema-provided values with the same name.

- **Documentation**
  - Clarified handling of single- and multiple-value indexed attributes, including precedence rules.

- **Tests**
  - Added coverage for multiple values, duplicate handling, array attributes, and precedence behavior.


KaveeshaPiumini:
Could we revisit the implications of adding VALUE to the primary key here? 

While this enables multiple values for the same identifier name, it also removes the previous PK-level guarantee that only one value could exist for a given (ENTITY_ID, DEPLOYMENT_ID, NAME), so the write/update logic now needs to enforce the system-vs-schema precedence consistently.

In particular, I see two cases that may cause incorrect indexing:
- A system attribute name is marked as overriding the schema value before confirming that it actually produces a usable value. If the system value is null/empty, a valid schema identifier could therefore be silently dropped.
- The partial update paths only update one source at a time, so system and schema values for the same name can now coexist even though the documented precedence says the system value should override the schema value.

Also, if we make VALUE part of the PostgreSQL B-tree PK and is TEXT, very large identifier values can hit PostgreSQL's index row-size limit.

PR #5393, opened 2026-09-14
Replace the agent create wizard with a flow-driven onboarding experience

Dilusha-Madushan:
### Purpose

The console created agents through a hand-rolled three step wizard (`AgentCreatePage`) that duplicated orchestration the flow engine already owns: step sequencing, attribute collection from the agent type schema, uniqueness checks and error routing. None of it was configurable, so a deployment could not add a step, reorder prompts or change validation without a code change.

This replaces that wizard with a flow-driven onboarding experience, so agent creation is defined by a flow definition in the same way user onboarding already is.

---

Agent creation in the console now runs through an administration flow resolved from the `flow.agentOnboardingFlow.defaultHandle` server configuration. Unlike user creation there is no non-flow fallback: the direct `POST /agents` path used by the old wizard has been removed from the console.

#### 💥 Impact

Deployments that provision their resources declaratively and do not include the new `default-agent-onboarding-flow` will find agent creation unavailable in the console. The page reports that the flow could not be resolved rather than failing silently, and the Agents list and
edit pages are unaffected.

Deployments using the shipped bootstrap resources are unaffected, as the flow and its server configuration entry are added there.

#### 🔄 Migration Guide

Add the agent onboarding flow to your declarative resource set and reference it from the server
configuration:

```yaml
# server configuration
flow:
  agentOnboardingFlow:
    defaultHandle: default-agent-onboarding-flow
```

---

### Approach

**Reuse the user onboarding pattern rather than inventing one.** `AgentOnboardPage` follows `UserCreatePage`: it resolves the configured flow, renders the components the engine returns, and posts actions back. This keeps the two onboarding experiences behaving alike.

**Share entity type resolution across categories.** Resolving which type to provision into (list, filter by allowed types, filter by organization unit, then auto select one or prompt) is identical for users and agents. That logic moved to `entity_type_resolver.go`, parameterised on `entitytype.TypeCategory`. `UserTypeResolver` and the new `AgentTypeResolver` delegate to it, so the user path renders exactly as before.

**Make the shared executors category aware through the `mode` node property.** `ProvisioningExecutor` and `AttributeUniquenessValidator` read the entity category from `mode`, and the flow error messages they raise are parameterised with `{{param(entity)}}` so a conflict on an agent reports "agent" rather than "user". `AttributeUniquenessValidator` ships as a single catalog entry, so its `mode` is exposed in the flow builder as a two value selection rather than a free text property.

**Two new executors.** `AgentTypeResolver` resolves the agent type and its organization unit. `OwnerResolver` validates a selected owner through the entity provider with an explicit category check, and offers a `USER_SELECT` input where the client fetches the candidates.

**Resolve the provisioning target from application configuration.** Where the target is not named by a resolver node, an agent's type now comes from the application's `allowedAgentTypes`, falling back to the default agent type when none is declared. These candidates are deliberately not narrowed by `allowSelfRegistration` the way the user candidates are, because an agent is provisioned on an administrator's behalf rather than by the entity itself.

**Default the agent logo in the provider.** The old wizard set a default avatar in the request body before calling the API. Since the flow collects no logo, that default moved into `defaultAgentMgtProvider` so an agent created through any path has something to render in a listing.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/4021

### Related PRs
- https://github.com/thunder-id/thunderid/pull/5200

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

https://github.com/user-attachments/assets/bbeb6339-6b04-4ce5-a5b8-f69926ccb142

## Summary by CodeRabbit

* **New Features**
  * Added flow-driven agent onboarding with organization unit, owner, agent type, details, and delegation steps.
  * Agent creation now displays client credentials, supports copying and revealing secrets, and offers “add another” actions.
  * Added agent provisioning with generated OAuth settings, logos, and optional model details.
  * Added onboarding templates, widgets, and flow-editor configuration options.
  * Schema enum values now populate selectable inputs in their declared order.

* **Bug Fixes**
  * Improved validation and error messaging for duplicate attributes, invalid owners, entity types, and organization units.
  * Added clearer handling when agent onboarding is unavailable or misconfigured.


backend/internal/agentmgtprovider/default_agent_mgt_provider.go
```diff
@@ -9,39 +9,135 @@ import (
...
+) []string {
+	if len(agent.AllowedUserTypes) > 0 || !delegated || p.entityTypeSvc == nil {
+		return agent.AllowedUserTypes
+	}
+
+	// Two is enough to tell "exactly one" from "more than one".
+	list, svcErr := p.entityTypeSvc.GetEntityTypeList(
+		security.WithRuntimeContext(ctx), entitytype.TypeCategoryUser, 2, 0, false)
+	if svcErr != nil || list == nil || len(list.Types) != 1 {
+		return nil
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check this with latest

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -658,48 +793,164 @@ func (p *provisioningExecutor) createUserInStore(nodeCtx *providers.NodeContext,
...
+			log.MaskedString(log.LoggerKeyEntityID, createdAgent.ID))
+	}
+
+	return createdAgent, nil
+}
+
+// hasRecordValues reports whether the flow collected any field the category carries on its record
+// rather than in its schema. A user has none, every user value being a schema attribute; an agent
+// has its name, description and logo.
+func hasRecordValues(ctx *providers.NodeContext, category entitytype.TypeCategory) bool {
```
thiva-k: This only checks name/description/logo, but owner/delegated/redirectUris are the same kind of record-level field (per the comment above `ownerKey`) - should they count here too?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
```diff
@@ -9,39 +9,151 @@ import (
...
+		GrantTypes:              []providers.GrantType{providers.GrantTypeClientCredentials},
+		TokenEndpointAuthMethod: providers.TokenEndpointAuthMethodClientSecretBasic,
+	}
+	if delegated {
+		config.GrantTypes = append(config.GrantTypes,
+			providers.GrantTypeAuthorizationCode, providers.GrantTypeRefreshToken)
+		config.ResponseTypes = []providers.ResponseType{providers.ResponseTypeCode}
+		config.RedirectURIs = redirectURIs
+		if len(config.RedirectURIs) == 0 {
+			config.RedirectURIs = []string{defaultDelegatedRedirectURI}
```
thiva-k: We silently fall back to `localhost:5173/agent-callback` when delegation is on but no redirect URI is given - should we reject instead of defaulting to a local URL for a real client?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
```diff
@@ -9,39 +9,151 @@ import (
...
 }
 
+// resolveAllowedUserTypes picks the user types a delegated agent may act for. A caller that named
+// none gets the first deployed user type, which lets the agent service derive the agent's token
+// attributes from that schema. Deriving them is only possible at creation, so defaulting to one
+// type is preferred over leaving the agent with no token attributes it can never regain.
+//
+// The listing is ordered by name, so the default is the first alphabetically rather than one chosen
+// for this agent. It is a starting point a caller replaces by naming the types itself.
+func (p *defaultAgentMgtProvider) resolveAllowedUserTypes(
```
thiva-k: Defaulting to the first user type alphabetically feels arbitrary for a delegated agent, and the delegation widget doesn't even expose a way to pick one. Was this intentional?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
```diff
@@ -9,39 +9,151 @@ import (
...
 	}
 }
 
+// SetAgentService injects the agent service. See AgentMgtProviderService.
+func (p *defaultAgentMgtProvider) SetAgentService(agentSvc agent.AgentServiceInterface) {
+	p.agentSvc = agentSvc
+}
+
 // CreateAgent provisions an agent through the agent service.
 func (p *defaultAgentMgtProvider) CreateAgent(
```
thiva-k: We used to reject agent creation with no owner, now it's just forwarded. With `OwnerResolver` on `FlowTypeRegistration` too, can't this create agents with an empty owner?

backend/internal/flow/executor/entity_type_resolver.go
```diff
@@ -0,0 +1,297 @@
...
+			filtered = append(filtered, entityType)
+		}
+	}
+
+	return filtered
+}
+
+// filterTypesByOU keeps the types whose own organization unit is an ancestor of, or equal to, the
+// selected one.
+func (r *entityTypeResolution) filterTypesByOU(ctx *providers.NodeContext,
```
thiva-k: This does one `IsParent` store call per candidate type, which is N ancestry walks for N types. Might be worth batching if the type list ever gets large.

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -815,45 +1068,128 @@ func splitTrimmed(s string) []string {
...
 	}
+	return types, nil
+}
 
-	// Fail if no user types have self-registration enabled
-	if len(selfRegEnabledSchemas) == 0 {
-		logger.Debug(ctx.Context, "No user types with self-registration enabled, cannot provision automatically")
-		return nil, nil
+// groupMemberTypeFor returns the group member type that represents an entity of the category.
+func groupMemberTypeFor(category entitytype.TypeCategory) group.MemberType {
```
thiva-k: We've got several separate if-agent-else-user helpers here instead of one lookup table, even though `attribute_uniqueness_validator.go` shows the generic pattern. Might be worth consolidating before a third category shows up.

backend/internal/flow/executor/provisioning_executor.go
thiva-k: Shouldn't owner/delegated/redirect uris be part of this too?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
thiva-k: We shouldn't silently fall back to a hardcoded url when delegation is on but no redirect URI is given. Shall we validate it in provisioning executor and get it from user input always?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
thiva-k: Defaulting to the first user type alphabetically seems wrong. Can we get this an user input from the new widget?

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
thiva-k: We used to reject agent creation with no owner, now it's just forwarded. With `OwnerResolver` on `FlowTypeRegistration` too, can't this create agents with an empty owner?

backend/internal/flow/executor/entity_type_resolver.go
thiva-k: This does one `IsParent` store call per candidate type, which is N ancestry walks for N types. Might be worth batching if the type list ever gets large.

backend/internal/flow/executor/provisioning_executor.go
thiva-k: Seems we added several separate if-else helpers here. Can we check of we can improve this to have one lookup table or something similar?

backend/internal/flow/executor/provisioning_executor.go
```diff
@@ -658,48 +793,164 @@ func (p *provisioningExecutor) createUserInStore(nodeCtx *providers.NodeContext,
...
+// way a boolean schema attribute is, accepting the casings and 1/0 forms a hand-authored flow may
+// carry. An absent or unparseable value is false.
+func collectedFlag(ctx *providers.NodeContext, key string) bool {
+	parsed, err := strconv.ParseBool(collectedValue(ctx, key))
+	return err == nil && parsed
+}
+
+// collectedValue reads a value the flow collected for key, preferring what the user just submitted
+// over what an earlier node left in runtime data.
+func collectedValue(ctx *providers.NodeContext, key string) string {
```
thiva-k, line 891: Shouldn't certain values be collected only from the runtimedata set by other executors and not user input? e.g. owner.

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
```diff
@@ -9,39 +9,151 @@ import (
...
-	ctx context.Context, agent *providers.Agent,
+	ctx context.Context, agent *providers.Agent, delegated bool,
 ) (*providers.Agent, *tidcommon.ServiceError) {
 	if agent == nil {
 		return nil, &ErrorInvalidRequestFormat
 	}
-	if agent.Owner == "" {
```
thiva-k, line 35: Why have we removed this check? Is this check being done somewhere else now?

Dilusha-Madushan: This check already exist in agent service.

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
Dilusha-Madushan: The initial plan was to support agents only in the “Act as Own” mode and introduce the Delegate mode later. That’s why most of the current configurations are hard-coded to mimic the template behavior.

However, I agree with your point. I’ll remove the hard-coded values for now and add a widget for this configuration, similar to the owner selection.

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: Yes, this is a valid concern. Instead of adding condition checks for specific attribute keys, we can have the executors write the resolved data into the runtime data using dedicated keys, rather than reusing the same keys as userInputs.

When reading the values, we can also give priority to the runtime data over `userInputs`. I think this approach should address the issue cleanly.

backend/internal/agentmgtprovider/default_agent_mgt_provider.go
Dilusha-Madushan: Yes, the issue I had previously was with adding conditional inputs—for example, making the redirect URI required when the user selects Delegate mode and optional otherwise. However, the current flow input structure doesn’t support this kind of conditional behavior, so I made it optional and provided a default value.

That said, I agree with your comment, so I’ll remove the default values and add non schema attribute checking in provisioning executor to get the redirect URI base on delegate mode input from user.

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: Yes, this is some missing part. I added those attributes in here and remove the description and logo_url for agent case since those are not direct attributed of the entity.

backend/internal/flow/executor/provisioning_executor.go
Dilusha-Madushan: `var categoryTraitsByCategory = map[entitytype.TypeCategory]categoryTraits{
	entitytype.TypeCategoryUser: {
		groupMemberType:                 group.MemberTypeUser,
		roleAssigneeType:                role.AssigneeTypeUser,
		attributeConflictCode:           user.ErrorAttributeConflict.Code,
		autoProvisionedAtAuthentication: true,
	},
	entitytype.TypeCategoryAgent: {
		groupMemberType:       group.MemberTypeAgent,
		roleAssigneeType:      role.AssigneeTypeAgent,
		attributeConflictCode: agentAttributeConflictCode,
		recordFields:          []string{nameKey, ownerIDKey, delegatedKey, redirectURIsKey},
		missingRecordFields:   missingAgentRecordFields,
	},
}
`
Follows the existing idpPropertyConfigs pattern in internal/idp/constants.go. Six helpers removed (groupMemberTypeFor, roleAssigneeTypeFor, attributeConflictCodeFor, recordFieldKeys, delegatedRecordFieldKeys, typeInputKey), plus two more category checks around auth-time auto-provisioning folded into one trait. Adding a category is now one map entry.

thiva-k:
IMO better to have a separate provisioning executor for agents. Having it in the existing user provisioning executor seemed to have made it more complex and many not be maintainable for new requirements/new entity categories. WDYT @ThaminduDilshan @senthalan ?

backend/internal/flow/executor/error_constants.go
```diff
@@ -744,11 +754,11 @@ var (
 		Code: "FET-1053",
 		Error: tidcommon.I18nMessage{
 			Key:          "flows.executor.errors.no_valid_user_types",
-			DefaultValue: "No valid user types available",
+			DefaultValue: "No valid {{param(entity)}} types available",
```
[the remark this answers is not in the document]
Dilusha-Madushan: @coderabbitai check this now

backend/internal/flow/executor/category_traits.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2026 The ThunderID Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package executor
```
thiva-k, line 4: Let's have only executor files here. Can we move this somewhere else appropriate?

backend/internal/flow/executor/identifying_executor.go
```diff
@@ -95,31 +96,31 @@ func (i *identifyingExecutor) IdentifyUser(ctx context.Context, filters map[stri
...
 	}
 
-	userID, err := i.entityProvider.IdentifyEntity(searchableFilter)
+	entityID, err := i.entityProvider.IdentifyEntity(searchableFilter)
 	if err != nil {
 		switch err.Code {
 		case entityprovider.ErrorCodeEntityNotFound:
-			logger.Debug(ctx, "User not found for the provided filters")
+			logger.Debug(ctx, "Entity not found for the provided filters")
 			execResp.Error = &ErrUserNotFound
```
thiva-k: These still say "user" but this runs on the agent path too now. Should we route them through errForEntityCategory like the rest?

backend/internal/flow/executor/category_traits.go
Dilusha-Madushan: Currently, we already have some non-executor helper code files, such as attribute_collector.go and credential_setter.go, within the executor section. Since these are helpers and not registered executors, moving them to utils.go would make that file quite lengthy.

Given the existing structure, what would be the preferred place to keep these helpers?

backend/internal/flow/executor/identifying_executor.go
Dilusha-Madushan: I notice this but didn't changed and planned to bring that fix in separate PR. But since you point out this I thought to add that fix also in here, and it went wider than the one line. Six sites inside IdentifyEntity set user-named errors, plus one in resolveExistingEntity, and all of them are reachable on the agent path via ProvisioningExecutor. Rather than word the messages generically, I reshaped ErrUserNotFound, ErrFailedToIdentifyUser and ErrAmbiguousUserIdentity to carry {{param(entity)}} and renamed them to ErrEntityNotFound, ErrFailedToIdentifyEntity and ErrAmbiguousEntityIdentity, matching how ErrEntityAuthFailed and the other category-aware errors are already written, and gave IdentifyEntity a required category parameter so they route through errForEntityCategory like the rest. FET-1001/1002/1003 are unchanged, since both the CIBA mapping and the console map on code. Callers that resolve a category pass it, so an agent lookup now reads "The agent could not be found"; the paths that are not scoped to one kind of entity pass a named categoryUnscoped constant rather than an implicit default, so it is visible at the call site which ones are still assuming a user. 

Two things worth flagging: adding the placeholder turned twelve surviving &ErrXxx references into placeholder leaks, so those are routed through the helper too; and mapFlowErrorToCIBAError was switching on the literal message text, which this would have broken silently, so it now matches on the error code instead, which also fixes a case that was already dead. Four of the six messages still render byte-identically for user flows; the other two changed only because the originals start with a capital "U" on the word being parameterised and TypeCategory is lowercase. Added TestIdentifyEntity_ErrorNamesTheCategory to cover both nouns.

backend/internal/flow/executor/category_traits.go
thiva-k: This should be fine

thiva-k (dismissed):

sahandilshan (approved):

PR #5406, opened 2026-09-15
Revoke lost access through role, group and scope administration flows

indeewari:
### Purpose

Removing a role assignment, deleting a role or group, changing what a role grants, or retiring a scope
all change what the *next* token carries and nothing about the tokens already issued. Until those
tokens expire, the principal keeps exercising access the change was meant to withdraw.

This adds two revocation dimensions for that gap and carries each of those changes out through an
ADMINISTRATION flow, so the revocation and the change run as one orchestrated sequence, the shape
application deletion already uses.

A token carries only `sub`, `aud` and `scope`: never a role id, a group id or a catalogue key. So the
dimensions are keyed on what a token actually holds:

- `entity.scope` denies one scope for one principal on one resource server
- `scope` denies one scope for every principal, for when the scope itself stops existing

Both are keyed by audience as well as scope, because a permission string is unique only within its
resource server: two servers can each define `license` and they are different scopes.

Six new flows, each following the configured-and-present rule the other administration handles do:
role assignment removal, role deletion, role permission change, group deletion, group membership
removal, and scope deletion. The client secret regeneration flow gains the same re-stamp node, which
it needed for the same reason.

### Approach

**One shape, six flows.** Every flow is `start → permission_validator → pre_<action> → revoke_scopes →
<action> → restamp_revocation → end`. Only nodes 3 and 5 vary. Node 3 is the only one that reads
caller input; node 5 reads only the trusted plan, so a caller cannot point the action at a different
target than the one that was revoked for.

**Revoke the whole path being cut, not the exact delta.** Computing what the principal still holds
afterwards would mean resolving a hypothetical post-change state, and a resolver disagreement there
under-revokes, which is the unsafe failure. The cost is one forced refresh for a principal who keeps a
scope by another route, and the boundary cutoff makes that self-healing.

**Close the window with a re-stamp node.** The cutoff is stamped before the action, which is the
ordering that keeps a failed action safe, but the grant is still held until the action commits, so a
token minted in between has an `iat` past the cutoff. Rewriting the same criteria afterwards advances
it. The write is the same idempotent upsert and only ever moves a boundary row forward, so the two
writes collapse to one row.

**Digest the parts rather than storing them.** `CRITERION_VALUE` is `VARCHAR(255)` while an identifier
is `VARCHAR(2048)` and a permission `VARCHAR(1000)`; the triple does not fit and the column cannot be
widened without exceeding the btree key limit on the lookup it backs. Every match here is an equality
test, so a SHA-256 digest loses nothing. Each part is length-prefixed before hashing so two different
triples cannot render alike, which a plain join would not guarantee: a resource server identifier has
no character allowlist, so `("a|b", "c")` and `("a", "b|c")` would otherwise collide.

**Refuse a change that would revoke without changing anything.** Removing an assignment the principal
never held, or a member the group never had, is a silent no-op in the store, so validating only the
ids would deny that principal every scope the path conveys, change nothing, report success, and leave
nothing to restore those tokens.

**Cap the fan-out.** Principals × scopes above `oauth.revocation.criteria.max_criteria` (10,000 by
default) refuses the whole change before writing anything. A half-written plan leaves some principals
revoked and others not, with no record of where it stopped.

**Write the plan's rows in one round trip.** A change can carry up to `max_criteria` rows, so the
criteria node batches them into as few `INSERT ... VALUES ... ON CONFLICT` statements as the row count
needs rather than paying a round trip per row. Every entry is validated before any row is written, and
a repeated `(type, value)` within one call keeps only its last occurrence, since one statement cannot
affect the same conflict target twice. `RevokeByCriteria` is unchanged for every existing caller — the
refresh grant, the RFC 7009 endpoint, the authorization service and session sign-out.

#### Scope and limits worth stating

- **Console-only.** `DELETE /roles/{id}` and the other native endpoints are unchanged and still revoke
  nothing. This is defence in depth for console operators, not an API-level guarantee.
- **"One forced refresh"** holds for a browser client with a live session. For an `offline_access`,
  mobile, machine or agent client with no session to fall back on, it is a full re-authentication.
- **Applications that map `sub`.** The dimension keys on the token's `sub`. An application that
  configures `subjectAttribute` (API/declarative only, no console exposure) issues tokens whose `sub`
  is a mapped attribute rather than the entity id, and those tokens will not match. This is inherited
  from the existing `subject` dimension behind user deletion and session revocation rather than
  introduced here, and is tracked separately.

### Related Issues
- Design discussion: thunder-id/thunderid#5210

### Related PRs
- Builds on #5179, which established the administration-flow shape for application deletion and client
  secret regeneration.

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added administration flows for role assignments, roles, permissions, groups, memberships, and scopes.
  * Management changes can revoke affected tokens before taking effect.
  * Scope revocation now supports individual principals and resource-server-wide scopes.
  * Added flow configuration options and console visibility for these operations.
  * Added safeguards for oversized revocation operations and protection of unrelated access.
  * Improved bulk processing for criteria-based revocations.

* **Bug Fixes**
  * Improved token audience handling for access and refresh tokens during scope revocation.


backend/internal/flow/executor/pre_scope_revocation_executor.go
```diff
@@ -0,0 +1,401 @@
...
+		},
+		targetInput: revocationInputRole,
+		argInputs:   []string{revocationInputPermissions},
+		notAllowed:  ErrRolePermissionRemovalNotAllowed,
+	}, maxCriteria, func(ctx context.Context, inputs map[string]string) (
+		*revocation.ScopeRevocationTarget, *tidcommon.ServiceError) {
+		// Parse the permission set here, before anything is revoked. The acting node decodes it again
+		// to use it, but by then the deny rows are written: a malformed payload would leave every
+		// holder's tokens denied and the role unchanged, with nothing to restore them.
+		if _, err := decodeRolePermissions(inputs[revocationInputPermissions]); err != nil {
```
ThaminduDilshan: This only checks the JSON shape. `UpdateRoleWithPermissions` still runs `validatePermissions` and `CanGrantPermissions`, which can refuse after all holders' scopes are already revoked, right? Shall we run those checks here before revoking?

backend/internal/role/admin_provider.go
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	losing := memberID
+	if losing == "" {
+		losing = groupID
+	} else {
+		// Refuse a principal that is not in the group. The removal itself is a silent no-op for a
+		// non-member, so without this check the flow would deny that principal every scope the group
+		// conveys while changing nothing, and report success. Nothing would restore those tokens.
+		isMember, svcErr := p.isDirectMember(ctx, groupID, memberID)
```
ThaminduDilshan: `RemoveGroupMembers` also checks `CanGrantMembership`, but that only runs in the acting node after the member's scopes are revoked. Shall we check it here as well?

backend/internal/system/security/service.go
```diff
@@ -121,6 +126,8 @@ func (s *securityService) Process(r *http.Request) (context.Context, error) {
 			TokenFamilyID: securityCtx.tokenFamilyID,
 			Subject:       securityCtx.revocationSubject,
 			AppKey:        securityCtx.revocationAppKey,
+			Audience:      securityCtx.revocationAudience,
+			Scopes:        securityCtx.permissions,
```
ThaminduDilshan: `BearerAuthenticator.Authenticate` doesn't pass `Audience` and `Scopes`, so the MCP endpoint doesn't enforce the new dimensions, right? Shall we add them there as well?

backend/internal/flow/executor/register.go
```diff
@@ -426,3 +490,46 @@ type applicationAdminProvider interface {
...
+type resourceAdminProvider interface {
+	ValidateDeleteAction(ctx context.Context, resourceServerID, resourceID, actionID string) (
+		*revocation.ScopeRevocationTarget, *tidcommon.ServiceError)
+	DeleteAction(ctx context.Context, resourceServerID, resourceID, actionID string) *tidcommon.ServiceError
+}
+
+// configuredMaxRevocationCriteria returns the configured fan-out ceiling for one administrative
+// revocation. A non-positive value, including when runtime config is not loaded, leaves the
+// preparatory executors on their built-in ceiling.
+func configuredMaxRevocationCriteria() int {
```
ThaminduDilshan: This reads the server runtime from the executor package and brings back the `IsServerRuntimeInitialized` guard we removed from the application package in #5179. Shall we pass this through the flow `Config` from `FromServerRuntime()`? Then we can remove `defaultMaxRevocationCriteria` as well, since `default.json` already sets it

backend/internal/flow/executor/criteria_revocation_executor.go
```diff
@@ -16,18 +16,43 @@ import (
...
+// window carries the scopes and its iat is past the cutoff. Rewriting the same criteria afterwards
+// moves the cutoff forward and sweeps those up. The write is the same idempotent upsert, and it only
+// ever advances a boundary row, so running it changes nothing else.
+func newCriteriaRevocationRestampExecutor(factory core.FlowFactoryInterface,
+	revoker revocation.CriteriaRevoker) *criteriaRevocationExecutor {
+	return newCriteriaRevocationExecutorFor(factory, revoker,
+		ExecutorNameCriteriaRevocationRestamp, true)
+}
+
+// newCriteriaRevocationExecutorFor builds one of the two registered names. They are separate names
```
ThaminduDilshan: Can't we use executor mode for this rather than registering a separate executor? It's the same executor behaving differently at runtime. `executors.json` already lists `ApplicationActionValidator` twice with different modes, so the palette can still show both

backend/internal/flow/executor/constants.go
```diff
@@ -37,11 +37,24 @@ const (
 	ExecutorNameOTPExecutor                  = "OTPExecutor"
 	ExecutorNamePreDelete                    = "PreDeleteExecutor"
 	ExecutorNameCriteriaRevocation           = "CriteriaRevocationExecutor"
+	ExecutorNameCriteriaRevocationRestamp    = "CriteriaRevocationRestampExecutor"
 	ExecutorNameSessionRevocation            = "SessionRevocationExecutor"
 	ExecutorNameUserDelete                   = "UserDeleteExecutor"
 	ExecutorNameApplicationActionValidator   = "ApplicationActionValidator"
 	ExecutorNameApplicationDelete            = "ApplicationDeleteExecutor"
 	ExecutorNameClientSecret                 = "ClientSecretExecutor"
+	ExecutorNamePreRoleAssignmentRemoval     = "PreRoleAssignmentRemovalExecutor"
```
ThaminduDilshan: Same as #5179, shall we drop the `Pre` prefix and the `Executor` part from these names? I.e. `RoleAssignmentRemovalValidator`, similar to `ApplicationActionValidator`

backend/internal/role/admin_provider.go
```diff
@@ -0,0 +1,603 @@
...
+		return nil, svcErr
+	}
+
+	return &revocation.ScopeRevocationTarget{EntityIDs: entityIDs, Scopes: scopes}, nil
+}
+
+// DeleteRole deletes the role, which unassigns it from every principal holding it.
+func (p *adminProvider) DeleteRole(ctx context.Context, roleID string) *tidcommon.ServiceError {
+	if svcErr := p.roleService.DeleteRole(ctx, roleID); svcErr != nil {
+		p.logger.Error(ctx, "Failed to delete role", log.String("roleId", roleID),
```
ThaminduDilshan: `roleService.DeleteRole` already logs server errors, and a client error like `ErrorImmutableRole` gets logged as an error here. Shall we remove these logs and only return? Check other methods in this file as well

backend/internal/role/admin_provider.go
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	return &revocation.ScopeRevocationTarget{EntityIDs: entityIDs, Scopes: scopes}, nil
+}
+
+// isAssigned reports whether the assignee holds the role in its own right.
+//
+// Only a direct assignment counts, for the reason isDirectMember counts only direct membership: a
+// principal holding the role through a group keeps that path, and this removal would not cut it.
+func (p *adminProvider) isAssigned(ctx context.Context, roleID, assigneeID string) (
```
ThaminduDilshan: Shall we order these methods as in the interface, with public methods first and private helpers after? Same comment from #5179

backend/internal/role/error_constants.go
```diff
@@ -184,6 +184,48 @@ var (
 				"Roles must be defined in declarative configuration files",
 		},
 	}
+	// ErrorGroupNestingTooDeep is returned when expanding a group's members descends further than the
+	// supported nesting depth. It is reported rather than truncated: a partial expansion would revoke
+	// for some members and silently miss others, and the caller could not tell which.
+	ErrorGroupNestingTooDeep = tidcommon.ServiceError{
+		Type: tidcommon.ClientErrorType,
+		Code: "ROL-1023",
```
ThaminduDilshan: Shall we keep error constants in ascending order? `ROL-1023` is placed before `ROL-1019` and `ROL-1020`. Also `ROL-1021` and `ROL-1022` are skipped

backend/internal/flow/executor/group_workflow_executors_test.go
```diff
@@ -0,0 +1,62 @@
...
+package executor
+
+import (
+	"github.com/stretchr/testify/mock"
+
+	"github.com/thunder-id/thunderid/internal/revocation"
+	"github.com/thunder-id/thunderid/pkg/thunderidengine/providers"
+)
+
+func (s *ScopeRevocationExecutorsTestSuite) TestGroupMembershipRemoval_RemovesTheMember() {
```
ThaminduDilshan: Same as #5179, shall we keep one test file per source file with its own suite? These tests reuse `ScopeRevocationExecutorsTestSuite` from `pre_scope_revocation_executor_test.go`. That file also tests the role acting executors, and `role_workflow_executors_test.go` tests the pre executor and the restamp executor

backend/internal/flow/config/config.go
```diff
@@ -38,6 +38,17 @@ type FlowSectionConfig struct {
 	UserDeletionFlow        FlowTypeConfig `json:"userDeletionFlow"`
 	ApplicationDeletionFlow FlowTypeConfig `json:"applicationDeletionFlow"`
 	SecretRegenerationFlow  FlowTypeConfig `json:"secretRegenerationFlow"`
+	// The administration flows for the authorization changes that must revoke before they act. Each
```
ThaminduDilshan: Shall we avoid adding this bulky comment? Same was raised for this file in #5179

backend/internal/flow/executor/constants.go
```diff
@@ -76,6 +89,26 @@ const (
 	userInputLoginHint         = "login_hint"
 	revocationInputSubject     = "subject"
 	revocationInputApplication = "targetApplicationId"
+	// revocationInputRole and revocationInputAssignee name the role and the principal a role
```
ThaminduDilshan: Do we need these comments? Same as #5179, let's avoid bulky comments. Check other places as well, this is there in almost every new file

backend/dbscripts/runtime_persistent/postgres.sql
```diff
@@ -20,8 +20,9 @@ CREATE UNIQUE INDEX idx_revoked_token_jti_deployment ON "REVOKED_TOKEN" (DEPLOYM
 CREATE INDEX idx_revoked_token_expiry_time ON "REVOKED_TOKEN" (EXPIRY_TIME);
 
 -- Table to store criteria-based (many-token) revocations: a generalized attribute deny list.
--- CRITERION_TYPE names the dimension ('token_family' today; subject/client/consent are future types)
--- and CRITERION_VALUE holds the revoked value (the tfid for 'token_family'). Part of the
+-- CRITERION_TYPE names the dimension ('token_family', 'subject', 'app.key', 'entity.scope',
```
ThaminduDilshan: Shall we revert this comment change? Listing the dimensions here means updating the schema script every time a dimension is added. Same concern as `query_constants.go` in #5179. Applies to `sqlite.sql` as well

Issue #5233, opened 2026-09-01
Support OIDC Back-Channel Logout


UdeshAthukorala:
## Problem

ThunderID currently supports **OpenID Connect RP-Initiated Logout 1.0**, allowing a Relying Party (RP) to request termination of an end-user session at ThunderID through the `end_session_endpoint`. Session termination today stops at ThunderID's own SSO session and the RP that initiated the request.

ThunderID does not provide a standards-based way to inform the other RPs that participated in that session. When a user has active sessions across multiple applications, ending the OpenID Provider (OP) session leaves local sessions at those applications active. The user appears to be signed out, but remains signed in wherever an RP maintains its own session, so coordinated Single Logout (SLO) is not achievable.

Browser-based propagation alone is not sufficient for this problem. It depends on the user agent successfully loading an iframe for every participating RP, and it fails silently when the browser is closed, when third-party cookies or iframes are blocked, or when logout is triggered outside a browser context such as an administrative session revocation or an expiring session.

ThunderID needs **OpenID Connect Back-Channel Logout 1.0**: direct server-to-server logout notifications, carrying a signed logout token, delivered from ThunderID to each participating RP without relying on the user's browser.

Front-Channel Logout is tracked separately and is out of scope here. Both mechanisms were originally proposed together in #5170; this issue covers the Back-Channel Logout half.

## Scope

This feature includes:

* supporting OpenID Connect Back-Channel Logout 1.0 as an OpenID Provider.
* allowing a client to register a back-channel logout endpoint and declare whether session information is required.
* tracking which RPs participated in a given OP session so that ThunderID can determine who must be notified.
* issuing an OIDC session identifier (`sid`) so that an RP can terminate the correct local session.
* generating and signing a standards-compliant `logout_token` for each notification.
* delivering logout notifications to registered endpoints, independently of the user agent.
* propagating logout when the OP session ends through RP-Initiated Logout or another supported termination path.
* advertising back-channel logout capabilities through OpenID Provider metadata.
* configuring back-channel logout for an application through the ThunderID Console and the management API.
* handling and recording delivery failures without blocking session termination.

This feature excludes:

* OpenID Connect Front-Channel Logout 1.0, tracked separately.
* OpenID Connect Session Management 1.0 iframe-based session status checking.
* SAML Single Logout.

## Actors

| Actor | Expected capability |
| --- | --- |
| End user | Initiates logout from an application or from ThunderID, and expects to be signed out of every participating application. |
[...]

PR #5409, opened 2026-09-15
Issue the OIDC sid claim in ID tokens and carry it across refresh

UdeshAthukorala:
### Purpose

ThunderID issues no `sid` claim, so a relying party has no identifier for the SSO session its ID token belongs to. This PR issues `sid` (OpenID Connect Front-Channel Logout 1.0 section 3, restated in Back-Channel Logout 1.0 section 2.4) as the first deliverable of #5233, per design discussion #5258.

- `sid` is the SSO session's `SESSION_ID`, one value per session, stable for its lifetime.
- It is not the session handle. The handle is the cookie value, a bearer credential; the id confers nothing.
- A flow without a Session node produces no `sid`. Nothing is synthesised.
- Access tokens never carry it.

### Approach

The value travels the path the token family id (`tfid`) already takes.

```mermaid
flowchart LR
    SE["Session node<br/>RuntimeData[ssoSessionId]"] --> AA["AuthAssert node<br/>assertion claim sid"]
    AA --> AC["Authorization code<br/>SessionID"]
    AC --> IDT["ID token<br/>sid"]
    AC --> RT["Refresh token<br/>sid"]
    RT --> IDT2["Refreshed ID token<br/>sid"]
    RT --> RT2["Rotated refresh token<br/>sid"]
```

- **Session node** publishes the session id on save (new `SaveCheckpointResult.SessionID`) and on load, where it comes from the session in force, not the replayed snapshot. The key is on the snapshot deny-list for the same reason `tfid` is.
- **AuthAssert node** stamps it as `sid` beside `tfid`. `sid` joins the reserved assertion claims so a schema attribute named `sid` cannot displace it.
- **Authorization service** reads it off the assertion into `AuthorizationCode.SessionID`. Unlike `tfid`, there is no fallback mint.
- **Token service** emits `sid` in the ID token when set, stores it on the refresh token, and restores it on validation.
- **Grant handlers** pass it to the ID token builder on both the code and refresh grants. It rides `TokenResponseDTO.SessionID` into refresh-token issuance, so `IssueRefreshToken` and its mock are unchanged.

Verified manually against a running server: two apps in one session share a `sid`, it survives refresh, differs after sign-out and re-login, matches the `SSO_SESSION` row, and is absent for a flow with no Session node.

### Related Issues
- Refs #5233

### Related PRs
- Design: #5258

### Checklist
    - [x] Unit Tests
    - [x] Integration Tests (`tests/integration/oauth/sso/sid_test.go`)

### Security checks

## Summary by CodeRabbit

- **New Features**
  - Added OIDC `sid` claims to ID tokens when an SSO session exists.
  - Preserved the same session identifier across authorization-code and refresh-token flows.
  - Prevented user attributes from overriding the authoritative `sid` claim.
  - Omitted `sid` when no SSO session was established.
  - Added integration coverage for shared sessions, refreshes, sign-out, and re-login.

- **Documentation**
  - Documented `sid` behavior in OpenID Connect and refresh-token guides.


backend/internal/oauth/oauth2/tokenservice/validator.go
```diff
@@ -167,6 +167,7 @@ func (tv *tokenValidator) ValidateRefreshToken(
 	actorSub, _ := extractStringClaim(claims, "act_sub")
 	jti, _ := extractStringClaim(claims, "jti")
 	tokenFamilyID, _ := extractStringClaim(claims, constants.ClaimTokenFamilyID)
+	sessionID, _ := extractStringClaim(claims, constants.ClaimSessionID)
```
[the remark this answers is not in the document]
UdeshAthukorala: Not applying this one. The refresh token is issued and signed by ThunderID, and the builder only ever writes `sid` as a string, so a present non-string value cannot come from a token that passes signature verification. The extraction deliberately mirrors `tfid` on the line above, which also ignores the extraction error. The strict presence check exists only for `dpop_jkt`, where silently dropping the value would remove sender-constraining; a missing `sid` only omits an optional claim.

backend/internal/flow/session/service.go
```diff
@@ -226,6 +228,7 @@ func (s *service) SaveCheckpoint(ctx context.Context, in SaveCheckpointInput) (S
 	s.logger.Debug(ctx, "Saved SSO checkpoint", log.String("checkpoint", in.Checkpoint))
 	return SaveCheckpointResult{
 		Handle:          target.HandleID,
+		SessionID:       target.SessionID,
 		Created:         created,
 		AuthenticatedAt: target.AuthenticatedAt,
```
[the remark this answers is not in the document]
UdeshAthukorala: This behaviour comes from #5404 (`AuthenticatedAt` on the result) and the earlier reuse-path touch logic; this PR only adds `SessionID` to the same return. On a failed touch the returned time matches what the session still stores, so `auth_time` stays consistent with later `max_age` checks. If the degrade-instead-of-fail choice should change, that belongs in a follow-up on #5404 rather than here.

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -186,6 +186,9 @@ var reservedAssertionClaims = map[string]bool{
 	oauth2const.ClaimSubjectID:     true,
 	oauth2const.ClaimSubjectType:   true,
 	oauth2const.ClaimCorrelationID: true,
+	// The session id addresses logout. A schema attribute named sid replacing it would point the ID
```
madurangasiriwardena: This comment does not express proper idea IMO

backend/internal/flow/executor/auth_assert_executor.go
```diff
@@ -313,6 +316,13 @@ func (a *authAssertExecutor) generateAuthAssertion(
 		jwtClaims[oauth2const.ClaimTokenFamilyID] = tokenFamilyID
 	}
 
+	// Carry the SSO session id (published by the Session node) so the authorization code, and in turn
```
madurangasiriwardena: Do we need to say anything about the protocol specific details here?

backend/internal/flow/executor/auth_assert_executor.go
UdeshAthukorala: Fixed in e744bf4

backend/internal/flow/executor/auth_assert_executor.go
UdeshAthukorala: Agreed, Fixed in e744bf4

backend/internal/oauth/oauth2/tokenservice/builder.go
```diff
@@ -596,5 +600,14 @@ func (tb *tokenBuilder) buildIDTokenClaims(ctx *IDTokenBuildContext) map[string]
 		claims[key] = value
 	}
 
+	// sid is owned by the session, never by the attribute merge above. Drop whatever an allow-listed
+	// user attribute of that name put there, then write the session id only when a session exists: a
+	// forged or synthesized sid would point a logout notification at a session the grant does not
+	// belong to, or promise one for a session that was never established.
+	delete(claims, constants.ClaimSessionID)
```
ThaminduDilshan: `sid` isn't in `ReservedAccessTokenClaimNames()` the way `tfid` is, so an app that allow-lists a user attribute named `sid` still gets it on the access token. The doc row we add says it's never present there. Shall we add it to the reserved set too?

backend/internal/flow/executor/session_executor.go
```diff
@@ -164,8 +170,13 @@ func (e *sessionExecutor) saveCheckpoint(ctx *providers.NodeContext, execResp *p
...
 	}
 	// The service declined the save because the freshly authenticated subject conflicts with the
-	// existing session's subject; degrade SSO without failing authentication.
+	// existing session's subject; degrade SSO without failing authentication. A session id carried
+	// forward from an earlier node names that other subject's session, so blank it: the engine merges
+	// RuntimeData by overwriting keys, so an empty value is what clears it, and the assertion treats
+	// empty as absent. The shipped check-then-load shape cannot reach this, because the SSO-Check node
+	// decides re-authentication before any load, but a graph that re-authenticates after a load can.
 	if result.Skipped {
+		execResp.RuntimeData[common.RuntimeKeySSOSessionID] = ""
```
ThaminduDilshan: The comment says the shipped shape cannot reach this. Do we need the blanking then? The handle carried on `RuntimeData` names the same other-subject session and isn't cleared either.

backend/internal/flow/session/model.go
```diff
@@ -27,7 +27,12 @@ const (
 // The write-once auth-event facts (completed steps + sanitized claim snapshot) live in the
 // sibling SessionContext (SESSION_AUTH_CONTEXT, 1:1 by session id), loaded only on the SSO path.
 type Session struct {
-	// SessionID is the internal primary key, never exposed to clients.
+	// SessionID is the primary key. It is never accepted from a client and no API takes it, but it is
+	// public: it is published to relying parties as the OIDC sid claim so an RP can name the session
+	// that ended, and every RP in the session sees the same value. Unlike HandleID it confers nothing:
```
ThaminduDilshan: Front-Channel Logout 1.0 section 2 defines `sid` as a session "at an RP", and `frontchannel_logout_session_supported` as identifying "the RP session with the OP". We issue one value per OP session instead. Two apps can already receive a different `sub` for the same user through `InboundClient.SubjectAttribute`, and a shared `sid` links them back together. I know #5258 settled on `SESSION_ID`. IMO worth revisiting before it's in the ID token for good. WDYT @UdeshAthukorala ?

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -283,6 +283,13 @@ const (
 	// family at once. Revocation-only and not a client-managed identifier: it rides the token JWTs
 	// but is not part of any client-facing API.
 	ClaimTokenFamilyID string = "tfid"
+	// ClaimSessionID identifies the SSO session an authentication belongs to. It is the session's id,
+	// minted when the session is established and stable for its lifetime, published to relying parties
+	// in the ID token as the OIDC sid claim (OpenID Connect Front-Channel Logout 1.0 section 3, restated
+	// in Back-Channel Logout 1.0 section 2.4) so an RP can name the session that ended. Absent when the
+	// login flow established no SSO session; it is never synthesized, because a sid no termination can
+	// reference would promise a logout notification that cannot come. Access tokens never carry it.
+	ClaimSessionID string = "sid"
```
ThaminduDilshan: Front-Channel Logout 1.0 section 3 conditions this on the OP advertising `frontchannel_logout_session_supported`, and Back-Channel 1.0 section 3 on `backchannel_logout_session_supported`. We advertise neither, so an RP can't discover the claim. Is that a later PR of #5233?

backend/internal/oauth/oauth2/tokenservice/builder.go
ThaminduDilshan: `sid` isn't in `ReservedAccessTokenClaimNames()` the way `tfid` is, so an app that allow-lists a user attribute named `sid` still gets it on the access token. The doc row we add says it's never present there. Shall we add it to the reserved set too?

backend/internal/flow/executor/session_executor.go
ThaminduDilshan: The comment says the shipped shape cannot reach this. Do we need the blanking then? The handle carried on `RuntimeData` names the same other-subject session and isn't cleared either.

backend/internal/flow/session/model.go
```diff
@@ -27,7 +27,9 @@ const (
 // The write-once auth-event facts (completed steps + sanitized claim snapshot) live in the
 // sibling SessionContext (SESSION_AUTH_CONTEXT, 1:1 by session id), loaded only on the SSO path.
 type Session struct {
-	// SessionID is the internal primary key, never exposed to clients.
+	// SessionID is the primary key and the session's public identifier: every application in the
```
ThaminduDilshan, line 30: Front-Channel Logout 1.0 section 2 defines `sid` as a session "at an RP", and `frontchannel_logout_session_supported` as identifying "the RP session with the OP". We issue one value per OP session instead. Two apps can already receive a different `sub` for the same user through `InboundClient.SubjectAttribute`, and a shared `sid` links them back together. I know #5258 settled on `SESSION_ID`. IMO worth revisiting before it's in the ID token for good. WDYT @UdeshAthukorala ?

backend/internal/oauth/oauth2/constants/constants.go
```diff
@@ -283,6 +283,9 @@ const (
 	// family at once. Revocation-only and not a client-managed identifier: it rides the token JWTs
 	// but is not part of any client-facing API.
 	ClaimTokenFamilyID string = "tfid"
+	// ClaimSessionID is the OIDC sid claim the id of the SSO session an authentication belongs to. Carried by ID and
+	// refresh tokens, never by access tokens, and omitted when no session was established.
+	ClaimSessionID string = "sid"
```
ThaminduDilshan, line 288: Front-Channel Logout 1.0 section 3 conditions this on the OP advertising `frontchannel_logout_session_supported`, and Back-Channel 1.0 section 3 on `backchannel_logout_session_supported`. We advertise neither, so an RP can't discover the claim. Is that a later PR of #5233?

docs/content/guides/protocols/oauth-oidc/openid-connect.mdx
```diff
@@ -45,6 +45,7 @@ The ID Token is a signed JWT delivered to the client at the token endpoint. It i
 | `auth_time` | When the user actually authenticated, included when available |
 | `nonce` | Echoed from the authorization request, clients **must** verify this |
 | `acr` | Authentication Context Class Reference, which authentication flow ran |
+| `sid` | Identifier of the SSO session the authentication belongs to. Applications that share one session receive the same value, and it stays the same across refresh. Omitted when the login flow established no session. Never present in access tokens |
```
ThaminduDilshan, line 48: "Never present in access tokens" doesn't hold yet. `sid` is missing from `ReservedAccessTokenClaimNames()` in `tokenservice/utils.go`, where `tfid` is listed, so the attribute merge in `buildAccessTokenClaims` still lets an app that allow-lists a user attribute named `sid` put one on its access token. Shall we add it there? The constant's own comment says the same thing.

backend/internal/flow/executor/session_executor.go
```diff
@@ -164,8 +170,11 @@ func (e *sessionExecutor) saveCheckpoint(ctx *providers.NodeContext, execResp *p
 		return err
 	}
 	// The service declined the save because the freshly authenticated subject conflicts with the
-	// existing session's subject; degrade SSO without failing authentication.
+	// existing session's subject; degrade SSO without failing authentication. A session id carried
+	// from an earlier node belongs to that other subject's session, so clear it. Setting it empty is
+	// what clears it, because the engine merges RuntimeData by overwriting.
 	if result.Skipped {
+		execResp.RuntimeData[common.RuntimeKeySSOSessionID] = ""
```
ThaminduDilshan, line 177: The handle carried on `RuntimeData` names the same other-subject session and isn't cleared with it. Shall we clear both here?

docs/content/guides/protocols/oauth-oidc/refresh-token.mdx
```diff
@@ -86,6 +86,7 @@ Access tokens issued earlier are unaffected and remain valid until they expire.
 | `exp` / `iat` | New values |
 | `cnf.jkt` (DPoP) | Preserved |
 | `auth_time` | Not carried into refreshed tokens |
+| `sid` (ID Token) | Preserved. A refreshed ID Token names the same SSO session as the original |
```
madurangasiriwardena, line 89: Bit confusing about the difference between Unchanged and Preserved. But that is an already existing issue.

docs/content/guides/protocols/oauth-oidc/openid-connect.mdx
madurangasiriwardena, line 48: Since this is about id tokens, we don't need to say "Never present in access tokens"

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
```diff
@@ -403,6 +408,8 @@ func (h *refreshTokenGrantHandler) IssueRefreshToken(
 		ClaimsLocales:        claimsLocales,
 		DPoPJkt:              dpopJktForRefresh(ctx, oauthApp),
 		TokenFamilyID:        tokenFamilyID,
+		// Taken from the response because both first issuance and rotation provide it there.
+		SessionID: tokenResponse.SessionID,
```
madurangasiriwardena, line 412: Can you check the indentation here?

backend/internal/oauth/oauth2/granthandlers/authorization_code.go
```diff
@@ -211,6 +212,7 @@ func (h *authorizationCodeGrantHandler) HandleGrant(ctx context.Context, tokenRe
 			ClaimsRequest:  authCode.ClaimsRequest,
 			Nonce:          authCode.Nonce,
 			CompletedACR:   authCode.CompletedACR,
+			SessionID:      authCode.SessionID,
```
thiva-k, line 215: We need to add this for the CIBA grant issued ID token too right?

backend/internal/oauth/oauth2/tokenservice/builder.go
```diff
@@ -596,5 +600,12 @@ func (tb *tokenBuilder) buildIDTokenClaims(ctx *IDTokenBuildContext) map[string]
 		claims[key] = value
 	}
 
+	// sid comes only from the session: drop any allow-listed user attribute of that name, then write
+	// the session id only when a session exists.
+	delete(claims, constants.ClaimSessionID)
```
thiva-k, line 605: Don't we have a list of reserved token claims shared with access token? If so shall we add it there

backend/internal/flow/session/model.go
UdeshAthukorala: Thanks for raising this. I looked at it from three angles.

1. On the spec. Section 2 does say "at an RP," but the same sentence goes on to say different sid values identify "distinct sessions at an OP," so honestly both readings are defensible. And in practice the shared reading is well-precedented: Keycloak, Duende IdentityServer, and Ory Hydra all hand the same OP session id to every client in the session.

2. On linkability as it stands. SubjectAttribute can only map sub to an attribute that's unique and required on the user type, such as email or username. Those are already real identifiers for the same person, so two apps can join their records straight through the directory without any help from sid. A shared sid doesn't add linkage that sub isn't already handing over.

3. On when a per-app sid would actually matter. Only once sub itself becomes pairwise. And that switch rewrites every application's sub, so RPs have to re-onboard their stored identities regardless, which means making sid per-application in that same change costs us nothing extra and gives RPs a single migration instead of two. Doing it now, by contrast, means carrying a per-participant id column on SSO_SESSION_PARTICIPANT from day one for a property sub doesn't even have yet.

So my opinion is to keep one sid per OP session in this PR, and that sid goes per-application at the same time as pairwise subjects, with the participant table as its home.

WDYT @ThaminduDilshan @coderabbitai ?

backend/internal/oauth/oauth2/constants/constants.go
UdeshAthukorala: Yes, this lands in the wiring PR. The spec defines backchannel_logout_session_supported as the OP being able to send sid in the logout token, and the ID token sid follows from that. So if we flip it to true now, before delivery actually exists, we'd be advertising notifications we can't send. That's why both back-channel discovery flags will go in together with the dispatcher and the oauth.logout.backchannel.enabled config, all tied to the same switch.

frontchannel_logout_session_supported is out of scope here. #5233 excludes front-channel logout, and we don't have a frontchannel_logout_uri to pair it with anyway.

backend/internal/flow/executor/session_executor.go
UdeshAthukorala: I left the handle alone on purpose. Unlike the session id, it never reaches a token or the browser. It's only read back by the Session node as the hint for a later join, where the subject check declines the other-subject session again, and the cookie only goes out through EngineData when a session is actually created. That's also exactly what this path did before this PR.

Clearing it would change the semantics. The next checkpoint would mint a new session for the new subject and replace the first subject's cookie while their session is still live. If that's the behaviour we want, I'd rather do it as its own change to the skipped-save logic than fold it into the sid PR.

WDYT? @ThaminduDilshan  @coderabbitai ?

docs/content/guides/protocols/oauth-oidc/openid-connect.mdx
UdeshAthukorala: Good catch. Added `sid` to `ReservedAccessTokenClaimNames()` next to `tfid` in 5fd8f9a, so the access token merge drops it the same way. The existing test that forges every builder-owned claim now covers it too.

docs/content/guides/protocols/oauth-oidc/refresh-token.mdx
UdeshAthukorala: Agreed, and it's an existing inconsistency in that table. For the row this PR adds I switched it to "Unchanged" in 5fd8f9a, since `sid` is copied as-is like `sub` and `iss`, while "Preserved" is used where a caveat follows.

docs/content/guides/protocols/oauth-oidc/openid-connect.mdx
UdeshAthukorala: Agreed, removed that sentence from the ID token row in 5fd8f9a.

backend/internal/oauth/oauth2/granthandlers/refresh_token.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/813ac1fd3fe8de450b3a7800b15d04ae5d4805b4..5fd8f9ad4b33f36e39af4b95d274db9b180e5c12

backend/internal/oauth/oauth2/granthandlers/authorization_code.go
UdeshAthukorala: Yes. Added it in 5fd8f9a: the CIBA callback now reads `sid` from the assertion, `MarkAuthenticated` stores it on the request record, and the CIBA grant passes it to the ID token and the token response so the refresh token carries it too, same as the authorization-code grant. There's also a new integration test that runs a CIBA request on a flow with a Session node and checks the ID token carries `sid`, and that a flow without one doesn't.

backend/internal/oauth/oauth2/tokenservice/builder.go
UdeshAthukorala: There wasn't one for ID tokens, and the access-token list has access-only claims (`grant_type`, `aci`, `cnf`, OU claims) so it can't be shared directly. In 5fd8f9a I added `builderOwnedIDTokenClaimNames()` (`auth_time`, `nonce`, `acr`, `sid`) next to the access-token one and used it in the ID-token merge instead of the `delete`. That also closes the same gap for `auth_time`, `nonce`, and `acr`, which an allow-listed attribute could previously override.

backend/internal/oauth/oauth2/tokenservice/utils.go
```diff
@@ -415,9 +415,21 @@ func ReservedAccessTokenClaimNames() map[string]bool {
...
 	reserved[constants.ClaimTokenFamilyID] = true
+	reserved[constants.ClaimSessionID] = true
 	return reserved
 }
 
+// builderOwnedIDTokenClaimNames returns the ID-token claims the builder writes itself, so a configured
+// attribute of the same name can never supply or replace one.
+func builderOwnedIDTokenClaimNames() map[string]bool {
+	return map[string]bool{
+		"auth_time":                 true,
```
ThaminduDilshan: We have `constants.ClaimAuthTime` for this. Shall we use it here, like `nonce` and `sid` below? A literal here and a literal in `buildIDTokenClaims` can drift apart and the claim silently stops being protected.

backend/internal/oauth/oauth2/granthandlers/ciba.go
```diff
@@ -249,6 +249,7 @@ func (h *cibaGrantHandler) issueTokens(ctx context.Context, record *ciba.CIBAAut
 
 	tokenResponse := &model.TokenResponseDTO{
 		AccessToken: *accessToken,
+		SessionID:   record.SessionID,
```
ThaminduDilshan: #5258 says CIBA carries no `sid`. Issuing it here looks right to me, but shall we correct the design doc so the back-channel PRs don't skip CIBA participants?

thiva-k (dismissed):

backend/internal/oauth/oauth2/tokenservice/utils.go
UdeshAthukorala: Fixed. Switched both the owned-claims list and buildIDTokenClaims to constants.ClaimAuthTime.

ThaminduDilshan (approved):

PR #5412, opened 2026-09-15
Add configurable audience validation for the API layer

### LINKED ISSUE #5427: Audience validation for REST and MCP is hardcoded and cannot be configured
### Current Limitation

#5272 unified MCP and REST authentication onto a single verification path but left the two surfaces with hardcoded, asymmetric audience behaviour:

- **MCP** requires `aud == <serverURL>/mcp`, derived at runtime.
- **REST** enforces no audience at all — a token minted for any resource server is accepted by the management API as long as it carries the required scope.

Neither is configurable by a deployment. A deployment that wants the management API to accept only tokens minted for it has no way to express that, and a deployment whose MCP resource identifier differs from its derived server URL has no way to correct it.

### Suggested Improvement

Make both configurable, with defaults that preserve current behaviour exactly:

```yaml
server:
  security:
    rest:
      # audience: "..."   # omitted -> audience NOT checked (REST authorizes by scope)
    mcp:
      # audience: "..."   # omitted -> derived from public_url; the check itself is mandatory
```

The asymmetry is intentional. The MCP authorization specification requires a server to verify tokens were issued for it, so there should be no toggle to disable that check — an absent value means "derive the audience", not "skip it". REST is not spec-bound here and authorizes by scope, so an absent value leaves the audience unchecked and lets deployments opt in rather than being locked out by an upgrade.

"Not configured" should never mean "not validated". An explicitly configured value must not be empty, and startup logs should state which behaviour is active for each surface so the resolved configuration is visible rather than inferred from a missing key.

One further constraint: a single variable should drive both MCP's audience guard and its published protected-resource metadata. If those diverge, a spec-compliant client reads the advertised resource, requests a token for it, and is then rejected by the guard — which presents as a broken client rather than a configuration error.

ImalshaD:
### Purpose

PR #5272 unified MCP and REST authentication onto a single verification path, but left the two surfaces with hardcoded, asymmetric audience behaviour:

* **MCP** required `aud == <serverURL>/mcp`, derived at runtime.
* **REST** enforced no audience at all — a token minted for any resource server was accepted by the management API as long as it carried the required scope.

Neither behaviour could be configured by a deployment. This PR makes both configurable, with defaults that preserve today's behaviour exactly.

```yaml
server:
  security:
    rest:
      # audience: "..."   # omitted → audience NOT checked (REST authorizes by scope)
    mcp:
      # audience: "..."   # omitted → derived from public_url; the check itself is mandatory
```

**No breaking changes:** omitting both keys reproduces the current behaviour on every surface.

### Approach

The behaviour is intentionally asymmetric because the two surfaces have different obligations.

* **MCP** — the MCP authorization specification requires a server to verify that tokens were issued for it, so there is deliberately no toggle to disable the audience check. An absent value means the audience is derived from the server's resource identifier rather than disabling the check.
* **REST** — REST is not spec-bound in this regard; authorization is by scope. An absent value leaves the audience unchecked, allowing deployments to opt in rather than being locked out by an upgrade.

"Not configured" never means "not validated." Both keys are strings. An absent key is valid and means what each surface defines above. An explicitly configured value must not be empty (`server.security.rest.audience` must not be empty; omit it to leave the audience unchecked) rather than silently leaving a gate unenforced.

Startup logs state which behaviour is active for each surface, making the resolved configuration visible in the logs instead of requiring it to be inferred from a missing configuration key.

### MCP Resource Identifier

One variable drives both MCP's audience guard and its published metadata.

`DefaultGuard` computes a single `mcpResourceIdentifier` and uses it both as the required audience and in the protected resource metadata. If these values diverge, a spec-compliant client reads the advertised resource, requests a token for it, and is then rejected by the guard — which presents as a broken client rather than a configuration error.

A test asserts both values from a single call, using a configured value unrelated to the server's own URL to ensure the configured value is actually being used.

### Installers

Helm exposes both keys under `configuration.server.security`. The MCP audience derives from `consoleClient.resourceIdentifier` (itself defaulting to `<publicUrl>/mcp`) via a new `thunderid.resourceIdentifier` helper, so a deployment never gains a separate `runtime.security.mcpAudience` configuration, following the existing `console.resourceIdentifier` pattern.

**Known gap:** `rest.audience` is intentionally not templated in OpenChoreo. An unset REST audience must omit the key entirely, and the OpenChoreo template cannot conditionally drop lines. Emitting `""` would fail startup, while emitting a derived value would silently turn enforcement on for every OpenChoreo deployment.

When required, the REST audience can be set via `configOverrides` by deploying the value inline.

### Related Issues

* Fixes #5427

### Related PRs

* Follows #5272 (unified MCP/REST bearer authentication), which introduced the shared authentication path this PR configures.

### Checklist

* [x] Followed the contribution guidelines.
* [x] Manual test round performed and verified.
* [x] Documentation provided. (`install/helm/README.md`, `inlinedeployment.yaml`, `install/helm/values.yaml`, `install/openchoreo/.../resourcetype.yaml`)

  * [ ] Ran Vale and fixed all errors and warnings — no documentation changes required.
* [x] Tests provided.

  * [x] Unit Tests
  * [ ] Integration Tests — see note below.
* [ ] Breaking changes. (Not applicable — defaults reproduce existing behaviour.)

  * [ ] Breaking changes section filled.
  * [ ] Breaking change label added.

### Security Checks

* [x] Followed secure coding standards in WSO2 Secure Coding Guidelines.
* [x] Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.

## Summary by CodeRabbit

* **New Features**
  * Added configurable audience validation for REST and MCP tokens.
  * REST audience checks are opt-in; when omitted, authorization continues to rely on scopes.
  * MCP audience validation remains mandatory and uses either a configured or automatically derived resource identifier.
  * Added REST and MCP audience settings across supported deployment methods, including environment-variable templates.
  * Added validation for audience formats, including whitespace normalization and invalid empty values.

* **Documentation**
  * Documented audience settings, defaults, validation rules, and configuration behavior in deployment and Helm references.


backend/cmd/server/main.go
```diff
@@ -106,6 +106,9 @@ func main() {
 	// Mount the MCP server's routes now that the revocation enforcer exists — DefaultGuard uses it
 	// to authenticate MCP requests with the same verification and revocation logic as the REST gate.
 	mcpGuard, mcpResourceMeta := mcp.DefaultGuard(jwtService, revocationEnforcer)
+	// A mismatched identifier surfaces as clients failing to authenticate, not as a config error.
+	logger.Info(ctx, "MCP resource identifier resolved",
```
rajithacharith: Do we need this info log?

backend/cmd/server/main.go
```diff
@@ -286,9 +289,21 @@ func createTLSListener(ctx context.Context, logger *log.Logger, server *http.Ser
...
-	middlewareFunc, err := security.Initialize(jwtService, revocationEnforcer)
+func createSecurityMiddleware(ctx context.Context, logger *log.Logger, cfg *config.Config,
+	mux *http.ServeMux, jwtService jwt.JWTServiceInterface,
+	revocationEnforcer revocationcache.EnforcerInterface) http.Handler {
+	// An unenforced gate must be visible in the logs, not inferred from a missing config key.
+	// Config validation already rejected an explicitly empty value, so a non-nil one enforces.
+	expectedAud := ""
+	if restAudience := cfg.Server.SecurityConfig.REST.Audience; restAudience != nil {
+		expectedAud = *restAudience
+		logger.Info(ctx, "REST audience validation is enabled", log.String("audience", expectedAud))
```
rajithacharith: Let's reword the log, 
And is it okay to say that this validation is there and value is this in the log

rajithacharith (approved):

PR #5419, opened 2026-09-16
Report an unresolvable user onboarding flow instead of redirecting

Dilusha-Madushan:
### Purpose

When the user onboarding flow cannot be resolved, the Add User page navigated to a "manual user creation" page. That page executed the same onboarding flow that had just been reported missing, and carried no handling for that failure.

The result was that a missing flow produced a silent redirect followed by a second, less specific failure. The fallback made the outcome worse than no fallback at all.

Three checks established that the fallback could never have worked:

- `UserCreatePage` contains no user-creation call. No `useCreateUser`, no POST to `/users`, no
  mutation. It renders the same `InviteUser` component, which pins `flowType: USER_ONBOARDING`.
- `UserCreatePage` contains no missing-flow handling. `isMissingOnboardingFlow` appears zero times.
- `UserCreatePage` was reachable only from this fallback. The **Add User** button calls
  `routes.add()`; `routes.addCreate()` had exactly one caller.

### Approach

**Report the condition in place rather than navigating away.** The Add User page now sets an unavailable state when the flow cannot be resolved, and returns early with a message naming the two things to check: that a user onboarding flow is configured, and that the flow it names exists.

The early return replaces the wizard rather than overlaying it, so no step remains interactive behind the message. A test asserts the step heading is gone, not merely covered.

**One state rather than two.** The agent onboarding page distinguishes `NOT_CONFIGURED` from `FLOW_MISSING`, because it resolves the configured handle itself before starting. The Add User page cannot: it detects the condition reactively from the flow execution error, and the backend returns a single code, `FLM-1003` "Flow not found", carrying no cause. Splitting the message would mean adding a server-config fetch to this page, which is new data-fetching behaviour rather than a fix, so the message covers both causes instead.

**Detection is unchanged.** `isMissingOnboardingFlow` already recognised the condition at both the `onError` and `onFlowChange` call sites. Only the response to it changed, so the matching logic keeps its existing coverage.

**Removal.** `UserCreatePage` and its test are deleted, along with the `index.ts` export, the lazy import and route in `App.tsx`, and `addCreate` from `RouteConfig.ts` and `useUserRoutes.ts`. Two stale references were cleaned up in passing: `App.test.tsx` mocked `../features/users/pages/UserCreatePage`, a path that stopped existing when the pages moved into the package, and `RouteConfig.test.ts` asserted the now-removed route.

**Tests.** Two existing tests asserted the redirect and were rewritten. Two were added: that the wizard is replaced rather than rendered alongside the message, and that an error which is *not* a missing flow leaves the wizard intact so a recoverable failure stays recoverable. The new tests were verified by mutation: removing the state assignment fails three of the four.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5413

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

### Security checks

## Summary by CodeRabbit

- **Bug Fixes**
  - User onboarding now shows a clear full-screen error when the configured flow cannot be found, instead of redirecting to manual user creation.
  - Other onboarding flow errors continue to display the wizard.

- **Changes**
  - Removed the separate manual user-creation page and route.
  - User creation is now handled through the standard user-add flow.

- **Documentation**
  - Added English translations for onboarding flow resolution errors.


Dilusha-Madushan:
image

this is how error shows now the flow is missing.

brionmario (approved):

PR #5444, opened 2026-09-17
Add the OIDC conformance test harness

Thumimku:
### Purpose

Runs ThunderID against the [OpenID Foundation conformance suite](https://gitlab.com/openid/conformance-suite)
nightly and on demand, so a regression in the OpenID Connect implementation is caught against
the specification rather than only against our own tests.

Passed Link: https://github.com/Thumimku/thunder/actions/runs/35217474856
screencapture-github-Thumimku-thunder-actions-runs-35217474856-2026-09-17-18_01_18

Conformance is asserted with the suite's own verdict rather than a local tally:
`POST /api/plan/{id}/certificationpackage`, the same check applied when preparing a
certification package. A module counts as failed only when it did not reach `FINISHED` or its
result was `FAILED`/`UNKNOWN`, so `PASSED`, `WARNING`, `REVIEW` and `SKIPPED` all pass.

Current state, measured:

| Profile | Plan | Modules | Result |
| --- | --- | --- | --- |
| `basic` | `oidcc-basic-certification-test-plan` | 35 | 35/35, certification-clean |
| `config` | `oidcc-config-certification-test-plan` | 1 | passes |

### Approach

**Profiles are configuration.** The `profiles` block of `plan-config.json` names each supported
plan and whether it needs a browser; the workflow offers the same names as a dropdown. Adding a
profile is an entry in that file plus one workflow option. `--plan <name>` still runs a plan the
file does not list. The block is stripped before the config is posted, since the suite rejects
fields it does not define.

**Clients are provisioned up front** (`client_registration=static_client`), so the suite does
not register and delete clients mid-run and no registration access token is needed. Three are
registered: two distinct relying parties, plus one using `client_secret_post`, since a client
may advertise only one token endpoint auth method.

**Clients are bound to an SSO-aware flow.** Dynamic client registration carries no flow
selection, so a registered client runs the deployment's default authentication flow. When that
flow does not consult an existing session, `prompt=none`, `id_token_hint` and `max_age` cannot
be answered and their modules fail. The harness creates its own flow from `sso-flow.json` and
binds each client to it through the application API. Both steps are idempotent, and
`--no-sso-flow` leaves the clients on the default flow.

**The login step is driven by Playwright.** The suite drives front-channel interaction with
HtmlUnit, which does not execute ES modules, and ThunderID's login page is a client-rendered
React app whose form is described by the flow at runtime. There is no static form to fall back
to, so `run_tests.py` polls the suite for pending URLs, signs in through Chromium, and reports
each URL back. The suite scores the module exactly as it would have.

**Screenshot-gated modules are completed.** Several modules end by asking a human to upload a
screenshot as certification evidence. Their assertions have already run at that point, so the
harness fills the placeholder and the module reaches `FINISHED` with a `REVIEW` result instead
of parking in `WAITING` until the timeout. A real certification *submission* still needs genuine
screenshots.

Two smaller points:

- `authorization_grant_profiles_supported` is declared in `allow_unexpected_metadata_fields`.
  It is a deliberate extension advertising the ID-JAG profile, and leaving the warning standing
  would mask a genuinely unexpected metadata field later.
- `patch_compose.py` converts mongo's `./mongo/data` bind mount to a named volume. The bind
  mount leaves a root-owned directory inside the workspace that survives `docker compose down -v`,
  which breaks any later step that walks the tree.

### Scope

Test tooling and CI only. No product code changes: the harness runs against an unmodified
ThunderID, and the 35/35 result was measured on the shipped default authentication flow.

The workflow patches `deployment.yaml` on a throwaway copy of the distribution to bind `0.0.0.0`,
set `public_url`, and enable `oauth.dcr.insecure` so the suite can register clients anonymously.
That last setting is test-only and must never be used in a real deployment.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5082

### Related PRs
- N/A

### Checklist
    - [ ] Unit Tests
    - [x] Integration Tests — the workflow itself; Basic OP passes 35/35 certification-clean.

## Summary by CodeRabbit

- **Tests**
  - Added end-to-end OpenID Connect conformance testing for ThunderID.
  - Added browser-assisted sign-in for front-channel certification scenarios.
  - Added automated setup for test users, clients, authentication flows, and server configuration.
  - Added configurable Basic Certification and discovery testing profiles.
  - Added certification verdict checks, result summaries, logs, and downloadable artifacts.

- **Chores**
  - Added scheduled and manually triggered CI execution.
  - Added readiness, connectivity, input, and configuration validation with clear failure reporting.


tests/oidc-conformance/setup_test_user.py
```diff
@@ -0,0 +1,204 @@
...
+# The bootstrap registers the system resource server with this literal identifier, regardless of
+# the hostname the server is configured on (see bootstrap/01-default-resources.yaml). A token
+# request must echo it exactly: there is no default resource server, so omitting `resource`
+# fails with invalid_target, and deriving it from the base URL fails the same way whenever the
+# conformance run uses a non-default hostname.
+SYSTEM_RESOURCE_IDENTIFIER = "https://localhost:8090/mcp"
+
+# ThunderID serves a self-signed certificate on loopback in this harness. The admin credentials
+# below are the throwaway CI defaults, and the whole exchange stays on the local machine.
+LOCAL_SELF_SIGNED = False
```
thiva-k: Seems these constants are duplicated in multiple places. Shall we define it in a central place?

tests/oidc-conformance/setup_test_user.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+def pkce_pair():
+    verifier = secrets.token_hex(32)[:43]
+    digest = hashlib.sha256(verifier.encode("ascii")).digest()
+    challenge = base64.urlsafe_b64encode(digest).decode("ascii").rstrip("=")
+    return verifier, challenge
+
+
+def get_admin_token(http, base_url, username, password):
```
thiva-k, line 41: Can't we get the admin token once and reuse, rather than doing it again for test user and client setup?

tests/oidc-conformance/setup_test_user.py
Thumimku: moved the function to common.py

tests/oidc-conformance/setup_test_user.py
Thumimku: not needed so its already getting reused, and there are two files reusing it. 
setup_test_user.py → one, in its main()
register_static_clients.py → one, in its main()

both of them will be two different process so to share the token i have to write the token in the disk and read again, i would prefer this way to keep the py files clean. Exactly two mints per conformance run

thiva-k (approved):

PR #5445, opened 2026-09-17
Trust a configured management token for import and the variable store

bhagyasakalanka:
### Purpose

Lets a deployment pipeline call this data plane's import and variable store APIs with a pre-shared
token, instead of first creating a machine-to-machine application and managing its credentials.

A pipeline is not a user. It has no interactive step in which to complete a grant, so requiring an
OAuth client has meant standing up an M2M application whose secret then has to be managed anyway.
This adds a second, deliberately narrow authentication mechanism for the small set of APIs where
that trade is worth making.

Configure it on the deployment:

```yaml
server:
  security:
    management_token: "<token>"
```

and present it as an ordinary bearer token:

```
curl -H "Authorization: Bearer <token>" https://localhost:8090/variables
```

Empty disables it, which is the default. A deployment gets this mechanism only by configuring one.
Token generation is out of scope here: the data plane trusts what it is given.

> **Independent of #5443.** `/variables` and `/secrets` appear in the trusted path list, and those
> routes arrive with the variable store in #5443. The two PRs can merge in either order: this
> authenticator matches paths by pattern and needs no route to exist, so until #5443 lands those two
> entries simply match nothing. `/import` works today.

### Approach

A new `AuthenticatorInterface` implementation registered ahead of the JWT authenticator.

**The comparison happens in `CanHandle`, not `Authenticate`, and that is the design.** A bearer value
that is not this token is left unclaimed, so it falls through to the JWT authenticator exactly as it
does today. Had the authenticator claimed the request and then failed the comparison, every OAuth
call to `/import` would have started returning 401 — the feature would have broken the path it was
meant to sit beside. The comparison is constant time, and the token is never copied into the
security context, because nothing downstream needs it and a credential that is never put somewhere
cannot be logged from there by accident.

**What it opens is a short list, not the management surface.** `/import`, `/import/**`, `/variables`,
`/variables/**`, `/secrets`, `/secrets/**`. The token is one long-lived string in a configuration
file: it cannot be scoped per caller, rotated per client, or revoked without a restart. What it
opens should therefore be only what a pipeline needs. A test enumerates `/users`, `/applications`,
`/organization-units`, `/user-types`, `/groups` and `/flows` and asserts the token is refused on
each, so widening the list later has to be a deliberate act that updates a failing test.

It authenticates as subject `management-token` holding the root system permission, which is what
carries it past the authorization step — both surfaces already require root. Revocation is correctly
a no-op: the enforcer returns early on an empty identity, so there is no lookup for a credential
that has no `jti` and no token family.

Tests cover the authenticator in isolation and through the real `Process` pipeline, since a
mechanism that authenticates but then fails authorization would otherwise pass unnoticed.

### Related Issues
- N/A

### Related PRs
- #5443 — the variable store this token opens alongside the import API. Independent of this PR; they
  can merge in either order.

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests

### Security checks

---

**Naming:** `management_token` follows `direct_auth_secret`'s shape. It was discussed as a "PAT", but
"personal access token" reads wrong for a value living in a server configuration file. Happy to
rename if the team prefers the original term.

**Verification:** `make lint_backend` 0 issues, mocks in sync, 146 backend packages pass. The manual round is
unchecked because no live request has been made against a running server; the mechanism is covered
by unit tests through the real security pipeline, not by an end-to-end call.

## Summary by CodeRabbit

* **New Features**
  * Added optional management-token authentication for the import, variables, and secrets APIs.
  * Authenticated management-token requests receive root system permissions on supported endpoints.
  * Added configuration support for setting the management token; leaving it empty disables this authentication method.
* **Security**
  * Invalid tokens and requests to unsupported paths continue through standard JWT authentication.
* **Documentation**
  * Documented the management-token configuration and supported API paths.


backend/pkg/thunderidengine/config/config.go
```diff
@@ -46,6 +46,21 @@ type SecurityConfig struct {
...
+	//     Authorization: Bearer <token>
+	//
+	// It authenticates as a principal holding the root system permission, and is trusted only on
+	// those paths; every other API stays behind OAuth. Empty disables it, which is the default: a
+	// deployment gets this mechanism only by configuring one.
+	//
+	// It is one long-lived string, so it cannot be scoped per caller or revoked without a restart.
+	// Treat it as a deployment credential, give it to as few callers as possible, and prefer an
+	// OAuth client wherever one can be obtained.
+	ManagementToken string `yaml:"management_token" json:"management_token"`
```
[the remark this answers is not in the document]
bhagyasakalanka: Added the configuration reference.

`server.security.management_token` is now documented in `docs/content/deployment/configuration.mdx`, in the same security table as `server.security.direct_auth_secret`, which is the closest existing precedent. The entry covers the empty default and that empty means disabled, how the credential is presented (`Authorization: Bearer <token>`), the exact paths it is trusted on, that it authenticates as a principal holding the root system permission, that any other bearer value falls through to the normal OAuth check, and the rotation limit: a single long-lived value that cannot be scoped per caller or revoked without a restart.

I did not add a separate guide under `docs/content/guides/` or an entry in `apis.mdx`. This is one configuration setting whose behaviour is fully described where every other `server.security.*` setting is described, and `apis.mdx` has no per-API content to add: it renders a combined OpenAPI spec generated from `api/*.yaml`. If the team would rather have a standalone guide, happy to write one.

docs/content/deployment/configuration.mdx
```diff
@@ -1092,6 +1092,7 @@ Controls server-wide security behavior that is not specific to any single authen
...
+| `server.security.management_token` | `""` (empty) | Pre-shared token that authenticates calls to this deployment's import API (`/import`, `/import/**`) and variable store (`/variables/**`, `/secrets/**`), so a deployment pipeline can push configuration without first obtaining OAuth client credentials. Presented as an ordinary bearer token: `Authorization: Bearer <token>`. It authenticates as a principal holding the **root system permission**, and is trusted **only on those paths**; every other API stays behind OAuth. A bearer value that is not this token is passed to the normal OAuth check, so tokens keep working on the same paths. Empty **disables it**, which is the default. Because it is a single long-lived value it cannot be scoped per caller or revoked without a restart, so treat it as a deployment credential, give it to as few callers as possible, and prefer an OAuth client wherever one can be obtained |
```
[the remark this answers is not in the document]
bhagyasakalanka: Fixed. The row now lists all six: `/import`, `/import/**`, `/variables`, `/variables/**`, `/secrets`, `/secrets/**`.

You were right that it was inconsistent with itself as well as incomplete: it spelled out both the root and the subpath for import, then gave only the subpath form for variables and secrets, even though `managementTokenPaths` has both and the test asserts a bare `GET /variables` is accepted.

Checked the corrected row against the source list programmatically rather than by eye, and the two sets now match exactly.

backend/pkg/thunderidengine/config/config.go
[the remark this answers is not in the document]
bhagyasakalanka: Both parts of this are now done.

The configuration reference you verified is in place, and the detail you flagged afterwards, that the row listed only `/variables/**` and `/secrets/**` while the implementation accepts the base paths too, is fixed: the row now lists all six, `/import`, `/import/**`, `/variables`, `/variables/**`, `/secrets`, `/secrets/**`. I checked the corrected row against `managementTokenPaths` programmatically rather than by eye, and the two sets match exactly.

That was discussed in the thread on `docs/content/deployment/configuration.mdx`, which you have already resolved, so this thread has nothing outstanding.

backend/pkg/thunderidengine/config/config.go
```diff
@@ -46,6 +46,21 @@ type SecurityConfig struct {
 	TrustedIssuer          TrustedIssuerConfig   `yaml:"trusted_issuer"           json:"trusted_issuer"`
 	SystemPermissionPrefix string                `yaml:"system_permission_prefix" json:"system_permission_prefix"`
 	TokenRevocation        TokenRevocationConfig `yaml:"token_revocation"         json:"token_revocation"`
+	// ManagementToken is a pre-shared token that authenticates calls to this deployment's import and
+	// variable store APIs, so a deployment pipeline can push configuration without first obtaining
+	// OAuth client credentials. It is presented as an ordinary bearer token:
```
rajithacharith: I'm thinking of can we use an bootstraped M2M application for importing this. 
If we are going with this approach thinking of using a header like APIKey instead of Authorization 
cc: @darshanasbg @senthalan

rajithacharith: So that, this implementation can be lead to the API Key feature in future

rajithacharith: As discussed, let's proceed with this approach, but let's not use Authorization header for this

bhagyasakalanka: Taken the header suggestion, not the M2M one. Pushed.

**Header.** The key now goes in its own `API-Key` header rather than `Authorization`. Beyond leaving room for a general API key feature, it removes a real awkwardness in the original: the key and an OAuth bearer token were competing for one header, and a request can legitimately carry both. With a dedicated header the two mechanisms are independent, and a caller holding a stale key plus a valid token is still served by the JWT authenticator.

I named it `API-Key` rather than `APIKey` to match the repo's existing `Flow-Secret` and `Attestation-Token`. Say the word if you would rather have it exactly as you wrote it.

**Hashing.** @bhagyasakalanka also asked for the credential to be split across the planes, so the deployment is now configured with the SHA-256 digest, not the key:

```yaml
server:
  security:
    management_api_key_hash: "<sha256-hex-digest>"
```

A data plane only needs to recognize a key, never to present one, so the key stays with whoever issues it and a leaked data plane configuration yields nothing a caller can use. The key carries 256 bits of entropy, so the digest is unsalted, reusing `cryptolib.HashToken` and `ValidateTokenHash` which the challenge token interceptor already uses this way. It is plain SHA-256 hex, so an operator can generate it with `printf '%s' "$KEY" | sha256sum` rather than a bespoke tool.

Three tests pin the new shape: configuring the plaintext key instead of its digest authenticates nothing, presenting the key in `Authorization` is refused, and presenting the digest as though it were the key is refused.

**On the M2M option.** I did not take it. Bootstrapping a machine application to import configuration reintroduces the credential-management step this exists to remove, and the pipeline still ends up holding a long-lived secret, just one with more machinery behind it. If the team would rather go that way it is a different PR rather than a change to this one.

cc @darshanasbg @senthalan

bhagyasakalanka: That was the deciding argument for the header, yes. Two things in the current shape are aimed at that future:

- The credential is presented as `API-Key: <key>`, which is the shape a general API key feature would use, so adopting one later is a matter of where keys come from rather than how callers present them.
- The deployment stores a digest, not the key. A per-key record in a store would hold the same thing, so moving from one configured key to many issued keys changes the lookup (`keyHash` becomes a store query) without changing verification.

What stays deliberately absent is issuance: no generation, no rotation, no per-key scoping. Those are the substance of an API key feature and they need a store and a management API, which is more than this should carry. The narrow path list (`/import`, `/variables`, `/secrets`) is the guard in the meantime, since one unscoped long-lived key should not open the whole management surface.

backend/pkg/thunderidengine/config/config.go
```diff
@@ -46,6 +46,27 @@ type SecurityConfig struct {
...
+	// therefore yields nothing a caller can present. Generate the pair with cryptolib.GenerateSecureToken
+	// and cryptolib.HashToken. The key carries 256 bits of entropy, so the digest is unsalted.
+	//
+	// A caller authenticates as a principal holding the root system permission, and only on those
+	// paths; every other API stays behind OAuth. Empty disables the mechanism, which is the default:
+	// a deployment gets it only by configuring one.
+	//
+	// It is one long-lived key, so it cannot be scoped per caller or revoked without a restart. Treat
+	// it as a deployment credential, give it to as few callers as possible, and prefer an OAuth client
+	// wherever one can be obtained.
```
rajithacharith, line 67: Let's reduce the comment size and mention what are the APIs which can be accessed with this API key

bhagyasakalanka: Done, both parts. It was 19 lines and never actually named the APIs, which was the more useful half of your comment:

```go
// ManagementAPIKeyHash is the SHA-256 hex digest of a key presented as `API-Key: <key>`, which
// authenticates a caller on these APIs and no others:
//
//     /import, /import/**
//     /variables, /variables/**
//     /secrets, /secrets/**
//
// Everything else stays behind OAuth. This is the digest, not the key, so a leaked configuration
// file yields nothing a caller can present. Empty disables it, which is the default.
//
// It is one long-lived key that cannot be scoped per caller or revoked without a restart, so
// prefer an OAuth client wherever one can be obtained. See the deployment configuration guide.
```

What came out was operator guidance that belongs in `docs/content/deployment/configuration.mdx`, which this PR already updates, rather than in a Go doc comment. The generation recipe went with it for the same reason.

I took the same pass over the other comments in the change, since the point applies to them too: `managementAPIKeyPaths` 9 lines to 4, the struct doc 6 to 3, `Authenticate` 4 to 2, the constructor 3 to 1, and `APIKeyHeaderName` 3 to 2. Net 45 lines of comment removed, 23 added.

Two I deliberately kept, because each describes a decision that looks like a mistake without it. `CanHandle` verifies the key rather than `Authenticate` doing it, so a request with a stale key still falls through to JWT instead of becoming a 401; moving that check is the obvious tidy-up and it breaks a working request. And the re-check inside `Authenticate` looks redundant but stops the method becoming a bypass if it is ever called directly.

Build clean, golangci-lint 0 issues, security and cmd/server tests pass.

backend/internal/system/security/management_api_key_authenticator.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if !a.trustsPath(r.URL.Path) {
+		return false
+	}
+
+	presented := utils.SanitizeString(r.Header.Get(constants.APIKeyHeaderName))
+	if presented == "" {
+		return false
+	}
+	return cryptolib.ValidateTokenHash(presented, a.keyHash)
```
rajithacharith: we only need to validate whether the API Key is presented in the CanHandle method. The validation needs to be moved to Authenticate

bhagyasakalanka: Done. `CanHandle` now checks presence only, and the digest comparison moved to `Authenticate`:

```go
func (a *managementAPIKeyAuthenticator) CanHandle(r *http.Request) bool {
	if a.keyHash == "" {
		return false
	}
	if !a.trustsPath(r.URL.Path) {
		return false
	}
	return utils.SanitizeString(r.Header.Get(constants.APIKeyHeaderName)) != ""
}
```

Worth being explicit that this changes behaviour rather than just moving a line, since the service picks the first authenticator whose `CanHandle` returns true and a failed `Authenticate` is a 401 with no second attempt.

Before, a request carrying a wrong API-Key plus a valid bearer token was left to the JWT authenticator and served. Now it is refused. I think your version is the better one: presenting a key is a claim to be authenticated by it, and quietly ignoring a bad credential because another one happens to be present is the kind of thing that reads fine and surprises someone later. Failing closed is easier to reason about.

A request presenting no API-Key header at all is unaffected, so OAuth still works on these paths. `TestOAuthStillWorksOnTheSamePaths` covers exactly that and still passes.

Five unit tests asserted the old semantics and are updated. One had its premise inverted, so it is renamed: `TestAWrongAPIKeyIsLeftToTheJWTAuthenticator` is now `TestAWrongAPIKeyIsRefusedRatherThanIgnored`, and it asserts both halves, that the request is claimed and that it then fails. The path rule is asserted through `CanHandle`, which is where it now lives, since `Authenticate` no longer re-checks it.

Also corrected the comment in the integration test that still described the old fallthrough.

Build clean, golangci-lint 0 issues, security and cmd/server tests pass.

rajithacharith (approved):

PR #5448, opened 2026-09-17
Add a session termination listener to the SSO session service

UdeshAthukorala:
### Purpose

Second deliverable of #5233 (OIDC Back-Channel Logout), per design discussion #5258 section 2. 
Back-channel logout needs to know, for every session that ends, which applications shared it. 
Today only `Terminate` reads that list, and only to revoke token families; `TerminateBySubject` never reads it, and once either transaction commits the participant rows are gone.

This PR adds the hook that observes termination. It is structural: nothing implements the listener yet, so behaviour is unchanged.

- `session.TerminationListener`, `TerminatedSession{SessionID, SubjectID, Participants, Reason}`, and `TerminationReason` (`sign_out`, `subject_revocation`), beside the existing `CriteriaRevoker` interface.
- `SetTerminationListener` on `session.Service`. Nil, the default, disables notification. A setter rather than an `Initialize` argument, because the eventual dispatcher depends on services built after the session service at the composition root.

### Approach

```mermaid
flowchart LR
    T["Terminate<br/>(sign-out)"] --> R1["read participants<br/>inside the transaction"]
    R1 --> D1["revoke families<br/>delete session rows"]
    D1 --> C1["commit"]
    C1 --> L["TerminationListener.OnTerminated<br/>after commit, panic-safe"]
    S["TerminateBySubject<br/>(administrative)"] --> R2["one bulk participant read<br/>before the transaction"]
    R2 --> D2["delete every session's rows"]
    D2 --> C2["commit"]
    C2 --> L
```

- **Listener runs after commit**, so it only hears about terminations that happened and can never fail or roll one back. A panic in it is recovered and logged.
- **`Terminate`** reads the participant list once inside its transaction and shares it between the revoker and the listener. When neither is wired the read is skipped, so the no-listener path is byte for byte what it was.
- **`TerminateBySubject`** captures participants with one bulk read before its transaction (new store method `ListBySessionIDs`, chunked `IN` clause, query `SSO-SESS-17`), groups them by session, and notifies once per session after commit. A failed read logs and proceeds without notifying: telling relying parties is best-effort, ending the sessions is not.
- **Index** `idx_sso_session_subject` on `SSO_SESSION (DEPLOYMENT_ID, SUBJECT_ID)` in both schema scripts. Listing a subject's sessions was an unindexed scan.
- The authentication-time touch query reused id `SSO-SESS-11`; it is renumbered to `SSO-SESS-16`.
- Mocks regenerated with `make mockery`.

Independent of #5409 (`sid`); the two touch different regions of `session/service.go`.

### Related Issues
- Refs #5233

### Related PRs
- Design: #5258
- #5409

### Checklist
    - [x] Unit Tests
    - [ ] Integration Tests (existing `tests/integration/oauth/sso` suite run as regression; no new behaviour to observe yet)

### Security checks

## Summary by CodeRabbit

* **New Features**
  * Added termination notifications for user sign-outs and administrative session revocations.
  * Notifications include the affected session, subject, participants, and termination reason.
  * Added support for retrieving participants across multiple sessions in batches.

* **Performance**
  * Improved administrative termination of all sessions for a subject with more efficient database lookups.

* **Tests**
  * Added coverage for notifications, participant retrieval, batching, errors, and transaction behavior.


backend/internal/flow/session/service.go
```diff
@@ -334,8 +380,19 @@ func (s *service) Terminate(ctx context.Context, handle, flowID string) (*Sessio
...
+	//
+	// The participant list is read once, inside the transaction, and serves both the revoker and the
+	// termination listener: it is the only record of who shared the session, and it is gone once the
+	// transaction commits. When neither is wired the read is skipped, since nothing would consume it.
+	var participants []Participant
 	if txErr := s.transactioner.Transact(ctx, func(txCtx context.Context) error {
-		if revErr := s.revokeSessionFamilies(txCtx, sess.SessionID); revErr != nil {
+		if s.criteriaRevoker != nil || s.terminationListener != nil {
+			var listErr error
+			if participants, listErr = s.store.ListBySessionID(txCtx, sess.SessionID); listErr != nil {
```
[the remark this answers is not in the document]
UdeshAthukorala: Fixed in a86fb29. A failed participant read is now fatal only when the revoker needs the list. With a listener alone, `Terminate` logs, completes the deletes, and skips the notification, matching `TerminateBySubject`. Two tests cover both paths.

backend/internal/flow/session/store.go
```diff
@@ -281,6 +281,43 @@ func (st *store) ListBySessionID(ctx context.Context, sessionID string) ([]Parti
...
+			results, queryErr := dbClient.QueryContext(ctx, buildListParticipantsBySessionIDsQuery(len(chunk)), args...)
+			if queryErr != nil {
+				return fmt.Errorf("failed to execute query: %w", queryErr)
+			}
+			for _, row := range results {
+				p, buildErr := buildParticipantFromRow(row)
+				if buildErr != nil {
+					return buildErr
+				}
+				result = append(result, p)
```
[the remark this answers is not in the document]
UdeshAthukorala: Fixed in a86fb29. When more than one chunk is read, the combined result is stable-sorted by session id then join time; the single-chunk path is unchanged and the caller's slice is not modified. The chunking test now returns rows out of global order and asserts the sorted result.

backend/dbscripts/runtime_persistent/postgres.sql
```diff
@@ -73,6 +73,10 @@ CREATE UNIQUE INDEX idx_sso_session_flow_execution ON "SSO_SESSION" (FLOW_EXECUT
 -- Index for absolute expiry on SSO_SESSION (supports cleanup)
 CREATE INDEX idx_sso_session_absolute_expires_at ON "SSO_SESSION" (ABSOLUTE_EXPIRES_AT);
 
+-- Index for listing a subject's sessions (per deployment), used when every session of a subject is
+-- terminated administratively.
+CREATE INDEX idx_sso_session_subject ON "SSO_SESSION" (DEPLOYMENT_ID, SUBJECT_ID);
```
madurangasiriwardena, line 78: Since this index is not required for the scope of this PR, shall we separate it out?

backend/internal/flow/session/service.go
```diff
@@ -397,9 +466,57 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
...
+	if s.terminationListener == nil {
+		return
+	}
+	defer func() {
+		if r := recover(); r != nil {
+			s.logger.Error(ctx, "Session termination listener panicked",
+				log.String("sessionId", ended.SessionID), log.Any("panic", r))
+		}
+	}()
+	s.terminationListener.OnTerminated(ctx, ended)
```
madurangasiriwardena, line 517: Since the OnTerminated is supped to work asynchronous work, ctx might not be available after this request is completed.

backend/internal/flow/session/service.go
```diff
@@ -63,6 +63,11 @@ type Service interface {
 	// nil when the subject holds no sessions.
 	TerminateBySubject(ctx context.Context, subjectID string) error
 
+	// SetTerminationListener installs the listener notified after Terminate or TerminateBySubject
+	// commits. Nil disables notification. Call it once during wiring, before serving requests; it
+	// is a setter because the listener depends on services built after this one.
+	SetTerminationListener(listener TerminationListener)
```
madurangasiriwardena, line 69: Since this in the public interface, anyone can update this field at any time. So shall we guard it and make sure unwanted updates to this field can be done.

madurangasiriwardena, line 69: On the other hand, we allow only one listener implementation with this design. Is it correct to assume that one is enough? Can there be situations in the future that there is a need for more than one listener?

backend/dbscripts/runtime_persistent/postgres.sql
UdeshAthukorala: Agreed, it's not needed for this change. Removed from this PR in c387315. I'll raise the index separately.

backend/internal/flow/session/service.go
UdeshAthukorala: Agrred. I fixed it to pass context.WithoutCancel(ctx) to the listener (c387315).

backend/internal/flow/session/service.go
UdeshAthukorala: Agreed, it shouldn't be on the interface at all. In c387315 I removed `SetTerminationListener` from `Service`. `Initialize` now returns a `TerminationHook` that only the composition root holds.

UdeshAthukorala: Agreed, something like a CAEP session-revoked emitter or an audit hook would want this event too. 
So made it additive in c387315.

backend/internal/flow/session/service.go
```diff
@@ -350,6 +439,14 @@ func (s *service) Terminate(ctx context.Context, handle, flowID string) (*Sessio
...
 
 	s.logger.Debug(ctx, "Terminated SSO session", log.String("flowId", sess.FlowID))
+	if notify {
+		s.notifyTerminated(ctx, listeners, TerminatedSession{
+			SessionID:    sess.SessionID,
+			SubjectID:    sess.SubjectID,
+			Participants: participants,
+			Reason:       TerminationReasonSignOut,
+		})
+	}
```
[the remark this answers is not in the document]
UdeshAthukorala: Checked this. Terminate and TerminateBySubject are only called from the sign-out and session revocation executors, and the flow engine never runs node execution inside a transaction, so there is no outer transaction for Transact to reuse here. Reuse is also keyed per database, so a transaction on another DB would not be picked up. The criteria revoker already runs inside this same transaction with the same assumption. Adding an after-commit hook to the transaction package for a caller that does not exist is out of scope for this PR.

WDYT @coderabbitai ?

backend/internal/flow/session/service.go
```diff
@@ -129,14 +131,81 @@ type CriteriaRevoker interface {
 	RevokeTokenFamily(ctx context.Context, tokenFamilyID string) error
 }
 
+// TerminationReason names the path that ended a session.
+type TerminationReason string
```
ThaminduDilshan, line 135: Shall we move these to a constants/ model file? Check whether any existing files exists fitting this

backend/internal/flow/session/service.go
```diff
@@ -129,14 +131,81 @@ type CriteriaRevoker interface {
...
+type TerminationListener interface {
+	OnTerminated(ctx context.Context, ended TerminatedSession)
+}
+
+// TerminationHook registers termination listeners on one session service. Initialize hands it to the
+// composition root, which needs it because listeners depend on services built after the session
+// service. Listeners are notified in registration order, each isolated from the others' panics. The
+// list is replaced atomically on every registration, so registering may race with terminations
+// already in flight.
+type TerminationHook struct {
```
ThaminduDilshan, line 170: Shall we define an interface for this and only expose the interface to the outside? Let's not expose unnecessary structs, constructor methods, etc to outside and keep them package internal

ThaminduDilshan, line 170: On a related note, better to move listener/ hook related code to a separate file moving it out from service.go

backend/internal/flow/session/store_constants.go
```diff
@@ -58,7 +61,7 @@ var (
 	// records a fact about an authentication that has already happened: losing a race with a
 	// concurrent slide must not discard it.
 	queryTouchAuthenticatedAt = model.DBQuery{
-		ID: "SSO-SESS-11",
+		ID: "SSO-SESS-16",
```
ThaminduDilshan, line 64: Is it intentional to change this id?

backend/cmd/server/servicemanager.go
```diff
@@ -575,15 +577,17 @@ func unregisterServices() {
...
+	criteriaRevoker flowsession.CriteriaRevoker, logger *log.Logger,
+) (flowsession.Service, flowsession.TerminationHook, flowsession.Config) {
 	cfg := readSessionConfig(ctx, svc, logger)
-	sessionService, err := flowsession.Initialize(dbprovider.GetDBProvider(), deploymentID,
+	sessionService, terminationHook, err := flowsession.Initialize(dbprovider.GetDBProvider(), deploymentID,
 		flowsession.NewTimeouts(cfg.IdleTimeoutSeconds, cfg.AbsoluteTimeoutSeconds,
 			cfg.ActivityRefreshIntervalSeconds), criteriaRevoker)
 	fatalOnError(ctx, logger, err, "Failed to initialize SSO session service")
-	return sessionService, cfg
+	return sessionService, terminationHook, cfg
```
ThaminduDilshan, line 590: What is the purpose of returning terminationHook to the service manager? Is it something coming in with a separate PR? Because currently we don't have any usages for this even it's return here from `initSessionService` method

backend/internal/flow/session/service.go
```diff
@@ -129,14 +131,81 @@ type CriteriaRevoker interface {
...
+		return errors.New("session termination listener must not be nil")
+	}
+	for {
+		current := h.svc.terminationListeners.Load()
+		var next []TerminationListener
+		if current != nil {
+			next = append(next, *current...)
+		}
+		next = append(next, listener)
+		if h.svc.terminationListeners.CompareAndSwap(current, &next) {
```
ThaminduDilshan, line 186: Do we need the atomic pointer and CAS loop here? Listeners are only added from the service manager at startup, before any request is served

backend/internal/flow/session/service.go
```diff
@@ -397,9 +498,64 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
...
+				Reason:       TerminationReasonSubjectRevocation,
+			})
+		}
+	}
 	return nil
 }
 
+// participantsForTermination returns the participants of the given sessions grouped by session id.
+// It returns nil, meaning nothing to notify, when no listener is wired or the read fails.
+func (s *service) participantsForTermination(ctx context.Context, sessions []Session) map[string][]Participant {
```
ThaminduDilshan, line 517: Shall we move these private methods after the exported ones? Same for `listeners()`. Unless they move to the new file with the hook code

backend/internal/flow/session/store.go
```diff
@@ -281,6 +283,52 @@ func (st *store) ListBySessionID(ctx context.Context, sessionID string) ([]Parti
...
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(sessionIDs) > participantsBySessionIDsChunkSize {
```
ThaminduDilshan, line 321: Do we need this sort? All rows of a session come from the same chunk, so they're already ordered by join time. The only caller groups them by session id anyway

backend/internal/flow/session/service_test.go
```diff
@@ -834,3 +835,303 @@ func (suite *ServiceTestSuite) TestDetachApplication_EmptyAppIDIsANoOp() {
...
+	runTx(m)
+	// One read serves both the revoker and the listener.
+	m.store.EXPECT().ListBySessionID(mock.Anything, "sess-1").Return(participants, nil).Once()
+	revoker.EXPECT().RevokeTokenFamily(mock.Anything, "tfid-a").Return(nil)
+	revoker.EXPECT().RevokeTokenFamily(mock.Anything, "tfid-b").Return(nil)
+	m.store.EXPECT().DeleteSession(mock.Anything, "sess-1").Return(nil)
+	m.store.EXPECT().Delete(mock.Anything, "sess-1").Return(nil)
+	committed := false
+	m.store.EXPECT().DeleteBySessionID(mock.Anything, "sess-1").RunAndReturn(
+		func(context.Context, string) error { committed = true; return nil })
```
ThaminduDilshan, line 917: `committed` is set inside the transaction callback, so this passes even if the listener is called inside the transaction. Shall we set it from the `Transact` mock after `fn` returns, like in `TestTerminateBySubject_NotifiesOncePerSessionAfterCommit`?

backend/internal/flow/session/service_test.go
```diff
@@ -834,3 +835,303 @@ func (suite *ServiceTestSuite) TestDetachApplication_EmptyAppIDIsANoOp() {
...
+	m.store.EXPECT().ListBySessionID(mock.Anything, "sess-1").Return([]Participant{{AppID: "app-1"}}, nil)
+	m.store.EXPECT().DeleteSession(mock.Anything, "sess-1").Return(errors.New("db down"))
+
+	_, err := svc.Terminate(context.Background(), "handle-abc", "flow-1")
+
+	suite.Require().Error(err)
+	listener.AssertNotCalled(suite.T(), "OnTerminated", mock.Anything, mock.Anything)
+}
+
+func (suite *ServiceTestSuite) TestTerminate_ListenerPanicDoesNotFailTermination() {
```
ThaminduDilshan, line 1004: Do we need this test? `TestTerminate_NotifiesEveryListenerDespiteAPanic` already covers a panicking listener

backend/internal/flow/session/service.go
```diff
@@ -397,9 +498,64 @@ func (s *service) TerminateBySubject(ctx context.Context, subjectID string) erro
...
+// listeners get a context that keeps the request's values but not its cancellation, because the
+// request ends as soon as its response is written while their work may not.
+func (s *service) notifyTerminated(ctx context.Context, listeners []TerminationListener, ended TerminatedSession) {
+	detached := context.WithoutCancel(ctx)
+	for _, listener := range listeners {
+		s.notifyOne(ctx, detached, listener, ended)
+	}
+}
+
+func (s *service) notifyOne(ctx, detached context.Context, listener TerminationListener, ended TerminatedSession) {
```
thiva-k, line 549: Add method level comments

backend/internal/flow/session/service.go
```diff
@@ -129,14 +131,81 @@ type CriteriaRevoker interface {
...
+const (
+	// TerminationReasonSignOut is a sign-out of one session.
+	TerminationReasonSignOut TerminationReason = "sign_out"
+	// TerminationReasonSubjectRevocation is an administrative termination of all the subject's sessions.
+	TerminationReasonSubjectRevocation TerminationReason = "subject_revocation"
+)
+
+// TerminatedSession describes a session the service has ended. Participants is captured before the
+// rows are deleted, so it is the only remaining record of which applications shared the session.
+type TerminatedSession struct {
```
thiva-k, line 146: Shall we move all the structs to model.go

thiva-k:
IMO we should do this with a generic event layer. Currently if we are to have a new event type we may need new interface, hooks, etc. right? WDYT @ThaminduDilshan

backend/internal/flow/session/store_constants.go
UdeshAthukorala: Yes, that's intentional. On main SSO-SESS-11 is used twice, once for the touch query and once for the participant delete. Since I was adding a new query here anyway I gave the touch query the next free id.
If you'd rather keep this PR strictly to the listener I can drop it and fix the duplicate in a separate PR.

backend/cmd/server/servicemanager.go
UdeshAthukorala: Yes, this is for the back-channel logout dispatcher PR. The dispatcher needs services that get built after the session service, so I can't pass it into Initialize. The idea is that the service manager holds on to the hook and registers the dispatcher through it later, similar to how the consent enforcer is wired. Since this PR only adds the seam, for now it's returned and discarded with a comment on what will use it.

backend/internal/flow/session/service.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service.go
UdeshAthukorala: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/store.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service_test.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service_test.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

backend/internal/flow/session/service.go
UdeshAthukorala: Fixed with: https://github.com/thunder-id/thunderid/compare/b03889a3d8ae240a83cb02c271b732732cfeda02..e1eb52ed8acd8ed506b605bab1a474d80bd35b71

PR #5456, opened 2026-09-18
 Notify Google Chat when the OIDC conformance run fails

Thumimku:
### Purpose

The nightly OIDC conformance run reports nowhere. A regression is only noticed by someone opening the Actions tab, and once found, the run tells you which modules failed but not why.

This makes a failure reach people, and makes it diagnosable without reproducing it:

- A Google Chat message when the run fails, matching how the other workflows here report.
- The suite's own log for each failed module, archived with the run's artifacts.
- A conformance badge in the README, so the current state is visible without going looking.
- 

oidc

### Approach

**Notification, failure only.** Modelled on the block in `nightly-build-validation.yml`: skip quietly when `GOOGLE_CHAT_WEBHOOK` is unset, build the card with `jq`, post with `curl -f`. It fires on failure only, as every other workflow here does. A nightly green tick is what the badge is for; a message a day is a message people stop reading.

The card names the profile and the failed modules, read back from the results file the run already writes. That file is the useful part and also the fragile part, since a run that dies before the suite starts never writes one, which is exactly how the last failure went. Every field degrades to a placeholder rather than failing the step that is trying to report a failure. Long module lists are truncated, because Chat rejects an oversized card.

The validated `--profile` value is exported through `GITHUB_ENV` so the card can name it. It is exported after the existing `^[a-z0-9-]+$` check, so nothing unvalidated reaches it.

**Failure logs.** `run_tests.py` already had a `get_log` method that nothing called. `GET /api/log/{id}` returns the entries the suite shows on its own results page, one per condition, carrying `result` (`FAILURE`, `WARNING`, `INFO`, `SUCCESS`, `REVIEW`), `src` (the condition class) and `msg`. Failed modules now have theirs written to `conformance-failures.log` in the run's artifact, keeping the `FAILURE` and `WARNING` entries and whatever evidence each condition attached.

The capture happens as each module resolves, not at the end, because the suite is torn down with `docker compose down -v` when the job finishes. The plan URL in the results file points at nothing by the time anyone reads the notification, so this is the copy that survives. `--failure-log ""` turns it off.

**Badge.** A `## Conformance` section with the OIDC badge, via shields.io like the four other badges in the README. The label is set explicitly rather than inherited from the workflow name, so it stays correct if the workflow is renamed, and a VC suite badge can join the same row later. The workflow is also renamed from `OIDC Conformance Test` to `OIDC Conformance`, which fixes its name in the Actions sidebar.

### Scope

CI and test tooling only. No product code, and no change to what the conformance run asserts or how it passes.

### Related Issues
- https://github.com/thunder-id/thunderid/issues/5082

### Related PRs
- Follows #5444, which added the conformance harness.

### Checklist
    - [ ] Unit Tests
    - [ ] Integration Tests

## Summary by CodeRabbit

- **New Features**
  - Added an OIDC Conformance badge to the README, linking to the conformance workflow.
  - Conformance runs now provide downloadable failure and warning logs.
  - Failed runs can send summary notifications to Google Chat when configured.

- **Improvements**
  - Failure logs redact sensitive credentials before being saved or uploaded.
  - Notifications include the selected profile, affected modules, commit details, workflow link, and failure date.
  - Long failure lists are shortened for easier reading.


README.md
```diff
@@ -8,6 +8,11 @@
 [![codecov.io](https://codecov.io/github/thunder-id/thunderid/coverage.svg?branch=main)](https://codecov.io/github/thunder-id/thunderid?branch=main)
 [![GitHub Release](https://img.shields.io/github/v/release/thunder-id/thunderid?color=blue)](https://github.com/thunder-id/thunderid/releases/latest)
 
+## Conformance
```
senthalan: Shall we not have this as a heading rather a normal test

Thumimku: Added via https://github.com/thunder-id/thunderid/pull/5456/commits/425a26607b2b1afbcf73fd06d68b5d5c4e0ff462

thiva-k (approved):

